PR #19, opened 2015-05-01
rdl enhancements and testing

grondo:
This PR includes a small RDL enhancements including testing and Travis CI support for flux-sched.

 * Add lua-based tests for RDL under `t/lua` using `fluxometer` directly from `flux-core`
 * Add `.travis.yml` file for Travis-CI builds. We build against `flux-core/master`
 * RDL Resource classes now support a list of default tags as a class attribute, empty by default. Default tags can be set on any type and is inherited by all subtypes. (Therefore setting default tags on the Resource type will set those tags on all resources in a hierarchy). Tags can be overridden in derived types. e.g.
 ```
  Resource:default_tags { "idle" }
  Hierarchy "default" { ... } 
 ```
Default tags can also be passed directly on the `uses` line, e.g.

 ```
 uses "Node" :default_tags { "mytag" }
 ```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,33 @@
+language: c
+
+compiler:
+  - gcc
+  - clang
+
+install:
+  - sudo sh -c 'echo "/usr/local/lib" >/etc/ld.so.conf.d/build'
+  - sudo apt-get update -q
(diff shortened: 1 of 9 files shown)


grondo:
Closing this. I will open a separate PR for travis support vs. any rdl enhancements/tweaks.

PR #22, opened 2015-05-08
rdl testing and small enhancements

grondo:
This PR adds some nominal RDL tests (for the Lua code) under `t/lua/*.t`, as well as some minor enhancements to RDL, including:

 * Set default tags on resource types, inherited by all base types. See `t/lua/t0003-default-tags.t` for examples.
 * A reference to loaded type is returned with `uses` function, so `local N = uses "Node"` now works.
 * Resource class is loaded as if by `uses` command, so the class is reinitialized with each new RDL file or string parsed.

### DIFF
--- a/rdl/RDL.lua
+++ b/rdl/RDL.lua
@@ -47,6 +47,7 @@ local function rdl_parse_environment ()
         pairs = pairs,
         tonumber = tonumber,
         assert = assert,
+        unpack = unpack,
         table = table,
         hostlist = hostlist
     }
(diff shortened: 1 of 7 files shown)


grondo:
OK, I've rebased this PR onto latest master after Travis enhancements were landed. Once we hear back from travis-ci, I would consider this open for review and merge.

grondo:
Ok, please review and consider merging this PR. Thanks.

PR #24, opened 2015-05-08
clang and build fixes

grondo:
This PR contains fixes for a couple warnings when building with `clang` in travis-ci (which are hard errors due to -Werror), and also rectifies an issue with commit 826a3b48f6906336ea392228a0efeedbcc2dcb48 wherein `check` was accidentally made the default target of `make` instead of `all`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,20 @@ include Makefile.inc
 
 SUBDIRS = rdl sched simulator t
 
-all:
+.PHONY: all clean $(SUBDIRS)
+all check clean: $(SUBDIRS)
 
+all: TARGET="all"
+clean: TARGET="clean"
 check: all
+	$(MAKE) TARGET="check"
 
-all clean install check:
-	for subdir in $(SUBDIRS); do make -C $$subdir $@; done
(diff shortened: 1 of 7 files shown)


grondo:
Let's wait to hear back from travis before merging this one. If there are any other travis-ci related buglets I'll add them to this PR.

grondo:
Hm, I do notice that while the build and some tests fail, `make check` is reporting success. I'll try to fix that here as well.

grondo:
Ok, I've added 2 more commits to this PR that 
 * remove the `for subdir in $(SUBDIRS)` loop that was hiding any errors in submake
 * Exit `make check` with non-zero status if any of the tests in `t` fail.

Still probably need to review the rest of the tests to make sure they'll propagate their errors as well.

grondo:
Ok, looks like I've found and fixed all clang build errors. This PR is probably ready to be reviewed and merged.

dongahn:
Great! Flux-sched can now rely on Travis for CI.  I plan to soon add some more basic scheduler service shardness tests as well.

PR #28, opened 2015-05-19
Alternate Resource Model

lipari:
Add an alternate resource representation implemented in C that supports a variety of resource connectivity models. Includes conversion of the sched framework and plugin API to the new model.  Also includes updates to track the latest changes to flux-core.


garlick:
Couple minor nits in the commit to update the API 8b22405 
* unchecked `asprintf ()` call in `sched/schedsrv.c::request_run()`
* zmsg leaked on failure in `simulator/simsrv.c::send_start_event()`

dongahn:
OK. I went over the code with Don. This really represents lots of work. Good job @lipari! There are a few minor nits came up (like having to conform to a recently added style -- an opaque type being a pointer to a type etc). But given we have far more core technologies we will have to push forward in the scheduler area, I propose we fold this in as is and move onto other more fundamental topics. I will wait until tomorrow just in case. If I don't hear from the others then, I will fold this in.

PR #29, opened 2015-05-29
Cosmetic

lipari:
Cosmetic changes to conform to new style recommendations in RFC 7 as well as more uniformity to the format of flux_log() messages.  No changes to the logic or behavior.


dongahn:
This looks good, Don. Once I hear from Travis, I will merge this in.

PR #30, opened 2015-06-04
Jsc sched

lipari:
Cut over to JSC.  Replaces the original implementation where job info was written directly to the KVS with the new Job Status and Control API now in flux-core.


lipari:
Need to fix clang warning

PR #31, opened 2015-06-04
Jsc sched

lipari:
Cut over to JSC. Replaces the original implementation where job info was written directly to the KVS with the new Job Status and Control API now in flux-core.


dongahn:
@lipari this looks good. I will merge this as is. As we spoke, there are Makefile changes and function pointer name changes (my codes) I would like to do. But I will quickly work on it after merging this and issue another PR.

PR #36, opened 2015-06-10
resrc: eliminate the hard-coded memory pool

lipari:
Use a resource's size field to determine whether the resource is a
pool.  This is a transitional commit that provides a practical replacement for a hard-coded placeholder.  I'll be looking into streamlining this area even further.


dongahn:
Looks good to me!

PR #38, opened 2015-06-26
resrc: redesign pool support

lipari:
Eliminate the dedicated pool structure and enhance the resrc structure
to support a size component.  This makes the resrc model conform with
the design described in RFC 4.  Update the tresrc test to match the
new model.  No changes were needed to the sched module.

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -42,28 +42,21 @@ static hostset_t hostset = NULL;
 typedef struct zhash_t resources;
 typedef struct zlist_t resource_list;
 
-typedef struct {
-    int64_t avail_items;
-    int64_t selctd_items;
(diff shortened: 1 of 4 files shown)


dongahn:
@lipari: as we talked on the phone, this seems very useful!

The only comment I have is that, as part of flux-core #223 (Flux should sanitize environment for launched programs), SLURM_NODELIST will be stripped away and e306f01 may break. 

I plan to add an option to export SLURM_NODELIST as part of flux-core #221 (support bootstrapping legacy mvapich v1 builds in Flux) from within wrexec*, but this may or may not be visible to when/where this code is used. 

I will merge this in, and we should follow up when these environment variables are stripped.

PR #39, opened 2015-07-07
Replace deprecated flux_sendmsg() in sched module

lipari:
And fixed a bug along the way.


dongahn:
Thanks for being on top of this, Don. Perhaps we might as well change the data type from zmsg_t to flux_msg_t? Also change the variable name from zmsg to just msg or fmsg?

lipari:
Good suggestions, Dong!  Done and re-pushed.

PR #41, opened 2015-07-15
Property support

lipari:


dongahn:
It seems travis-CI fails to build because it couldn't resolve zhash_set_destructor, a function newly used in this commit. 

From LC environment (e.g., cab) where we have czmq 3.0.0, this function is declared in zhash.h but the czmq 3.0.2 used in Travis, this destructor is not defined in that header file. Instead, it seems 3.0.2 introduced a new header file (zhashx.h) and there zhashx_set_destructor is defined. 

I there any reason we use czmq 3.0.2  instead of 3.0.2 in  travis-dep-builder.sh?

garlick:
Try including <czhq.h> instead of the headers for each class that's being used?

dongahn:
@garlick I think it is included that way already. I think this is likely a version mismatch between czmq on LC and czmq on Travis...

grondo:
The version on LC is a 3.0.0 release candidate, I wonder (I hope) that is why the zeromq developers changed interfaces on a micro version update??

I'm also seeing that they changed the prototype for `zlist_compare_fn` between LC and czmq-3.0.2

on LC 3.0.0-rc1:
```
//  @interface
//  Comparison function for zlist_sort method
typedef bool (zlist_compare_fn) (void *item1, void *item2);
```

in czmq-3.0.2
```C
/  @warning THE FOLLOWING @INTERFACE BLOCK IS AUTO-GENERATED BY ZPROJECT!
//  @warning Please edit the model at "api/zlist.xml" to make changes.
//  @interface
// Comparison function e.g. for sorting and removing.
typedef int (zlist_compare_fn) (
    void *item1, void *item2);
```

garlick:
Sigh.  I just checked and that function was there in 3.0.0 (not release candidate) and gone in 3.0.1.

The stuff in zhashx.h is a different class.

dongahn:
@grondo: this explains the whole picture then... Perhaps time to update czmq to 3.0.2 on LC instead?

grondo:
I'm not sure what to do here -- this is going to make grabbing dependencies difficult for users...

garlick:
Here is the offending [commit](https://github.com/zeromq/czmq/commit/505dc87bc6301c8dfee558f6106678f1b4944efc)

lipari:
I have a workaround.  There is an alternate means to free the hash values.  I'll make the change and squash all the attempts into one commit.

grondo:
Ok, (not to co-opt this PR for too much packaging discussion) this tells me that the developers don't really intend libczmq to be an systemwide installed library used by multiple projects. We might need to either pull a version of the library into our tree (perhaps with git-submodule), or extricate usage of czmq interfaces from our sources.

garlick:
At least we caught this with testing, and are eliminating czmq types from our public interfaces so if it really goes of the rails we can do something about it internally.

dongahn:
@lipari: This now fails with the same issue @garlick hit with his argv PR... You might need a macro to handle this one. czmq at least defines the version macros so we can take advantage of these? 

#define CZMQ_VERSION_MAJOR 3
#define CZMQ_VERSION_MINOR 0
#define CZMQ_VERSION_PATCH 0

grondo:
@dongahn I have a "fix" (workaround) in my travis-ci updates branch already

grondo:
Ok, I've fixed up the travis hiccups in PR #44. once that is merged @lipari, you should be able to rebase and checks should likely pass.

dongahn:
Looks good. Thanks!

lipari:
And thanks to you and @grondo !

PR #42, opened 2015-07-23
modules: update mod_main() prototype per RFC 5

garlick:
This change goes along with flux-framework/flux-core#289 where the prototype of `mod_main()` changed to use argc, argv style argument passing rather than a zhash_t.

I repeated that silly `zhash_fromargv()` several times thinking that people would probably want to revise the way arguments are handled in those modules one at a time;  feel free to move that somewhere shared if you want to keep it.

### DIFF
--- a/sched/schedsrv.c
+++ b/sched/schedsrv.c
@@ -835,6 +835,25 @@ static int job_status_cb (JSON jcb, void *arg, int errnum)
     return action (ctx, j, ns);
 }
 
+static zhash_t *zhash_fromargv (int argc, char **argv)
+{
+    zhash_t *args = zhash_new ();
+    int i;
+
+    if (args) {
+        for (i = 0; i < argc; i++) {
+            char *key = xstrdup (argv[i]);
+            char *val = strchr (key, '=');
+            if (val) {
+                *val++ = '\0';
+                zhash_update (args, key, xstrdup (val));
+                zhash_freefn (args, key, free);
+            }
+            free (key);
+        }
+    }
+    return args;
+}
 
 /******************************************************************************
  *                                                                            *
@@ -842,12 +861,15 @@ static int job_status_cb (JSON jcb, void *arg, int errnum)
  *                                                                            *
  ******************************************************************************/
 
-int mod_main (flux_t h, zhash_t *args)
+int mod_main (flux_t h, int argc, char **argv)
 {
     int rc = -1;
     ssrvctx_t *ctx = NULL;
     char *schedplugin = "sched.plugin1";
+    zhash_t *args = zhash_fromargv (argc, argv);
 
+    if (!args)
+        oom ();
     if (!(ctx = getctx (h))) {
         flux_log (h, LOG_ERR, "can't find or allocate the context");
         goto done;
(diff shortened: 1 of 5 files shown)


dongahn:
Travis complains about the comparator function used in the scheduler simulator. Probably changing

> bool compare_job_t (void *item1, void *item2)

to

> int compare_job_t (void *item1, void *item2)

should fix the issue. I don't know why it just fails just now, though. @SteVwonder, is this your code? Does this make sense?

grondo:
Hm, I'm inclined to merge this because without it flux-sched tests are segfaulting when the sched module is loaded:
```
Core was generated by `/g/g0/grondo/git/f/src/broker/.libs/lt-flux-broker --size=1 --rank=0 --sid=2055'.
Program terminated with signal 11, Segmentation fault.
#0  0x00002aaaab7a9355 in ?? () from /usr/lib64/libczmq.so.1
Missing separate debuginfos, use: debuginfo-install czmq-3.0.0.rc1-1.ch5.3.x86_64 json-c-0.11-12.ch5.3.x86_64 libgcc-4.4.7-11.el6.x86_64 libsodium-0.4.5-3.el6.x86_64 libstdc++-4.4.7-11.el6.x86_64 libuuid-2.17.2-12.18.el6.x86_64 lua-5.1.4-4.1.el6.x86_64 munge-libs-0.5.11-1.ch5.1.1.x86_64 openpgm-5.1.118-3.el6.x86_64 zeromq4-4.0.5-2.ch5.3.x86_64
(gdb) where
#0  0x00002aaaab7a9355 in ?? () from /usr/lib64/libczmq.so.1
#1  0x00002aaaab7a9523 in zhash_lookup () from /usr/lib64/libczmq.so.1
#2  0x00002aaae022daab in load_rdl (ctx=0x2aaae8009070, args=0x1)
    at schedsrv.c:347
#3  0x00002aaae022ecc3 in mod_main (h=0x2aaae8000ab0, args=0x1)
    at schedsrv.c:865
#4  0x000000000040c23d in module_thread (arg=0x644090) at module.c:133
#5  0x00002aaaabc6e9d1 in start_thread (arg=0x2aaae0edf700)
    at pthread_create.c:301
#6  0x00002aaaac8008fd in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:115
(gdb) quit
```

Shortly thereafter I'll submit a PR to fix all the travis weirdness (or workaround it) which should get the builds passing again.

@garlick, think it is OK to merge?

dongahn:
@lipari's patch should fix this czmq 3.0.2 issue. But he just hit another problem totally unrelated to his changes. If we figure that one out, I think I can merge Don's PR. Then, rebasing this PR to the new master will fix this issue as well and at that point I can merge this one in as well. Make sense?

grondo:
@dongahn, no his patch won't work since new `zlist_sort` expects traditional compare return values (< 0, 0, >0)

Can we please merge this one first, then I can make sure all the travis failures are resolved in my branch, then we can look at @lipari's PR

dongahn:
OK. I will merge this one first.

grondo:
Thanks!!

PR #44, opened 2015-07-24
travis updates and build fixes

grondo:
The travis-ci build had a few issues which should be addressed here:
 * We weren't getting the latest flux-core/master because the git repo was being updated via `git-pull`, which was failing due to changes in the working tree. We now use a hard reset to ensure we're building against latest flux-core
 * Latest flux-core now has python dependencies and these are now installed
 * Latest flux-core removes the libuuid dependency which broke linking in `resrc`. Since resrc directly requires libuuid it should specify that on link line
 * We need to support two styles of comparator functions for zlist in simulator, depending on czmq version. (Hopefully I got the compare function correct enough, it only has build testing)


grondo:
Thanks!!

PR #45, opened 2015-08-03
Issue 43

lipari:
Removes the temporary solution that was put in when the module main function prototype changed to pass arguments in argc, argv format rather than a zhash_t (core PR 289).  While the changes to the simulator reflect the mods to sched, I can only guarantee that the simulator code compiles.


dongahn:
Looks good otherwise!

dongahn:
Nicely done. Thanks!

PR #47, opened 2015-08-17
resrc: Move the resource request code into its own module

lipari:
Breaks the close association of a resource request to a resource tree.
Will allow a resource request to include other graph associations
(e.g., flow) independently from its support for a resource tree.
This commit introduces no functional changes to the code.


resrc/resrc_reqst.c
```diff
@@ -0,0 +1,397 @@
...
+#include <stdbool.h>
+#include <string.h>
+#include <assert.h>
+#include <czmq.h>
+
+#include "rdl.h"
+#include "resrc_reqst.h"
+#include "src/common/libutil/xzmalloc.h"
+
+typedef struct zlist_t resrc_reqst_list;
```
dongahn: @lipari : sorry I'm so fixated on this pattern: a  chain of incomplete types. 

I just want to understand what's going on here better to convince myself. Perhaps @trws or other project members can help me out. I see the following patterns many times in the implementation files for resrc and the scheduler plugin:

`typedef struct zlist_t analias` or `typedef struct zhash_t analias`

And in the header files (e.g., resrc_reqst.h) for public APIs:

`typedef struct analias another_alias`

I agree that we want to hide the implementation details like zhash_t or zlist_t from the public header files, but i just can't wrap my head around this syntax.

For example, zlist_t is actually an alias to struct _zlist_t from czmq.h and the typedef in the implementation file appears to me like 

`typedef struct struct _zlist_t alias`

Or a new forward declaration to an incomplete type struct zlist_t, in which case there could be a type safety issue...

Even though this compiles and works, I'm wondering if there is a better syntax that allow people like me to understand...

trws: I actually put these in initially.  If you look in one of my branches at an alternate version, you'll see the more normal pattern, like this:

```c
// In header
struct req_list; // forward-declare type
typedef struct req_list resrc_reqst_list;
// In implementation
struct req_list {
  // ...
};
```

It's chained to other types in this way because the external header file needs to not contain a direct reference to a zlist.  Even so, the implementation file can't contain the type definition, because that's defined in the czmq library, so the type has to be forwarded onto the definition of the struct by czmq.  Nowhere in our code, or a user's code, can they create a zlist-sized type, because it's an incomplete type everywhere, but the type-system still checks them all correctly short of casting them through another pointer type.

dongahn: Hmmm... I guess I understand the normal pattern you showed. But I just have a hard time to see where  `struct resrc_reqst_list` declared, which the header file typedefs -- unless there is a syntax that I don't understand, which is more than a possibility :-) 

And I'm also unsure if 

'typedef struct zlist_t resrc_reqst_list` in the header file forward-declares zlist_t because czmq declares `struct _zlist_t` and `zlist_t`, not `struct zlist_t`.

resrc/resrc_reqst.c
```diff
@@ -0,0 +1,397 @@
...
+    }
+}
+
+/***********************************************************************
+ * Resource request list
+ ***********************************************************************/
+
+resrc_reqst_list_t resrc_reqst_new_list ()
+{
+    return (resrc_reqst_list_t) zlist_new ();
```
dongahn: FYI -- @lipari and I just did some experiment by compiling the code without the typecasting in this routine and the compilation fails with "incompatible pointer type" warning message.  

My suspicion is:

`typedef struct resrc_reqst_list *resrc_reqst_list_t` in the header file creates an alias (resrc_reqst_list_t) to an incomplete type which never gets resolved. I think this should be addressed and @lipari will take a look at this.

lipari: I propose a minor mod of casting to a void* in the .h file to look like this:

-typedef struct resrc_reqst_list *resrc_reqst_list_t;
+typedef void *resrc_reqst_list_t;

and removing the following line from the .c file:

-typedef struct zlist_t resrc_reqst_list;

@dongahn worries that casting to a void * is dangerous because it won’t allow compile time type checks (e.g., in a very rare case, if somehow the compiler picks up a list_new () with a different return type, the compiler won’t be able to detect this.)  He would like a third opinion before merging.

trws: This is in no way an improvement.  The struct is not being overridden in 
this version of the RDL, but it is there so that other implementations 
of the interface can define it with a real structure and then use the 
field information for those structures.  My test implementation with 
sqlite did this, and switching to `void *` would just force more 
implementations to cast everywhere.

Why is the current version a problem?

dongahn: My main problem with the current version is it is confusing. 

In the implementation: 
-> typedef struct zlist_t resrc_reqst_list; 

I initially thought `resrc_reqst_list` was somehow typedefing `zlist_t` but realized `struct zlist_t` isn't defined elsewhere so it would be a forward declaration to an incomplete type. czmq.h declares `struct _zlist_t` and `zlist_t`, not `struct zlist_t`.

Then in the header file, the same name `resrc_reqst_list` is used (not sure if this is intended or not) in a typedef:
-> typedef struct resrc_reqst_list *resrc_reqst_list_t;

So in the end, if I am not mistaken, there is no association between `struct resrc_reqst_list ` in the header file and `resrc_reqst_list` in the implementation. And then the implementation doesn't seem to associate `resrc_reqst_list_t` with `zlist_t` either. I think the net effect is, we use `resrc_reqst_list_t` as an alias to an incomplete type and the implementation still has to explictly cast `zlist_t *` to `resrc_reqst_list_t * ` or else it won't compile. 

In this sense, `void *` is better because it at least enhances readability. But then I know ppl can feel strongly about the use of `void *` so I wanted to have a 3rd opinion.

grondo: As I understand it, the current version is only a problem because
```C
typedef struct zlist_t resrc_reqst_list;
```
is very confusing because there is no such thing as `struct zlist_t`.
`zlist_t` is a typedef for `struct _zlist_t` as seen [here](https://github.com/zeromq/czmq/blob/master/include/czmq_library.h#L119).
It would seem like the compiler should complain about this usage.
Can you `typedef zlist_t resrc_reqst_list`?

trws: Having it typedefed to the same base as zlist_t, which as you point out @dongahn is `struct _zlist_t`, or to zlist itself if it will let you would be great.  I'm fine with use of `void` and `void *` where they're warranted, but for defining an abstract type like this having an overridable struct name is preferable to me, even if this particular implementation doesn't use it.

dongahn: I am not sure if the compilers will let you do this, either. @lipari, did you try this? I think a good test would to remove the explicit casting in resrc_reqst_new_list, and see if you can compile.

lipari: Making the mod @grondo suggested compiles:
-typedef struct zlist_t resrc_reqst_list;
+typedef zlist_t resrc_reqst_list;

dongahn: Sorry for being annal. But does it compile without the explicit casting in resrc_reqst_new_list? I think I am hallucinating :-), but it seems  

`typedef struct resrc_reqst_list *resrc_reqst_list_t`

in the header file shouldn't make the association between `resrc_reqst_list` and `struct resrc_reqst_list`... I need to go back to my C book...

lipari: Commenting this out of the .h file:
//typedef struct resrc_reqst_list *resrc_reqst_list_t;
does not compile:

trws: You're not hallucinating, it won't.

There is no good way to alias a struct of unknown size and contents to 
another struct name in C, in that there is no way that makes it 100% 
transparent.  As far as I know, the only way to resolve this all the way 
through with correct type checking and no need for casts is to turn it 
into this:

```c
//header
typedef struct resrc_reqst_list resrc_reqst_list_t;

//impl
struct resrc_reqst_list {
	zlist_t *list;
};

resrc_reqst_list_t resrc_reqst_new_list ()
{
	resrc_reqst_list_t ret = {zlist_new()};
	return ret;
}
```

The downside is that now you either have to write `l.list` everywhere, 
or cast that to `zlist_t *`, either works thanks to the structure of the 
type.

dongahn: @trws and @lipari: Thank you folks! Since @trws' proposal agrees with the second suggestion of @lipari, I would say go for it :-)

dongahn:
Hi @lipari: This is not part of this particular PR, but I noticed that `void *` is used to pass an object to destroy in two functions:

> resrc/resrc.h:60 jobid_destory (void *object) 

Where/how jobid_destory used? 

> resrc/resrc.h:101 void resrc_resource_destroy (void *object)

Here, you probably want to use `resrc_t` instead since it is an opaque type already. 

Once you give a chance to look at these, I will merge it unless others have more feedback/inputs. But going forward, here are two potential items for you to consider. 

1. Can we put some const to function return types so that compilers can check const correctness for us? For example, `char *resrc_name (resrc_t resrc)` in `resrc.c` unless the intent is to change the name of resource from the caller, `const char *resrc_name (resrc_t resrc)` could be a bit safer... 

2. We can consider removing JSON from our public interfaces as well as with other flux-core interfaces are evolving. 

Thank you for the work.

garlick:
Maybe off topic here but the names in resrc.c are a little awkward:
```
bool resrc_match_resource (resrc_t resrc, resrc_t sample, bool available)
void resrc_print_resources (resources_t resrcs)
void resrc_stage_resrc (resrc_t resrc, size_t size)
int resrc_allocate_resources (resources_t resrcs_in,
                              resource_list_t resrc_ids_in, int64_t job_id)
etc
```

Any reason not to go with  `resrc_match`, `resrc_destroy`, `resrc_stage`, `resrc_print`, etc?

lipari:
@dongahn: I removed vestigal jobid_destroy().  I left resrc_resource_destroy() as is, because it is passed to zhash_freefn() and must comply with the required signature.  I will work on adding the const's where appropriate in later work.
@garlick; We don't really have any guidance on function naming in RFC7, but if we did, I'd be happy to comply.

dongahn:
@lipari I am merging this one in. My guess is, once you pull in PR 50, travis will like flux-sched again...

PR #50, opened 2015-08-22
update flux-core API usage after changes

garlick:
kvsitr_t  is now a struct not a pointer to struct.

kvs callback typedefs changed names.  Eliminate places in simulator
where this was being cast, as doing so is a bit dangerous - changes
in prototype would be hidden.  Case in point, these functions are
expected to return an integer result and were declared with return
type void.

Use the (deprecated) jsc _obj interfaces


dongahn:
Ok. This should go in right away. @garlick, can you rebase this to the new head?

garlick:
Yup, just did.

dongahn:
Thanks @garlick.

PR #53, opened 2015-08-26
sched: Fix Issue 52

lipari:
Inserted a minimum requirement of one node into the resource request
logic.

### DIFF
--- a/sched/schedsrv.c
+++ b/sched/schedsrv.c
@@ -616,6 +616,7 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job)
      * Require at least one task per node, and
      * Assume (for now) one task per core.
      */
+    job->req->nnodes = (job->req->nnodes ? job->req->nnodes : 1);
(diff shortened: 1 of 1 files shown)


dongahn:
So this was a divided-by-zero problem then :-) The mod makes sense to me. But do we want to print out a warning message about "the resource shapes are not fully specified"?

lipari:
In this case, I see it as a convenience to omit the full spec (nodes in this example).  While other omissions would require correction before being accepted, omitting a node count is excusable. This is consistent with our existing RM products.

dongahn:
Ok. I propose to merge the fix unless others disagree.

dongahn:
Based on the discussion going on in another discussion thread, my impression is there is no harm to merge this in while more changes may come in the future?

lipari:
Agreed.  If nothing else, it eliminates the subject of this issue.  Any further improvements in various node/core scenarios will appear as the design evolves.

dongahn:
Thanks.

Issue #48, opened 2015-08-18
Merging the emulator to the current scheduler framework service and new RDL


dongahn:
Stephen, Suraj and I have just discussed this. To facilitate future scheduling development and research (e.g., dynamic scheduling), we will make it a priority to merge the emulator codes into the latest flux-sched. I am creating this issue to capture the communications needed for this effort.

SteVwonder:
My current plan is to do this in two steps.  Merge the latest emulator code into the master with one PR, then do another PR that makes the scheduler framework service compatible with the emulator and EASY backfilling (through the introduction of extra schedule_jobs functions).

The 1st PR should be pretty straight forward.  It is just taking sometime to squash down 9 months of commits (where each intermediate commit compiles and runs).

SteVwonder:
For the 2nd PR, does it make sense to keep most of the shared emulator code in the `simulator` directory and just have the scheduler framework `#include "simulator.h"`?

Also, I plan to change the `simulator` directory name to `emulator` to correct the misnomer.

dongahn:
The plan makes sense to me. 

Ideally, it would be best if the scheduler framework doesn't know it does its operation for a real environment or an emulated environment. At first glance, it feels the scheduler framework doesn't need to use any functions that belong to the emulator... or if it does the emulator can help faking it. 

Perhaps, we can make it such that one passes "emulation-mode" as an argument to the framework service module, which then uses his flag to decide whether it has to interact with `rexec`. E.g., if this flag is one, we won't launch a program right?

Now, are there emulator specific actions that scheduler framework should do (besides disabling the interaction w/ rexec)?

dongahn:
BTW, if JSC and things need to be adjusted for this (which I suspect we will have to), let's work together to make this happen.

SteVwonder:
I agree on the `emulation-mode` flag, but I believe that the scheduler framework does need to "call out" to shared emulator code.  Particularly to register new events with the emulator manager.

As far as modifications that don't require calling out to shared emulator code:
* As you mentioned, it would also need to communicate with the emulated rexec rather than the real rexec
  * This is easy to fake by unloading rexec and replacing it with the emulated version
* It also needs to queue up resource and job events and handle them when triggered by the emulation manager rather than handling them immediately.
  * Should be easy to do by checking the `emulation-mode` flag and queueing the events in a zlist

[...]

PR #54, opened 2015-08-27
Merge in long-running emulator branches

SteVwonder:
So this isn't 100% merge ready.  I am still waiting to hear back about my test job data so that I can include my sharness tests.  I am also now noticing a few things that I want to go back and clean up (missing copyrights at tops of files, etc.)

But I figured I would put this up now in case anyone wants to take a glance at it.


rdl/RDL/types/Node.lua
```diff
@@ -39,12 +39,25 @@ function Node:initialize (arg)
...
+    if arg.tags ~= nil and arg.tags['max_bw'] ~= nil then
+	   bw_per_socket = arg.tags.max_bw / num_sockets
+    end
+
     local sockid = 0
     for _,c in pairs (arg.sockets) do
         self:add_child (Socket{ id = sockid, cpus = c,
-                                memory = arg.memory_per_socket })
+                                memory = arg.memory_per_socket,
+                                tags = { ["max_bw"] = bw_per_socket, ["alloc_bw"] = 0 }})
```
SteVwonder: I placed this here since it made my life much easier, but I can understand why we wouldn't want it in the master like this.  If we want, I can go back and push this functionality into the emulator code rather than in the lua RDL.

SteVwonder:
I should also mention that this is a step towards closing #48

SteVwonder:
It seems that hostlist isn't included in the dependency build for travis.  I will add it and submit a PR to the flux-core repo.

SteVwonder:
@grondo, it seems that flux cannot find the lua hostlist installed in the travis environment.  What is the best way to fix this?  Modify the "--printenv" flag for travis-dep-builder.sh to include $HOME/local/.... in the LUA_PATH?

grondo:
Yes, the flux-sched .travis.yml needs to be updated to use
```
eval $(./src/test/travis-dep-builder.sh --printenv)
```

I thought I'd done that, sorry! I'm not sure how the other RDL tests pass in travis though, so maybe I'm not quite understanding root cause yet.

grondo:
Nice set of tests BTW!

SteVwonder:
For whatever modifications I make to this PR, should I just add them as additional commits on top, or can I rebase, thus modifying the existing commits and changing the SHA1s.  From what I have gathered from previous PRs, the rebasing of your personal fork/branch is ok with the Flux team.

SteVwonder:
Also, thanks :)  I'm not sure why "default" keeps getting printed out.  But I will track that down later.  I also plan to remove the hacky while loop with a python script that watches for the "sim.complete" event and then spits out the scheduling order.

grondo:
Yes, PR is fine to rebase. Encouraged in fact.

dongahn:
I think adding them as additional commits will do and rebase should be ok too..

rdl/RDL/types/Node.lua
dongahn: Either @grondo or @lipari comment on this?

dongahn:
OK. I'm done with my rather superficial review for now. Only action items from my comments above would be the long-line wrapping and  typo removal. Other than the RDL extension, I think this PR is contained nicely within the emulator logic, so once done, I think this can go in with too much trouble. @SteVwonder: what other changes did you plan to make as part of this PR? Let me know when you are happy with this PR.

SteVwonder:
If @lipari and @grondo are ok with the additions to the RDL, then nothing else to add to this PR other than running clang-format on all my emulator code.  At some point I will get around to renaming everything from simulator->emulator, but I think the 2nd part of the merge (integration w/ sched framework) is more important currently.

dongahn:
Sounds all good. Thanks.

garlick:
For new code that is going in, could we make sure it's not using the deprecated functions from flux-core?  A new PR should not contribute to that problem IMHO.

SteVwonder:
@garlick, great point.  Shouldn't be too bad of a fix.  Working on that now.

rdl/RDL/types/Node.lua
grondo: Ah, sorry I meant to comment here but I needed to review the usage more to understand what's going on here.

I wonder if creating a new type derived from Node or Socket would work here, but in general we should probably think about how to easily add custom attributes/tags to resources.

lipari:
If @grondo approves the changes to the RDL, so do I.

SteVwonder:
@dongahn, this is now ready to be merged.

dongahn:
OK. Thanks!

PR #57, opened 2015-09-01
Eliminate dependence on the flux internal library

lipari:
Address Issue 56.  A straightforward approach of copying over from
flux-core the non-flux utilities that flux-sched products and tests
depend on.  At this point, there is no change in the use of those
utilities.


garlick:
Before pressing ahead with this approach, what about discussing the one @grondo came up with in flux-framework/flux-core#336?  Both approaches definitely have their upsides and downsides.

lipari:
@garlick, I'd like to weigh in as strongly neutral on this.  The request to break the reliance on flux-internal was proposed at our team meeting yesterday and I volunteered to make it so.  At the time, we did not consider the secret sources solution as an alternative.

garlick:
Oh sorry, I wasn't there so apologies if you guys came to a consensus on this.

grondo:
Actually we didn't really have any good ideas on *how* to actually accomplish this. I can't actually remember why we felt the need to get this done right away -- @lipari do you remember?

lipari:
As I recall, we saw this as a precursor to converting the flux-sched to use Autotools.  We could then build flux-sched based off an installed flux core library.

dongahn:
If @liparti wants to look at auto tools support I think this is necessary with a caveat that we will move over to the magic common utility library later when it comes.

Auto tools support is something that we will need but I am not sure if this is the right time because sched is still in flux.

Perhaps we should discuss benefits to merge this pr and timing for autotools. Similarly, what's the harm just going with it now even without autotools.

@liparti do you plan to look at auto tools support right away?

In any case I can see the argument either way but I am inclined to merge this PR just because we have it.

garlick:
I'm not very keen on the code duplication.

I understand the desire to autoconfiscate.  If now is the time what about implementing something like
```
configure --with-flux-core-srcdir=path --with-flux-core-builddir=path
```
that would work with the current "check out both trees" method of building and a future `flux-core-sources` RPM package as discussed in  flux-framework/flux-core#336?

If now is not the time, then I would argue against merging.  If we end up doing the above and sources have diverged in the mean time, pain and suffering will result.

grondo:
I agree with @garlick above, especially since there could be many updates to the duplicated code. (Though I feel bad for the work @lipari put into this already)

Another alternative would be to package up utility code used between projects into an inter-project library, but that would be a lot of work, requires thought about APIs and symbols, and generally I'm not even sure I like it any more than compiling against sources. 

If we want to move forward with RPM sources package, let me know and I can make that a priority.

lipari:
Don't worry about the effort to create this PR.  I understand the objections.  This was not the right step to improve the flux-sched build system.  @grondo, the RPM sources solution sounds good, but I'm not sure how much priority to place on the job.

PR #59, opened 2015-09-03
support for reliable jsc

dongahn:
Note that this PR should not be merged before PR #386 of flux-core is merged. This PR contains initial support for reliable jsc including a sharness test case. 

To facilitate testing, a new utility called `flux-waitjob` is added so that a shareness test can wait until all of the jobs are complete, which are submitted through `flux-submit`. 

In addition, to get notified of the “submitted” event, flux-submit now emits wreck.state.submitted as well. 

Finally, this PR also contains other minor fixes and a workaround (i.g., removing the use of posix.signal from within flux-submit).

### DIFF
--- a/sched/Makefile
+++ b/sched/Makefile
@@ -5,7 +5,7 @@ LIBS = $(FLUX_LIBS) $(RESRC_LIBS)
 
 CONF=../conf/hype.lua
 
-BUILD = schedsrv.so schedplugin1.so
+BUILD = schedsrv.so schedplugin1.so flux-waitjob
 
 all: $(BUILD)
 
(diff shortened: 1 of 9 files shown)


SteVwonder:
This was perfectly timed @dongahn.  On top of helping with the sched framework merge, the flux-waitjob will make my tests much cleaner.
Just to confirm I understand how this works, calling `jsc_notify_status` sets up a callback that will be run for every state change of every job.  Additionally, I assume that those state changes will arrive in order ("complete" will not be received before "running", etc).

dongahn:
Thank you for the feedback @SteVwonder. Yes, that's correct. If that's not the case, I need to know about it :-)

Issue #63, opened 2015-09-08
not ok 5 - jobs scheduled in correct order


garlick:
I'm finding that `test 5 - jobs scheduled in correct order` is consistently failing for me in the following tests

**t2000-fcfs**

Expected:  `1 2 3 4 5 6 7 8 9 10 11 12`
Actual: `1 2 4 10 11 12`

**t2001-fcfs-aware**

Expected: `1 2 3 4 5 6 7 8 9 10 11 12`
Actual: `1 2 4 10 11 12`

**t2002-easy**

Expected: `1 2 3 6 7 8 9 10 11 4 5 12`
Actual: `1 2 4 5 6 7 9 11 10 8 12 3`

This occured on current master of flux-sched:  f912cb256b1efb191c0f196fdba4a4494e5760a0 and flux-framework/flux-core@ad0f54b452791231eb9952e6f74298e21cc2d796.
I also tried one PR back in flux-framework/flux-core@8d00634cbb60a3f49d58f4595860c43ac6299248 

This is on RHEL 6/TOSS 2 based desktop.

dongahn:
@garlick: I reported this to @SteVwonder last week and he plans to address these as part of his next PR.

SteVwonder:
If you want, we can remove those tests from the HEAD of flux-sched while we are waiting on my next PR.  I am not sure why they are failing all of a sudden given that @dongahn's changes shouldn't have affected the simulator.  Either way, I will make sure they work again in the "new and improved" scheduler framework service.

dongahn:
@SteVwonder: I think @garlick is saying these tests are failing on a desktop, not on Travis. This aggrees with my experience: they fail on cab while passing on Travis.

garlick:
I am fine either way - just wanted to make sure I reported the failure in case it was new.

garlick:
This just came up in our meeting and I _thought_ the we decided that the failing tests had been fixed.  However I just pulled master and noted that the tests are still disabled from commit 22250a2ddc3b58cb3f1aa75077e4f2d56c9368b7

When I uncommented the fcfs test, it failed on my RHEL6 based desktop.

Maybe the discussion was referring to a branch that hasn't been committed yet?

dongahn:
Ah, I misspoke then. I mistakenly thought that getting 

>
# passed all 4 test(s)

are all that we need to get the pass and didn't realized this test case was disabled that way. 

[...]

PR #64, opened 2015-09-09
Resrc deserialize

lipari:
As promised, the deserialization code.  The sched build fails however with the "jobs scheduled in correct order" problem.

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -394,7 +394,9 @@ int resrc_to_json (JSON o, resrc_t *resrc)
 {
     int rc = -1;
     if (resrc) {
-        Jadd_str (o, resrc_name (resrc), resrc_type (resrc));
+        Jadd_str (o, "type", resrc_type (resrc));
+        Jadd_str (o, "name", resrc_name (resrc));
+        Jadd_int64 (o, "size", resrc_size (resrc));
         rc = 0;
     }
     return rc;
@@ -819,6 +821,7 @@ resrc_t *resrc_new_from_json (JSON o)
 {
     JSON jpropso = NULL; /* json properties object */
     JSON jtagso = NULL;  /* json tags object */
+    const char *name = NULL;
     const char *type = NULL;
     int64_t ssize;
     json_object_iter iter;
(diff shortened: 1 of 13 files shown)


SteVwonder:
Given that the issue is with the tests/simulator not your code, I think we should just comment out the simulator tests from t/Makefile.  I suggested this in #63, but it didn't seem to be affecting Travis at the time, so we held off.  Something happened after the JSC merge, but I don't know what.  Since the simulator is going to change with my next PR, it probably isn't worth the effort to fix it right now (I will make sure they work with my changes).

SteVwonder:
Also, thanks for getting this done so quickly.  This helps me a lot :)

lipari:
Two things:  while the contents of the serialized resource data saved to the job record in the kvs was increased somewhat, there are more fields which could be included.  I held off doing so to minimize the noise, but if you find fields are necessary, they can easily be added to both the serialization and de-serialization operations.
The test that is failing is t1000-jsc.t.  I'm looking into that right now and when I have a fix, I'll add it to this branch/PR.

lipari:
I have pushed an update to this PR with the fix and the job ordering tests commented out.  The build succeeded.

dongahn:
@SteVwonder: if you come back online sooner than me, could you please review this in particular since we will be the initial user of this functionality?

SteVwonder:
I looked over the code, and it looks good.  It seems to have everything I need for IO-Aware scheduling.

Just to give it a "test drive", I rebased my current work on top of @lipari's branch, and the serialization method seems to work just fine.  The updated format ends up in the KVS, etc.

Thanks again Don :smiley:

dongahn:
@lipari: this looks great and should be merged ASAP. Sorry for the delay. Do you know why you got this Travis error? Could you rebase to see if the problem goes away?

lipari:
I did actually rebase the branch.  Of the two travis builds of mine, the clang succeeded but the gcc hung when it went to execute this test:
ok 1 - jsc: expected job-event sequence for single-job scheduling.
I have yet to delve into the cause.  The test runs fine out of my home directory, so it is a mystery what changes under travis.

lipari:
Issued a rebuild request and it passed!  Still need to discover the source of what I expect is a timing problem, but at least this should allow us to proceed with the next round of PR's.

dongahn:
@lipari, we will see if we can find a reproducer for the jsc. But for now, I am merging this in.

Issue #62, opened 2015-09-07
Need temporal allocations in resrc


SteVwonder:
@lipari, already has the idea sketched out (which he shared with @surajpkn and I).
My summary from our short meeting is:
*  Expand the `resrc_allocate` and `resrc_release` functions to include a start and end time
  * Either (or both) of these times could be -1, meaning infinity
* Include a function that allows you to determine if a resource is allocated at a particular time
  * This would perform an intersection check for every allocation

surajpkn:
@lipari please feel free to delegate any part of the implementation to me. Since I cannot really proceed to implement the dynamic scheduling without the above, I can work on parts myself too (if it wouldn't really require too much of background knowledge on what you already did or is easy to catch up on).

SteVwonder:
I would also offer my help, but I will be unavailable next week.  So if this still needs work after next week, I can jump in and help too.

lipari:
First off, this issue seems to reflect the needs first described in [Issue 25](https://github.com/flux-framework/flux-sched/issues/25).

Adding time to the resource model has a potential of becoming complicated unless we can think of a means to restrict the requirements or simplify the design.

Starting with the basics as laid out in [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc), any resource allocated to a job needs to be annotated with that job's id.  A more significant complication is the hierarchical nature of the composite resource pool also from RFC 4.  That RFC calls for annotating the resource with not only a job it is allocated to, but the job id's of all ancestors.

This was the mechanism we devised to allow a scheduler to determine whether any given resource could be allocated exclusively to a job.  E.g., if a "rack" resource was mostly idle, but there was one job running on one core on one node of the rack, that job id would be annotated at the rack level, and all intervening resources down to the core itself.  A scheduler in this example could tell that the rack could not be allocated to a job that requested exclusive access to a rack.

The "allocs" and "reservtns" members of the resrc structure were intended to implement this design.  These two members were intended to capture all the job's that were either allocated to the resource or had a reservation on the resource.

The first enhancement to this initial implementation was the size component.  In theory, every resource has a size, and that size is nominally one.  A memory resource for example could have a size of 32 for every GB of memory we can individually allocate to a job.  Hence, the model should eventually (it does not now) support the ability to assign 8 "size" units to job 5 while simultaneously allocating 16 units to job 17.

[...]

PR #66, opened 2015-09-15
Time component to resrc

surajpkn:
More than a PR, this is ping to get a feedback from everyone on the extending the resrc with time component (mainly for @lipari to see if the resrc extension is close to what he envisioned and @SteVwonder to check if this does seem to suit his needs). If its good, I will continue working on it after feedback and bring it to a mergable state. Otherwise, we can close the PR and I can rework differently.

Here is a summary of the changes and logic:
1. resrc is extended with another hash table "twindow". When allocated/reserved for future with jobs, this will store job_id as key and a json string as value - { "starttime" :  epoch_start_time, "endtime" : epoch_end_time }. Note that this is valid only for nodes and cores at this point. 
2. When root loads the RDL from configuration, all the resources have a default entry in twindow for job_id 0 : { "starttime" : epoch_sched_start_time, "endtime" : UNIT64_MAX }
3. flux_res_t of a job now also has a walltime component to it along with nodes, cores and corespernode. By default, this is set to 3600 seconds during fill_resource_req (sidenote: for walltime support, flux-submit and the JSC needs extension).
4. For resrc created for match_resource purpose will have twindow with key 0 and { "walltime" : walltime }. This is used by resrc_walltime_match called from within resrc_match_resource to check if that resource does is free for the required walltime.
5. When the resources are allocated for the job, the resource is updated with the twindow information. This should ideally be done in resrc_allocate_resource (by providing also a walltime parameter to it). But I didn't want to change the existing functions before I have go, and so created similar function resrc_walltime_update() which does the job. 

Let me know your feedback, so that I can work on this further. Once a consensus is reached, I can expand/modify this and also provide utility functions, for example, that will look up if a resource is free at a particular time, or the next gap in reservations etc (depending upon the need).

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -507,7 +507,7 @@ bool resrc_walltime_match (resrc_t *resrc, resrc_t *sample)
 
     /* fix job start and end time from now */
     jstarttime = time_now;
-    jendtime = time_now + jwalltime;
+    jendtime = jstarttime + jwalltime;
     tstarttime = TIME_MAX;
 
     /* Iterate over resrc's twindow and find if it sample fits from now */
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -489,6 +489,7 @@ bool resrc_walltime_match (resrc_t *resrc, resrc_t *sample)
     int64_t jwalltime;
     int64_t tstarttime;
     int64_t rstarttime;
+    int64_t lendtime;
 
     char *json_str_walltime = NULL;
     char *json_str_window = NULL;
@@ -509,6 +510,15 @@ bool resrc_walltime_match (resrc_t *resrc, resrc_t *sample)
     jstarttime = time_now;
     jendtime = jstarttime + jwalltime;
     tstarttime = TIME_MAX;
+
+    /* If jendtime is greater than the lifetime of the resource, then falst */
+    json_str_window = zhash_lookup (resrc->twindow, "0");
+    JSON lt = Jfromstr (json_str_window);
+    Jget_int64 (lt, "endtime", &lendtime);
+    Jput (lt);
+    if (jendtime > (lendtime - 10)) {
+        return false;
+    }
 
     /* Iterate over resrc's twindow and find if it sample fits from now */
     json_str_window = zhash_first (resrc->twindow);
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -61,13 +61,13 @@ struct resrc {
     zhash_t *tags;
     zhash_t *allocs;
     zhash_t *reservtns;
+    zhash_t *twindow;
 };
 
 
 /***************************************************************************
  *  API
  ***************************************************************************/
-
 char *resrc_type (resrc_t *resrc)
 {
     if (resrc)
@@ -190,6 +190,7 @@ resrc_t *resrc_new_resource (const char *type, const char *name, int64_t id,
         resrc->reservtns = zhash_new ();
         resrc->properties = zhash_new ();
         resrc->tags = zhash_new ();
+        resrc->twindow = zhash_new ();
     }
 
     return resrc;
@@ -211,6 +212,7 @@ resrc_t *resrc_copy_resource (resrc_t *resrc)
         new_resrc->reservtns = zhash_dup (resrc->reservtns);
         new_resrc->properties = zhash_dup (resrc->properties);
         new_resrc->tags = zhash_dup (resrc->tags);
+        new_resrc->twindow = zhash_dup (resrc->twindow);
     }
 
     return new_resrc;
@@ -235,6 +237,7 @@ void resrc_resource_destroy (void *object)
         zhash_destroy (&resrc->reservtns);
         zhash_destroy (&resrc->properties);
         zhash_destroy (&resrc->tags);
+        zhash_destroy (&resrc->twindow);
         free (resrc);
     }
 }
@@ -312,6 +315,15 @@ static resrc_t *resrc_add_resource (zhash_t *resrcs, resrc_t *parent,
                 zhash_freefn (resrc->tags, iter.key, free);
                 Jput (jtago);
             }
+        }
+
+        /* add twindow */
+        if ((!strncmp (type, "node", 5)) || (!strncmp (type, "core", 5))) {
+            JSON j = Jnew ();
+            Jadd_int64 (j, "start", epochtime ());
+            Jadd_int64 (j, "end", TIME_MAX); 
+            zhash_insert (resrc->twindow, "0", (void *)Jtostr (j));
+            Jput (j);
         }
 
         while ((c = rdl_resource_next_child (r))) {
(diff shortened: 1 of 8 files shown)


lipari:
Looks great!  And here are a couple thoughts...

I agree that adding a walltime argument to resrc_tree_list_allocate() and resrc_tree_allocate() will be the way to go long term.

Also, as an optimization, you might make the twindow member of the resrc struct a list ordered by start time.  This way, you can do away with searching the whole list each time to find the soonest start time.

But, yes, you have the right idea of adding time to the resource struct.  That will provide the foundation for rendering a time-ordered series of resource trees that represent available resources at every job start or terminate event.

surajpkn:
Great. I am glad to know that this is in the right direction. I'll make the said changes.

dongahn:
@surajpkn: thank you for doing this work. This looks like a great start to me as well. 

This may be a question to @lipari and @SteVwonder, but what are the common operations that will be performed on this time interval support? I am asking this question because I don’t have hands-on experiences on the scheduling algorithms that use this support and so trying to wrap my head around what would be an efficient data structure for these operations. What are the operations that the start-time-ordered list will optimize? From @SteVwonder’s previous work, I know you need to be able to sort your running jobs list according to their complete time to quickly find the right reservation time, but this optimization would require different data structure support than the start-time sort within resrc. 

For your side note for #3, adding wall-time clock support to JSC and `flux-submit` should be  straightforward. For JSC, one can add JSC_RDESC_WTIME attribute and have it store the wall clock time along with the number of nodes and cores in the KVS. But then, we will also need wreck to enfornce this limit. 

Also, one can extend flux-submit with `-t <time>`  or`--time=<time>` as well. 

Finally, the Travis testing is failing with 

```
resrc.c:35:40: fatal error: src/common/liblsd/hostlist.h: No such file or directory
```

This issue is mentioned in the #400 of flux-core. What was the resolution for this issue?

surajpkn:
@dongahn if you have a list of reservations sorted according to the start time, it will be easy to match resources for backfill and also slacktime scheduling. Because, the question that will asked at that point is, "can I use this resource to start a job which has this walltime?". With the sorted list, this is faster to answer rather than an unsorted hash table using which one has to compute the earliest start time reservation that the resource has (which is what the code does right now). 

Through this, if you want to find the earliest time when a resource will become available, it depends on the resource type.
1. A core is never shared between jobs. So simply finding the completion time of a reservation at the end of the list would suffice. 
2. A node could be shared between jobs - some cores for 1 job and the others for a second one (although I am not sure if FLUX already supports this. does it?). In that case, one will have to look through every reservation and find the maximum. 

From what you mention about @SteVwonder's requirements, it seems he requires the "running jobs" to be sorted according to completion time, to find the right reservation time. If his algorithm really depends on the "running jobs", then we should think about having the r_queue in schedsrv.c sorted according to completion time. Or if it indeed will use the resrc information, the case 1 and 2 that I mentioned above applies. I am pretty sure that his backfilling algorithm can take advantage of the sorted list. But I don't know about the others (like IO aware scheduling, etc.). 

But yes, before I change the data structure, it would be good to have @SteVwonder's comments.

lipari:
One miscellaneous thought to throw on the discussion...
The last two bullets of [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc) explains that a resource that is allocated exclusively to a job has all of its children also allocated.  So, that means that we can eventually economize and not need to visit children of an exclusively allocated resource, to assign a job or see whether a job is assigned.

dongahn:
@surajpkn: please rebase to the master and see if the Travis problem has been fixed.

surajpkn:
The build just seems to have stopped with some strange error. I am not able to trigger it again. Can someone with write access do it? (i guess you need to have write access to the master to do be able to retrigger it directly?)

dongahn:
I suspect the build filled because of the same jsc test issue @lipari mentioned in PR #64. I should merge that first and see if it has a fix for this build issue as well.

surajpkn:
@dongahn Do you want me to extend JSC for walltime and do a PR to flux-core?

dongahn:
Yes, if you can do that, that be great! This should be a straightforward change. 

At some point, we should have a better way to manage the name space of our job attributes (currently Macros), but I don't think we are there yet as our job schema isn't defined.

SteVwonder:
I can look at this more in depth this weekend, but at a quick glance, this is great!

The high-level operation that I expect to use the most will be "give me the state of the RDL at time X".  This will require doing an intersection between time X and the allocations in the hashtable/list.  If the number of allocations in the list is small, its probably fastest to just loop through them.  If there are a lot of allocations, @trws, mentioned using interval trees (https://en.wikipedia.org/wiki/Interval_tree).

Any particular reason why this is limited to nodes & cores?  Also, to expand to other resources (memory, IO) we will need to add "count" to the allocation entry.

surajpkn:
After some digging, I found interval trees (insert item + find item) to be faster than a list (append item + sort list + find item), even for a small list with 6 items. I think it is better to use interval trees since it can avoid a lot of sorting. 
I found this http://ideone.com/fhsfFH - a very simple interval tree code. This can be easily augmented for our purposes. If everyone thinks its a good idea, I can go ahead with this one. 

@SteVwonder I actually assumed that `flux_res_t` will be expanded with more resources such as memory and IO. So the walltime is also applicable to them. Not true? or are you talking about separate walltime for each resource?

resrc/resrc.c
```diff
@@ -314,6 +317,15 @@ static resrc_t *resrc_add_resource (zhash_t *resrcs, resrc_t *parent,
             }
         }
 
+        /* add twindow */
+        if ((!strncmp (type, "node", 5)) || (!strncmp (type, "core", 5))) {
```
SteVwonder: @surajpkn, i was referring to this piece of code.  Why limit the twindow to only nodes and cores?  Why not also include the other resources?

SteVwonder:
I played around with [this implementation](https://github.com/ekg/intervaltree) of interval trees and came to the same performance conclusions.  I think it is a worthwhile improvement if you want to do it, but I would say that it is necessary at this point.

trws:
Just as a thought, if you're doing an interval tree, it might be worth building on a red-black-tree or something that will handle self-balancing for you and all of that.  There's a reasonable example of how to augment a red-black into an interval tree [here](http://www.bowdoin.edu/~ltoma/teaching/cs231/fall07/Lectures/augtrees.pdf), or there's the trie-based option we discussed the other day.  I'm not sure if we have a generic balanced tree in utils yet, but it is almost certain to come up.  Unless I find a good graph/network framework to use I'll be pulling one in shortly myself.

surajpkn:
@SteVwonder Because I was not sure if we wanted to have walltimes associated with resources such as switches. Maybe it makes sense to remove that line and eliminate only walltime checks for those resources that don't require it.

lipari:
@surajpkn, I guess I would advise you to focus on your research and use the existing solution if is suffices.  As I just summarized in [Issue 62](https://github.com/flux-framework/flux-sched/issues/62), the entire temporal support needs to be reexamined and we'll need to support a timeline for multiple jobs per resource and per resource size allocation.

surajpkn:
Yes, I reckon that after reading #62. I would go with the existing hash table solution for now.

surajpkn:
@lipari Could we push these changes into the master and later modify them to handle the cases you mentioned or you want me to work on my own branch? It could help Stephen too for the time.

lipari:
I'm fine with that if @dongahn is.

surajpkn:
I am with @dongahn and he is OK with it.

PR #71, opened 2015-09-23
Eliminate flat resources

lipari:
The original design included a hash table of all the resources.  The thinking was that this table would provide an alternate means to search and update resources that could perform better than a tree search.  This facility has gone unused and now serves no purpose other than to unnecessarily complicate the code

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -213,7 +213,7 @@ resrc_t *resrc_new_from_json (JSON o, resrc_t *parent, bool physical)
     Jget_str (o, "name", &jname);
     if (!(Jget_int64 (o, "id", &id)))
         id = 0;
-    name = xasprintf ("%s%ld", jname, id);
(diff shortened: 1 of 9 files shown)


surajpkn:
This eases many things for me. I am looking forward to this being merged.

dongahn:
Otherwise, the change looks good to me. Since this will help @surajpkn, I would like to merge this in ASAP. Because this deprecates an existing capability, one thing I want to make sure is @SteVwonder's work  doesn't depend on the flat resrc search, either. Once @SteVwonder gives his ack,  I will merge.

SteVwonder:
Changes look good to me.  My work does not depend on it either.

garlick:
29b3b42 touches some uses of asprintf with unchecked return code.

lipari:
Thanks, @garlick.  Updated PR just pushed.

dongahn:
Thanks Don!

Issue #74, opened 2015-09-25
Deserialized id and name not consistent.


surajpkn:
There seems to two not-so-important issues with serialize/deserialize RDL. 

1. Serialization does not include any of the job's ids. - Easily solved by adding one line to include an id to `resrc_to_json()`. Where this is important is when resources are passed down to child instances, they need to have consistent ids. 

2. The resource name is wrong after deserialization. 
Example:  resrc->name = hype201, id=201. 
Serialization (including id): "name" : hype201, "id" : 201
Deserialzaiton: resrc->name = hype201201, id = 201

The id is concatenated with name. Thats of course because the deserialization code uses the same function that generates resources from the conf file. 

This is not important, but I would just like to point it out. `resrc_new_from_json()` makes that concatenation. It may be a good idea to not do such a concatenation inside `resrc_new_from_json()` so that it stays generic. For the purpose of loading RDL from file, the json object passed to `resrc_new_from_json()` can be modified to already consist of the right name with the concatenation.

lipari:
I need to understand item 1 better.  Currently the serialized resources are stored in the kvs under the job's id.  It seems redundant to add the job ids to the items serialized.  If you are needing to attach the allocated job ids to the serialized resources, can you just adding them to the json object:
```
        if (zhash_size (resrc->allocs)) {
            JSON jallocs = Jnew_ar ();
            size_t *size_ptr = zhash_first (resrc->allocs);
            while (size_ptr) {
                JSON jobo = Jnew ();
                Jadd_int64 (jobo, "job", *size_ptr);
                json_object_array_add (jallocs, jobo);
                size_ptr = zhash_next (resrc->allocs);
            }
            json_object_object_add (o, "allocs", jallocs);
        }
```
I will fix item 2.

surajpkn:
@lipari I am so sorry I messed up item 1. What I actually meant was node's ids and not job's ids.

lipari:
The node id is listed under the "name" key:
```
lwj.1.rdl = [ { "type": "node", "name": "hype201", "size": 1, "children": [ { "type": "socket" ...
```
Where else should the node id be?

surajpkn:
Sure. But I just assumed concatenating id to non-number name was only to create the hostnames right in these configs we use. If thats the very design, it suffices of course. 
Nevertheless, I include the id key in serialization to just ease its extraction.

lipari:
[...]

PR #75, opened 2015-09-28
resrc: modify the serialized resources

lipari:
Change the name to only include the name without the id appended.  Add
id, path, and uuid to the list of serialized (and deserialized)
resource members.
Addresses https://github.com/flux-framework/flux-sched/issues/74

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -37,6 +37,7 @@
 
 struct resrc {
     char *type;
+    char *path;
     char *name;
     int64_t id;
     uuid_t uuid;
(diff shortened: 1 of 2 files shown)


surajpkn:
Thank you @lipari. Is there already a lookup function that uses path+name to search in root_resrc?

lipari:
There used to be.  Do you want me to put it back in?
Take a look at https://github.com/flux-framework/flux-sched/blob/78dc2f4fb8148f19c36a2e73ccdf261607df9917/resrc/resrc.c.  All the resources used to be entered into a zhash_t hash table which was returned by resrc_generate_resources().  The key for this hash table was the full path name and the value was a pointer to the resrc_t.

surajpkn:
Oh I see. I will stick to using the ids and hash table of ids -> resrc pointers for my purposes. Since I already coded a couple of functions this way, its convenient for me. As discussed in the meeting, we can defer the decision to bring it back or not based on all other needs as well. Thanks a lot again!

dongahn:
Looks great. Thanks!

Issue #76, opened 2015-09-29
Travis fails with hostlist


dongahn:
I'm getting Travis CI build failures. I initially thought this was because of my changes in the waitjob_fix branch of my fork. But then when I sync'ed the master of my fork with the upstream sched master, I notice I'm still getting the same failure. Travis CI  didn't complained about @lipari's latest PR so I'm kind of baffled. 

```
ok 3 - resoure file readable
loading RDL: /home/travis/build/dongahn/flux-sched/rdl/RDL.lua:30: module 'hostlist' not found:
	no field package.preload['hostlist']
	no file '/usr/local/share/lua/5.1//hostlist.lua'
	no file '/usr/local/share/lua/5.1//hostlist/init.lua'
	no file '/home/travis/.luarocks/share/lua/5.1//hostlist.lua'
	no file '/home/travis/.luarocks/share/lua/5.1//hostlist/init.lua'
	no file '/usr/share/lua/5.1//hostlist.lua'
	no file '/usr/share/lua/5.1//hostlist/init.lua'
	no file './hostlist.lua'
	no file '/usr/local/share/lua/5.1/hostlist.lua'
	no file '/usr/local/share/lua/5.1/hostlist/init.lua'
	no file '/usr/local/lib/lua/5.1/hostlist.lua'
	no file '/usr/local/lib/lua/5.1/hostlist/init.lua'
	no file '/usr/share/lua/5.1/hostlist.lua'
	no file '/usr/share/lua/5.1/hostlist/init.lua'
	no file '/home/travis/build/dongahn/flux-sched/rdl/hostlist.lua'
	no file '/home/travis/flux-core/src/bindings/lua/hostlist.lua'
	no file './hostlist.lua'
	no file '/usr/local/share/lua/5.1/hostlist.lua'
	no file '/usr/local/share/lua/5.1/hostlist/init.lua'
	no file '/usr/local/lib/lua/5.1/hostlist.lua'
	no file '/usr/local/lib/lua/5.1/hostlist/init.lua'
	no file '/usr/share/lua/5.1/hostlist.lua'
	no file '/usr/share/lua/5.1/hostlist/init.lua'
	no file '/usr/local/lib/lua/5.1//hostlist.so'
	no file '/home/travis/.luarocks/lib/lua/5.1//hostlist.so'
	no file './hostlist.so'
	no file '/usr/local/lib/lua/5.1/hostlist.so'
[...]

PR #77, opened 2015-09-30
jsc/etc: beefing up the jsc tests and adjusting for a recent lua binding churn

dongahn:
This PR contains fixes for two distinct issues: 1) intermittent jsc test failures and 2) Travis CI build failures because `hostlist` and `posix` have moved to the flux namespace (e.g., `src/bindings/lua/flux/hostlist.so`). 

1) is captured in PR #64 along with other issues. I suspect the issue is due to some less precise synchronization mechanisms used between `flux-waitjob` and subsequent `flux-submit`s within the test script (`t/t1000-jsc.t`). To make synchronization air-tight, better synchronization support has been added within `flux-waitjob` (along with its own test case --- `t/t0002-waitjob.t`) and `t/t1000-jsc.t` has also been improved.  

2) is captured in #76. To debug this better in the future, I've made a very minor modification to  `.travis.yml` such that I can see the environment variables that are exported by the `flux-core`'s deb builder script. And the other fixes are trivial changes in a sharness test script, Makefiles and lua files: e.g., `require 'hostlist'` to `require 'flux.hostlist'`. BTW, the reason that Travis CI has not detected these failures is that its test have been run with a cache. So occasional deleting of Travis caches should be encouraged to detection a condition like this.

I've ran this PR on Travis CI for 10 times and all succeeded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
   - eval `luarocks path`
   - wget https://raw.githubusercontent.com/flux-framework/flux-core/master/src/test/travis-dep-builder.sh
   - bash ./travis-dep-builder.sh --cachedir=$HOME/local/.cache
-  - eval $(./travis-dep-builder.sh --printenv)
+  - bash ./travis-dep-builder.sh --printenv
   - pip install --user cffi coverage
 
 script:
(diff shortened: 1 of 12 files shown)


dongahn:
I just merged #75 so let me rebase.

PR #79, opened 2015-10-06
Issue 61

lipari:
Add ability to read resources generated by hwloc

### DIFF
--- a/resrc/Makefile
+++ b/resrc/Makefile
@@ -1,8 +1,8 @@
 include ../Makefile.inc
 
 SUBDIRS = test
-CFLAGS = $(COMMON_CFLAGS) $(FLUX_CFLAGS) $(RDL_CFLAGS)
-LIBS = $(FLUX_LIBS) $(RDL_LIBS) -luuid
+CFLAGS = $(COMMON_CFLAGS) $(FLUX_CFLAGS) $(RDL_CFLAGS) -I/usr/include/libxml2
+LIBS = $(FLUX_LIBS) $(RDL_LIBS) -luuid -lxml2
 
 BUILD = libresrc.so.0
 
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -28,6 +28,8 @@
 #include <string.h>
 #include <assert.h>
 #include <czmq.h>
+#include <libxml/parser.h>
+#include <libxml/tree.h>
 
 #include "rdl.h"
 #include "resrc.h"
@@ -572,7 +574,7 @@ ret:
     return resrc;
 }
 
-static resrc_t *resrc_add_resource (resrc_t *parent, struct resource *r)
+static resrc_t *resrc_add_rdl_resource (resrc_t *parent, struct resource *r)
 {
     JSON o = NULL;
     resrc_t *resrc = NULL;
@@ -582,7 +584,7 @@ static resrc_t *resrc_add_resource (resrc_t *parent, struct resource *r)
     resrc = resrc_new_from_json (o, parent, true);
 
     while ((c = rdl_resource_next_child (r))) {
-        (void) resrc_add_resource (resrc, c);
+        (void) resrc_add_rdl_resource (resrc, c);
         rdl_resource_destroy (c);
     }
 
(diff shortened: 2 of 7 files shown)


garlick:
> This test can only read the resources of the node invoking the test.  So, the resource request construction had to be simplified to eliminate multiple nodes and properties that were only present in the sample RDL config file.

Would it make sense to capture/construct hwloc data for testing, as was done with the RDL config file, and use it rather than relying on hwloc of the test system?  In flux-framework/flux-core#272 there was talk about providing dummy hwloc data for test.  I'm not sure what was decided there?

Rather than parsing XML directly, is there some way to bulk load it into a hwloc topology and then parse it using the hwloc API?  (Would that be safer than hardwiring the XML layout?)

grondo:
> Rather than parsing XML directly, is there some way to bulk load it into a hwloc topology and then parse it using the hwloc API? (Would that be safer than hardwiring the XML layout?)

That is an excellent idea, and @trws already has shown how to do that easily with `flux lstopo` (though it is python, probably not difficult to adapt to a shell script, it is a small program.)

This could be powerful for testing different "real" topologies and the hwloc reader support in sched by supplying xml from lots of systems.

trws:
Sounds good to me.  The code that actually does the bulk loads is in resource_hwloc already. I don't have it hooked up with an option to override the system detection, but it would be maybe a half-hour thing to add if it's desirable.  it's also possible to construct logical topologies from a number of real topologies, as @grondo mentioned that's what the lstopo message/command does.

lipari:
These are all good ideas and I intend to add an enhancement that allows one to create a fake hwloc topology object from a synthetic config file (if @trws doesn't beat me to it).

As far as parsing the resources back into a resrc_tree, the walking of the XML was one of three options and the one I went with in the interest of simplicity.  A second option would be to use the hwloc api to walk a hwloc topology object similar to what is currently done in resource-hwloc/resource.c.  This could be easier/cleaner or not, I didn't try it.  The third option would have been to walk the resource.hwloc.by_host entries in the kvs and I really wanted a solution that could be tested without requiring a flux instance.

I guess I'd recommend going with the XML solution for now, if for no other reason than we'll have the work available for posterity.  If the hwloc API parsing option turns out to be the better choice, I can always switch over to it.

dongahn:
I second @garlick's idea of capturing hwloc-based resource data for later testing although it shouldn't be a part of this PR. 

This can be an extremely powerful technique for testing scheduling scalability and effectiveness. Say, we capture this data from the entire Catalyst system during DAT and use our emulation environment for testing as well as exploring the effectiveness of various advanced scheduling that we are and need to be working on. In particular, scheduling complexity would be a function of the number of resource items and jobs so the real data with lots of resource details will also help us study one important unexplored dimension. I.e., current hierarchical scheduling doesn't take into account the resource representation granularity in our hierarchical scheme, but we probably need to look at it at some point and this data can help reveal the needs. Of course, the hand-configured RDL does provide this purpose (as we have been using) but real topologies makes a lot of sense to me in terms of how it augments our current RDL approach.

dongahn:
The code assumes the resource data will be coming from hwloc when `rdl-conf`is not present, which is simple and should work well now. But if one wants to extend this, this can be a problem. 

Since we are talking about a multi-reader option, would it make sense to add a new option, say `rdl-conf-mode=[rdl|resrc|hwloc]` instead? In case a mode needs an additional path as an argument, this can be `rdl-conf-mode=resrc=path`?

dongahn:
With @SteVwonder's patch merged, we now have 5 different options (`rdl-conf`, `rdl-resource`, `in-sim`, `plugin` and `backfill`) and some of them would not be compatible. Does it make sense to add some checks?

dongahn:
@lipari: I am done with my review for now. This work is significant! I agree with you that we want to go with your xml-based solution for now to make some progress and work on some enhancements as part of future work.

I think we can "use" hwloc's native API to walk its native objects to be more efficient but we probably don't want to solely rely hwloc's operations (i.e., tree-walk abstraction, etc) in providing `resrc`'s abstraction. What I would like to see is your `resrc` ultimately provides the level of abstraction on resrc objects, which is matched with that of hwloc, so that our system works well no only with hwloc read option but also with rdl option.

In general, I like the various tree-walking, -query abstraction of hwloc, and I would like to see more of these abstractions (e.g., retrieve all of the resource of the type A, walk from me toward the root...) from reserc. I believe a richer resrc abstraction can go a long way towards enabling both scheduling and mapping/binding policies.

And I'd love to see these abstractions being firmed up with real use cases. As a first step, perhaps we can see if resrc makes it easy to write various task mapping policies? I think allowing to build a hierarchical resource id space can be very effective in this area, as you and I briefly chatted before. We can discuss more, I guess.

dongahn:
@lipari: thank you for breaking the comments into manageable pieces and creating new issues. With these, I'm merging.

PR #80, opened 2015-10-08
flux-rdltool: update optparse_t usage

garlick:
optparse_t was redefined as a struct rather than a pointer to one
per RFC 7.  The single use in flux-sched also needs to be updated.


grondo:
Ack sorry I missed hitting up sched after that change!

PR #82, opened 2015-10-09
test: print PASS/FAIL after each sharness test

garlick:
I got a little lost trying to figure out what sharness test belonged to what output, especially the ones that repeat the same tests for different scheduler algs.  This prints PASS or FAIL: and the name of the test after each test is run similar to the automake test driver.


dongahn:
Thanks Jim. Merging.

PR #86, opened 2015-10-13
resrc: plug various memory leaks

lipari:


dongahn:
Looks good to me. I will merge. This patch will help @SteVwonder.

PR #87, opened 2015-10-14
Consistent options

lipari:
Addresses the concern that, with the addition of the new simulator options, users could stipulate incompatible sched module load options.  Also, includes the replacement of a deprecated function call.


dongahn:
Looks good! Like we discussed, let's put in a few test cases later to verify that the emulator also works with the new hwloc-based resources! I'm merging this for now.

Issue #88, opened 2015-10-15
Migrate backfill scheduling support from the sched framework module to its own plugin.


lipari:
The latest enhancements that added backfill scheduling were helpful and carefully done so as to not impact the existing FCFS operation.  However, the vision for the schedsrv framework service was to provide the foundation for a variety scheduling algorithms.  Backfill is the first such alternative algorithm to be developed.  This issue is to move backfill scheduling out of the sched framework and into its own plugin.

dongahn:
This sounds good, @lipari! I look forward to pushing this ideas forward, along with  @SteVwonder and @surajpkn who actually have hands-on experiences of implementing the backfill algorithms.

lipari:
Closed with the merge of [PR 91](https://github.com/flux-framework/flux-sched/pull/91)

PR #91, opened 2015-10-28
sched:  extract the backfill scheduling code from the sched framework

lipari:
A new backfillplugin1 was added as an alternative to the original FCFS
schedplugin1 plugin.  The EASY_BACKFILL definition in this plugin
controls whether the backfill is EASY or conservative.  It's currently
defined to be EASY so as to function as an exact replacement.

3 new API's were added to the sched framework: sched_loop_setup(),
allocate_resources(), and reserve_resources().

Includes many changes across sched as well as resrc.  resrc now
includes support for time-based as well as now-based resource
matching, allocations and reservations.  The backfill scheduler uses
the time-based functionality while the schedplugin1 FCFS uses the
now-based versions.

The stage_tree_list() was removed and instead staging was moved to the
selection function.  As size elements of a resource are selected, they
are also staged.

Finally, the t2002-easy.t test was modified to load the new plugin.
The hype-test.csv file of input jobs was not modified and the
easy_expected order remains the same.

Addresses [Issue 88](https://github.com/flux-framework/flux-sched/issues/88)

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -44,13 +44,14 @@ struct resrc {
     int64_t id;
     uuid_t uuid;
     size_t size;
+    size_t available;
     size_t staged;
     resource_state_t state;
     resrc_tree_t *phys_tree;
     zlist_t *graphs;
     zhash_t *properties;
     zhash_t *tags;
-    zhash_t *allocs; // TODO: merge with twindow
+    zhash_t *allocs;
     zhash_t *reservtns;
     zhash_t *twindow;
 };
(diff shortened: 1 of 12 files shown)


surajpkn:
Want to just point that I need to rebase on this before I can proceed with my next changes. So early merge requested! :-)

dongahn:
OK. I will make this a priority.

dongahn:
@lipari and I had a code review together because this represents pretty significant changes. We've seen some minor points to address, but other than these this PR looks solid! Once he addresses them, this PR can be merged from my perspective. Actually, this should be merged ASAP given that @surajpkn depends on this.

lipari:
Ok, all suggestions made during the code review have been made and revised commit has been pushed.

dongahn:
:+1:

PR #96, opened 2015-11-25
sched: replace direct read of kvs with rpc to resource-hwloc

lipari:
Leverages resource-hwloc's topo rpc instead of reading the kvs
directly.

### DIFF
--- a/sched/schedsrv.c
+++ b/sched/schedsrv.c
@@ -432,33 +432,37 @@ static int load_resources (ssrvctx_t *ctx, char *path, char *uri)
                       ctx->rctx.root_uri, path);
         }
     } else if ((ctx->rctx.root_resrc = resrc_create_cluster ("cluster"))) {
-        char    *buf = NULL;
-        char    *key;
-        int64_t i = 0;
+        JSON o = NULL;
+        const char *buf = NULL;
+        const char *json_str = NULL;
+        flux_rpc_t *rpc;
         size_t  buflen = 0;
 
-        rc = 0;
-        while (1) {
-            key = xasprintf ("resource.hwloc.xml.%"PRIu64"", i++);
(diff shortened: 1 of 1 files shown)


garlick:
Apologies if this has already been discussed but why not read this from the KVS?

Does the single RPC directed to FLUX_NODEID_ANY achieve what the old code did by iterating through the ranks pulling in XML from each one?

dongahn:
My guess is resrc_generate_xml_resources has the walking logic to populate resrc_t objects. 

Actually, I have been looking at the latest addition to schedsrv.c and found that the code base is getting increasingly complex with various support like multiple reader options and emulation mode. 

I have been trying to factor out things and one of the abstraction is resource reader class called rsreader. 

I have to believe that the new bulk loading can be the basis for a placeholder method in it called rsreader_hwloc_bulkload.

@lipari, I plan to post a prelim PR today and if you don't mind, can you please look at the new code base before we proceed with this change?

lipari:
Utilizing the resource-hwloc's topo rpc is similar to using the jsc to report job status.  It allows the owner, in this case the resource-hwloc, to own and deliver info that it stores in the kvs.

I will eventually need to target each resource-hwloc specifically and get it to return the resources it sees.  But for now they all report the same inventory of all the resources across the instance.

lipari:
> @lipari, I plan to post a prelim PR today and if you don't mind, can you please look at the new code base before we proceed with this change? 

Sure.

garlick:
Since `resource-hwloc` already does the work to store data instance wide hwloc data in the KVS, it would be less scalable to `flux_rpc_multi()` than to read the data directly from the KVS.

There's no reason modules can't have "contracts" using the KVS as intermediary.

lipari:
Good point.  I will withdraw the PR.

PR #98, opened 2015-11-25
sched: remove sched's flux-submit

lipari:
Instead use the new flux-submit that was added to flux-core in
https://github.com/flux-framework/flux-core/pull/480


grondo:
Nice! Though not strictly necessary, the `-x $tdir/sched` could also be removed from the test scripts' use of `flux submit` as well.

(I was working on a similar PR but `t1000-jsc.t` started failing for me with these changes.. not sure what I did wrong)

lipari:
> Nice! Though not strictly necessary, the -x $tdir/sched could also be removed from the test scripts' use of flux submit as well.

Done.  The tests passed, so the `-x $tdir/sched` vestiges failed to grab my attention.

grondo:
Thanks! Certainly I ACK this one!

dongahn:
Looks great. Thanks.

Issue #69, opened 2015-09-17
Need better mapping/scheduling between tasks and nodes/cores


dongahn:
This was discussed a bit as part of #58, but I wasn't sure if a new issue has been created for this. As is, even if you launch 4 brokers, and then submit each job requiring 2 nodes back to back to back (e.g., my stress case), sched maps these 2-node jobs only to the first two ranks, never using the other two ranks.

This is just to discuss if there are some better mapping/scheduling that sched can do as a short-term improvement? 

@lipari: what is the semantics of `find_resources` and `select_resources` of the `schedplugin1`  plugin? Does it simply treat all of the resources as shared and grab the first available resources regardless they have running jobs? Are there any other scheduler plugins in the works with more smarts? 

Also, if there have been some updates to `wreckrun` with respect to job-to-resource mapping, it seems worthwhile to match that in the `sched-flux` as well. If wreckrun launches jobs back to back what is the current semantics? Anyone has issue numbers that capture some of these questions, mind listing the numbers here?

grondo:
`flux-wreckrun` itself currently ignores actual resources and distributes tasks evenly across *ranks* when running without `sched` (see `alloc_tasks_hack`). This code was only meant as a stand in for actual resources assignment when run with `--immediate`, but when `sched` was removed from core it became the default.

`wrexecd` simply checks `lwj.<id>.rank.<rank>.cores` and launches that many tasks on the current rank. A user of the `wreck` service outside of `flux-wreckrun` could assign those as they see fit.

Thus, if `flux-wreckrun` launches jobs back to back, in the default mode, it will run two identical jobs on the same two ranks. Current thinking is that we shouldn't spend too much time fixing up wreck prototype at this point, but if there is something quick we need to get real work done I am not opposed to that. But clearly, at this point `flux-wreckrun` or any other front-end tool won't be able to reliably tell which resources to pick to avoid overlap. We need some sort of scheduling or resource update/query support for that.

dongahn:
Thanks @grondo. Yeah we need scheduling and sched is supposed to do that, filling the right information into  '...rank.cores'


I think a part o the problem is that I often use a rdl file that doesnt acrurately describe the resources under the bootstrap environment. 

I think we should reinvested this after we have the multiple RDL reader option implemented, which is captured in #33 and #58.

I don't know if anyone is looking at this currently but this seems a pretty interesting item to look at if I can have some spare time.

[...]

PR #100, opened 2015-11-26
rs2rank: add support to link resrc to broker rank, etc

dongahn:
This is not ready to be merged. But before I turn into a pumpkin for a week, I wanted to post this to get some early feedback. This PR has two distinct items for your review:

- Fix the issue #69 where schedsrv incorrectly targets the first
k consecutive broker ranks (starting from rank 0) to launch a job
of k nodes, regardless of what nodes it is allocated to
and what broker ranks manage these nodes. For example, even if two
jobs are allocated back to back to cab[1-2] and cab[3-4], the old
scheme will end up launching both of the jobs into cab[1-2], assuming
rank 0 and 1 are running on these nodes. The detail is in the commit message. 

- Add a bridge layer into schedsrv: With the latest emulation integration,
emulation-specific code has been sprinkled through schedsrv, potentially making
the code error prone. This layer centralizes where specialized
code for either mode is used so that hopefully can improve code
maintainability and reduce errors.

This PR needs a related flux-core PR for testing, which I will post soon.

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -41,6 +41,7 @@ struct resrc {
     char *type;
     char *path;
     char *name;
+    char *digest;
     int64_t id;
     uuid_t uuid;
     size_t size;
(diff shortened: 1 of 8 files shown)


dongahn:
Travis CI doesn't seem to like it. I won't have time to look at the error now but the code should be okay to be reviewed with the build error. I will fix the error when I come back from my vacation.

springme:
Happy Thanksgiving, Dong!



Sent with Good (www.good.com)

grondo:
> Travis CI doesn't seem to like it. I won't have time to look at the error now but the code should be okay to be reviewed despite the build error. I will fix the error when I come back from my vacation.

@dongahn, the error is most likely because your flux-core PR has not yet been merged, and in Travis this PR will build against flux-core/master. Nothing to worry about over your vacation!

dongahn:
Closing this and I will submit a new PR from a different branch.

PR #102, opened 2015-12-14
resrc: Fix bug in resource state assignment

lipari:
The allocation and reservation logic uses the resource state to decide
whether to do time or now-based allocations/reservations.  This commit
fixes some places where the state was being improperly set.


dongahn:
Thanks. I'm merging it.

PR #103, opened 2015-12-15
Issue 101

lipari:
Address [Issue 101](https://github.com/flux-framework/flux-sched/issues/101) by:
* support multiple job allocations per resource
* support a request for exclusive access to a resource

PR #104, opened 2015-12-16
resrc: Change resrc_match_resource() to match a request

lipari:
The old argument to resrc_match_resource() matched a specified
resource against a sample resource.  The sample resource was the
resource part of a resource request object.  This proved to be
combersome as we proceed with the next round of changes.  So, now we
pass the whole request to resrc_match_resource().

This complies with a request made in https://github.com/flux-framework/flux-sched/pull/103#issuecomment-164945235 to break [PR 103](https://github.com/flux-framework/flux-sched/pull/103) into two pieces.

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -34,6 +34,7 @@
 #include "rdl.h"
 #include "resrc.h"
 #include "resrc_tree.h"
+#include "resrc_reqst.h"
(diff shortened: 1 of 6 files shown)


dongahn:
I will be making some in-line comments to the code.

dongahn:
Unless I hear otherwise, I will merge this by COB today.

PR #107, opened 2016-01-05
Exclusive mods

lipari:
Fix some minor bugs in requesting a resource size greater than one.  Also, use a new flux_res_t member to replace hard coded behavior.


dongahn:
@lipari: I think this is fine. I will merge this in and rebase my PR on it. Thanks.

PR #108, opened 2016-01-09
resrc: add to_string to print a resrc's state to a string

dongahn:
Add resrc_to_string so that a resrc's state can be printed to a string which then can be directed to `flux_log ()`. For now, the debug print function (`resrc_print_resource ()`) prints the state stdout, which gets lost when used in a comms module (e.g., `schedsrv`). `resrc_print_resource ()` is now a thin wrapper on top of `resrc_to_string ()`.


dongahn:
Rebased.

PR #110, opened 2016-01-15
build/test: Update flux commandline options

grondo:
flux(1) utility command line options --module-path (-M), --lua-path (-L),
--lua-cpath (-C), and --python-path (-P) will be removed from upstream
flux-core. This commit removes use of those options from flux-sched
build and test infrastructure, in preference for environment variables
that perform a a similar function.

This change should be applied immediately after flux-framework/flux-core#516
is merged. It cannot be applied beforehand because `flux(1)` does not currently
support extending `FLUX_MODULE_PATH` from the environment.

If we'd rather not temporarily break flux-sched, we could break this into parts

 1. flux-core: Add `FLUX_MODULE_PATH` environment support to `flux(1)`
 2. flux-sched: Merge this PR
 3. flux-core: Remove -M, -P, -L, -C

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -50,5 +50,8 @@ KVS_LIBS = -Wl,-rpath,$(FLUX_BUILDDIR)/src/modules/kvs/.libs \
 #
 # When we invoke flux we need to provide additional LUA paths for RDL
 #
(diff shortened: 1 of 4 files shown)


grondo:
Just to reiterate, Travis-CI won't pass until flux-framework/flux-core#516 has been merged. (Or a subset with support for `FLUX_MODULE_PATH` has been)

dongahn:
The change looks good to me. Merging as [flux-framework/flux-core#516)[flux-framework/flux-core#516] has already been merged.

grondo:
Thanks!

Issue #83, opened 2015-10-11
Add autotools support


dongahn:

dongahn:
Create a new issue for autotools support as I closed #3. There we have addressed all issues except for autotools support.

garlick:
Fixed by #115.

Issue #56, opened 2015-09-01
Eliminate from flux-sched products the dependency on libflux-internal


lipari:
This is the beginning step toward streamlining flux-sched's build system.  This leaves a dependence on the flux-core library only.

SteVwonder:
What is our plan for shortjson.h?
I know copying the version in flux-core into flux-sched has been discussed.
Did we ever agree on a solution?

lipari:
https://github.com/flux-framework/flux-sched/pull/57 indeed copies shortjson.h into flux-sched/common to make things simple.  Long term, we can ferret out all the places that rely on the shortjson.h defs and replace them with the json-c equivalents, if such a cleansing is desired.

SteVwonder:
I now see the discussions happening in #57 and flux-framework/flux-core#336

SteVwonder:
Ok. Thanks @lipari!

lipari:
While we discussed this a while ago, I thought I would bring it up again as a way to embed the flux-core code base into flux-sched.  This would allow us to build flux-sched using the latest commits to flux-core.  There are three similar git tools to do this:

* https://git-scm.com/book/tr/v2/Git-Tools-Submodules#_git_submodules
* https://git-scm.com/book/tr/v2/Git-Tools-Advanced-Merging#_subtree_merge
* https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt

Of these three, `git subtree` appears to be favored in various tutorials I have seen.  My understanding is that, once we have flux-core embedded in the flux-sched project, we could add Autotool components to flux-sched and it would recursively build flux-core.  I also believe this allows flux-core to remain an independent project:  flux-sched would depend on flux-core and not vice versa.

lipari:
Copying in the contents of an email thread that discussed a different approach where flux-sched would not contain flux-core and instead pull in flux-core sources...

@garlick said:
On the build farm, the flux-core RPM would be installed into the mock root that is being used to build flux-sched's RPM.

In travis-ci, an RPM would not be used.  flux-core would be "side installed" (make install prefix=my-side-dir) and used by flux-sched the same way that flux-core uses its dependencies.  See src/test/travis-dep-builder.sh.

In both of the above cases, flux-sched would use normal mechanisms like pkg-config to locate flux-core.  It may need some help from environment variables if non-system locations are used, e.g. PKG_CONFIG_PATH, LD_LIBRARY_PATH.

@grondo referenced his work here:
[...]

PR #115, opened 2016-02-02
add Autotools support to flux-sched project

lipari:
Replaced the Makefiles with Makefile.am's and added the Autotool support.  Rather than clone the associated files from flux-core, the effort started from scratch using more current conventions.  You can invoke `autoreconf -i` the first time and then `autoreconf` each time after that.  An `autogen.sh` script was added to resemble the flux-core system.  `autogen.sh` wraps `autoreconf -i` and does some cleanup.

Used the Autotool version available on our existing TOSS 2 systems. The plan is to update the tools to the TOSS 3 versions, once we abandon development on TOSS 2.

flux-sched is based on the flux-core library.  flux-sched can be built against an installed flux-core package.  However, flux-sched can be built against any side-installed flux-core by setting the PKG_CONFIG_PATH environment variable to the location of the side-installed lib/pkgconfig directory.

flux-sched requires additional utilities from flux-core that are not include in the flux-core library.  Rather than copy/paste these extra utilities into flux-sched, we opted to point to the flux-core source files and developed a mechanism for identifying where those extra source files can be found.

The `configure` script offers one option and it is the means of identifying the location of the source to the utilities.
 * `--with-flux-install=` allows you to build flux-sched using installed flux-core libraries.  The value of this option provides the location of the installed source tree.

Addresses:  https://github.com/flux-framework/flux-sched/issues/56 and https://github.com/flux-framework/flux-sched/issues/83


lipari:
This is a preliminary PR for review.  There were some minor changes to source code to that appeared under -Werror.  If desired, I can pull those changes into a separate commit.  Also, I will enable the silent mode once the dust settles.

grondo:
Separating some of this work into other commits, if easy, would be helpful in a review. Thanks! I'm going to take a closer look, but this represents a heroic effort on your part. :+1:

grondo:
Some quick comments: These are all really minor.

 - Consider using `AC_CONFIG_AUXDIR` and `AC_CONFIG_MACRO_DIR` so move autoconf products and macros into subdir and remove clutter from top-level srcdir.
 - Maybe the `config` script and `.build` link could be removed, and we could figure out how to add a `./configure --with-flux-core=/path/to/flux/builddir` options?
 - Consider an `autogen.sh` -- even if it just calls `autoreconf -i` (to match flux-core usage)
 - Re-enable`AM_SILENT_RULES([yes])`? If you want a verbose build you could always use `make V=1`
 - Newer(?) automake is complaining that `subdir-objects` should be used project-wide:
```
automake: warning: possible forward-incompatibility.
automake: At least a source file is in a subdirectory, but the 'subdir-objects'
automake: automake option hasn't been enabled.  For now, the corresponding output
automake: object file(s) will be placed in the top-level directory.  However,
automake: this behaviour will change in future Automake versions: they will
automake: unconditionally cause object files to be placed in the same subdirectory
automake: of the corresponding sources.
automake: You are advised to start using 'subdir-objects' option throughout your
automake: project, to avoid future incompatibilities.
```

garlick:
Yeah this is great @lipari.

Note we just dumped META in pr flux-framework/flux-core#526.  You might check out that change as it increases safety in the release generation process IMHO.

dongahn:
Thanks @lipari! 

@grondo makes good points. I will appreciate have break a single commit to smaller topical commits for review. 

One quick comment. I think you can use EXTRA_DIST rule to "make dist" pick up some testing scripts and testing data files that are not automatically picked up. E.g., you can specify this rule in t/Makefile.am to distribute sharness scripts and testing data.

grondo:
> One quick comment. I think you can use EXTRA_DIST rule to "make dist" pick up some testing scripts and testing data files that are not automatically picked up. E.g., you can specify this rule in t/Makefile.am to distribute sharness scripts and testing data.

`make distcheck` is a good way to ensure dist tarball is properly formed. It will test `make dist`, build from the resulting tarball using a VPATH build (a good check for builddir vs srcdir), run `make check` and then `make clean` to ensure the builddir is properly cleaned.

lipari:
> @grondo makes good points. I will appreciate have break a single commit to smaller topical commits for review.
I'm conflicted on this.  While it would be easy to divide the commits into separate packages, e.g. the rdl, resrc, sched, and simulator, none of these intermediate commits would compile until the last one.  That seems wrong to me in that it prevents the use of git bisect.
Unfortunately, this is a wholesale revamp of the flux-sched project as opposed to independent changes to existing functionality as typical PR's contain.

grondo:
@lipari: Good point that splitting some of this out would break intermediate builds. (I was only suggesting that, when possible, small logical commits are easier for review). All you could have done was maybe convert to autotools in one commit without `-Werror`, update code to fix errors in compile, then add `-Werror`. I'm not sure the work is worth it in this case.

garlick:
That's a good point although it could at least be broken up functionally like
* Add configure.ac, autogen.sh, config/* (no effect on building)
* Add Makefile.am's (still no effect)
* Remove Makefile, Makefile.inc (now autotools take over)

Then at least you don't have to sift through all the deletions to find the new content in a review.

lipari:
> @lipari: Good point that splitting some of this out would break intermediate builds. (I was only suggesting that, when possible, small logical commits are easier for review). All you could have done was maybe convert to autotools in one commit without -Werror, update code to fix errors in compile, then add -Werror. I'm not sure the work is worth it in this case.

Ok, understood.  And though I neglected to say this, [PR 116](https://github.com/flux-framework/flux-sched/pull/116) contained only the fixes to the source code that will eliminate the errors.  The new push of this branch will not include any code mods.

@garlick, I will follow the plan you proposed.

garlick:
Another thought is if you're bringing in chunks wholesale from another source (such as INSTALL, lua.m4, etc, add them in separate commits and cite the source in the commit message.  That helps reviewers tell the difference between new work, work imported wholesale, and work imported then modified (in the latter case, modifications should be in commit(s) following an import commit).

dongahn:
> make distcheck is a good way to ensure dist tarball is properly formed. It will test make dist, build from the resulting tarball using a VPATH build (a good check for builddir vs srcdir), run make check and then make clean to ensure the builddir is properly cleaned.

Yes, good point. We should probably have `distcheck` as part of our CI as well. For this pass, though, I am okay if we just make sure all autotool targets like this are correct. Then this CI change can come later, which I expect to be pretty minor.

lipari:
Ok, just re-pushed the branch that addresses the suggestions:
* broke the one major commit into smaller pieces
* added AC_CONFIG_MACRO_DIR to configure.ac
* replaced the old .build solution with a new configure --with-flux-core=/path/to/flux/builddir option
* added a autogen.sh script to be consistent with flux-core
* re-enabled AM_SILENT_RULES
* removed the inclusion of the META macro

I experimented with adding the subdir-objects as an option to AM_INIT_AUTOMAKE, but it resulted in automake/libtool trying to deposit intermediate objects under the /usr/lib64/flux-core/build/src directories.  So, I will revisit this another day, if/when it becomes a problem.

I experimented with replicating the PACKAGE_VERSION and PACKAGE_NAME system from flux-core described above in [PR 526](https://github.com/flux-framework/flux-core/pull/526), but I will defer that work to the subsequent effort to create a flux-sched distribution.

There are still some issues to work out (e.g., [Issue 545](https://github.com/flux-framework/flux-core/issues/545)) when building for an installed flux on TOSS 3, so there will be another pass at this once the issues are resolved.

garlick:
Very nice improvement in the structure of the PR!

Could you get away with just configure czmq not zeromq?   Maybe just a one liner in your configure.ac, e.g. 
```
PKG_CHECK_MODULES([CZMQ], [libczmq >= 3.0.0])
```
I don't think you're using zeromq directly, correct?

lipari:
@garlick, you are correct.  Just pushed a rebase that replaces the X_AC_ZEROMQ macro with your suggested PKG_CHECK_MODULES([CZMQ]).  I removed the no-longer-necessary m4/x_ac_zeromq.m4 as well.

garlick:
On AC_INIT, we updated the one we were using in flux-core to:
```
AC_INIT([flux-core],
        m4_esyscmd([git describe --always | awk '/.*/ {printf "%s",$1; exit}']))
```
`--always` was added so that it generates something even if there are no tags in the repo.  Maybe that was what was preventing you from using it?    See flux-framework/flux-core@167676689d52863f4cf5f1e77d47baaa895bd182

lipari:
>  Maybe that was what was preventing you from using it?

It was indeed.  Just pushed a new configure.ac with the suggested AC_INIT arguments.

garlick:
running autogen, I hit this:
```
simulator/Makefile.am:72: `%'-style pattern rules are a GNU make extension
```
Are these make convenience targets to start flux and load modules seven needed?  They seem vaguely inappropriate in automakefiles.

I'd suggest moving this block to its own m4 file (defined as a macro), as this may be useful to steal for other framework projects that depend on core.   It would be nice to avoid hardwiring library paths and library names here, but I'm not sure what the right approach is.
```
#
# Use the specified path to flux-core if it has been given.
# Otherwise, look to the installed locations.
#
AC_ARG_WITH([flux-core],
    [AS_HELP_STRING([--with-flux-core=[path_to_src]],
        [specify location of flux-core project])],
    [FLUX_BUILDDIR=${withval} FLUX_SRCDIR=${FLUX_BUILDDIR}
     FLUX_CORE_CFLAGS="-I${FLUX_SRCDIR} -I${FLUX_SRCDIR}/src/include"
     FLUX_CORE_LIBS="-L${FLUX_BUILDDIR}/src/common/.libs -lflux-core \
      -L${FLUX_BUILDDIR}/src/modules/libjsc/.libs -lflux-jsc \
      -L${FLUX_BUILDDIR}/src/modules/kvs/.libs -lflux-kvs"
     FLUX="${FLUX_BUILDDIR}/src/cmd/flux"
     AC_SUBST(FLUX_CORE_CFLAGS)
     AC_SUBST(FLUX_CORE_LIBS)
    ],
    [PKG_CHECK_MODULES([FLUX_CORE], [flux-core],
      [FLUX_BUILDDIR=/usr/lib64/flux-core/build
       FLUX_SRCDIR=/usr/lib64/flux-core/build
       FLUX=/usr/bin/flux
      ],
      AC_MSG_ERROR([No installed flux package or --with-flux-core specified]))
    ]
)

AC_SUBST(FLUX_BUILDDIR)
AC_SUBST(FLUX_SRCDIR)
AC_SUBST(FLUX)
```

garlick:
I think setting `AC_CONFIG_AUX_DIR([config])` will prevent autom4te.cache, config.h, etc from being created at the root level, cluttering things up a bit.

lipari:
> I think setting AC_CONFIG_AUX_DIR([config]) will prevent autom4te.cache, config.h, etc from being created at the root level, cluttering things up a bit.

I don't follow you.  The AC_CONFIG_AUX_DIR([config]) is the same thing used in flux-core's configure.ac.  Furthermore, after running flux-sched's autogen.sh script, I see a root directory populated with (mostly) the same contents as flux-core.  The config/ directory is not present initially, but autoreconf -i creates and populates it.

garlick:
Sorry, mainly I wanted to call your attention to the auto* products left at the root level which might be tidier to move to config.  I think maybe the AC_CONFIG_AUX_DIR setting is only part of the solution; you can see the flux-core autogen.sh script does some tidying:
```sh
echo "Running aclocal ... "
aclocal -I config
echo "Running libtoolize ... "
libtoolize --automake --copy
echo "Running autoheader ... "
autoheader
echo "Running automake ... "
touch ChangeLog
automake --copy --add-missing
echo "Running autoconf ... "
autoconf
echo "Cleaning up ..."
mv aclocal.m4 config/
rm -rf autom4te.cache
echo "Now run ./configure."
```

lipari:
Understood.  When I created the flux-sched version of autogen.sh to wrap `autoreconf -i`, I "borrowed" the same clean-up measures from flux-core's `autogen.sh`:
```
echo "Running autoreconf -i ... "
autoreconf -i
echo "Cleaning up ..."
mv aclocal.m4 config/
rm -rf autom4te.cache
echo "Now run ./configure."
```
Note that flux-sched's .travis.yml executes `autoreconf -i` directly, not `autogen.sh`.

garlick:
I think I had myself confused back there, sorry!  Never mind.

lipari:
Ok. Latest push has cleaned up Makefile.am files and the flux-core macros moved into a new ax_flux_core.m4 file.  Thanks for all your suggestions, BTW.

garlick:
This "else" leg that runs if the user doesn't configure `--with-flux-core=srcdir` really shouldn't hardwire these paths
```
  [PKG_CHECK_MODULES([FLUX_CORE], [flux-core],
       [FLUX_BUILDDIR=/usr/lib64/flux-core/build
        FLUX_SRCDIR=/usr/lib64/flux-core/build
        FLUX=/usr/bin/flux
       ],
```
I think you could find FLUX with AC_CHECK_PROG.

I'm not too sure what to do about the others.

lipari:
> I'm not too sure what to do about the others.

I just pushed an update with  AC_CHECK_PROG(FLUX) replacing the hard coded FLUX assignment.  I'm inclined to defer action on the others until a decision is made on how to address [Issue 545](https://github.com/flux-framework/flux-core/issues/545)

garlick:
This is looking pretty nice.  DId you miss removing the start and dump targets from `rdl/Makefile.am` or were those left in on purpose?

lipari:
>  DId you miss removing the start and dump targets from rdl/Makefile.am or were those left in on purpose?

Ah, that was an oversight.  Rebased branch just pushed.

garlick:
Why does configure run twice here?  (once when I run it outright, again when I run make)?
```
$ module load flux-core 
$ ./autogen.sh
Running autoreconf -i ... 
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `config'.
libtoolize: copying file `config/ltmain.sh'
configure.ac:24: installing `config/compile'
configure.ac:15: installing `config/config.guess'
configure.ac:15: installing `config/config.sub'
configure.ac:11: installing `config/install-sh'
configure.ac:11: installing `config/missing'
rdl/Makefile.am: installing `config/depcomp'
Cleaning up ...
Now run ./configure.
$ ./configure
[snip]
config.status: executing depfiles commands
config.status: executing libtool commands

  flux-sched version 9e8b56a
  Prefix...........: /usr/local
  Debug Build......: 
  C Compiler.......: gcc -std=gnu99
  CFLAGS...........: -g -O2
  CPPFLAGS.......... 
  FLUX.............: flux
  FLUX_BUILDDIR....: /usr/lib64/flux-core/build
  FLUX_CORE_CFLAGS.: -I/opt/flux-core/include  
  FLUX_CORE_LIBS...: -L/opt/flux-core/lib/flux -ljsc -llive -lmodctl -lmrpc -lkvs -lbarrier -lflux-core  
  FLUX_SRCDIR......: /usr/lib64/flux-core/build
  LDFLAGS..........: 
  LIBS.............: -lxml2 -luuid -lhwloc 
  Linker...........: /usr/bin/ld -m elf_x86_64

$  make
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/sh /home/garlick/proj/flux-sched/config/missing --run aclocal-1.11 -I m4
 cd . && /bin/sh /home/garlick/proj/flux-sched/config/missing --run automake-1.11 --gnu
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/sh /home/garlick/proj/flux-sched/config/missing --run autoconf
/bin/sh ./config.status --recheck
running CONFIG_SHELL=/bin/sh /bin/sh ./configure --no-create --no-recursion
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
[snip]

**Comment by garlick** [comment]
Incidentally, building against the installed flux-core/0.1.0, the build fails due to the hardwired FLUX_SRCDIR (or FLUX_BUILDDIR) as that should be `/opt/flux-core/build` on this system (TOSS 2/RHEL6)

```
$ make
make  all-recursive
make[1]: Entering directory `/home/garlick/proj/flux-sched'
Making all in rdl
make[2]: Entering directory `/home/garlick/proj/flux-sched/rdl'
Making all in .
make[3]: Entering directory `/home/garlick/proj/flux-sched/rdl'
  CC     cpuset_la-lua-cpuset.lo
  CC     cpuset_la-cpuset-str.lo
  CCLD   cpuset.la
  CC     rdl.lo
rdl.c:39:36: error: src/common/liblsd/list.h: No such file or directory
rdl.c:40:39: error: src/bindings/lua/json-lua.h: No such file or directory
rdl.c:41:41: error: src/common/libutil/xzmalloc.h: No such file or directory
```

**Comment by garlick** [comment]
You are doing some nice work here @lipari.  I am leaving for the day but I'll try to keep on top of it over the weekend if you continue to work on it, or if not there's always next week.

**Comment by lipari** [comment]
@garlick, it turns out that using the `autoreconf` script, `aclocal.m4` needs to stay in the root directory.  The `configure` script re-run was caused by the `autogen.sh` script moving `aclocal.m4` it into the `config` directory.  I thought of messing with environment variables to get around this, but I decided it does not introduce that much clutter that it bothers me.  Perhaps someone else will know a clever way to move it and not trigger the `configure` script to re-run.
In thinking more about the hard coded, installed source code locations, I decided to add another option to `configure` (via the `ax_flux_core.m4` macro).  `--with-flux-install` will allow you provide the location of the installed source tree, whether it is on TOSS 2 or 3.
The downside to this is that now you *have* to provide an option to `configure`.  That makes sense because you either need to tell it the location of the installed source tree or your cloned flux-core repo.  Admittedly, there might be a clever way to automate this someday and find the associated external source files automagically.  But for now, this will work.
Rebased branch was pushed.  And, I edited the PR description above to reflect these latest changes.

**Comment by garlick** [comment]
> it turns out that using the autoreconf script, aclocal.m4 needs to stay in the root directory. 

Good catch.

This is turning out nice, but  `--with-flux-core` and `--with-flux-install` seem off.  It seems like an _installed_ flux-core should be required, and found in the usual way (without options), and the sources that have to be found in the installed build directory might require an option like `--with-flux-core-build` temporarily until we can find a way for installed flux-core to provide the path, provide those sources another way, or eliminate the dependency.

**Comment by lipari** [comment]
> It seems like an installed flux-core should be required, and found in the usual way (without options), and the sources that have to be found in the installed build directory might require an option like --with-flux-core-build temporarily until we can find a way for installed flux-core to provide the path, provide those sources another way, or eliminate the dependency.

That's actually what happens with the latest changes.  When you specify --with-flux-install, it uses pkg-config to get the CFLAGS (albeit null) and LIBs for flux-core.  But that is insufficient for flux-sched which needs those other utility source files.  And because we haven't come up with a rule on where to find those across TOSS distros, --with-flux-install is how to convey the location.

Admittedly, it is a short term solution until we wean flux-sched from those extra utilities.  Then --with-flux-install can go away and just plain `configure` can use the installed flux-core libraries.

**Comment by garlick** [comment]
I guess then my comment boils down to
* eliminate `--with-flux-core` (always require flux-core to be _installed_)
* rename `--with-flux-install` to `--with-flux-core-builddir`

**Comment by lipari** [comment]
A brief recap of the two scenarios...  The first scenario is building flux-sched against a cloned repo of flux-core.  This is the way we have been building flux-sched since the beginning and required the .build link to point to where the flux-core tree can be found.  This scenario will continue to be required whether or not an official flux-core package is installed on the machine.

The second scenario builds against an installed flux-core package.  PKG_CONFIG macro and AC_CHECK_PROG will identify the flux-core libraries and the flux command, but flux-sched needs to pull in utilities not included in libflux-core.  At the moment, there is no package that contains a library of these utilities and there is no rule for finding these source files.  So, the developer needs to specify to configure where they reside on his/her machine.

The goal is to minimize the necessary options to `configure`.  But for each of these scenarios above, a path needs to be supplied.  And, unless we hard code the location of the source tree to the utilities, running `configure` alone with no options to build against the installed flux-core package will fail without the knowledge of where the source to the utilities is.

I'm fine with renaming the `--with` options to `configure` along the lines you propose.  However, I don't see how we can avoid having to specify either the location of the cloned flux-core repo or the location of the source to the utilities.

**Comment by garlick** [comment]
> This scenario will continue to be required whether or not an official flux-core package is installed on the machine.

Oh, I had assumed this would not be required anymore.  Even if flux-sched is to be built against flux-core master in travis-ci or during development, it could still be built from a side installed location (e.g. ~/local with PKG_CONFIG_PATH set to point there, see `travis-dep-builder.sh`).  If that's not done then it's hard to restrict sched to only the parts of flux-core that are exported.  That's the path we want to be on IMHO.

**Comment by garlick** [comment]
Couple thoughts - appreciate your take on this @grondo:
* since `build` is installed by the spec file, sched cannot be built from a side installed, unpackaged core.  Should the build directory be created by `make install`?
* Could we dispense with the `--with-flux-core-builddir` option if the installed build location were made available as a broker attribute?  (Somehow that feels wrong but I cant't say exactly why so maybe not?)

**Comment by grondo** [comment]
> since build is installed by the spec file, sched cannot be built from a side installed, unpackaged core. Should the build directory be created by make install?

I can see what you mean, but ewww, no. For now what is the problem with pointing `flux-sched` to the builddir from which you did `make install`? I thought this was temporary anyway. I guess you are considering a case where a site isn't using packaging, and only does `make install`?

> Could we dispense with the --with-flux-core-builddir option if the installed build location were made available as a broker attribute? (Somehow that feels wrong but I cant't say exactly why so maybe not?)

That feels really wrong, but it does have some nice properties. If nothing else is wrong with it though, it doesn't seem like we should assign a broker attribute to some temporary workaround... I could be just being a stick in the mud though.

It sure would be interesting though to require flux-framework projects to be built under a flux instance.. kind of like that old `flux get` idea...

**Comment by grondo** [comment]
I don't see such a problem with requiring `./configure --with-flux-core-builddir=` for now. This will be a pain point until we can pull common code into flux-sched, export interfaces from flux-core, or find some other solution. Since it will bug people we'll be more likely to look for solutions. Once we have a solution, we can just remove the `--with-flux-core-builddir` option from the `flux-sched` build and put it behind us.

**Comment by lipari** [comment]
FWIW, the change I'm currently testing eliminates the `--with-flux-core=` option as Jim suggested.  Developers will have the option of:
* building against an installed flux-core and use `--with-flux-core-builddir=` to point to the root of the extra sources, or
* creating a side install of flux-core, set PKG_CONFIG_PATH to where the side installed .pc files are, and use `--with-flux-core-builddir=` to point to the root of the extra sources.

**Comment by garlick** [comment]
Heh, I thought you might have an opinion :-)

> I guess you are considering a case where a site isn't using packaging, and only does make install

I was urging @lipari to build flux-sched in travis by checking out the flux core master, and installing it to  $HOME/local like the other dependencies.

**Comment by lipari** [comment]
Ok, the latest push successfully instructs travis to build and install the flux-core master branch and then build flux-sched against the installed library.  It copies the all the flux-core source files into a local directory and passes this directory to as the argument to `configure --with-flux-core-builddir` 

The lua tests are currently commented out.  I will next integrate the changes @grondo has provided in [PR 554](https://github.com/flux-framework/flux-core/pull/554)

**Comment by garlick** [comment]
It would be good to split out into standalone commits:
* remove old makefiles
* update travis.yml

**Comment by garlick** [comment]
Looking good.  I did just hit this in `make check`.  I assume you meant you had commented out the failing tests in your working copy?
```
make[3]: Entering directory `/home/garlick/proj/flux-sched/rdl/test'
loading RDL: /home/garlick/proj/flux-sched/rdl/RDL.lua:30: module 'flux.hostlist' not found:
        no field package.preload['flux.hostlist']
        no file '/home/garlick/proj/flux-sched/rdl/flux/hostlist.lua'
        no file '/home/garlick/proj/flux-sched/../flux-core/../share/lua/5.1/flux/hostlist.lua'
        no file './flux/hostlist.lua'
        no file '/usr/share/lua/5.1/flux/hostlist.lua'
        no file '/usr/share/lua/5.1/flux/hostlist/init.lua'
        no file '/usr/lib64/lua/5.1/flux/hostlist.lua'
        no file '/usr/lib64/lua/5.1/flux/hostlist/init.lua'
        no file '/home/garlick/proj/flux-sched/rdl/.libs/flux/hostlist.so'
        no file '/home/garlick/proj/flux-sched/../flux-core/../lib64/lua/5.1/flux/hostlist.so'
        no file './flux/hostlist.so'
        no file '/usr/lib64/lua/5.1/flux/hostlist.so'
        no file '/usr/lib64/lua/5.1/loadall.so'
        no file '/home/garlick/proj/flux-sched/rdl/.libs/flux.so'
        no file '/home/garlick/proj/flux-sched/../flux-core/../lib64/lua/5.1/flux.so'
        no file './flux.so'
        no file '/usr/lib64/lua/5.1/flux.so'
        no file '/usr/lib64/lua/5.1/loadall.so'
lt-trdl: rdllib_open: No such file or directory
FAIL: trdl
```

**Comment by garlick** [comment]
Never mind!  I reconfigured against the installed builddir and it works fine.  (I had configured against a checked out source dir)

**Comment by lipari** [comment]
Ok.  I *am* going to be taking another pass that the LUA paths, so things will change slightly...

**Comment by lipari** [comment]
> It would be good to split out into standalone commits:
>    remove old makefiles
>    update travis.yml

Done.

**Comment by garlick** [comment]
Just ran a quick test with a "side installed" flux-core and things seem to be good at least to get sched built and checked!
```
# build flux-core
./configure --prefix=$HOME/local
make
make install
mkdir -p $HOME/local/lib/flux-core
make clean
cp -r . $HOME/local/lib/flux-core
```

And then
```
# build flux-sched
export LD_LIBRARY_PATH=$HOME/local/lib
export CPPFLAGS=-I$HOME/local/include
export LDFLAGS=-L$HOME/local/lib
export PATH=$HOME/local/bin:$PATH

export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig
export BUILDDIR=$HOME/local/lib/flux-core/build

./configure --with-flux-core-builddir=$BUILDDIR --prefix=$HOME/local
make
make check
```

There are some issues with the install target which I'll comment on shortly.

**Comment by garlick** [comment]
Doing a `make install` after building sched with `--prefix=$HOME/local` I note:

* comms modules should be installed to `$libdir/flux/modules` not `$libdir`.
* includes should be installed to `$includedir/flux/sched` not `$includedir`
* subcommands should be installed to `$libexecdir/flux/sched` not `$bindir`(*)

(*) We might need to add this directory to FLUX_EXEC_PATH

**Comment by garlick** [comment]
Couple more nits:
* Replace relative source paths like `-I../resrc` with `-I${top_srcdir}/resrc` 
* Replace relative build paths like `../resrc/librsrc.la` with `${top_builddir}/resrc/librsrc.la`
* Comms modules should not be versioned, and should not export all their symobols.  In core we define this in configure.ac:
```
fluxmod_ldflags="-Wl,--no-undefined -avoid-version \
    -export-symbols-regex '^mod_(main|name|service)\$\$' \
    --disable-static -shared -export-dynamic"
AC_SUBST(fluxmod_ldflags)
```
then for each module use
```
kvs_la_LDFLAGS = $(fluxmod_ldflags) -module
```

**Comment by dongahn** [comment]
I am following @garlick and will run some testing.

**Comment by garlick** [comment]
Also: the installed core libs are all prefixed with `libflux-`, so probably 
* `librdl.so` should be `libflux-rdl.so`
* `libresrc.so` should be `libflux-resrc.so`
* `libsim.so` should be `libflux-sim.so`
* `libcpuset.so` should be `libflux-cpuset.so`

One other thing: `libforeign.la` should probably be built just once, not repeatedly in resrc/test, resrc, rdl/test, rdl, sched, simluator, and simulator/test.

**Comment by grondo** [comment]
> And then
> # build flux-sched
> export LD_LIBRARY_PATH=$HOME/local/lib
> export CPPFLAGS=-I$HOME/local/include
> export LDFLAGS=-L$HOME/local/lib
> export PATH=$HOME/local/bin:$PATH

I'm probably late to the game, but shouldn't `CPPFLAGS` and `LDFLAGS` be handled by `flux-core.pc`?

Say there are multiple `flux-core` installed and you want to build flux-sched against any one of them. What is the strategy to build against any one version of flux-core? i.e. do you have to set all these variables, or can we just set `PKG_CONFIG_PATH`? I'm not critiquing, just asking.

**Comment by garlick** [comment]
> shouldn't CPPFLAGS and LDFLAGS be handled by `flux-core.pc`?

Oh right of course.  PKG_CONFIG_PATH to build; PATH and LD_LIBRARY_PATH to run.

**Comment by grondo** [comment]
> subcommands should be installed to `$libexecdir/flux/sched` not `$bindir`(*)
> (*) We might need to add this directory to FLUX_EXEC_PATH

I forsee that as potentially problematic. Sorry if we've already been down this road before, but is there some reason we can't make a single exec-dir instead of one per framework project? I worry about automatic maintenance of the `FLUX_EXEC_PATH`...

**Comment by garlick** [comment]
> is there some reason we can't make a single exec-dir instead of one per framework project?

I don't remember how we got where we are but I see no reason why we couldn't 
rename `${libexec}/flux/core` to `${libexec}/flux/cmd` or similar and use it for all flux projects.

**Comment by garlick** [comment]
> rename `${libexec}/flux/core` to `${libexec}/flux/cmd`

Or maybe just drop the last directory, since subcommands are prefixed with `flux-`

**Comment by grondo** [comment]
> Or maybe just drop the last directory, since subcommands are prefixed with flux-

I slightly prefer keeping the subdir just so that `flux(1)` subcommands are kept separate from other things like utility scripts and daemons used internally (without having to worry about *not* naming them flux- something).  However, I only slightly prefer it. :-)

**Comment by garlick** [comment]
Looks like libforeign is built with different content in different places.  This is the union:
```
$(FLUX_BUILDDIR)/src/bindings/lua/json-lua.c \
$(FLUX_BUILDDIR)/src/common/liblsd/list.c \
$(FLUX_BUILDDIR)/src/common/libutil/log.c \
$(FLUX_BUILDDIR)/src/common/libutil/optparse.c \
$(FLUX_BUILDDIR)/src/common/libutil/xzmalloc.c
$(FLUX_BUILDDIR)/src/common/libutil/base64.c \
$(FLUX_BUILDDIR)/src/common/libutil/jsonutil.c \
$(FLUX_BUILDDIR)/src/common/libtap/tap.c \
```
Plus the header-only `shortjson.h` is included.

**Comment by dongahn** [comment]
Reading the README.md, I have to guess the scope of this PR is to support only a subset of make targets. `make`, `make check` in particular.

The committed README says `make install` isn't supported. It should also probably say other targets are not support including `make dist` and `make distcheck`

```
hype356{dahn}138: make dist
{ test ! -d "flux-sched-e12e6c2" || { find "flux-sched-e12e6c2" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -fr "flux-sched-e12e6c2"; }; }
test -d "flux-sched-e12e6c2" || mkdir "flux-sched-e12e6c2"
 (cd rdl && make  top_distdir=../flux-sched-e12e6c2 distdir=../flux-sched-e12e6c2/rdl \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[1]: Entering directory `/g/g0/dahn/stage/flux-sched/rdl'
cp: cannot stat `.//g/g0/dahn/local/build/src/bindings/lua/json-lua.c': No such file or directory
make[1]: *** [distdir] Error 1
make[1]: Leaving directory `/g/g0/dahn/stage/flux-sched/rdl'
make: *** [distdir] Error 1
```

Perhaps, the latest comments from @garlick should be moved to an issue ticket so that they can later be fixed?

**Comment by dongahn** [comment]
SLURM session section of the README file stated

>  Also, if you have run `make start` or `make load`, exit now.

This can be confusing as neither start nor load is a valid make target any longer. Perhaps this sentence should just drop?

**Comment by dongahn** [comment]
This seems to build an archive the static library *and* a shared library version of some of the components that are me anted to be dynamically loaded (e.g., scheduler plugins and scheduler framework service comms module). Probably better only to build the SO version for these components...

**Comment by garlick** [comment]
I would be OK with ignoring the install target in this PR, and would be willing to help fix it up in a follow on PR after this is merged.  It would probably be a good idea in this PR to:
* fix up the relative paths
* set correct comms module ldflags
* centralize the libforeign
to minimize churn in the commit log.

On libforeign:  I'd suggest maybe just adding a "libutil" in sched that is built from the union of the foreign sources, and then reference that from the various places that need it.

**Comment by dongahn** [comment]
@garlick: sounds good to me, and I came across @lipari in the hallway and he's already working on this. 

For other make targets, we will need to make them (in particular, install and dist - also DESTDIR support) a priority because we need them to test and build an fllux-sched rpm package: distribution issue [#5]( https://github.com/flux-framework/distribution/issues/5)

**Comment by garlick** [comment]
Sounds good @dongahn.  I'm also keen to get rid of the "builddir" hack in flux-core, and was planning to see what I could do to eliminate it in a near term PR to sched by pulling some bits in (trivial and/or not actively developed in core), and weaning sched off other bits that it makes minimal use of.

I think if the "libutil" solution is adopted here, then this can be done with minimal sched churn.

**Comment by dongahn** [comment]
@garlick: oh awesome. Yes, I thought builddir (currently mandatory option) was more of a stopgap. So if you can pull this off, IMO this will streamline things a lot.

**Comment by garlick** [comment]
Nice progress @lipari.

I think the travis failures are a result of tightening up the args for building installed shared libraries with `--no-undefined`.  When building those (e.g. libflux-resrc.so) you'll need to explicitly link all their deps including the transitive ones from any library built statically.  libtool will build all noinst libraries like libforeign.la statically.

**Comment by lipari** [comment]
Thanks, @garlick.  There will be a lot of attempts forthcoming.  Everything builds fine in my hype environment and I only learn of problems once travis attempts to build.  Stay tuned...

**Comment by lipari** [comment]
Latest push incorporates most of the feedback:
* README.md updates
* relative path replaced by ${top_srcdir} variables
* renamed library names with "flux-" prefixes
* gathered the separate foreign library builds into one
* adopted the flux-core fluxmod_ldflags for versioning, etc.
* started the conventions that will support make install

**Comment by garlick** [comment]
This looks pretty good!  I'll test against the rpm of flux-core-0.2.0 that I built for RHEL 7 in the morning.

**Comment by garlick** [comment]
Couple nits:
* flux-waitjob should be installed `${fluxcmddir}` not `${bindir}`
* resrc.h, resrc_tree.h, and simulator.h are installed, but use `JSON` typedef from shortjson.h, which is not installed
* libflux-rdl.so, libflux-resrc.so are installed into `${libdir}/flux` but should probably be installed into `${libdir}` like flux-core libs
* libflux-sim.so is installed into `${libdir}/flux/modules` but should be installed into `${libdir}`
*backfillplugin1.so, schedplugin1.so are still being versioned

**Comment by garlick** [comment]
Couple more:

* libforeign.la shouldn't need `$(LUA_LIB)` or any LDADDs for that matter as its a noinst built statically
* The `nodist_` prefix should be added to `libforeign_la_SOURCES` in foreign/Makefile.am

**Comment by garlick** [comment]
And (from running make distcheck with DISTCHECK_CONFIGURE_FLAGS set to the builddir option):
* cpuset-str.h is not included in distribution; add to `flux_cpuset_la_SOURCES` if it is not to be installed
* add $(LUA_LIB) to `libflux_sim_la_LIBADD`
* scheduler.h is not included in distribution; add to `fluxschedinclude_HEADERS` if it is to be installed
* add rs2rank.h, rsreader.h to `fluxschedinclude_HEADERS` if these are to be installed

That gets up to the point of tests failing with missing RDL.lua

**Comment by garlick** [comment]
Taking a step back, should the sched package be exporting public api's at this point?   Maybe all the headers and libs should be noinst?

**Comment by dongahn** [comment]
> add rs2rank.h, rsreader.h to fluxschedinclude_HEADERS if these are to be installed

Those are some of the internal interfaces to schedsrv and as such shouldn't be installed.

> Taking a step back, should the sched package be exporting public api's at this point? Maybe all the headers and libs should be noinst?

I think in generally most of the libs/headers are internal use. I can see why at one point we want to make resrc as a library so that this can be used by others as well but were are not there yet.

However, there are scheduling plugins that should be dl_opened by schedsrv so they should be installed. Even than, perhaps its own directory?

Also, all of the scheduler emulator comms modules need to be installed similarly to schedsrv.

**Comment by lipari** [comment]
New push with most of the above suggestions followed.  Except...
> libforeign.la shouldn't need $(LUA_LIB) or any LDADDs for that matter as its a noinst built statically

It builds fine on hype/upgrade, but travis complains:
../foreign/.libs/libforeign.a(json-lua.o): In function `lua_is_json_null':
/home/travis/local2/build/src/bindings/lua/json-lua.c:45: undefined reference to `lua_touserdata'
...
> The nodist_ prefix should be added to libforeign_la_SOURCES in foreign/Makefile.am 

This didn't fly:
foreign/Makefile.am:8: variable `noinst_libforeign_la_SOURCES' is defined but no program or
foreign/Makefile.am:8: library has `noinst_libforeign_la' as canonical name (possible typo)

The location of the sched products and which should be installed is still somewhat up in the air.  I don't think we need to decide today.

**Comment by grondo** [comment]
One other comment (if this hasn't been addressed already). I noticed (by being lazy) that `--with-flux-core-builddir` doesn't tolerate a relative path (causes build failures in libforeign). This could be fixed with
something like:
```diff
--- a/m4/ax_flux_core.m4
+++ b/m4/ax_flux_core.m4
@@ -41,7 +41,7 @@ AC_DEFUN([AX_FLUX_CORE], [
       [specify location of installed flux-core build directory])
     ],
     [PKG_CHECK_MODULES([FLUX_CORE], [flux-core],
-      [FLUX_BUILDDIR=${withval}
+      [FLUX_BUILDDIR=$(cd ${withval} && pwd)
        AC_SUBST(FLUX_BUILDDIR)
        AC_CHECK_PROG(FLUX,[flux],[flux])
        FLUX_PREFIX=`pkg-config --variable=prefix flux-core`
```

Though it might also be wise to test for existing of `${withval}` as a directory first.

Also, I was able to get `automake` to shut up about subdir objects with the following:
```diff
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ AC_CONFIG_SRCDIR([NEWS])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([m4])
-AM_INIT_AUTOMAKE
+AM_INIT_AUTOMAKE([subdir-objects])
 AM_SILENT_RULES([yes])
 
 #LT_PREREQ([2.4.2])
```

garlick:
> foreign/Makefile.am:8: variable noinst_libforeign_la_SOURCES' is defined but no program...

should be `nodist_`.  Without this make dist will fail.  Here's the little script I'm using to test this:
```
PREFIX=$HOME/local
export PKG_CONFIG_PATH=${PREFIX}/lib/pkgconfig
export BUILDDIR=${PREFIX}/lib/flux-core/build
CONFIG_BUILDDIR="--with-flux-core-builddir=$BUILDDIR"
./configure $CONFIG_BUILDDIR --prefix=$PREFIX
make distcheck DISTCHECK_CONFIGURE_FLAGS="$CONFIG_BUILDDIR"
```

garlick:
Same with references to shortjson.h (it is not included in the distribution and so needs to be referenced with `nodist_`)

I  was planning to recommend that we pull in shortjson.h in a future PR.  Maybe it would be easier to just do that now?

garlick:
Er, I misspoke on shortjson.h.  If you don't pull it in, it should not be referenced at all in any HEADERS or SOURCES in sched Makefile.am's.

garlick:
scheduler.h, rs2rank.h, and rsreader.h are still not included in the dist.  If they are not going to be installed then they need to be included in the _SOURCES for the scheduler/simulator modules that require them.

garlick:
In summary, this diff gets things built up to that RDL.lua error in make check that I mentioned before.
```diff
diff --git a/foreign/Makefile.am b/foreign/Makefile.am
index 06c18a2..7e32162 100644
--- a/foreign/Makefile.am
+++ b/foreign/Makefile.am
@@ -5,7 +5,7 @@ AM_CPPFLAGS = $(JSON_CFLAGS) $(LUA_INCLUDE) $(CZMQ_CFLAGS) \

 noinst_LTLIBRARIES = libforeign.la

-libforeign_la_SOURCES = \
+nodist_libforeign_la_SOURCES = \
     $(FLUX_BUILDDIR)/src/bindings/lua/json-lua.c \
     $(FLUX_BUILDDIR)/src/common/liblsd/list.c \
     $(FLUX_BUILDDIR)/src/common/libtap/tap.c \
@@ -14,4 +14,4 @@ libforeign_la_SOURCES = \
     $(FLUX_BUILDDIR)/src/common/libutil/log.c \
     $(FLUX_BUILDDIR)/src/common/libutil/optparse.c \
     $(FLUX_BUILDDIR)/src/common/libutil/xzmalloc.c
-libforeign_la_LIBADD = $(LUA_LIB)
\ No newline at end of file
+libforeign_la_LIBADD = $(LUA_LIB)
diff --git a/resrc/Makefile.am b/resrc/Makefile.am
index 27c2b89..8314f56 100644
--- a/resrc/Makefile.am
+++ b/resrc/Makefile.am
@@ -7,8 +7,7 @@ SUBDIRS = . test

 lib_LTLIBRARIES = libflux-resrc.la

-fluxschedinclude_HEADERS = resrc.h resrc_tree.h resrc_reqst.h \
-    $(FLUX_BUILDDIR)/src/common/libutil/shortjson.h
+fluxschedinclude_HEADERS = resrc.h resrc_tree.h resrc_reqst.h

 libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_reqst.c
 libflux_resrc_la_CFLAGS = -I$(top_srcdir)/rdl
diff --git a/sched/Makefile.am b/sched/Makefile.am
index 5e2f7b3..7c828bb 100644
--- a/sched/Makefile.am
+++ b/sched/Makefile.am
@@ -12,7 +12,7 @@ fluxmod_LTLIBRARIES = schedsrv.la \
                   schedplugin1.la \
                   backfillplugin1.la

-fluxschedinclude_HEADERS = schedsrv.h
+fluxschedinclude_HEADERS = schedsrv.h rs2rank.h rsreader.h

 schedsrv_la_SOURCES = schedsrv.c rs2rank.c  rsreader.c
 schedsrv_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/resrc
diff --git a/simulator/Makefile.am b/simulator/Makefile.am
index 511b4c2..b7d23b8 100644
--- a/simulator/Makefile.am
+++ b/simulator/Makefile.am
@@ -12,7 +12,7 @@ lib_LTLIBRARIES = libflux-sim.la
 fluxmod_LTLIBRARIES = simsrv.la sim_execsrv.la submitsrv.la sim_sched_fcfs.la \
     sim_sched_fcfs_aware.la sim_sched_easy.la

-fluxschedinclude_HEADERS = simulator.h
+fluxschedinclude_HEADERS = simulator.h scheduler.h

 libflux_sim_la_SOURCES = simulator.c
 libflux_sim_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
```

garlick:
Maybe add `noinst_SCRIPTS = RDL.lua` to rdl/Makefile.am?

Something screwy is going on, maybe a quoting problem in a couple places?
```
$ find . -name \$*
./foreign/$(FLUX_BUILDDIR)
./simulator/test/$(top_srcdir)
```

lipari:
> Also, I was able to get automake to shut up about subdir objects with the following:
> +AM_INIT_AUTOMAKE([subdir-objects])

While that does squelch the automake complaints, it generates a fatal failure later on:
```
Making all in foreign
make[2]: Entering directory `/g/g0/lipari/flux-sched/foreign'
Makefile:454: /g/g0/lipari/local/build/src/bindings/lua/.deps/json-lua.Plo: No such file or directory
Makefile:455: /g/g0/lipari/local/build/src/common/liblsd/.deps/list.Plo: No such file or directory
Makefile:456: /g/g0/lipari/local/build/src/common/libtap/.deps/tap.Plo: No such file or directory
```
I would have had to build flux-core and copy all its products to $HOME/local/build instead of just the source files.

lipari:
> Something screwy is going on, maybe a quoting problem in a couple places?
> $ find . -name \$*
> ./foreign/$(FLUX_BUILDDIR)
> ./simulator/test/$(top_srcdir)

That problem went away with latest push.

grondo:
> While that does squelch the automake complaints, it generates a fatal failure later on:

Oops, sorry. I think you told me that before and I forgot! I'll poke at that one some more perhaps.

lipari:
@grondo, where do think RDL.lua should go?  In `luadir = ${prefix}/share/lua/5.1` ?

grondo:
If installed flux-sched components still use the `RDL.lua` interface, then yes `RDL.lua` should go into luadir for now. (Might be better to eventually move this stuff to `flux/RDL.lua` in the future, though)
So you can list that one file under `dist_lua_SCRIPTS` I think.

Also, everything under `RDL/*` also should go into `$luadir/RDL/*` (though I'm sure you already realized that).

You might have to define a `rdlluadir = $(luadir)/RDL` Then liat all RDL lua files as `nobase_dist_rdllua_SCRIPTS =`. (See src/bindings/lua/Makefile.am for examples). Sorry I don't know a better way, if you figure it out let me know!

grondo:
> > Also, I was able to get automake to shut up about subdir objects with the following:
> > +AM_INIT_AUTOMAKE([subdir-objects])
> 
> While that does squelch the automake complaints, it generates a fatal failure later on:

Ugh. We may be hitting the bug described [here](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=13928), which is triggered by using a `$(var)` in `_SOURCES` line as we're doing to point to the `$(FLUX_ BUILDDIR)` (Shouldn't this actually be $(FLUX_SRCDIR) since we're building from sources, not built objects?)

If the patches proposed below go into automake before `subdir-objects` becomes the default we should be ok, though we have to put up with the warnings. Other alternatives might be to copy in the sources required, or the link to them (though I could forsee other problems with those workarounds).

https://lists.gnu.org/archive/html/automake-patches/2015-01/msg00010.html

garlick:
Are we wasting too much time on this `$(FLUX_BUILDDIR)` thing given
* we are talking about a pretty small amount of code
* I had promised to follow this PR with one that trims this down further and pulls in the trivial/not maintained in core stuff?

What if @lipari just copies in what he needs in this PR and we trim it down in the next?

grondo:
Yeah, @garlick, that is a good proposal at this point actually.

I wasn't necessarily focusing on `$(FLUX_BUILDDIR)`, but we want to make sure `subdir-objects` works since `automake` claims it will at some point be the default. The use of `$(FLUX_BUILDDIR)` in `_SOURCES` just happened to be the cause of the problem.

lipari:
Latest push addresses distribution of lua components per above discussion.  I don't have anything more planned to add at this point.

dongahn:
time to merge? @garlick or @grondo, you've been following this more closely than me. If you ack, i will push the button.

garlick:
I can't figure out what is screwed up in my environment, but I can't seem to get past make check with this.  I'm bombing out here:
```
PASS: t0001-basic.t
flux: flux_open: No such file or directory
flux: flux_open: No such file or directory
/usr/bin/lua: /home/garlick/local/share/lua/5.1/fluxometer.lua:58: module 'flux.alt_getopt' not found:
        no field package.preload['flux.alt_getopt']
        no file '/home/garlick/local/share/lua/5.1/fluxometer/flux/alt_getopt.lua'
stack traceback:
        [C]: in function 'require'
        /home/garlick/local/share/lua/5.1/fluxometer.lua:58: in main chunk
        [C]: in function 'require'
        ./lua/t0001-rdl-basic.t:5: in main chunk
        [C]: ?
FAIL: lua/t0001-rdl-basic.t
flux: flux_open: No such file or directory
flux: flux_open: No such file or directory
```
after that, more of the same....

grondo:
@garlick looks like `flux/alt_getopt.lua` didn't get installed in your environment? Let me peek at this from my end, since this looks like an installed `fluxometer.lua` fail...

garlick:
Seems to be though maybe something is wrong with the configured lua paths...
``` 
$ find $HOME/local -name alt_getopt.lua
/home/garlick/local/share/lua/5.1/flux/alt_getopt.lua
```

grondo:
@garlick, can you try something like 
```diff
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -9,8 +9,7 @@ T_LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) \
 #
 TESTS_ENVIRONMENT = \
     LUA_PATH="$(abs_top_srcdir)/rdl/?.lua;$(FLUX_PREFIX)/share/lua/5.1/?.lua;$(LUA_PATH);;" \
-    LUA_CPATH="$(abs_top_builddir)/rdl/.libs/?.so;$(FLUX_PREFIX)/lib64/lua/5.1/?.so;$(LUA_CPATH);;" \
-    FLUXOMETER_LUA_PATH="$(FLUX_PREFIX)/share/lua/5.1/?.lua"
+    LUA_CPATH="$(abs_top_builddir)/rdl/.libs/?.so;$(FLUX_PREFIX)/lib64/lua/5.1/?.so;$(LUA_CPATH);;"
```

`FLUXOMETER_LUA_PATH` is really only meant to be used internally to the `fluxometer.lua` code (since we re-exec the test script in `fluxometer.lua`, we have to make sure we pick up the same fluxometer.lua and conf.lua that we started with)

With this change I was able to get tests to run after ensuring I ran `flux keygen` with installed version of flux...

garlick:
Thanks @grondo that fixed it.  (@lipari please pick up that change so sched can work with installed flux-core)

garlick:
OK, make distcheck fails here:
```
../../t/t0002-waitjob.t: line 8: ../../t/sharness.sh: No such file or directory
FAIL: t0002-waitjob.t
../../t/t1000-jsc.t: line 8: ../../t/sharness.sh: No such file or directory
FAIL: t1000-jsc.t
../../t/t1001-rs2rank-basic.t: line 9: ../../t/sharness.sh: No such file or directory
```
Looks like sharness.sh should be added to t/Makefile.am EXTRA_DIST.

grondo:
> Looks like sharness.sh should be added to t/Makefile.am EXTRA_DIST.
￼
Also add `sharness.d` which should copy all contents of that dir into the distribution

grondo:
Oops that one is already there, sorry!

garlick:
Ah, this is a bit of a problem.
```
#
# variables
#
dn=`dirname $0`
tdir=`readlink -e $dn/../`
schedsrv=`readlink -e $dn/../sched/.libs/schedsrv.so`
rdlconf=`readlink -e $dn/../conf/hype.lua`
```
Those paths should be relative to $FLUX_BUILD_DIR and $FLUX_SOURCE_DIR.  Hmm I'm not sure sharness is setting those in sched...

grondo:
> Those paths should be relative to $FLUX_BUILD_DIR and $FLUX_SOURCE_DIR. Hmm I'm not sure sharness is setting those in sched...

But it is relative to `dirname $0`, which should be directory of the tests or `top_srcdir/t`

However, I can't get VPATH build to work (I assume I'm seeing the same issue). I'll try to take a better look at it.

lipari:
Latest push removes FLUXOMETER_LUA_PATH from t/Makefile.am and adds missing sharness.sh to t/Makefile.am EXTRA_DIST

grondo:
> However, I can't get VPATH build to work 

I should say `make check` in VPATH build doesn't work..

lipari:
I tried to fix the flux-sched/t tests so that they could be run manually (`./t0001-basic.t`) in addition to `make check`.  Hence I tried to avoid relying on the variables in the Makefile.  I see some of the tests leverage ${SHARNESS_TEST_SRCDIR} and maybe that should replace all the `..`'s

grondo:
> I tried to fix the flux-sched/t tests so that they could be run manually
> Hence I tried to avoid relying on the variables in the Makefile

Oh yeah, that is perfect and is probably how sharness-driven tests should work.

I can't get the tests to run this way though as they don't find the installed flux (which I installed out of PATH). Are there instructions for running this way?

lipari:
> I can't get the tests to run this way though as they don't find the installed flux (which I installed out of PATH). Are there instructions for running this way?

I'm afraid you are pushing the envelope.  I was planing to break this work into two efforts:  build and installing/packaging.  I have yet to battle test the install portion

dongahn:
Isn't ${SHARNESS_TEST_SRCDIR} set when a test source in sharness.sh though and most tests source it before setting those shell variables?

lipari:
Yes.  ${SHARNESS_TEST_SRCDIR} is set in sharness.sh.  However, many of the tests don't use it and just look for needed objects using relative paths.

dongahn:
ah. sorry i misread your comment.

grondo:
@lipari, I meant tests don't find the installed `flux(1)` when I run `./t*.t`. Then I did find that if you have PKG_CONFIG_PATH set correctly it works (iff LUA_PATH LUA_CPATH are set)

lipari:
> I can't get the tests to run this way though as they don't find the installed flux (which I installed out of PATH). Are there instructions for running this way?

I can run these tests individually after setting these variables:
```
export LUA_PATH="${HOME}/flux-sched/rdl/?.lua;${HOME}/local/share/lua/5.1/?.lua;;"
export LUA_CPATH="${HOME}/flux-sched/rdl/.libs/?.so;${HOME}/local/lib64/lua/5.1/?.so;;"
```

lipari:
@grondo. Sorry, I should have mentioned that:  setting PKG_CONFIG_PATH is critical.

grondo:
Yeah, that's fair.

However, I wonder (in general for framework project tests) if you find the `flux(1)` you are testing against with `pkg-config` we could grab the correct LUA_PATH, LUA_CPATH, etc via the output of `/path/to/flux env`?

grondo:
I should say I meant the above as future work, not something to fix right now.

lipari:
> I wonder (in general for framework project tests) if you find the flux(1) you are testing against with pkg-config we could grab the correct LUA_PATH, LUA_CPATH, etc via the output of /path/to/flux env?
> I should say I meant the above as future work, not something to fix right now.

Yes.  That's a good idea.

grondo:
Heh, actually something like this might work:
```
$ PATH=/tmp/flux/bin:$PATH
PKG_CONFIG_PATH=/tmp/flux/lib/pkgconfig flux ./t0001-basic.t
```

lipari:
Yes, (after unsetting LUA_PATH and LUA_CPATH) that does work!

grondo:
Ok, I do see places that reference `$SHARNESS_TEST_SRCDIR` which should reference `$SHARNESS_TEST_BUILDDIR`. `make distcheck` will fail until this is resolved. I can work on a suggested patch a bit later this evening.

garlick:
>  if you find the flux(1) you are testing against with pkg-config

How does that work?

grondo:
> How does that work?

Currently the testsuite uses `pkg-config` to grab the `$prefix` of the flux-core installation.

I'm currently moving some stuff around so that there is not as much divergence between `sharness.d/flux-sharness.sh` from flux-core and flux-sched. (per-project modifications should go into the project specific sharness files, `sched-sharness.sh` in this case).

I thought during this update I'd simplify the way the flux-sched tests work and just use the `flux(1)` found in `PATH`. The only drawback is you might use inadvertently use a different `flux` than you built with if `PKG_CONFIG_PATH` is set to a different flux than PATH.

However, it would nicely allow you to run tests by hand with *any* flux via `/path/to/my/flux ./t...`

Does that sound ok? I'll propose something later today on top of this PR.

garlick:
Sounds reasonable to me.  

So `$(pkg-config --variable=exec_prefix)/bin` then?

Since one could configure flux-core with `--bindir` set to something differen, maybe we should add a `bindir=@bindir@` to flux-core.pc.in and use that?

grondo:
I was saying my proposed changes would avoid using pkg-config to try to find the `flux(1)` binary. The tests would just use the first one found in `PATH` by default. The `t/Makefile` would ensure `make check` uses the correct `flux(1)` by setting `PATH` appropriately if necessary.

garlick:
Ooops, right, OK sounds good.

grondo:
I wonder if your idea to add `bindir` to `flux-core.pc.in` would still be valuable, though. In fact, perhaps adding other variable directories somehow would be a good idea. (e.g. the `fluxcmddir` would be useful for framework projects to be able to query, since they have to install their subcmds there...)

grondo:
Ok, I'm working on a redo of the flux-sched testsuite on top of this PR. If you're interested the work is on grondo/flux-sched "testsuite" branch [here](https://github.com/grondo/flux-sched/commits/testsuite)

This branch also requires flux-framework/flux-core#575 be merged.

The basic changes are:

 * consolidate flux-sched specific environment and other setup to `sharness.d/sched-sharness.sh`
 * Copy back "upstream" `flux-sharness.sh` from flux-core. 
 * set up environment such that explicit paths to modules and commands don't need to be used in tests
 * move duplicate code from tests into `sched-sharness.sh`
 * Fix up and use srcdir vs builddir so tests work in VPATH build
 * remove `set_instance_size()` requirement for tests (instead get size from `flux getattr size`)
 * Testsuite now runs via the first `flux(1)` binary found in path. This means running tests by hand can be done without setting `PKG_CONFIG_PATH` or `LUA*` 
 * `make check` sets `PATH` appropriately such that the `flux(1)` binary used for the build is used during testing.

lipari:
Nice work!  I'm glad to see that you pulled in flux-sharness.sh from flux-core.  That will save time in the long run for anyone having to study the differences and determine whether they were relevant.  I also appreciate that you consolidated the duplicate code into sched-sharness.sh.

I was playing with t1001-rs2rank-basic.t on Friday and could not reproduce the sporadic failures I had seen last week.  Hence, I was planning on removing the TRAVISHAPPY null variables - effectively reverting b23ef6f1a3.  I don't know how best to handle this from a git perspective.  I could certainly force push a new branch with that commit missing, but I think that would orphan your "testsuite" branch.

Option B is to just merge the autotools_conversion + testsuite branch as is, and create a new commit that removes the TRAVISHAPPY shunts.

Otherwise, good to go AFAIAC.

dongahn:
@lipari: if you force push the change, @grondo can simply rebase to it?

If that can cause conflict issues, the option B also sounds reasonable.

lipari:
> @lipari: if you force push the change, @grondo can simply rebase to it?

@grondo , are you ready for option A?

grondo:
@lipari, whatever is easiest for you!

 I'm fine with you pushing the changes, I'll just rebase.
(I actually only put the test_set_prereq in there so I could test my changes)

grondo:
Then I'd propose that you just fast-forward merge my branch onto this PR and push, to make sure everything checks out...

lipari:
Option A pushed.  Alas, the marginality of t1001-rs2rank-basic.t has returned.  @grondo proceed with your rebase.  I'd rather troubleshoot the new version of the test than the old one.

grondo:
Ok, I rebased on your latest. I added a commit on top that disables the entire `rs2rank-basic` test unless `TRAVISHAPPY=t` is set in the environment. We can drop that commit once we figure out any problem with the test.

lipari:
@grondo, thank you for the assist.  Rebased to your testsuite and re-pushed - and build succeeded.

dongahn:
Time to merge?

grondo:
`make distcheck` still appears to be failing this AM. I think all that needs to be added to `EXTRA_DIST` is the `m4` directory.

lipari:
> make distcheck still appears to be failing this AM. I think all that needs to be added to EXTRA_DIST is the m4 directory.

So done, and pushed.  Though the following was successful for me on hype before I added m4 to EXTRA_DIST:
```
make distcheck DISTCHECK_CONFIGURE_FLAGS="--with-flux-core-builddir=/g/g0/lipari/local/build"
```

grondo:
> Though the following was successful for me on hype before I added m4 to EXTRA_DIST:

Hm, does it make sense that the `m4` directory would be required for the dist tarball? It seemed to be required on my ubuntu system or else `tap-driver.sh` was missing. Maybe it would only be required on newer automake?

lipari:
Comparing to flux-core, only config/tap-driver.sh is included in EXTRA_DIST.

grondo:
Yeah, I guess autotools would already know about the other stuff in m4

lipari:
Re-pushed with just m4/tap-driver.sh in EXTRA_DIST.

garlick:
distcheck is working for me with the latest flux-core master installed into ~/local.  (It is not working with the installed flux-core-0.2.0-2 on toss2 but perhaps that is expected?)

Here are a couple of install related issues:
* libs and includes for non-published APIs should probably be "noinst_"
* sched modules that are not comms modules (do not follow RFC 5) should not be installed into `$libdir/flux/modules` where `flux module` searches.   They should probably have their own directory like $libdir/flux/sched-plugin`.

Otherwise seems pretty good!

dongahn:
> sched modules that are not comms modules (do not follow RFC 5) 

Assuming you mean the scheduling plugins (IOW, not the `schedsrv` comms module itself), yes they are not intended to be comms modules. I agree it would be good if they are installed in a separate directory. 

I know @lipari previously wanted to tackle `install` make target as a separate PR. @lipari, do you still want to do it that way or do you want to address it in this PR?

lipari:
I'm ok with adjusting the install location of the sched plugin location.  But @garlick 's `noinst_` recommendation leaves me guessing.  At this point, the whole project is a candidate to be installed and I have not thought about which components should be withheld.

garlick:
Couple of quick thoughts (my $0.02):
* at this stage, install what's required to _run_ the scheduler (and its plugins) from an installed package
* if API's are not stable/documented yet, and you expect collaborators to work within the flux-sched project for now, then don't install headers or API libraries
* it's will be less churn/noise in the git history to fix this stuff now and rebase than to merge and then fix.

grondo:
> It is not working with the installed flux-core-0.2.0-2 on toss2 but perhaps that is expected?

Yeah, `make check` for flux-sched depends on the `FLUX_*_PREPEND` environment variables which didn't make it into 0.2.0.

lipari:
@garlick, W/r/t the include files, are you reconsidering your [comment above](https://github.com/flux-framework/flux-sched/pull/115#issuecomment-184865294)?  I see this as a work in progress with ultimate locations still in flux, so to speak.  Right now, everything is included and this can be pared down in due time.  Do you have specific changes you'd definitely like to see?  I will have a re-push with the relocated plugins shortly.

garlick:
> @garlick, W/r/t the include files, are you reconsidering your comment above?

Well...that was about the install directories, not the question of whether to install.  I  think it was @dongahn that (later?) suggested that these shouldn't be installed.

Whatever you guys decide - I'm just advocating for getting the install paths stable in this PR rather than doing it a couple days later in another one that will pollute the history and potentially conflict with fixups such as the one I plan to work on to drop `with-flux-builddir`.

lipari:
Latest push has sched plugins relocated to `$libdir/flux/sched-plugin`

dongahn:
Merging this after I talked to @garlick and @lipari. Thank you @lipari,  this represents a huge effort. Thank you @garlick and @grondo to help advance this flux-sched PR to this level!

PR #117, opened 2016-02-02
Improved LONGTEST usage in testsuite

grondo:
I was getting OOMs and EAGAIN errors from `fork` on a small VM when trying to run `make check` in flux-sched, due to the 64 rank flux session being started in `t1002-rs2rank-64ranks.t`, even when LONGTEST is not set. 

It seemed the intent here (and in `t1003-stress.t`) was to skip all tests if `--long` or `TEST_LONG` is not set, so, in addition to setting `LONGTEST` prereq in one place, this PR puts the stanza:
```sh
if ! test_have_prereq LONGTEST; then
    skip_all='LONGTEST not set, skipping...'
    test_done
 fi
```

At the top of these functions, which results in skipping these entire files, thus avoiding unnecessary launch of a flux session.

### DIFF
--- a/t/sharness.d/sched-sharness.sh
+++ b/t/sharness.d/sched-sharness.sh
@@ -9,6 +9,10 @@ sched_start_jobid=0
 sched_end_jobid=0
 sched_test_dir=""
 
+if test "$TEST_LONG" = "t"; then
+    test_set_prereq LONGTEST
+fi
+
 #
 # Internal subroutines
 #
--- a/t/t1002-rs2rank-64ranks.t
+++ b/t/t1002-rs2rank-64ranks.t
@@ -8,8 +8,9 @@ the nodes allocated to a job.
 '
 . `dirname $0`/sharness.sh
 
-if test "$TEST_LONG" = "t"; then
-    test_set_prereq LONGTEST
+if ! test_have_prereq LONGTEST; then
+    skip_all='LONGTEST not set, skipping...'
+    test_done
 fi
 
 tdir=`readlink -e ${SHARNESS_TEST_SRCDIR}/../`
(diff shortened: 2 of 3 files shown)


dongahn:
Nice refactoring! I will merge this when Travis gives us the green light. 

At some point, I would like to discuss ways not do "forks"  so that we can test the rest of the system using a small fraction of compute nodes (without having to hit resource limit issues like you are having.

grondo:
I will rebase on master once @liparis latest fixes go in (I don't think travis will work until then)

grondo:
> At some point, I would like to discuss ways not do "forks" so that we can test the rest of the system using a small fraction of compute nodes (without having to hit resource limit issues like you are having.

Cool, do we have an issue open already? I'm not really following about not doing "fork", but I'm sure there is some good discussion here.

dongahn:
@grondo: I opened an issue before and then closed it for now: flux-core [#502](https://github.com/flux-framework/flux-core/issues/502)

As Tapasya and I will start to work on critical path analysis instrumentation, we will know which components we are most interested in observing, and based on that, we may be able to suggest some logic to be emulated. 

For example, for "fork" emulation. I was thinking, if I pass an "emulation-test" option, wreck would do everything except for "process forking" (or even wreckd forking) so that we don't have to be bogged down by various resource limits. The main goal would be to run as many brokers as possible on a node and get some realistic scaling data. 

I should be much more well versed once we do CPA work.

dongahn:
For my own note, the order of work
1) @lipari xasprintf
2) me merging 1
3) @grondo rebase
4) me merging 3

dongahn:
@gronodo: I just merged 1. please rebase and push.

grondo:
Thanks, @dongahn. I've rebased and pushed, but I had trouble testing on my platform tonight, so wait for travis to report back to be sure.

PR #126, opened 2016-02-24
Add code-coverage target for travis, other travis-ci cleanup

grondo:
This PR adds the `./configure --enable-code-coverage` support to flux-sched and uses this to activate coveralls.io code coverage tracking via travis.

Additionally, `gcc` and `clang` travis targets now use `make distcheck` instead of just `make check`.

Still working on a `make install` test if we'd like that, but might need to tweak sharness testing to load the installed sched instead of in-tree sched modules, utils, etc. Ok to merge this first if you want.


grondo:
Oh, I also forgot to mention this includes the update to use `subdir-options` flag for automake.

lipari:
@dongahn, I have reviewed the changes in this PR and it looks good to me.  You're free to merge when you complete your review.

PR #127, opened 2016-02-25
Issue 122

lipari:
Address the issues described in Issue #122.

Wired in the use of existing version.map files and added two such files where needed.
Without the leakage, some more libutil's needed to be added to various LIBADD and LDADD definitions.

PR #128, opened 2016-02-25
schedsrv: fix build_hwloc_rs2rank synchronization

garlick:
This allows sched to be loaded on the broker command line, without arguments (arguments are a bit of a problem for size > 1 as discussed in flux-framework/flux-core#584)

It replaces `kvs_get_string()` with `kvs_watch_once()` with initial value of NULL, so that the response from the KVS is delayed until the value is non-NULL, as may be the case if resource-hwloc has not completed its initialization on all ranks.

This also fixes a leak of the `kvs_get_string()` result on all but the last iteration.

### DIFF
--- a/sched/schedsrv.c
+++ b/sched/schedsrv.c
@@ -502,40 +502,67 @@ static void setup_rdl_lua (flux_t h)
     flux_log (h, LOG_DEBUG, "LUA_CPATH %s", getenv ("LUA_CPATH"));
 }
 
+/* Block until value of 'key' becomes non-NULL.
+ * It is an EPROTO error if value is type other than json_type_string.
+ * On success returns value, otherwise NULL with errno set.
+ */
+static json_object *get_string_blocking (flux_t h, const char *key)
+{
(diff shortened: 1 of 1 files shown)


garlick:
oops, rebasing on current master.  Hmm, coveralls seems to be unhappy.

dongahn:
This looks good. When does coveralls declare a "failure?" When the coverage ratio simply goes down?

dongahn:
Maybe, you had a very small amount of increase in the exception branches... Does it really make sense to make coverall this sensitive? I wonder if they have some threshold control...

grondo:
Oh, sorry my fault you have to manually set the coveralls acceptance. Don't
worry about it for now I will fix this evening.

dongahn:
I see. :-) But the coverage check is COOL! I will merge @lipari's change first. And if/when @garlick rebase, i will merge this in.

dongahn:
I just merged PR #127.

garlick:
Just rebased on current master.

dongahn:
Great. Thanks.

PR #129, opened 2016-02-25
Plug leaks

lipari:
This commits some memory leak fixes from @SteVwonder submitted a number of weeks ago.  The code has undergone so many mods since the common base that it was easier to just cherry pick the relevant changes that remained.


SteVwonder:
Looks good!  Thanks for pulling these in.

PR #130, opened 2016-02-25
update waitfile.lua

lipari:
Copies over the latest waitfile.lua from flux-core with changes that
appeared in PR https://github.com/flux-framework/flux-core/pull/580


dongahn:
Thanks @lipari!

PR #133, opened 2016-02-26
travis-ci: add install target for testing

grondo:
This PR adds support for an "install" target for travis-ci testing. This target builds flux-sched with `--prefix=$HOME/local2` (same prefix as flux-core), does `make install`, and then runs `make check` against the *installed* flux-sched components via a newly supported `FLUX_SCHED_TEST_INSTALLED` env var.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ compiler:
   - gcc
   - clang
   - gcc+coverage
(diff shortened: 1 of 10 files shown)


lipari:
Thanks for doing this, @grondo!

PR #139, opened 2016-03-09
testsuite: fix non-POSIX warning from automake

### LINKED ISSUE #136: t/Makefile.am:17: if $(FLUX_SCHED_TEST_INSTALLED: non-POSIX variable name
I'm getting this warning from autoreconf on current master 23ca224e5cbb2b809bbe59be71b8c5ffca3eea6e
```
$ ./autogen.sh
Running autoreconf -i ... 
t/Makefile.am:17: if $(FLUX_SCHED_TEST_INSTALLED: non-POSIX variable name
t/Makefile.am:17: (probably a GNU make extension)
Cleaning up ...
Now run ./configure.

grondo:
Fix use of non-POSIX $(if ...) usage in t/Makefile.am to keep automake
from complaining. The replacement uses POSIX variable substitution to
substitute the build-tree LUA_PATH and LUA_CPATH to the PREPEND variables
iff FLUX_SCHED_TEST_INSTALLED *is not* set.

When FLUX_SCHED_TEST_INSTALLED *is* set, then the value of that variable
(`t` or maybe `1`) is substituted, which still isn't quite right, but at least
the build-tree Lua modules will not be used. Unfortunately, this was the
best that we could do at this time.

If this isn't acceptable, then we can take some other approach like discussed in #136,
or we could add the `-Wno-portability` flag to automake, since we probably don't support
non-GNU systems anyway. (However, it might be useful to keep portability warnings for
the rest of the build files)

Fixes #136

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -14,12 +14,16 @@ T_LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) \
 #  $FLUX_SCHED_TEST_INSTALLED is set in the current environment,
 #  skip the export of the PREPEND variables.
 #
+#  (XXX: the only way I've found to portably do this is to substitute
+#   the required path iff FLUX_SCHED_TEST_INSTALLED is not set, otherwise we
+#   substitute the *value* of FLUX_SCHED_TEST_INSTALLED, which might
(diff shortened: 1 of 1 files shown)


lipari:
Good by me.

lipari:
Close #136

Issue #144, opened 2016-03-15
support dynamic sched plugin loading


garlick:
Just wanted to float this idea -

If sched implements the `lsmod` request handler as discussed in RFC 5, `flux module list sched` can be used to list its loaded plugins.

Similarly, if `rmmod` and `insmod` handlers are implemented, `flux module remove sched.fcfs` and `flux module load sched.backfill` (say) could work.   Of course then sched has to handle "plugin not loaded" sensibly  (like pausing/restarting scheduling without losing the queue).

I'm not sure how sched guys fell about this or how much work would be involved, but it would be kind of neat to be able to swap out the scheduling algorithm on the fly.  I'd be happy to supply a PR to implement the stubbed out handlers if this is of interest.

lipari:
Sure.  That sounds like a nice enhancement.  I'll take you up on your offer to implement the stubbed out handlers.
Will we be able to preserve the automatic default loading behavior?  Right now, you get the sched_fcfs by default if you don't specify a plugin.

garlick:
OK, I'll get that started.  I'll just stub the handlers and you can decide what to do with the requests and the default behavior.  My suggestion would be to not wire a default into sched, but instead have a default dynamically loaded in the sched rc1 script.  (Seems like it would be as easy to support "no plugin at initialization" as "no plugin after rmmod")

lipari:
>  My suggestion would be to not wire a default into sched, but instead have a default dynamically loaded in the sched rc1 script.

Sounds good.

garlick:
Looks like this is complete?  Closing.

PR #147, opened 2016-03-31
Test for Dynamic Sched Plugin Loading

lipari:
Miscellaneous follow-up to the work done to address Issue #144.  Includes a plugin replacement test.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -1187,7 +1187,8 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
     resrc_tree_list_t *selected_trees = NULL;
     struct sched_plugin *plugin = sched_plugin_get (ctx->loader);
 
-    assert (plugin != NULL); /* XXX FIXME: handle this case */
+    if (!plugin)
+        return rc;
 
     /*
      * Require at least one task per node, and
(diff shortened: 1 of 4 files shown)


sched/sched.c
```diff
@@ -1187,7 +1187,8 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
     resrc_tree_list_t *selected_trees = NULL;
     struct sched_plugin *plugin = sched_plugin_get (ctx->loader);
 
-    assert (plugin != NULL); /* XXX FIXME: handle this case */
+    if (!plugin)
```
dongahn: Thanks Don. 

Can this check be placed earlier? If plugin is null, I don't see why we bother doing extra logic before reaching this point.

dongahn:
Also, it seems we may also need to firm up the upper level call (action) *at some point*. As is, the action fails on bad state transitions, but not on an error code returning form the down calls like schedule_jobs.

sched/sched.c
lipari: When you say, "earlier", do you mean before schedule_job() is called?  The check is already at the start of schedule_job().

sched/sched.c
```diff
@@ -1293,7 +1294,8 @@ static int schedule_jobs (ssrvctx_t *ctx)
         (int64_t) ctx->sctx.sim_state->sim_time : epochtime();
 
     resrc_tree_release_all_reservations (resrc_phys_tree (ctx->rctx.root_resrc));
-    assert (plugin != NULL); /* XXX FIXME: handle this case */
+    if (!plugin)
```
dongahn: Oh I'm sorry. I was looking at this statement.

lipari: The placement of the plugin check is intended.  The resrc_tree_release_all_reservations() call will remove any resource reservations.  This can happen even if there is no plugin loaded.  The code would work fine if the plugin check was moved earlier, I wanted to be super cautious and cancel those resource reservations w/o delay.  If we ever add some sort of syncing of the resources in memory to a persistent storage, this could be a worthwhile effort.

dongahn: Thank you for the explanation. I'm OK with this then.

dongahn:
Looks good!

PR #148, opened 2016-03-31
Issue 143:  eliminate shortjson.h dependency from installed headers

lipari:
Moved shortjson.h #includes from header files to source files

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,17 +53,14 @@ export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig:$PKG_CONFIG_PATH
 git clone <flux-sched repo of your choice>
 cd ~/flux-sched
 ./autogen.sh
-./configure
+./configure --prefix=$HOME/local
 make
 make check
+make install
 ```
 
-`make install`, `make dist`, and `make distcheck` are not yet
-supported in flux-sched.  When you run make in ~/flux-sched, the
-products remain local; they do not populate anything in ~/flux-core.
-
-To actually exercise a functioning sched module in a comms session,
-follow these steps.
+To exercise a functioning sched module in a comms session, follow
+these steps.
 
 ##### Flux comms session
 
@@ -84,7 +81,7 @@ $HOME/local/bin/flux start -s3
 cd to ~/flux-sched and load the sched module specifying the
 appropriate rdl configuration file:
 ```
-flux module load sched rdl-conf=$HOME/flux-sched/conf/hype.lua
+flux module load sched rdl-conf=$HOME/flux-sched/conf/hype.lua plugin=sched.fcfs
 ```
 
 Check to see whether the sched module loaded:
--- a/rdl/flux-rdltool.c
+++ b/rdl/flux-rdltool.c
@@ -32,6 +32,7 @@
 #include <stdarg.h>
 #include <string.h>
 #include <getopt.h>
+#include <json.h>
 
 #include "rdl.h"
 
--- a/rdl/rdl.c
+++ b/rdl/rdl.c
@@ -31,6 +31,7 @@
 #include <assert.h>
 #include <dlfcn.h>
 
+#include <json.h>
 #include <lua.h>
 #include <lualib.h>
 #include <lauxlib.h>
--- a/rdl/rdl.h
+++ b/rdl/rdl.h
@@ -7,7 +7,6 @@
 
 #include <stdlib.h>
 #include <stdbool.h>
-#include <json.h>
 
 /*
  *  Forward declarations:
--- a/resrc/Makefile.am
+++ b/resrc/Makefile.am
@@ -6,9 +6,9 @@ AM_CPPFLAGS = -I$(top_srcdir) \
 
 SUBDIRS = . test
 
-lib_LTLIBRARIES = libflux-resrc.la
+noinst_LTLIBRARIES = libflux-resrc.la
 
-fluxschedinclude_HEADERS = resrc.h resrc_tree.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_reqst.h
 
 libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_reqst.c
 libflux_resrc_la_CFLAGS = -I$(top_srcdir)/rdl
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -34,6 +34,7 @@
 #include <libxml/parser.h>
 #include <libxml/tree.h>
 
+#include "src/common/libutil/shortjson.h"
 #include "rdl.h"
 #include "resrc.h"
 #include "resrc_tree.h"
--- a/resrc/resrc.h
+++ b/resrc/resrc.h
@@ -6,7 +6,6 @@
  */
 
 #include <uuid/uuid.h>
-#include "src/common/libutil/shortjson.h"
 
 #define TIME_MAX INT64_MAX
 
@@ -104,7 +103,7 @@ void resrc_resource_destroy (void *object);
 /*
  * Create a resrc_t object from a json object
  */
-resrc_t *resrc_new_from_json (JSON o, resrc_t *parent, bool physical);
+resrc_t *resrc_new_from_json (json_object *o, resrc_t *parent, bool physical);
 
 /*
  * Return the head of a resource tree of all resources described by an
@@ -122,7 +121,7 @@ resrc_t *resrc_generate_xml_resources (resrc_t *host_resrc, const char *buf,
 /*
  * Add the input resource to the json object
  */
-int resrc_to_json (JSON o, resrc_t *resrc);
+int resrc_to_json (json_object *o, resrc_t *resrc);
 
 /*
  * Print details of a specific resource to a string buffer
--- a/resrc/resrc_reqst.c
+++ b/resrc/resrc_reqst.c
@@ -32,6 +32,7 @@
 #include <assert.h>
 #include <czmq.h>
 
+#include "src/common/libutil/shortjson.h"
 #include "rdl.h"
 #include "resrc_reqst.h"
 #include "src/common/libutil/xzmalloc.h"
--- a/resrc/resrc_reqst.h
+++ b/resrc/resrc_reqst.h
@@ -95,7 +95,7 @@ resrc_reqst_t *resrc_reqst_new (resrc_t *resrc, int64_t qty, int64_t size,
 /*
  * Create a resrc_reqst_t object from a json object
  */
-resrc_reqst_t *resrc_reqst_from_json (JSON o, resrc_reqst_t *parent);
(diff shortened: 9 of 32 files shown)


lipari:
Ok, rebasing to the current master is done and pushed.  This PR is now ready for review.

dongahn:
@lipari:

Thank you for the work. I looked at the issue #148, and I was unsure if this PR fully addresses the issue. So here are some clarification questions:

It is good to eliminate shortjson.h from the headers! But now the resrc interfaces have "struct json_object" instead. Assuming that's a json-c's type, the users of this interface will have to bring in json-c header. And I thought our general direction is not to have dependencies on this specific implementation of json so that we can easily migrate to a different implementation if necessary. (--> I'm also a culprit as I still haven't found time to sort this out for JSC)

In any case, IMHO, we should hide this json type as our own opaque type similarly to what you have been doing for czmq container types. In general, I sort of like @garlick's proposal in the issue ticket, just not to install these headers until the issues around the public interfaces are fully reworked?

Still, this could have been an intermediate step towards this goal, though, in which case please let me know your steps along the way.

garlick:
Yes probably if we're going to expose a json paramter in a public API function, it should be encoded as a string so the user can have a different json implementation.  Or if the intent is not for the user to look inside the json, it could be wrapped in an opaque type.

lipari:
The insertion of "struct json_object*" should not require json.h to define what a json_object is.  It is just a pointer to a struct - which could, in theory, be any struct.  Admittedly, this is a compromise to avoid having to change the API's.  It sounds like you both deem this to be too onerous.  So, I'm willing to call this a partial fix, and keep Issue 143 open until a more elegant solution is devised.

dongahn:
@lipari: on that note, I'm okay to merge this in so that you can proceed w/ other changes. @garlick, okay with this?

garlick:
The shortjson.h remove is required if you're going to publish these headers, so that should go in.

I'm "meh" on the `struct json_object *` change but I guess it doesn't hurt anything.

Keep in mind we're probably going to want to tag and package this stuff pretty soon so if these interfaces are not yet ready for public consumption, they should probably be `noinst_`.

lipari:
@garlick, I just pushed additional commits to this PR that remove the resrc header/library from the installed products.  It is ready for review.  Ignore the travis test failure.  It is intermittent and I have opened a new [Issue](https://github.com/flux-framework/flux-sched/issues/149) to track.

garlick:
Sorry it took so long to get to this!  Here are a few comments:

`libflux-sim.so` is installed.  Since its headers are `noinst_`, this should be a also.

`scheduler.h`, `rs2rank.h`, and `rsreader.h` appear to be internal headers since no corresponding library is installed, and so should be `noinst_`.

I'm confused about the simulator plugins.  They probably do not belong in modules/sched if they don't extend sched.  They all have a MOD_NAME of `sim_sched` internally.  Perhaps these should not be installed util that's sorted.  (which probably means the other simulator comms modules should not be installed either).

I would probably revisit the first commit now that the headers aren't installed and change `struct json_object *` to `json_object *`.

lipari:
Latest push addresses @garlick 's suggestions, with on exception.  I opted not to `noinst` the following line in `simulator/Makefile`:
```
fluxmod_LTLIBRARIES = simsrv.la sim_execsrv.la submitsrv.la
```
The `fluxmod` association leverages the `fluxmod_ldflags` variable defined in configure.ac.  If I made these three modules `noinst`, I would have had to replicate that functionality manually.  Not only would this have been messy, it is something that would have had to be undone once the three modules are ready for prime time.

garlick:
This is looking good, although it seems the simulator plugins (now internal) are only being built statically.  Tests are passing so I assume they are not loaded by any of the tests?

There's a trick needed to build these modules dynamically and not install them.  Basically this will do it:
```diff
diff --git a/simulator/Makefile.am b/simulator/Makefile.am
index dec2630..5c15fc8 100644
--- a/simulator/Makefile.am
+++ b/simulator/Makefile.am
@@ -62,7 +62,7 @@ sim_sched_fcfs_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 sim_sched_fcfs_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-sim_sched_fcfs_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+sim_sched_fcfs_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowhere
 
 sim_sched_fcfs_aware_la_SOURCES = sim_sched_fcfs_aware.c scheduler.c \
     scheduler.h
@@ -70,13 +70,13 @@ sim_sched_fcfs_aware_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 sim_sched_fcfs_aware_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-sim_sched_fcfs_aware_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+sim_sched_fcfs_aware_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowher
 
 sim_sched_easy_la_SOURCES = sim_sched_easy.c scheduler.c scheduler.h
 sim_sched_easy_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 sim_sched_easy_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-sim_sched_easy_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+sim_sched_easy_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowher
 
 EXTRA_DIST = sim_version.map
```
You can use the same trick to build the comms modules dynamically but not install them, e.g.
```diff
diff --git a/simulator/Makefile.am b/simulator/Makefile.am
index 5c15fc8..e4c6ee2 100644
--- a/simulator/Makefile.am
+++ b/simulator/Makefile.am
@@ -24,10 +24,9 @@ tsched_lib_LDADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/src/common/libutil/libutil.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
 
-fluxmod_LTLIBRARIES = simsrv.la sim_execsrv.la submitsrv.la
-
 noinst_LTLIBRARIES = libflux-sim.la \
-    sim_sched_fcfs.la sim_sched_fcfs_aware.la sim_sched_easy.la
+    sim_sched_fcfs.la sim_sched_fcfs_aware.la sim_sched_easy.la \
+    simsrv.la sim_execsrv.la submitsrv.la
 
 noinst_HEADERS = simulator.h
 
@@ -42,20 +41,20 @@ simsrv_la_SOURCES = simsrv.c
 simsrv_la_CFLAGS = $(AM_CFLAGS) $(AM_CPPFLAGS)
 simsrv_la_LIBADD = $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-simsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+simsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowhere
 
 sim_execsrv_la_SOURCES = sim_execsrv.c
 sim_execsrv_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 sim_execsrv_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-sim_execsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+sim_execsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowhere
 
 submitsrv_la_SOURCES = submitsrv.c
 submitsrv_la_CFLAGS = $(AM_CFLAGS)
 submitsrv_la_LIBADD = $(top_builddir)/resrc/libflux-resrc.la libflux-sim.la \
     $(FLUX_CORE_LIBS) $(JSON_LIBS) $(CZMQ_LIBS)
-submitsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module
+submitsrv_la_LDFLAGS = $(AM_LDFLAGS) $(fluxmod_ldflags) -module -rpath /nowhere
 
 sim_sched_fcfs_la_SOURCES = sim_sched_fcfs.c scheduler.c scheduler.h
 sim_sched_fcfs_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
```

lipari:
New push with all suggestions taken.

lipari:
Well, this is strange.  I just pushed a rebase of my issue_143 branch to github.  You can see it here:
https://github.com/lipari/flux-sched/commit/debf95639c96fefbfb63351d3e9c9c31d6c481a8
Yet, github doesn't appear to have replaced the new commits in the https://github.com/lipari/flux-sched/commits/issue_143 commit history.

To diagnose, I cloned a fresh copy of my repo, checked out the issue_143 branch, and saw the modified commit history starting with the commit that amended the comment.

I'd like to add to this branch a fix for [Issue 150](https://github.com/flux-framework/flux-sched/issues/150).  Maybe when I push that, it will override the old history.

grondo:
The issue currently lists that it is "Checking for ability to automatically merge..." I wonder if github is experiencing technical difficulties right now. You could try rebasing again on top of upstream/master and do a force push.. see if that pushes any new objects.

lipari:
Well... It could be that github is having trouble right now.  Repushing a rebased history of 4 new commits (with no actual changes) nudged my *last* commit into visibility.  That should be fine to merge, but there is currently a mismatch between what my cloned repo reports for the issue_143 contents and what github displays https://github.com/lipari/flux-sched/commits/issue_143.  So, we should probably wait. :(

lipari:
Wait time over!  They are now in sync!

garlick:
Wait for fix to #150?

lipari:
No, go ahead and merge.  I'll submit a separate PR for #150.

Issue #150, opened 2016-04-04
gcc 5.2.1 compilation issue


garlick:
On ubuntu 15.10, gcc 5.2.1 found this:
```
make[1]: Entering directory '/home/garlick/proj/flux-sched/simulator'
  CC       sim_sched_fcfs_la-scheduler.lo
scheduler.c: In function 'update_job_cores':
scheduler.c:1174:13: error: logical not is only applied to the left hand side of comparison [-Werror=logical-not-parentheses]
             < 0) {
             ^
cc1: all warnings being treated as errors
Makefile:973: recipe for target 'sim_sched_fcfs_la-scheduler.lo' failed
make[1]: *** [sim_sched_fcfs_la-scheduler.lo] Error 1
make[1]: Leaving directory '/home/garlick/proj/flux-sched/simulator'

```
It needs a fix like this:

```diff
diff --git a/simulator/scheduler.c b/simulator/scheduler.c
index afc3dae..5db99bc 100644
--- a/simulator/scheduler.c
+++ b/simulator/scheduler.c
@@ -1170,8 +1170,7 @@ int update_job_cores (ctx_t *ctx,
 
     if (imanode) {
         if (!(key = xasprintf ("lwj.%"PRId64".rank.%"PRId64".cores", job->lwj_id,
-                               *pnode))
-            < 0) {
+                               *pnode))) {
             flux_log (ctx->h, LOG_ERR, "%s key create failed", __FUNCTION__);
             rc = -1;
             goto ret;
```

garlick:
Fixed with #151 merge

PR #151, opened 2016-04-05
Fix gcc 5.2.1 compilation issue in simulator

lipari:
Address flux-sched Issue #150


lipari:
@garlick , would you do the honors of pressing the button?  The change matches the suggestion you made.

garlick:
Thanks!

PR #152, opened 2016-04-11
Generate hwloc resources

lipari:
Use the hwloc API to walk a topology and populate a resource tree.  Replaces the xml parsing that was formerly done.


resrc/resrc.c
```diff
@@ -677,79 +675,48 @@ static char *lowercase (char *str)
...
-                }
-                break;
-            }
-            xmlFree (prop);
-            cur = cur->next;
-        }
-        if (!name)
+        hwloc_name = hwloc_obj_get_info_by_name (obj, "HostName");
+        if (!hwloc_name) {
+            fprintf (stderr, "Failed to get HostName for node\n");
```
dongahn: Does fprintf (stderr...) print the error message when the resrc is used within a broker?

resrc/resrc.h
```diff
@@ -112,11 +113,11 @@ resrc_t *resrc_new_from_json (json_object *o, resrc_t *parent, bool physical);
...
 /*
- * Return the head of a resource tree of all resources described by an
- * xml serialization
+ * Return the head of a resource tree of all resources described by a
+ * hwloc topology
  */
-resrc_t *resrc_generate_xml_resources (resrc_t *host_resrc, const char *buf,
-                                       size_t length, const char *sig);
+resrc_t *resrc_generate_hwloc_resources (resrc_t *host_resrc,
+                                         hwloc_topology_t topo, const char *sig);
```
dongahn: Are you feeling comfortable to have a dependency to hwloc data type in resrc API? Do we expect that we will ever want to change the underlying implementation of the topo? 

What if we have a system where hwloc isn't implemented? Will we want to provide our own straw man implementation in which case this API can be in the way.

I am just asking these questions as you will ultimately want to make resrc as a public library to be installed as part of flux and I'm sure we will ask these questions at that point anyway. 

I don't recall what was the decision within flux-core using hwloc data type in a public API.

resrc/resrc.c
lipari: I don't believe any of the printf statements work within a broker.  It is an artifact of designing to resrc system to be Flux agnostic.  But none of the printf statements are necessary for normal operation.  They are instead useful for diagnosing problems when `make check` is invoked.

dongahn: OK. 

This is just comment -- IOW, I'm not asking you to do this as part of this PR. 

But can resrc get a ile descriptor from the client code and use it to print things out? I remember I had to introduce `resrc_to_string` for better debugging of resrc under flux. But in general, if there is a general way for resrc to pass debug info to the broker log, this will be helpful.

resrc/resrc.h
lipari: I'm open to suggestions.  But with these changes, we now support resrc_generate_hwloc_resources() as part of the resrc API.  If we someday decide to make generating a resrc tree from an hwloc topology optional, then that would be the time to separate all the hwloc code into its own optional module.

resrc/resrc.c
lipari: I could *kind of* see this working... Add a `FILE *stream` argument to functions that invoke fprintf() - resrc_generate_hwloc_resources() for example.  But then any invoking code (like the sched module) would need to set up a service that read from that stream and invoked flux_log() on everything it received.
The other thing I could do is just take the fprintf(stderr, ) statements out of resrc entirely and establish the condition resrc offers no error reporting to a running flux instance.

dongahn: At the end of the data flux_log () will use a file descriptor and I am wondering if there is any way we can pass that to resrc.

lipari: Yes, I thought of that.  But flux_log() calls flux_vlog() and that thing creates a JSON object and passes this in a message to the "cmb.log" topic.

lipari: Wait a minute... just had another idea.  I could add a char** argument to resrc_generate_hwloc_resources().  If it returned null, that ptr would contain the reason.  Then sched could invoke flux_log() and pass it the string.  How does that sound?

resrc/resrc.c
```diff
@@ -677,79 +675,48 @@ static char *lowercase (char *str)
...
-        prop =  xmlGetProp (nodePtr, (const xmlChar *) "os_index");
-        if (prop) {
-            id = strtol ((char *) prop, NULL, 10);
-            xmlFree (prop);
         }
+        name = xstrdup (hwloc_name);
+    } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_NODE)) {
+        type = xstrdup ("numanode");
+        name = xasprintf ("%s%"PRId64"", type, id);
+    } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_SOCKET)) {
```
dongahn: I'm really asking... I once had trouble using HWLOC_OBJ_NUMANODE so decide to stick with HWLOC_OBJ_SOCKET. 

My suspicion was HWLOC_OBJ_NUMANODE is a macro replacing HWLOC_OBJ_SOCKET in a newer version of hwloc. If so, do hwloc will keep guaranteeing backward compatibility for HWLOC_OBJ_SOCKET?

lipari:
@dongahn, new push that addresses your comments.

dongahn:
@lipari: this looks good to me! The only question is have is that the id field of each resrc object created from hwloc now gets the logical_index. I believe this is the right way to go because local index has much better properties like sibling resources get contiguous id numbers and this should also be relatively consistent over time (unless the system gets a significant firmware update).

But I'm curious to know where is the id field has been used and if there would be any side effect. I couldn't find any by looking through the source but I thought I should ask.

dongahn:
Ah, I actually have two more minor comments in my note.

1. If you want to further hide hwloc data type later in case we may want to support a system without this package. I think we can consider borrowing the factor method OOP concept  (https://en.wikipedia.org/wiki/Factory_method_pattern) and provides a straw man hwloc-like abstraction.

2. Don't we need safety check whether the version of the loaded hwloc library match with the version that was used to compile flux? (In particular the code is supposed to work with >= 2.x as well? 

I found 

```
#include <hwloc.h>
#if HWLOC_API_VERSION >= 0x00020000
  /* headers are recent */
  if (hwloc_get_api_version() < 0x20000)
    ... error out, the hwloc runtime library is older than 2.0 ...
#else
  /* headers are pre-2.0 */
  if (hwloc_get_api_version() >= 0x20000)
    ... error out, the hwloc runtime library is more recent than 2.0 ...
#endif
```
from https://www.open-mpi.org/projects/hwloc/doc/v1.11.2/a00030.php

lipari:
@dongahn:
 * While there will probably be a need for the resrc model to support physical and logical indices someday, the original RDL only offers one index.  Rather than include a model revision in this PR, I opted for the logical index because I found the physical indices were mysteriously reporting -1 for the cache objects.  Until we get to the day when both are required in our model, this was what I decided to go with.  I'd be happy to change it to the physical os index if you feel it makes better sense.
* After reading your reference to the OOP page, I don't know what to do short term to address your concern.  Do you feel that preparing for solution different from hwloc is something we need to address in this PR?
* I am familiar with the hwloc_get_api_version() API and considered using it.  However, newer versions of hwloc.h do not have backward compatibility with the hwloc_obj_type_t enums.  So the code that compares the type against the enums would not have compiled unless I made them #ifdef's.
* These #ifdef's are intended to support a range of hwloc versions and free us from having to lock into the specific version that was used to build flux-core.

dongahn:
@dongahn<https://github.com/dongahn>:

  *   While there will probably be a need for the resrc model to support physical and logical indices someday, the original RDL only offers one index. Rather than include a model revision in this PR, I opted for the logical index because I found the physical indices were mysteriously reporting -1 for the cache objects. Until we get to the day when both are required in our model, this was what I decided to go with. I'd be happy to change it to the physical os index if you feel it makes better sense.

If there is no side effect with logical_index in the current sched code base, we should stick with logical_index as the id. I was basically asking if you know of any side effect now; it seems the answer is no. So this is good.

Later, there will be a question as to how a scheduler will request a specific core set (e.g., core[0-2]) in a way consistent with how wreck sees them. But I don’t think that we have wreck protocol that supports this. (Though I haven’t been able to follow recent updates)


  *
     *   After reading your reference to the OOP page, I don't know what to do short term to address your concern. Do you feel that preparing for solution different from hwloc is something we need to address in this PR?

No, no short term request here. I just mainly referred to this as a distance future work if needed.


  *
     *   I am familiar with the hwloc_get_api_version() API and considered using it. However, newer versions of hwloc.h do not have backward compatibility with the hwloc_obj_type_t enums. So the code that compares the type against the enums would not have compiled unless I made them #ifdef's.

  *
     *
     *   These #ifdef's are intended to support a range of hwloc versions and free us from having to lock into the specific version that was used to build flux-core.

Yes, these #ifdefs are good.  I raised this question just because there can be cases where this code is compiled with, say, hwloc version 2 header (and thus the version-2 branch of #ifdelf) but somehow version 1.x hwloc is picked up at runtime. (Assuming we use a shared library version of hwloc). In this case, we can have pretty nasty errors: that typical header-library version mismatch problem…! I think we can punt this for now and revisit this for all the code that use hwloc later. Or you can have one-time simple check like:

if (HWLOC_API_VERSION != hwloc_get_api_version ())
bad ();

to disallow no version mismatch since we don’t exactly know the backward compatibility guarantee of this package?

lipari:
Thanks for the clarification on the hwloc version.  I understand now; it's a fine idea.  I'll have an updated PR in the morning.

lipari:
Just pushed the hwloc version check.  Ready for review.

dongahn:
Thanks. This works for me. I will wait until the COB today in case others have comment. If not, I will merge.

Issue #153, opened 2016-04-21
Change verbose level to INFO when RDL spec != hwloc


dongahn:
Currently, when the hwloc resource info does not match with the specified RDL file, `sched` reports 

```
[1461268835.654123] sched.err[0]: RDL(/nfs/tmp2/dahn/FLUX-20160421/flux-sched/conf/hype.lua) inconsistent w/ hwloc!
```

But this is not a fatal failure because `sched` will tolerate it and rebuild the `resrc` tree based on the hwloc resource info instead. I think @tpatki was once confused about this before. And looking at this after a while doesn't seem right to me either. 

What do folks think changing the verbose level for this condition to INFO?

lipari:
INFO sounds good to me.  I would amend the message to state that (in addition to the RDL file not matching hwloc), that the resources discovered by hwloc would be used.

lipari:
In thinking about this more, is this message even needed.  If the tests will never run on hype, why bother trying to read a hype.lua file?

dongahn:
I think the message is needed since the message can potentially help debug. But in terms of test cases, you have a good point. Perhaps, what we should do is not to pass an RDL file (e.g., hype.lua) to `sched` load line for all but a few tests that deliberately test if `sched` can tolerate RDL file != hwloc info?

grondo:
There should be at least one test that loads an example RDL file if we continue to support that. I think the `hype.lua` file was just an example meant to sanity check readers.

dongahn:
@grondo: good point. Yes, we do have some test coverage for these cases. So we can safely remove the RDL file from other subtests...

```
ok 4 - rs2rank: works with a matched RDL
[1461271183.176294] sched.err[0]: RDL(/nfs/tmp2/dahn/FLUX-20160421/flux-sched/t/data/hwloc-data/004N/exclusive/cab.hwloc2.lua) inconsistent w/ hwloc!
ok 5 - rs2rank: works with an inconsistent RDL (fewer cores)
[1461271184.204055] sched.err[0]: RDL(/nfs/tmp2/dahn/FLUX-20160421/flux-sched/t/data/hwloc-data/004N/exclusive/cab.hwloc3.lua) inconsistent w/ hwloc!
ok 6 - rs2rank: works with an inconsistent RDL (fewer nodes)

PASS: t1001-rs2rank-basic.t
```

dongahn:
PR #157 ready for review.

PR #155, opened 2016-04-22
Issue153: Better resource unmatch test and etc

dongahn:
Fix Issue #153. 

- Remove rdl-config=RDL-file option from the tests that are not designed to test a RDL-hwloc inconsistency
- Improve the verbosity for RDL != hwloc condition
- Add fail-on-error option to the sched module for better testing

### DIFF
--- a/sched/rsreader.c
+++ b/sched/rsreader.c
@@ -196,10 +196,11 @@ done:
     return rc;
 }
 
-int rsreader_link2rank (machs_t *machs, resrc_t *r_resrc)
+int rsreader_link2rank (machs_t *machs, resrc_t *r_resrc, char **err_str)
 {
     int rc = -1;
     size_t ncnt = 0;
+    char *e_str = NULL;
     resrc_tree_t *rt = NULL;
     int nsocks = 0, ncs = 0;
     const char *digest = NULL;
@@ -216,12 +217,19 @@ int rsreader_link2rank (machs_t *machs, resrc_t *r_resrc)
          */
         digest = rs2rank_tab_eq_by_count (machs, resrc_name (resrc_tree_resrc (rt)),
                                           nsocks, ncs);
-        if (!digest)
+        if (!digest) {
+            e_str = xasprintf ("%s: Can't find a matching resrc for <%s,%d,%d>",
+                __FUNCTION__, resrc_name (resrc_tree_resrc (rt)), nsocks, ncs);
             goto done;
+        }
         resrc_set_digest (resrc_tree_resrc (rt), xasprintf ("%s", digest));
     }
     rc = 0;
 done:
+    if (err_str)
+        *err_str = e_str;
+    else
+        free (e_str);
     return rc;
 }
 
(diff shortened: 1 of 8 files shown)


grondo:
@dongahn, nice work!

One question:  You mention that sched "sets name to hostname" in hwloc reader mode. RDL itself is not supposed to be node-centric and therefore in Lua-based rdl config code `rdl_resource_name()` does not differentiate between "node" types vs other types, and will return `name+id` for *any* resource that has an ID field. (e.g. `{ name: "core", id: 1 }` returns `rdl_resource_name() == "core1"`)

I just wanted to make sure this doesn't further affect sched if it is currently behaving node-centric.

dongahn:
@grondo: Thank you for the quick response and good question!

@lipari should chime in for this question. If I read hwloc reader code correctly in [here](https://github.com/flux-framework/flux-sched/blob/master/resrc/resrc.c#L679), the name of resrc is actually set to name+id where id is the logical index of hwloc_object_t type except for the node type. For node, resrc sets it to hostname -- and we had some issues when it didn't do it that way: PR #109.  

I think it will be good if we go through resrc to make sure it builds its internal fields consistently across different reader modes: rdl and hwloc. To me, it is okay as far as the name always returns the same information regardless of the reader mode. You may have a different opinion though.

lipari:
@grondo, the resrc code mirrors the conventions established with [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc) as well as the rdl code itself.  When generating a resrc_t object from an RDL file, I opted to use the rdl_resource_json() instead of the rdl_resource_name() call.  I'd have to check, but if the "name" field is getting an id appended, that is where it comes from.

dongahn:
FYI -- I've created Issue #157. To discuss general (in)consistency issues on the semantics of resrc fields between different reader modes, I'd suggest we take it to #157.

Let's keep this PR to discuss specifics relevant to this particular resource match test failure patch.

dongahn:
Please don't merge. I will adapt this PR after @lipari's PR from Issue #157 is merged.

lipari:
[PR 161](https://github.com/flux-framework/flux-sched/pull/161) submitted.

dongahn:
OK. Rebased to the latest master and forced a push.

As agreed, with PR #161, I took off the hostname support from this PR. The rest of the changes seem work out of box on the latest HEAD of master.

lipari:
@dongahn, looks good!  Only two suggestions:
* RDL_CONF_DEFAULT in t/sharness.d/sched-sharness.sh is vestigial.  Is there a need/plan to use it somewhere/sometime?
* In isolation, the name of the `--fail-immediate` option seems confusing.  Is it an option to make the sched plugin fail at launch?  How about naming the option `--rdl-conf-reqrd` or `--rdl-conf-only`?

dongahn:
Thanks!

> RDL_CONF_DEFAULT in t/sharness.d/sched-sharness.sh is vestigial. Is there a need/plan to use it somewhere/sometime?

I believe this needs to be removed. I will change and do a push. 

> In isolation, the name of the --fail-immediate option seems confusing. Is it an option to make the sched plugin fail at launch? How about naming the option --rdl-conf-reqrd or --rdl-conf-only?

The intent of this option is to fail the `sched` on any soft error for testing purposes beyond this rdl unmatch error. Without such an option, it becomes difficult to add a test case like the new "unmatch failure test" into shardness. I am open to other naming suggestions but the both of your suggestions seems too rdl centric...

lipari:
I see.  How about `--fail-on-error` then?

dongahn:
I like `--fail-on-error` better!

dongahn:
OK. Forced a push.

PR #156, opened 2016-04-23
runlevel: Add runlevel support

dongahn:
If sched is not installed into flux-core's $prefix,
one should set FLUX_RC_EXTRA environment variable to flux-sched's $prefix/etc/flux
so that flux start can automatically locate and execute sched's runlevel 1 and 3.

In addition, users can set FLUX_SCHED_OPTIONS if they want flux
to load in the sched module with non-default options.
(e.g., FLUX_SCHED_OPTIONS="plugin=sched.backfill" will cause
the flux session to load the sched module with the backfil scheduling plugin.

For LLNL's /opt rpm package, I plan to add FLUX_RC_EXTRA to its module script.
Also, module load flux-sched-xxx.rpm will print out available load options
of sched to console to help users to choose options to play with.

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -31,6 +31,7 @@ test_size_large() {
 #
 test_under_flux() {
     size=${1:-1}
+    personality=${2:-full}
     log_file="$TEST_NAME.broker.log"
     if test -n "$TEST_UNDER_FLUX_ACTIVE" ; then
         cleanup rm "${SHARNESS_TEST_DIRECTORY:-..}/$log_file"
@@ -49,6 +50,18 @@ test_under_flux() {
     fi
     if test -n "$SHARNESS_TEST_DIRECTORY"; then
         cd $SHARNESS_TEST_DIRECTORY
+    fi
+
+    if test "$personality" = "minimal"; then
+        export FLUX_RC1_PATH=""
+        export FLUX_RC3_PATH=""
+    elif test "$personality" != "full"; then
+        export FLUX_RC1_PATH=$FLUX_SOURCE_DIR/t/rc/rc1-$personality
+        export FLUX_RC3_PATH=""
+        test -x $FLUX_RC1_PATH || error "$FLUX_RC1_PATH"
+    else
+        unset FLUX_RC1_PATH
+        unset FLUX_RC3_PATH
     fi
 
     TEST_UNDER_FLUX_ACTIVE=t \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 
 ACLOCAL_AMFLAGS = -I m4
 
-SUBDIRS = src rdl resrc simulator sched t
+SUBDIRS = src rdl resrc simulator sched etc t
 
 EXTRA_DIST= \
 	conf \
--- a/configure.ac
+++ b/configure.ac
@@ -74,6 +74,11 @@ AC_SUBST(schedplugindir)
 AS_VAR_SET(fluxschedincludedir, $includedir/flux/sched)
 AC_SUBST(fluxschedincludedir)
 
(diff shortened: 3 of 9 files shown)


dongahn:
Apparently I need to work on the install test in travis...

garlick:
Glanced at this briefly - I guess sched is loaded now by default and tests expect to load/unload it.  You may want to follow a similar pattern as the core tests which now have their own rc scripts in t/rc.

dongahn:
Yes. That's what I figured as well. 

BTW, I am thinking to add another environment variable to sched's rc1 script so that users can load in a hwloc file if they want, (using hwloc reload command). Thoughts?

grondo:
Maybe that is something that could be in a user or site specific rc file that runs after flux-core and flux-sched rc files have completed? At least, I'm not sure it belongs in the sched rc1 script since hwloc support is part of core. If the hwloc reload needs to occur before sched module is loaded, then we'd need some way to have the optional rc file inserted after core and before sched's rc1 files.

garlick:
Maybe there should be a way via a variable in core's RC to substitute a file for the real hwloc data?  As opposed to loading the real and then reloading the file?

grondo:
That sounds reasonable to me for now, but I'm getting a little worried about the direction this is going (config by hundreds of environment variables...)

grondo:
I wonder if broker attrs could be used for some of this stuff? @garlick I know you wanted to have attr namespaces perhaps before going that route, but for me it would be slightly better than env vars (even though I perhaps initially suggested that to @dongahn), especially if you could eventually specify lots of attrs via a file ...

dongahn:
Well I don't think I need a serious solution for fake hwloc reloading inserted somewhere during runlevel1.

I don't think there will be that many uses of wanting to load a fake hwloc info to begin with. So i am okay to kick the can down the road. i have to think initially many users will simple do 

module load flux-core
module load flux-sched

optionally setting FLUX_SCHED_OPTIONS (in particular if they want to load backfill scheduler...)


and flux start ... to bootstrap an instance within the slurm allocation to experiment with it.

For other advaned stuff like fake hwloc, it should be ok for them to make some extra manual setup efforts  for now IMHO.

dongahn:
ok. I fixed the the problems with FLUX_SCHED_TEST_INSTALLED and forced a push.

t/sharness.d/flux-sharness.sh
```diff
@@ -51,6 +52,18 @@ test_under_flux() {
         cd $SHARNESS_TEST_DIRECTORY
     fi
 
+    if test "$personality" = "minimal"; then
+        export FLUX_RC1_PATH=""
+        export FLUX_RC3_PATH=""
+    elif test "$personality" != "full"; then
+        export FLUX_RC1_PATH=$FLUX_SOURCE_DIR/t/rc/rc1-$personality
```
lipari: Unlike flux-core, FLUX_SOURCE_DIR is not currently defined in flux-sched.

dongahn: Ah. Great you caught this @lipari. Do we have an equivalent variable that points to the flux-core source tree? This was a wholesale copy and paste and it would be good to use the identical file for this though.

lipari: The latest version of the flux-sched's configure.ac replaces the old dependence on the flux-core source tree with FLUX_PREFIX - which will point to the root of the installed flux-core.
However, the personality cases are for flux-core-specific tests (kvs, testenv, wreck).  I will create flux-core issue to address desire to use the latest flux-sharness.sh against its use by non-flux-core projects.

dongahn:
Perhaps this should be merged as is for now since FLUX_SOURCE_DIR doesn't break anything. Then, if we agree on a better protocol to control flux-sharness.sh as part of flux-core #654, we can pick it up later?

PR #165, opened 2016-05-25
test: add CTS-1 hwloc data

garlick:
@dongahn requested this in flux-framework/distribution#12

Taken from opal compute node with `lshw -xml` (as non-root).

Tell me where it should land and I'll adjust the path.


dongahn:
@garlick: Thank you for the xml file. I can take this into the `sched` source tree, but my thinking was to use this in a scalability testing repo. Didn't we once plan to build up some testing rig outside our regression tests? This way, we can run our regular tests (e.g., ones in `t` directory) as part of CI testing and complementing them with larger stress/scalability tests that we can run from time to time (e.g., DST?) In this regard, I will probably want to move my stress sharness test into the scalability testing rig if we decide to build one.

garlick:
@trws created [flux-scalability](https://github.com/flux-framework/flux-scalability) but we've not done anything with it yet.

Should this go there?  I am fine with resubmitting it in that project.

dongahn:
This will be added to flux-scalability as part of CTS-1 sched testing phase II. Closing.

PR #169, opened 2016-06-09
Remove old, unused sim_sched code

SteVwonder:
Remove the old scheduler code that used to run under the simulator.  The simulator now interfaces with the sched module, making this code unnecessary.


lipari:
Very nice!  Thank you.

PR #170, opened 2016-06-10
Copy gitignore over from flux-core

SteVwonder:
My 'git status' is becoming rather unruly with all of the untracked files (mainly from autotools).  Any objections to pulling flux-core's .gitignore into flux-sched?


SteVwonder:
Looks as though something in flux-sched relies on flux-mrpc, which was just removed in flux-framework/flux-core#689.  I'll create an issue.

garlick:
I don't see an actual user in sched - maybe it is a travis caching issue?

PR #172, opened 2016-06-14
logging:  Bring over the latest flux-core logging changes

lipari:
This is a periodic duty of keeping the few flux-core/src/common files
that flux-sched copied in sync with flux-core


lipari:
@garlick , as the one responsible for the logging changes and in the absence of @dongahn, it would be great if you could review this PR and merge when all issues are addressed.

garlick:
Looks great to me and travis is green so merging!

lipari:
Thanks!

PR #173, opened 2016-06-14
Versatile Scheduling

lipari:
Replace tree list searches with tree searches

Modifies the resrc search code to search a resource tree instead of a resource tree list.  This has two benefits:
    
* Can now search for an isolated resource (like cores) without regard to its parent resource (like nodes).  Formally, we had to define a request that include one node and multiple cores to find resource composites of node+cores.  Now we can just search for cores.
    
* The code now conforms to the RFC 4 requirement:
"The job ID for a job that is allocated a resource in a composite hierarchy MUST be annotated not only to the resource, but to each parent up the tree of those resources allocated to the enclosing
instance"

Also in this PR are minor mods that prepared the way for the tree search changes.

### DIFF
--- a/resrc/test/tresrc.c
+++ b/resrc/test/tresrc.c
@@ -53,47 +53,38 @@ u_int64_t get_time() {
         + (t.tv_usec - start_time.tv_usec);
 }
 
-static void test_select_children (resrc_tree_t *rt)
+/*
+ * Select some resources from the found tree
+ * frt is found resource tree
+ * prt is selected resource tree parent
+ */
+static resrc_tree_t *test_select_resources (resrc_tree_t *frt, resrc_tree_t *prt,
+                                            int select)
 {
-    if (rt) {
-        resrc_t *resrc = resrc_tree_resrc (rt);
-        if (strcmp (resrc_type(resrc), "memory")) {
-            resrc_stage_resrc (resrc, 1);
-        } else {
+    resrc_tree_t *selected_res = NULL;
+
+    if (frt) {
+        resrc_t *resrc = resrc_tree_resrc (frt);
+
+        if (!strcmp (resrc_type(resrc), "core")) {
+            if (resrc_id (resrc) == select)
+                resrc_stage_resrc (resrc, 1);
+            else
+                goto ret;
+        } else if (!strcmp (resrc_type(resrc), "memory"))
             resrc_stage_resrc (resrc, 100);
-        }
-        if (resrc_tree_num_children (rt)) {
+
+        selected_res = resrc_tree_new (prt, resrc);
+        if (resrc_tree_num_children (frt)) {
             resrc_tree_t *child = resrc_tree_list_first (
-                resrc_tree_children (rt));
+                resrc_tree_children (frt));
             while (child) {
-                test_select_children (child);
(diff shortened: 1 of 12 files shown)


lipari:
This PR awaits @dongahn's return, though others are encouraged to review and comment.

resrc/test/tresrc.c
```diff
@@ -53,47 +53,38 @@ u_int64_t get_time() {
...
+    if (frt) {
+        resrc_t *resrc = resrc_tree_resrc (frt);
+
+        if (!strcmp (resrc_type(resrc), "core")) {
+            if (resrc_id (resrc) == select)
+                resrc_stage_resrc (resrc, 1);
+            else
+                goto ret;
+        } else if (!strcmp (resrc_type(resrc), "memory"))
             resrc_stage_resrc (resrc, 100);
```
SteVwonder: Unchanged by this commit, but why is memory reserved in chunks of 100?

sched/sched.c
```diff
@@ -1212,75 +1223,107 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
...
+        Jadd_int64 (req_res, "req_qty", job->req->nnodes);
+        nreqrd = job->req->nnodes;
+
+        /* Since nodes are requested, make sure we look for at
+         * least one core on each node */
+        if (job->req->ncores < job->req->nnodes)
+            job->req->ncores = job->req->nnodes;
+        job->req->corespernode = (job->req->ncores + job->req->nnodes - 1) /
+            job->req->nnodes;
+        if (job->req->node_exclusive) {
```
SteVwonder: I'm not sure I fully comprehend what's going on here.  What does req_size do?

SteVwonder: More generally, how do req_qty, req_size, and exclusive combine?  It seems to me that only 2 of the 3 are necessary (e.g. req_qty and exclusive OR req_qty and req_size).  Sorry, I know this didn't change with this PR, but I can't seem to get these attrs straight in my head.

lipari: Every resource has a size element.  Size would be 1 for most resources like a node and core and it represents a singular unit.  Memory is a resource, but more accurately, a pool of resources, though "pool" is an overloaded concept.  As such, you can ask for and be allocated portions of that "pool".  That is what size is for.

A resource request could contain a composite, like one core and 100 bytes of memory on one node.  The exclusive flag stipulates the requested ownership.  A core would be requested exclusively when the user doesn't want any other job sharing that core.  A node however, would be more likely shared or exclusive, depending on the user's request.

Size and exclusivity are orthogonal.  You could ask for a node of size 1 and stipulate either exclusive or shared access.  Either would be supported.

lipari:
Just rebased and pushed.  Ready for eyes...

dongahn:
@lipari walked me through this change and this looks good to me -- although I won't have time to review the code with a fine-toothed comb before my travel next week. @SteVwonder: if you are ok with this change, I am fine. Please let me know.

SteVwonder:
@dongahn, after a quick, second look-through, it seems good to me.

Sidenote: At some point this summer, I will attempt to rewrite my I/O-aware scheduling plugin using this new setup.  I think this redesign will make it much easier.

dongahn:
Sounds good.

Issue #159, opened 2016-04-26
Providing a runtime option to select conservative/hybrid backfill algorithms?


dongahn:
This is just an idea. 

As I was trying to package things up, I thought it might be useful to make not only EASY backfill, but also conservative backfill algorithms as a **runtime** option. This way, users can try most popular online scheduling algorithms: FCFS, EAST/Hybrid/Conservative backfill. 

I have to believe all of the algorithms is already in the code, but it is currently statically determined [here]( https://github.com/flux-framework/flux-sched/blob/master/sched/sched_backfill.c#L44). 

How about we devise some ways to pass "backfill_depth=K" where K can vary from 1 to unlimited. If it is set to 1, sched will do EAST backfill. If it is set to unlimited, the algorithm becomes conservative backfill. If K is set to be some number, it becomes hybrid. 

I was thinking to pass that as a new option to sched load. But then, if there is a way to pass a option to sched.backfill load, this would even be better.

Thoughts?

SteVwonder:
I like this idea.  It would avoid creating a different plugin for every minor variation of a particular scheduler.

To make this work, do we want to store the options as local to the plugin (some local struct), or store them in the sched and pass them into the plugin functions with every call?  It seems to me that allowing the options to be parsed and stored in the plugin is more flexible and cleaner.  I am just not sure the best way to go about implementing that.

Since the plugin is loaded directly by the sched module (and not via a flux load call), I don't see how it would be possible to directly pass arguments to the plugin.  The only thing that I can think of is to add a "setup_plugin" function to each plugin that takes in the the argv of the sched module (and is called after the plugin is loaded).  This would allow the plugin to make a pass over the arguments passed to sched for arguments relevant to it (the plugin).  The problem with this is namespace collisions.

garlick:
see `plugin.c::insmod_cb()` - the sched modules can be loaded by `flux module load` and arguments are available in that function to do with what you like.

dongahn:
@SteVwonder, @garlick: I agree that passing this at `flux module load` time is a way to go. FYI -- With latest runlevel support, a scheduling plugin can be selected by setting that as a load time option to an environment variable: `FLUX_SCHED_OPTIONS` before `flux start`

@SteVwonder: if you can quickly implement this, this can be used as part of Phase III testing of our [CTS-1 test plan](https://github.com/flux-framework/distribution/issues/14). I know this can also help you w/ your summer dynamic-scheduling investigation. Thanks.

[...]

PR #174, opened 2016-06-16
sim & sched: fix segfaults when running in sim

SteVwonder:
Minor fixes to some free'ing that causes a segfault when running under the simulator.
The problem only occurs when the simulator is being shutdown.


SteVwonder:
Ran across these while writing the code for #159.

I can include this commit in with a larger PR that will include my code for #159, or we can just merge it now.  I'm fine with either.

lipari:
Looks good!  Thanks.

Issue #168, opened 2016-06-09
Broader verification of scheduler correctness


SteVwonder:
In #159, @dongahn mentioned: 

> do you have an idea as to how we can test correctness of various scheduling algorithms for CTS-1 testing? Perhaps we can bring in some well-known workloads w/ known schedules?`

Opening as a separate issue to isolate the discussion here.

SteVwonder:
We currently have a very simple test using the simulator that launches 10 jobs and verifies their schedule under FCFS and EASY.  So the framework is there to do the testing (assuming the simulator is still functioning under master, which is a separate issue that I'll be address this summer), but now we need some more extensive tests.  I will look into this to see what I can find.

At minimum, we should be able to find emulated/simulated schedulers that run EASY, Conservative, and the hybrid algorithms.  We could then throw the same set of jobs to the Flux implementation and the other implementation, comparing their schedules.

At best, there is (as @dongahn mentioned) a set of workloads and associated schedules that already exist.

dongahn:
Agreed.

SteVwonder:
@dongahn and I spoke on the phone today, and the idea of using SLURM's scheduler to verify Flux's scheduler came up.  We are unsure of the exact scheduling policies that SLURM supports.  @lipari's [slides](https://slurm.schedmd.com/slurm_ug_2012/SUG-2012-Scheduling.pdf) (specifically slide 34) make it seem that 4 years ago only EASY backfilling was supported by SLURM.   @lipari, do you know what scheduling policies are currently supported by SLURM?  Does it now support conservative & hybrid backfilling?

dongahn:
Thanks @SteVwonder.

Context for @lipari:  an easy verification mechanism of our backfill policy implementations will significantly add to reaching our [CTS-1 scheduler milestone](https://github.com/flux-framework/distribution/issues/14). 

Verifying FCFS is a no brainer -- we can look at each job's `starting-time` timestamp within `lwj.$jobid` and make sure they are ordered. But I have no idea how easily I can verify the correctness of backfill algorithms. I would like to cover EASY, CONSERVATIVE, and Hybrid with a relatively deep depth (currently set to 100).

lipari:
The Slurm settings to enable backfill behavior which will serve as a reference are:
* PriorityType=basic
* SchedulerType=sched/backfill
* SchedulerParameters=bf_...

[...]

PR #175, opened 2016-06-16
Add runtime option to select backfill algorithm

SteVwonder:
Addresses issue #159 

To pass options through to the plugin when launching sched, use plugin-opts.  This list is comma separated. (e.g. `flux module load sched plugin=sched.backfill plugin-opts=reserve-depth=1,foo=bar`)

To pass options through to the plugin when loading just the plugin, specify them as a space separated list. (e.g. `flux module load sched.backfill reserve-depth=1 foo=bar`)

Examples of launching with various backfilling algorithms:
- Conservative
  - `flux module load sched.backfill reserve-depth=-1`
- EASY
  - `flux module load sched.backfill reserve-depth=1`
- Hybrid
  - `flux module load sched.backfill reserve-depth=4`
- Pure
  - `flux module load sched.backfill reserve-depth=0`

### DIFF
--- a/sched/plugin.c
+++ b/sched/plugin.c
@@ -87,6 +87,11 @@ static struct sched_plugin *plugin_create (flux_t h, void *dso)
         flux_log (h, LOG_ERR, "can't load reserve_resources: %s", dlerror ());
         goto error;
     }
+    plugin->process_args = dlsym (dso, "process_args");
+    if (!plugin->process_args || !*plugin->process_args) {
+        flux_log (h, LOG_ERR, "can't load process_args: %s", dlerror ());
+        goto error;
+    }
     plugin->dso = dso;
     return plugin;
 error:
@@ -213,7 +218,12 @@ static void insmod_cb (flux_t h, flux_msg_handler_t *w,
     }
     if (sched_plugin_load (sploader, path) < 0)
         goto done;
+
+    if (sploader->plugin->process_args (sploader->h, argz, argz_len) < 0) {
+        goto done;
+    }
     rc = 0;
+
 done:
     if (flux_respond (h, msg, rc < 0 ? errno : 0, NULL) < 0)
         flux_log_error (h, "%s: flux_respond", __FUNCTION__);
--- a/sched/plugin.c
+++ b/sched/plugin.c
@@ -56,40 +56,50 @@ static void plugin_destroy (struct sched_plugin *plugin)
 static struct sched_plugin *plugin_create (flux_t h, void *dso)
 {
     int saved_errno;
+    char *strerr = NULL;
     struct sched_plugin *plugin = malloc (sizeof (*plugin));
+
     if (!plugin) {
         errno = ENOMEM;
         goto error;
     }
     memset (plugin, 0, sizeof (*plugin));
+    dlerror (); // Clear old dlerrors
+
     plugin->sched_loop_setup = dlsym (dso, "sched_loop_setup");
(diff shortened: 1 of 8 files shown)


garlick:
The arg passing mechanism looks good to me.  Nice work!

grondo:
Nice work. My only comment would be that, eventually, you might want to support the strings "conservative", "EASY", "hybrid", "pure" as a convenience options over the `reserve-depth=N` argument. (If that even makes sense, if it doesn't then feel free to ignore me).

Also, it would be interesting to be able to query the active algorithm at runtime (admittedly I didn't check to see if you already can). Certainly not necessary in this PR though.

SteVwonder:
@grondo & @garlick: I meant to ask you guys what you think the best way to query the plugin is.  Is it currently possible to hook into `flux module info` calls via a callback?  If so, I could print the current plugin parameters when a user runs `flux module info sched.backfill`.

However it is done, this would also make testing in sharness more complete (I currently don't test that the reserve-depth is correctly parsed).

SteVwonder:
@grondo:  that is a good idea.  And if the user specifies "hybrid", we can accept an additional parameter that defines the depth (since hybrid is a depth x where 0 < x < INT_MAX).  If the querying of the plugin is reasonably easy, I'll implement that first, then implement this.

SteVwonder:
Also, I should state that I have kicked the can on verifying the correctness of these backfilling algorithms.  Broader correctness checks has been documented in #168.

grondo:
I haven't looked in  a bit, but I don't think the module RFC defines any kind of query mechanisms for parameters passed to, or active within, a loaded module. That might be interesting to try to define. In other modules I've just made a `service.info` or `service.query` topic for which the modules replies with service-specific information. For that to work you need a command to send to query and unmarshall the data though... I'm sure @garlick will have some thoughts here...

garlick:
In this case the sched modules aren't "messaging actors", e.g. they don't have their own reactor or thread, so that might be a bit tricky to get done via the module RFC and the "standard" module command.

Maybe sched could use a purpose built status interface/query mechanism down the road?  My guess is that's an exercise for later once use cases develop...

lipari:
@SteVwonder, Your PR looks good to me.  It is good to merge once @dongahn checks it over.

dongahn:
@SteVwonder: Nice work and thank you again for taking on this. Only nit is the error checking with dlsym(). Man on dlsym(3) says: 

> Since the value of the symbol could actually be NULL (so that a NULL
       return from dlsym() need not indicate an error), the correct way to
       test for an error is to call dlerror(3) to clear any old error
       conditions, then call dlsym(), and then call dlerror(3) again, saving
       its return value into a variable, and check whether this saved value
       is not NULL.

I know you were just following the upstream patterns. But while you are at it, could you quickly change the error checking with dlerror() and push?

SteVwonder:
@dongahn: I am missing the motivation for wanting to differentiate between a missing symbol and a symbol that is NULL.  In either case, the sched module will be unable to use that dynamically loaded function.  So it seems reasonable to me to error out in both cases.

dongahn:
@SteVwonder: I believe the current code should just work fine for all cases. I was just suggesting this to  make our code a bit more readable as dlerror() is a more standard way of handling dlopen errors.

SteVwonder:
Ok.  I just pushed an additional commit.  Let me know if that is what you were thinking.

dongahn:
Thanks. Looks good.

PR #179, opened 2016-08-13
travis updates, and NEWS for 0.2.0

grondo:
This PR updates flux-sched travis-ci with new dependencies from flux-core, adds gcc-4.9 and clang-3.8, as well as a new `cppcheck` target for testing (along with a couple fixes required for cppcheck to pass, these should be vetted by @SteVwonder).

Additionally, a `deploy:` section is added to auto-upload travis releases on tags.

Finally, a `NEWS.md` file is created with content for the upcoming `0.2.0` release. (Actually I could split that out into a separate PR if people want)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,3 +82,24 @@ after_failure:
  - find . -name t[0-9]*.output -print0 | xargs -0 -I'{}' sh -c 'printf "\033[31mFound {}\033[39m\n";cat {}'
  - find . -name *.broker.log -print0 | xargs -0 -I'{}' sh -c 'printf "\033[31mFound {}\033[39m\n";cat {}'
 
+before_deploy:
+  # Get anchor (formatted properly) and base URI for latest tag in NEWS file
+  - export ANCHOR=$(sed -n '/^flux-sched version/{s/\.//g; s/\s/-/gp;Q}' NEWS.md)
+  - export TAG_URI="https://github.com/${TRAVIS_REPO_SLUG}/blob/${TRAVIS_TAG}"
+  - export TARBALL=$(echo flux-sched*.tar.gz)
+  - ls -l $TARBALL
+  - echo "Deploying tag ${TRAVIS_TAG} as $TARBALL"
+
+deploy:
+  provider: releases
+  skip_cleanup: true
+  file: $TARBALL
+  prerelease: true
+  body: "View [Release Notes](${TAG_URI}/NEWS.md#${ANCHOR}) for flux-sched ${TRAVIS_TAG}."
+  api_key:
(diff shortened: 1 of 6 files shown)


grondo:
Also, added in support for codecov.io as in flux-core

SteVwonder:
:+1: Changes look good to me

grondo:
Oops, don't merge yet the `cppcheck` check is sometimes checking all the dependencies source as well

grondo:
Ok, I moved dep-builder run outside of the builddir and things are better now I hope.

dongahn:
Nice work @grondo!

A nit: [e72daa](https://github.com/flux-framework/flux-sched/pull/179/commits/3216cdccf4e1681128833fe6a8e4af4c27e72daa) has a dependency to NEWS.md which don't exit yet?

dongahn:
Getting a failure from codecov/patch.  Covering error branches is not always practical. Is there anyway we can adjust codecov's target behavior?

grondo:
Ah yeah, I'll reorder commits  so use of NEWS.md comes after its addition. 

Also, I'll check on making codecov.io coverage error into a warning.

Thanks!

grondo:
I can change the target for codecov "patch" commit status to be `0%`, however there isn't a fast rule that we can't merge commits that have failed CI checks, so I think it is ok to merge when coverage test fails, but we at least know why.

Therefore, it might be nice to get a "red flag" when coverage of added code is not at least something like 50%. In most cases it will be obvious if the code is all error handling or otherwise can't be tested, and we can ignore the `codecov/patch` error.

Does that sound reasonable? Like I said we could also set target to 0% if you want, I think/hope codecov/patch will still comment and we'll notice if the coverage is unexpectedly low and we can at least note the number during review.

dongahn:
I think 50% target makes sense. With 0%, people in particular like me :-) won't even pay attention. At least the red flag will motivate us to take a look and ensure the reason for low coverage is sensible (like this time.)

grondo:
Ok, @dongahn, I made the suggested changes -- I split the NEWS->NEWS.md update into two commits, and left the update for 0.2.0 as the final commit.

dongahn:
Thank you so much @grondo. Unless you have something else to do for this, I will merge, when we get the result from all the checks.

grondo:
Great then I will try a tag. warning though, I've been having trouble with Travis all day today

grondo:
Hm, the threshold for codecov didn't seem to take,  but we could look into this later.

lipari:
I reviewed the PR and it looks great!  Thanks again, @grondo!  @dongahn, merge when you're ready.

PR #189, opened 2016-09-02
Graph work

lipari:
Introduce flow graph support to the resrc
    
Expanded the resrc module to support associations to hierarchies beyond just the physical, composite hierarchy.  Multiple hierarchies are now represented by the new flow_graph module.

Yet to do is expand the sched module to exercise scheduling of the flow graphs.  However, I wanted to provide an early preview of the changes.

### DIFF
--- a/conf/hype.lua
+++ b/conf/hype.lua
@@ -12,3 +12,53 @@ Hierarchy "default" {
                }
     }
 }
+
+Hierarchy "power" {
+    Resource{ "powerpanel", name="ppnl", size = 1000, children = {
+      Resource{ "pdu", name = "pdu1", id = "1", size = 500,
+        children = { ListOf{ Node,
+                       ids = "201-299",
+                       args = { basename = "hype",
+                                sockets = { "0" },
+                                size = 10 }
+                     }
+                   }
+        },
+      Resource{ "pdu", name = "pdu2", id = "2", size = 500,
+        children = { ListOf{ Node,
+                      ids = "300-354",
+                      args = { basename = "hype",
+                            sockets = { "0" },
+                            size = 20  }
+                     }
+                   }
+        }
+      }}
+}
+
+Hierarchy "bandwidth" {
+    Resource{ "filesystem", name="pfs", size = 90000, children = {
+      Resource{ "gateway", name = "gateway_node_pool", size = 90000, children = {
+        Resource{ "core_switch", name = "core_switch_pool", size = 144000, children = {
+          Resource{ "switch", name = "edge_switch1",  id = "1",size= 72000,
+            children = { ListOf{ Node,
+                         ids = "201-299",
+                         args = { basename = "hype",
+                                  sockets = { "0" },
+                                  size = 4000 }
+                           }
+                       }
+            },
+          Resource{ "switch", name = "edge_switch2",  id = "2", size= 72000,
+            children = { ListOf{ Node,
+                         ids = "300-354",
+                         args = { basename = "hype",
+                                  sockets = { "0" },
+                                  size = 4000  }
+                           }
+                       }
+            }
+        }}
+      }}
+    }}
+}
--- a/rdl/RDL/types/Node.lua
+++ b/rdl/RDL/types/Node.lua
@@ -41,6 +41,7 @@ function Node:initialize (arg)
           basename = basename,
           id = id,
           name = name,
+          size = arg.size,
           properties = arg.properties or {},
           tags = arg.tags or {}
         }
--- a/resrc/Makefile.am
+++ b/resrc/Makefile.am
@@ -8,9 +8,9 @@ SUBDIRS = . test
 
 noinst_LTLIBRARIES = libflux-resrc.la
 
-noinst_HEADERS = resrc.h resrc_tree.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h
 
-libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_reqst.c
+libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c
 libflux_resrc_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 libflux_resrc_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/src/common/libutil/libutil.la \
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -1044,36 +1044,36 @@ bool resrc_match_resource (resrc_t *resrc, resrc_reqst_t *request,
                            bool available)
 {
     bool rc = false;
-    char *rproperty = NULL;     /* request property */
-    char *rtag = NULL;          /* request tag */
+    char *rproperty = NULL;                             /* request property */
+    char *rtag = NULL;                                  /* request tag */
+    resrc_t *reqst_resrc = resrc_reqst_resrc (request); /* request's resrc */
 
-    if (!strcmp (resrc->type, resrc_reqst_resrc (request)->type)) {
-        if (zhash_size (resrc_reqst_resrc (request)->properties)) {
+    if (reqst_resrc && !strcmp (resrc->type, reqst_resrc->type)) {
+        if (zhash_size (reqst_resrc->properties)) {
             if (!zhash_size (resrc->properties)) {
                 goto ret;
             }
             /* be sure the resource has all the requested properties */
             /* TODO: validate the value of each property */
-            zhash_first (resrc_reqst_resrc (request)->properties);
+            zhash_first (reqst_resrc->properties);
             do {
-                rproperty = (char *)zhash_cursor (
-                    resrc_reqst_resrc (request)->properties);
+                rproperty = (char *)zhash_cursor (reqst_resrc->properties);
                 if (!zhash_lookup (resrc->properties, rproperty))
                     goto ret;
-            } while (zhash_next (resrc_reqst_resrc (request)->properties));
+            } while (zhash_next (reqst_resrc->properties));
         }
 
-        if (zhash_size (resrc_reqst_resrc (request)->tags)) {
+        if (zhash_size (reqst_resrc->tags)) {
             if (!zhash_size (resrc->tags)) {
                 goto ret;
             }
             /* be sure the resource has all the requested tags */
-            zhash_first (resrc_reqst_resrc (request)->tags);
+            zhash_first (reqst_resrc->tags);
             do {
-                rtag = (char *)zhash_cursor (resrc_reqst_resrc (request)->tags);
+                rtag = (char *)zhash_cursor (reqst_resrc->tags);
                 if (!zhash_lookup (resrc->tags, rtag))
                     goto ret;
-            } while (zhash_next (resrc_reqst_resrc (request)->tags));
+            } while (zhash_next (reqst_resrc->tags));
         }
 
         if (available) {
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -1193,21 +1193,19 @@ static int resrc_allocate_resource_now (resrc_t *resrc, int64_t job_id)
     size_t *size_ptr;
     int rc = -1;
 
-    if (resrc && job_id) {
-        if (resrc->staged > resrc->available)
-            goto ret;
+    if (resrc->staged > resrc->available)
+        goto ret;
 
-        id_ptr = xasprintf ("%"PRId64"", job_id);
-        size_ptr = xzmalloc (sizeof (size_t));
-        *size_ptr = resrc->staged;
-        zhash_insert (resrc->allocs, id_ptr, size_ptr);
-        zhash_freefn (resrc->allocs, id_ptr, free);
-        resrc->available -= resrc->staged;
-        resrc->staged = 0;
-        resrc->state = RESOURCE_ALLOCATED;
-        rc = 0;
-        free (id_ptr);
-    }
+    id_ptr = xasprintf ("%"PRId64"", job_id);
+    size_ptr = xzmalloc (sizeof (size_t));
+    *size_ptr = resrc->staged;
+    zhash_insert (resrc->allocs, id_ptr, size_ptr);
+    zhash_freefn (resrc->allocs, id_ptr, free);
+    resrc->available -= resrc->staged;
+    resrc->staged = 0;
+    resrc->state = RESOURCE_ALLOCATED;
+    rc = 0;
+    free (id_ptr);
 ret:
     return rc;
 }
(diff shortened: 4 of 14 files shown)


SteVwonder:
This looks really good!

Are there plans to support temporal flow allocations, that is to say, are there plans for resrc_flow_allocate_in_time and resrc_flow_available_at_time functions?  As far as I can tell, that is the only feature that is still missing before we can port the IO-Aware scheduling done for the HPDC paper to this model.

resrc/resrc.c
```diff
@@ -445,6 +454,41 @@ resrc_tree_t *resrc_phys_tree (resrc_t *resrc)
...
+}
+
+int resrc_twindow_insert (resrc_t *resrc, const char *key, void *item)
+{
+    int rc = zhash_insert (resrc->twindow, key, item);
+    zhash_freefn (resrc->twindow, key, free);
+    return rc;
+}
+
+int resrc_graph_insert (resrc_t *resrc, const char *type, resrc_flow_t *flow)
```
dongahn: a nit: probably "name" is better than "type" because a resrc object can belong to multiple graphs of the same type? E.g., a CN is connected to multiple storage hierarchy (lscratchb, lscratcha, etc.)

dongahn:
Nice work, @lipari! 

Before looking at the code with a fine toothed comb, just wanted to generate a bit of verbal exchanges. One high level observation I have is that there is enough of commonality between resource tree and flow, which then made me wonder if we want to refactor the common graph operation and data structure as the "Base" class and then derive tree and flow classes from that.

If we borrow some of OOP design patterns this way, I can see the users of resrc "dynamically cast" each resrc object as a vertex in a particular graph only when they want to manipulate the resrc object in that specific graph context while keeping all of the common operations like graph traversal and what not as generic graph API, if this makes sense.

resrc/resrc.c
lipari: Yes, I originally had "name" but switched it to "type" to be more fundamental.  But you make a good point, so I will change it to "name".

lipari:
@dongahn , rebased, repushed, and ready for review when green.

lipari:
> Are there plans to support temporal flow allocations, that is to say, are there plans for resrc_flow_allocate_in_time and resrc_flow_available_at_time functions?

It's actually already there.  The existing resrc_match_resource() function was extended to visit all of a resource's associated graphs and make sure each requested graph flow is available.  For each requested flow amount, the resrc_flow_available() function runs a resrc_available() for now-based or resrc_walltime_match() for time-based check to see whether enough capacity is available to service the job request.

Similarly, resrc_stage_resrc() was extended to visit each requested flow graph and stage the requested amount.  It recursively visits all parents to stage the requested flow amount from each source.

Finally, the existing resrc_allocate_resource()/resrc_reserve_resource() was extended to allocate/reserve now or in_time any staged graph resource.  Here also, each flow parent is visited to allocate/reserve the staged amount of flow.

dongahn:
Thanks @lipari. 

I will review this code shortly. Meanwhile, here are some of the references for the visitor/matcher design patterns which I think can help manage increasing complexity. Please feel free to move this to a separate issue if you want.

As discussed offline, I believe providing matcher and visitor patterns from resrc and the friends can make resrc APIs more ease to use and robust, and make it be better adaptable when @trws's job spec design comes along, as mentioned in this [comment](https://github.com/flux-framework/rfc/pull/53#issuecomment-234016452)


[ROSE's AstPrePostProcessing class](http://rosecompiler.org/ROSE_HTML_Reference/classAstPrePostProcessing.html) is one of the examples. It is an abstract class that defines` preOrderVisit` and `postOrderVisit` interfaces for an Abstract Syntax Tree (AST). One can override (e.g., C++ lingo -- meaning, implement) these interfaces in his or her AST analysis class to plug their analysis into the system. Then, this class takes care of an AST traversal and lets the user-defined analysis to be invoked `on post visit` and `on pre visit` during this depth first traversal. 

I actually used this interface for the STAT project. One example code can be found in STAT's TemporarlOderAnalysisAPI in our internal bitbucket repo ([here](https://lc.llnl.gov/bitbucket/projects/STAT/repos/temporalorderingapi/browse/src/TO_LexicographicalOrdering.C#160)).  I used this pattern to implement a way to identify all of the lexical scopes in a C/C++ source file, which contain a specific program point or points. This static information,  along with the runtime stack trace, then allowed me to temporally order points of execution among MPI processes (i.e., determining relative progress). 

Our case can be viewed in a similar way. Our scheduling plugins traverse each resrc hierarchy and act on a vertex that they wants to operate on. But plugins don' t need to implement raw traversal logic by themselves.  Since our language is C, we can't use a OOP concept. But I think a callback that gets invoked on each node visit can bring us similar benefits.  

If we abstract out data structures and visitor/match patterns from resource details (e.g., node type, PDU type, switch type and etc), a scheduler plugin writer can focus on scheduling operations like matched/unmatched, allocate/deallocate etc while leaving the traversal across potentially many different resource hierarchy to the visitor pattern provider.  Good separations of concerns. 

If we were to do this, I would see if we can turn some of the tree traversing logic in our current scheduler plugin to use such visitor/matcher patterns. 

BTW, Clang also has Matcher patterns [in here](http://clang.llvm.org/docs/LibASTMatchersReference.html) though I don't have hands-on experience.

trws:
Clang matchers are, for the record, one of the most awesome features ever written for manipulating C/C++.

Anyway, actual on-topic comment, the classic [twalk](https://www.gnu.org/software/libc/manual/html_node/Tree-Search-Function.html) style pattern could work well for this, but I'd probably want to simplify it to a DFS only where the function pointer's return value informs the walker as to whether to process children, skip children and continue, or abort completely.  I say DFS specifically because a DFS from the root of the hardware tree, restricted to the hardware tree, can most rapidly identify branches to skip in that ordering IIRC.

lipari:
Spawned a [new issue](https://github.com/flux-framework/flux-sched/issues/193) to address the pattern matching suggestions above.

resrc/resrc.c
```diff
@@ -1159,10 +1159,28 @@ bool resrc_match_resource (resrc_t *resrc, resrc_reqst_t *request,
     return rc;
 }
 
-void resrc_stage_resrc (resrc_t *resrc, size_t size)
+int resrc_stage_resrc (resrc_t *resrc, size_t size)
```
dongahn: Maybe I can use this opportunity to clarify on the concept of staging. I also see the later changes to stage function. What does staging exactly do?

rdl/RDL/types/Node.lua
```diff
@@ -41,6 +41,7 @@ function Node:initialize (arg)
           basename = basename,
           id = id,
           name = name,
+          size = arg.size,
```
dongahn: I was initially confused as to why Node should have size but now I understand. Seems this needs to be fleshed out in RFC/4 at some point though.  I already see your [rfc#56}(https://github.com/flux-framework/rfc/issues/56). Nice work.

dongahn:
@lipari: I enjoyed reviewing this code and talking to you :-) Nice work and concept! I believe this will be a good platform to explore more advanced scheduling techniques before further factoring things for ease-of-use. I do believe, though, porting @SteVwonder's IO aware schedulers can bring an interesting perspective for this. I'm not asking for that as part of this PR of course. 

This is also a future reference, but one thing that I realized is, the visitor patterns we discussed above may need to be a bit more extensive. W/ this hierarchy overlay scheme, a general resource allocation pattern with each flow hierarchy would look like:

- find a node resource
- walk towards the root in the physical hierarchy to find the first ancestor node that has a link to the target flow resource
- jump to that flow resource node and allocate the request amount 
- walk towards the root and allocate the requested amount for each and all of the ancestor nodes

Hope this complexity can be hidden behind some traversal APIs.

dongahn:
Since we only have strict trees for now, fetching all ancestor vertices of a node should be cheap though. In fact your resrc indexing should just do the trick?

Up tree operations should be scalable too as you only have to deal with log (vertices). DAG should also be reasonable to handle. General graph is questionable...

lipari:
> What does staging exactly do?

Staging decouples the selection process from the allocation/reservation operations.  As the candidate resources that `find()` returns are being evaluated by `select()`, each selected resource is "staged".  If enough resources are found to meet the request criteria, those staged resources are allocated.  However, if `select()` fails to find all the requested resources, those staged resources are reserved.

lipari:
Regarding resrc_table_destroy (void), it is an internal table.  In theory, it will be empty once the last resource is destroyed.  However, I had neglected to actually call it in the right places.  New push on the way...

resrc/Makefile.am
```diff
@@ -8,9 +8,9 @@ SUBDIRS = . test
 
 noinst_LTLIBRARIES = libflux-resrc.la
 
-noinst_HEADERS = resrc.h resrc_tree.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h
 
-libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_reqst.c
+libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c
```
lipari: I'm thinking the following satisfied that need:
```
-noinst_HEADERS = resrc.h resrc_tree.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h
```

rdl/RDL/types/Node.lua
lipari: Yes.  With the introduction of `resrc_flow`  I'm re-purposing the original resrc_t structure to model each vertex in the flow graph.  Hence, the RDL's Node class needed to be given to ability to represent a size > 1.  The Node's "size" will represent current draw when used as the `flow_resrc` of a power flow graph.

lipari:
> Since we only have strict trees for now, fetching all ancestor vertices of a node should be cheap though. In fact your resrc indexing should just do the trick?
Up tree operations should be scalable too as you only have to deal with log (vertices). DAG should also be reasonable to handle. General graph is questionable...

My goal at this point was to design a solution that addressed the need expressed in [Issue 56](https://github.com/flux-framework/rfc/issues/56).  This is a complex undertaking that extends the scheduling find/select/allocate operations to every graph associated with a resource.

Now that this is done, we may find it doesn't scale well.  At that point, we can look for shortcuts that provide the essential functionality without the exhaustive visitations in the current design.

lipari:
New push with the above changes turned green.  Ready for re-review.

dongahn:
> Staging decouples the selection process from the allocation/reservation operations. As the candidate resources that find() returns are being evaluated by select(), each selected resource is "staged". If enough resources are found to meet the request criteria, those staged resources are allocated. However, if select() fails to find all the requested resources, those staged resources are reserved.

Thank you for the clarification. Neat! I have some other feedback regarding the use of `find` and `select` and I will post that to Issue #193.

resrc/Makefile.am
dongahn: Ah. You are absolutely right. Sorry for the false alarm.

dongahn:
> Regarding resrc_table_destroy (void), it is an internal table. In theory, it will be empty once the last resource is destroyed. However, I had neglected to actually call it in the right places. New push on the way...

I guess my main concern was on the interface itself. If resrc_table is internal, can we live with not exposing it to the public. I also have a slight preference to mark the definition of `resrc_hash` static [here](https://github.com/lipari/flux-sched/blob/8c89091314e6a7701a045fb51ed64781a5ffdde7/resrc/resrc.c#L65).

dongahn:
> My goal at this point was to design a solution that addressed the need expressed in Issue 56. This is a complex undertaking that extends the scheduling find/select/allocate operations to every graph associated with a resource.

> Now that this is done, we may find it doesn't scale well. At that point, we can look for shortcuts that provide the essential functionality without the exhaustive visitations in the current design.

Great direction! I am a bit preoccupied with scalability at the moment so looked at the current paradigm through a perf/scalability lens. Again, I will comment mine to other issue(s).

rdl/RDL/types/Node.lua
dongahn: Yeah I figured this one out. Cool!

lipari:
> I guess my main concern was on the interface itself. If resrc_table is internal, can we live with not exposing it to the public. I also have a slight preference to mark the definition of resrc_hash static here. 

At the moment, every resource and resource tree that a client uses can be destroyed and its memory reclaimed.  The `resrc_hash` table will be empty once the last resource is destroyed.  Leaving it around would not be that much of a leak.  Nonetheless, `resrc_table_destroy()` is the way clients can say, "I'm done with all those resources I was using and don't expect to need them again.  Now get rid of that internal table you kept".

grondo:
> , resrc_table_destroy() is the way clients can say, "I'm done with all those resources I was using and don't expect to need them again. Now get rid of that internal table you kept".

A more typical way to handle this same approach is to have a `resource_library_init (void)` and `resource_library_fini (void)` or calls that initialize and destroy arbitrary internal state, with big warnings that these calls are not thread-safe. The benefit is that you abstract knowlege of the internal state away from the user, and internal state implementation could be changed without turning the function names into a misnomer.

I would argue in many cases it is preferable to return state or context object to the user, as it allows more than one state (resource tables in this case?) to be created per process, and MT safety is pushed off onto the caller. Something like `resrc_ctx_t * resrc_context_create (void);` and `void resrc_context_destroy (resrc_ctx_t *ctx);`

However, I have not looked in depth at the implementation here and perhaps it does not make sense in this case. Just wanted to throw in my 2cent.

lipari:
@grondo , I edited my earlier comment with a similar suggestion for `resrc_fini()`.  At this point, a `resrc_init()` would be a no-op; and functionally unnecessary.  As far as MT enhancements, I'll create a separate issue to track that idea.

lipari:
New push with the latest change turned green. Ready for re-review.

trws:
WRT the table, is there a strong reason for it to be an internal rather than a returned construct? Do we necessarily want to restrict resrc to managing exactly one pool of resources per address space?

lipari:
The `resrc_hash` is a temporary solution to link the same resource that is present different hierarchies.  Once we have the new canonical job spec work in place, we can rely on a single resource UUID to tie a node from the physical composite hierarchy to the same node resource in the power or bandwidth hierarchy.  This PR, leverages the RDL config reader and doesn't support a common UUID for the same resource.

conf/hype.lua
```diff
@@ -12,3 +12,53 @@ Hierarchy "default" {
...
+      Resource{ "pdu", name = "pdu1", id = "1", size = 500,
+        children = { ListOf{ Node,
+                       ids = "201-299",
+                       args = { basename = "hype",
+                                sockets = { "0" },
+                                size = 10 }
+                     }
+                   }
+        },
+      Resource{ "pdu", name = "pdu2", id = "2",
```
dongahn: I just noticed this. Should pdu2 also have a size field (e.g., `size = 500`)?

conf/hype.lua
```diff
@@ -12,3 +12,53 @@ Hierarchy "default" {
...
+        children = { ListOf{ Node,
+                      ids = "300-354",
+                      args = { basename = "hype",
+                            sockets = { "0" },
+                            size = 20  }
+                     }
+                   }
+        }
```
dongahn: Just an on-topic comment (i.e., not asking a change for this PR): 

In terms of resource specification in a condensed manner, seems it would be extremely helpful if I can express the quantities or attributes of a nested resource as a function of the quantities/attributes of the nesting resources. 

In many cases, IDs of the children may be the only thing that falls into this category, and I have to believe we can leverage the regularity of the ID space on a k-ary tree for this... If id increases monotonically from the root to the fringes of the hierarchy in a breadth first manner, for example, children ids of a parent = { ((parent_id + 1) * k - 1),  ((parent_id + 1) * k - 1 + 1), ..., (((parent_id + 1) * k - 1 + (k-1)) }, assuming the id space starts from 0. Some variations of a scheme like this should cover lots of basis I think.

resrc/resrc.c
```diff
@@ -61,11 +61,24 @@ struct resrc {
     zhash_t *twindow;
 };
 
+zhash_t *resrc_hash = NULL;
```
dongahn: If you are going to push another push, could that PR also mark this `static`? I know the symbol won't leak because of other linker trick, but this way source code itself can tell the reviewer about the intent of the table (Internal)?

conf/hype.lua
lipari: Good catch!

conf/hype.lua
lipari: Yes, I agree that would be helpful for certain resource hierarchies.

resrc/resrc.c
lipari: Another good catch!  Rebased commits just pushed.

dongahn:
@lipari: Thank you for the work again! Merging.

Issue #182, opened 2016-08-22
A linear job search performs poorly for HTC load


dongahn:
We need an optimization for the linear job search scheme as captured in [flux-distribution#14](https://github.com/flux-framework/distribution/issues/14): specifically, [this code](https://github.com/flux-framework/flux-sched/blob/master/sched/sched.c#L389). Once an optimization is done, please rerun some tests.

dongahn:
The scheduler currently has three job queues: pending, running and complete queues. And it only performs very basic operations on these queues: find a job across all three queues; append a job to a queue; and finally move a job from one queue to another (e.g., from the pending queue to running queue).

An obvious optimization is to add indexing based on job id. 

Given the scheduler only needs limited operations on our queues, an easy solution would be to add a `zhash_t` object and index each job into it on the first enqueue event (i.e., into the pending queue). Then, the remaining change will be to remove a job from this hash table when it is reaped from the complete queue -- which isn't implemented yet (So I will mark it as TODO in the code); and make sure the table is initialized/finalized properly. 

This will add a small run overhead on each job append, but will turn O(N) to O(1) for `q_find_job()` where N is the number of jobs in the queues. And also slight increase in memory of course.

Alternatively, you can have a lookup table per queue, but that seems an overkill at this point and adds higher overheads as this will incur one additional delete on each job move.

Also one can introduce a composite data structure which abstracts out these operations (traditional enqueue + constant find). But I propose that we do this if and when more complex operations are needed on our job queues.

dongahn:
This optimization alone doesn't seem to help much:

- Jobs Executed Per Second (JEPS) at <2 nodes, 1 broker per node, unit sizing policy, non-mpi sleep, FCFS>: 1.77

I realized that `q_move_to_rqueue()` and `q_move_to_cqueue()` can also perform poorly as they use `zlist_remove()`, which also uses  a linear search as in [here](https://github.com/zeromq/czmq/blob/master/src/zlist.c#L282). This needs to be optimized too. Now that I think about this, these may not be that bad w/ FCFS scheduling. Perhaps, #183 is a bigger fish to fry.

lipari:
Closed via #190

lipari:
Close via #190

Issue #183, opened 2016-08-22
Quantify the impact of and optimize the schedule loop's job iteration scheme


dongahn:
We may need an optimization for [this code](https://github.com/flux-framework/flux-sched/blob/master/sched/sched.c#L1396)  as captured in flux-distribution#14: specifically, this code. Once an optimization is done, please rerun tests.

dongahn:
To avoid premature optimization, we will probably want to look at the performance impacts of this on conservative and hybrid backfill algorithms. In terms of peeling this onion in a right way, it would be good to add optimization for #182 first and see the impact of this issue on conservative and hybrid after that.

lipari:
As suggested in [distribution #14](https://github.com/flux-framework/distribution/issues/14#issuecomment-241521858), I recommend adding to the sched comms module a setting called "queue-depth" that will impose a limit on the number of jobs considered in each scheduling loop.  This would apply for both the FCFS and backfill plugins.

dongahn:
@lipari:

This is reasonable, but as you and I discussed (before I realized I had a meeting with Tapasya), my main question is how the queue depth should play with the reservation depth, which we currently use for hybrid backfill scheduling. If that makes sense. 

Should queue-depth take precedence over reservation depth? Then, if queue-depth < reservation depth in hybrid backfill, we print out an error or warning message: this policy is not fully supported? Or would it be better to use an alternative parameter which tells the scheduler to consider x number of jobs after reservation depth is satisfied? As you argue, the downside of this is, it may not work with FCFS as its concept for reservation is implicit. But at least it won't change the policy.

In any case, one thing we will probably want to do is to introduce a max for the pure conservative backfill mode. (Make it also configurable). We probably need to understand what that max should be in a stress load like  [distribution #14](https://github.com/flux-framework/distribution/issues/14).

The other reasonable optimization is not to schedule for each and every job/resource event.

dongahn:
OK. I did a quick test for this by just limiting the schedule requested jobs to 100 and this makes a pretty substantial difference (2x):

Jobs Executed Per Second (JEPS) at <2 nodes, 1 broker per node, unit sizing policy, non-mpi sleep, FCFS>: 3.5.

dongahn:
In order to reason about the performance in a more comprehensive way, I am pretty convinced that I will need a performance model for our scheduler. I will spend some time to come up w/ one.

lipari:
[...]

PR #190, opened 2016-09-03
optimizations: add queue-depth and better job search support

dongahn:
This PR adds two enhancements to address some of the performance issues discovered from the phase I of our CTS scheduler scalability milestone work [flux-distribution#14](https://github.com/flux-framework/distribution/issues/14).

 - Queue-depth support that allows sched to only consider a limited, configurable number of jobs for scheduling on each scheduling event;
 - Faster job search using an job index instead of a linear search over all of the job queues.

More details can be found in the individual commit messages. Fix Issue #182 and #183.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -112,6 +112,7 @@ typedef struct {
 /* TODO: Implement prioritization function for p_queue */
 typedef struct {
     flux_t        h;
+    zhash_t      *job_index;          /* For fast job lookup for all queues*/
     zlist_t      *p_queue;            /* Pending job priority queue */
     zlist_t      *r_queue;            /* Running job queue */
     zlist_t      *c_queue;            /* Complete/cancelled job queue */
@@ -258,6 +259,7 @@ done:
 static void freectx (void *arg)
 {
     ssrvctx_t *ctx = arg;
+    zhash_destroy (&(ctx->job_index));
     zlist_destroy (&(ctx->p_queue));
     zlist_destroy (&(ctx->r_queue));
     zlist_destroy (&(ctx->c_queue));
@@ -283,6 +285,8 @@ static ssrvctx_t *getctx (flux_t h)
     if (!ctx) {
         ctx = xzmalloc (sizeof (*ctx));
         ctx->h = h;
+        if (!(ctx->job_index = zhash_new ()))
+            oom ();
         if (!(ctx->p_queue = zlist_new ()))
             oom ();
         if (!(ctx->r_queue = zlist_new ()))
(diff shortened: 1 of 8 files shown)


sched/sched.c
```diff
@@ -425,27 +430,27 @@ static int append_to_pqueue (ssrvctx_t *ctx, JSON jcb)
         flux_log (ctx->h, LOG_ERR, "failed to append to pending job queue.");
         goto done;
     }
+    key = xasprintf ("%"PRId64"", jid);
+    if (zhash_insert(ctx->job_index, key, job) != 0) {
+        flux_log (ctx->h, LOG_ERR, "failed to index a job.");
+        goto done;
+    }
+    /* please don't free the job using job_index; this is just a lookup table */
```
lipari: Don't you need a `zhash_freefn (ctx->job_index, key, free)` here?

sched/sched_backfill.c
```diff
@@ -360,11 +360,27 @@ int process_args (flux_t h, char *argz, size_t argz_len)
...
+        errno = EINVAL;
+    } else if (reservation_depth > sp->queue_depth) {
+        /* Conservative backfill (-1) will still be limited by the queue-depth
+         * but we just treat queue-depth as the limit for it
+         */
+        flux_log (h, LOG_ERR, "Reservation depth (%d) cannot be supported",
+                  reservation_depth);
+        flux_log (h, LOG_ERR, "Smaller than queue depth (%ld)",
+                  sp->queue_depth);
+        rc = -1;
```
lipari: While not a showstopper, it's nice to report one error on one line instead of spilling over to a second line.  It helps when pouring through the logs to spot a problem.  I would suggest, "Ignoring reserve-depth value (%d) - greater than queue-depth (%d)"

dongahn:
@lipari: yes this makes sense. I will update it shortly.

dongahn:
Hmmm looking at the code, it actually returns -1 with EINVAL. So we are not exactly ignoring it. Since this is an error condition, perhaps simply, "LOG_ERR: reserve-depth value (%d) - greater than queue-depth (%d)?" Then, the upper level codes currently fails on receiving -1 from this routine with "failed to process args for..."

lipari:
It's a judgement call as to whether this merits a full-blown error condition.  Yes, a reservation depth greater than queue depth is invalid.  However, for the sake of convenience, I would just emit the warning and allow it to run.  But I can see it both ways, so... your call.

dongahn:
@lipari: Thanks. Agreed that this is a judgement call. 

I have a slight preference not to recover from it because it is pretty easy for the user to reload the scheduler and its plug-in with valid arguments under flux.  For my taste, this seems  a bit better than limping along with a undefined policy. Push is coming.

Issue #185, opened 2016-08-24
Faster submit-event handling for high throughput loads


dongahn:
This problem was discovered as part of [distribution #14](https://github.com/flux-framework/distribution/issues/14#issuecomment-242176848). To avoid the head-of-line blocking issue caused by high throughput workloads, `sched` needs to optimize ways to handle job creation events: [code](https://github.com/flux-framework/flux-sched/blob/master/sched/sched.c#L1452).

garlick:
Just wanted to point out that the code in `src/modules/kvs/kvs.c` for rate limiting commits might be similar to what you need here.  Look for `min_commit_window` and `commit_timestamp`.

In a nutshell, the heavy weight commit operation is only performed in the message handler for the commit request if `min_commit_window` seconds have elapsed since the last one.  Otherwise the request is queued and a timer is started with a timeout computed so the callback is run `min_commit_seconds` from the previous commit operation.  The timer runs the commit operation (consuming all enqueued commits), disables the timer, and updates the time stamp.

An optimization is to use the (recently added) `flux_reactor_now()` function to get the timestamps, as it saves making a time fetching system call many times in the critical path for precision you don't need.

dongahn:
Thanks @garlick. Yes, this sounds exactly like what I need for this optimization.

dongahn:
@garlick: We chatted about this a bit. A condition that I thought could lead to a deadlock situation with an event-based-only scheme is the following. I am probably too close to the problem so I will appreciate another pair of eyes here:

- a job submitted and the timer is installed
- other jobs are submitted before the timer callback is invoked, but the scheduling for these jobs is delayed
- When the timer expires, `schedule_jobs` will be invoked. But this callback won't be able to iterate through all of the jobs in the pending queue because of the [`queue-depth` optimization] (https://github.com/flux-framework/flux-sched/issues/183#issuecomment-241752491), which we need to limit the overhead of `schedule_jobs.` We could consider context-sensitive queue-depth schemes in which `schedule_jobs` uses essentially unbounded queue depth, but we don't probably want to make it too heavyweight either.

At this point, I wasn't clear what this invoked timer callback has to do. It can install the timer again, but then when the callback is invoked again, it seems the callback wants to perform some state check whether the queue state has changed or not. I thought about a couple of ways to do a state check (e.g., job count, etc), but none of them was a correct solution but an approximation.

[...]

PR #191, opened 2016-09-07
optimization: delayed scheduling support 

dongahn:
This is an experimental PR. My apology that this PR contains other optimization items as they are pending for @lipari's review. Commit [22236](https://github.com/flux-framework/flux-sched/commit/22236289334b1c64c2ea6ec64dce5d390b415d2b) is the main commit that I can use your review.  It implements the techniques we discussed at Issue #185. 

Two things:
- This currently breaks a test case: `t2002-easy.t.` I don't understand why changing the timing of invoking schedule_jobs changes the order of a schedule under EASY backfill... There could be some integration I need to do further to make our emulator happy. Since I am not super familiar enough with the emulator code, I thought I could use @SteVwonder's review for this. 

- It is not yet clear whether this actually improves the performance of scheduling with high throughput workloads. I am running some tests with my sched scalability testing rig.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -61,6 +61,10 @@
 #if ENABLE_TIMER_EVENT
 static int timer_event_cb (flux_t h, void *arg);
 #endif
+static void ev_prep_cb (flux_reactor_t *r, flux_watcher_t *w,
+                        int revents, void *arg);
+static void ev_check_cb (flux_reactor_t *r, flux_watcher_t *w,
+                         int revents, void *arg);
 static void res_event_cb (flux_t h, flux_msg_handler_t *w,
                           const flux_msg_t *msg, void *arg);
 static int job_status_cb (const char *jcbstr, void *arg, int errnum);
@@ -114,6 +118,7 @@ typedef struct {
     flux_t        h;
     zhash_t      *job_index;          /* For fast job lookup for all queues*/
     zlist_t      *p_queue;            /* Pending job priority queue */
+    bool          pq_state;           /* schedulable state change in p_queue */
     zlist_t      *r_queue;            /* Running job queue */
     zlist_t      *c_queue;            /* Complete/cancelled job queue */
     machs_t      *machs;              /* Helps resolve resources to ranks */
@@ -121,6 +126,9 @@ typedef struct {
     rdlctx_t      rctx;               /* RDL context */
     simctx_t      sctx;               /* simulator context */
     struct sched_plugin_loader *loader; /* plugin loader */
+    flux_watcher_t *before;
+    flux_watcher_t *after;
+    flux_watcher_t *idle;
 } ssrvctx_t;
 
 
@@ -133,6 +141,7 @@ typedef struct {
 static inline void sched_params_default (sched_params_t *params)
 {
     params->queue_depth = SCHED_PARAM_Q_DEPTH_DEFAULT;
+    params->delay_sched = SCHED_PARAM_DELAY_DEFAULT;
     /* set other scheduling parameters to their default values here */
 }
 
@@ -142,17 +151,28 @@ static inline int sched_params_args (char *arg, sched_params_t *params)
     char *argz = NULL;
     size_t argz_len = 0;
     const char *e = NULL;
+    char *o_arg = NULL;
+    long val = 0;
 
     argz_create_sep (arg, ',', &argz, &argz_len);
     while ((e = argz_next (argz, argz_len, e))) {
         if (!strncmp ("queue-depth=", e, sizeof ("queue-depth"))) {
-            char *o_arg = strstr(e, "=") + 1;
-            long val = strtol(o_arg, (char **) NULL, 10);
+            o_arg = strstr(e, "=") + 1;
+            val = strtol(o_arg, (char **) NULL, 10);
             if ((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN))
                    || (errno != 0 && val == 0))
                 rc = -1;
             else
                 params->queue_depth = val;
+        } else if (!strncmp ("delay-sched=", e, sizeof ("delay-sched"))) {
+            if (!strncmp ((strstr(e, "=") + 1), "true", sizeof ("true")))
+                params->delay_sched = true;
+            else if (!strncmp ((strstr(e, "=") + 1), "false", sizeof ("false")))
+                params->delay_sched = false;
+            else {
+                errno = EINVAL;
+                rc = -1;
+            }
         } else {
            errno = EINVAL;
            rc = -1;
@@ -256,6 +276,37 @@ done:
     return rc;
 }
 
+static int adjust_for_sched_params (ssrvctx_t *ctx)
+{
+    flux_reactor_t *r = NULL;
+    int rc = 0;
+    if (ctx->arg.s_params.delay_sched && !ctx->sctx.in_sim) {
+        if (!(r = flux_get_reactor (ctx->h))) {
+            rc = -1;
+            goto done;
+        }
+        if (!(ctx->before = flux_prepare_watcher_create (r, ev_prep_cb, ctx))) {
+            rc = -1;
+            goto done;
+        }
+        if (!(ctx->after = flux_check_watcher_create (r, ev_check_cb, ctx))) {
+            rc = -1;
+            goto done;
+        }
+        /* idle watcher makes sure the check watcher (after) is called
+           even with no external events delivered */
+        if (!(ctx->idle = flux_idle_watcher_create (r, NULL, NULL))) {
+            rc = -1;
+            goto done;
+        }
+        flux_watcher_start (ctx->before);
+        flux_watcher_start (ctx->after);
+    }
+done:
+    return rc;
+}
+
+
 static void freectx (void *arg)
 {
     ssrvctx_t *ctx = arg;
@@ -277,6 +328,12 @@ static void freectx (void *arg)
         zlist_destroy (&(ctx->sctx.timer_queue));
     if (ctx->loader)
         sched_plugin_loader_destroy (ctx->loader);
+    if (ctx->before)
+        flux_watcher_destroy (ctx->before);
+    if (ctx->after)
+        flux_watcher_destroy (ctx->after);
+    if (ctx->idle)
+        flux_watcher_destroy (ctx->idle);
     free (ctx);
 }
 
@@ -290,6 +347,7 @@ static ssrvctx_t *getctx (flux_t h)
             oom ();
         if (!(ctx->p_queue = zlist_new ()))
             oom ();
+        ctx->pq_state = false;
         if (!(ctx->r_queue = zlist_new ()))
             oom ();
         if (!(ctx->c_queue = zlist_new ()))
@@ -306,6 +364,9 @@ static ssrvctx_t *getctx (flux_t h)
         ctx->sctx.jsc_queue = NULL;
         ctx->sctx.timer_queue = NULL;
         ctx->loader = NULL;
+        ctx->before = NULL;
+        ctx->after = NULL;
+        ctx->idle = NULL;
         flux_aux_set (h, "sched", ctx, freectx);
     }
     return ctx;
@@ -415,7 +476,7 @@ static int plugin_process_args (ssrvctx_t *ctx, char *userplugin_opts)
  *                                                                              *
  *******************************************************************************/
 
-static int append_to_pqueue (ssrvctx_t *ctx, JSON jcb)
+static int q_enqueue_into_pqueue (ssrvctx_t *ctx, JSON jcb)
 {
     int rc = -1;
     int64_t jid = -1;
@@ -432,6 +493,7 @@ static int append_to_pqueue (ssrvctx_t *ctx, JSON jcb)
         flux_log (ctx->h, LOG_ERR, "failed to append to pending job queue.");
         goto done;
     }
+    job->enqueue_pos = (int64_t)zlist_size (ctx->p_queue);
     key = xasprintf ("%"PRId64"", jid);
     if (zhash_insert(ctx->job_index, key, job) != 0) {
         flux_log (ctx->h, LOG_ERR, "failed to index a job.");
@@ -455,9 +517,22 @@ static flux_lwj_t *q_find_job (ssrvctx_t *ctx, int64_t id)
     return j;
 }
 
+static int q_mark_schedulability (ssrvctx_t *ctx, flux_lwj_t *job)
+{
+    if (ctx->pq_state == false
+        && job->enqueue_pos <= ctx->arg.s_params.queue_depth) {
+        ctx->pq_state = true;
+        return 0;
+    }
+    return -1;
+}
+
 static int q_move_to_rqueue (ssrvctx_t *ctx, flux_lwj_t *j)
 {
     zlist_remove (ctx->p_queue, j);
+    /* dequeue operation should always be a schedulable queue operation */
+    if (ctx->pq_state == false)
+        ctx->pq_state = true;
     return zlist_append (ctx->r_queue, j);
 }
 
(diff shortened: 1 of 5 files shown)


garlick:
I had a look at 2223628 and the overall logic looks good!

I didn't see offhand whether a job state change event was triggering schedule_jobs?  Presumably if a job finishes you might need to run?

dongahn:
@garlick: thanks. 

We took care of that job event type very early in our project. When a job completes, that ultimately generates a resoure event and the corresponding callback invoked schedule_jobs.

garlick:
> that ultimately generates a resoure event and the corresponding callback invoked schedule_jobs.

So maybe it just needs to set pq_state now?  Otherwise if many jobs end at once you'll run schedule_jobs over and over when once would suffice?

dongahn:
@garlick: that's a good point! i dont know if that's a performance issue currently but it does seem logical to extend delay scheduling support for that event as well.

 Perhaps what should be done is to make this optimazation a configurable scheuling parameter so that users can customize it according to their workload, and i can explore that as one if the scheduling parameters as well.

SteVwonder:
In the case of what @garlick mentioned, this will affect the ordering of jobs scheduled.  Since when `schedule_jobs` eventually does run, it will have more available resources to work with.  I'm not advocating against it; I think it is a great idea! We will just need to update the test cases accordingly.

SteVwonder:
@dongahn: can you post the observed orders for completed jobs and the output log as a gist?  The observed orders should be an output file of the test case (if you can't find it, I can probably reverse engineer it from the output log).

dongahn:
@SteVwonder: [here is the gist](https://gist.github.com/dongahn/d24c71201b2a288092238959e3f44fa3).

dongahn:
Somehow the observed order is FCFS...

dongahn:
Forced a push: I found a bug in the previous commit. 

It turned out the performance issue above simply goes back to kvs performance degradation with respect to increasing numbers of lwj directories.

As this optimization made job enqueue operatinos go much faster, `flux submit` was also made faster. This submit vs. schedule rate difference then generally resulted in higher numbers of lwj directions (with the job state being submitted), which wreck purge cannot purge. 

When I throttled flux submit by 10 msec per submit, the throughput went up:

```
hype356{dahn}190: cat trash-directory.t0001-sched-scale-20829/*.in*
job-group-id,used-nodes,job-count,JEPS
0,1,1000,26.35337493441962647569
1,1,1000,21.61533798444970970060
2,1,1000,19.87112265464368364671
3,1,1000,18.86497874407384990223
4,1,1000,18.12353355693665741653
5,1,1000,17.36162020445356461928
6,1,1000,16.91140456757420140880
7,1,1000,16.28635609378198487652
8,1,1000,16.10580145785849404185
9,1,1000,16.22012474022050967622
```

With this, I have to think this optimization is generally good to have. If we can address the easy backfill test case issue above, I will add some test cases so that this can go in.

grondo:
Sounds like we'll need to address submission scalability in flux-core sometime soon.

Would changing the `lwj.` directory to 1 level prefix tree help do you think? Or using the least significant couple digits (or suffix tree I suppose) would cause lwj's to be distributed across parent dirs immediately instead of filling a dir and moving to the next directory. Not sure which would be better for performance....

dongahn:
@grondo: Changing lwj data scheme within kvs was exactly what I was thinking about this morning. Thank you for starting this discussion. Let me move this discussion to a separate issue because I believe this can have a pretty good mileage.

SteVwonder:
@dongahn: I pulled your branch, rebased on master (to get the autotools fixes), and I cannot reproduce the problem.  Unless I am running the test incorrectly, I believe your latest commit (the one that delays res.freed events) fixed the problem.  Am I correct in assuming that the delayed scheduling is on by default (thus the t2002-easy.t test is run with it on)?

dongahn:
@SteVwonder: Sorry Stephen, it is actually off by default because of this problem. If you change a macro in `scheduler.h`, you should be able to reproduce this, I hope.

sched/sched.c
```diff
@@ -1573,18 +1608,20 @@ static int action (ssrvctx_t *ctx, flux_lwj_t *job, job_state_t newstate)
...
         }
-        flux_msg_t *msg = flux_event_encode ("sched.res.freed", NULL);
-        if (!msg || flux_send (h, msg, 0) < 0) {
-            flux_log (h, LOG_ERR, "%s: error sending event: %s",
-                      __FUNCTION__, strerror (errno));
-        } else {
-            flux_msg_destroy (msg);
-            flux_log (h, LOG_DEBUG, "Released resources for job %"PRId64"",
-                      job->lwj_id);
+        if (!ctx->arg.s_params.delay_sched) {
```
SteVwonder: With this

SteVwonder:
So the reason the test is hanging is because the simulator relies on the 'sched.res.freed' event, which isn't being emitted under delayed scheduling (as @dongahn mentioned above).  So I added `in_sim` to the conditional blocking that, which lead to the next issue....

The idle watcher is invalidating one of the assumptions of the simulator: that a module only handles events when it has the "token" (AKA when it is triggered by the simulation).  With delayed scheduling, the scheduler module will run whenever it is idle.

Specifically, when the delay scheduling is set and a simulation is running, the scheduler will move a job from the rqueue to the cqueue (setting pq_state to true), fire off the res.freed event (setting up the next event in the DES), and then return control to the simulator module.  In normal simulator execution, the res.freed event would be queued up and handled once the simulator module returns control to the scheduler (via the `trigger_cb` function).  In the case of delayed scheduling + simulation, the idle watcher kicks after the scheduler returns control of the simulation back to the simulator module.  The idle watcher cb causes a `schedule_jobs` loop to run (since the pq_state is set to true) even though the scheduler module doesn't have the "token".

IMO, the best solution here is to not combine the simulation and delayed scheduling.  Delayed scheduling makes a lot of sense in production where a huge stream of jobs can overload the scheduler if it is constantly running the scheduler loop.  In the simulator, the cost (in terms of time) of running the scheduling loop is not factored into the simulation time, since doing so is hard to do correctly (we tried it originally, against David Jefferson's advice, and it didn't go over well).  Given this, I think that simulating the effects of overloading the scheduler would be extremely difficult.   If some particular nuance of the delayed scheduling is interesting, we could potentially fake the effects.  For example, if the effects of batching the job submissions together is interesting, we could simulate this within the simulator by batching job submissions that are close in time (AKA without having to directly simulate the idle watcher).

dongahn:
@SteVwonder: Interesting! I'm wondering if this explains both of the failure modes I observed, though:

Initially I used the idle watcher for coalescing job submit/creation events and EASY backfill test failed with a different scheduling order. The observed order was FCFS not EASY and the code change for this was [this line](https://github.com/flux-framework/flux-sched/pull/191/commits/22236289334b1c64c2ea6ec64dce5d390b415d2b#diff-c0818a4e1636a85919d4b8e597378d54L1615). 

In this case, the original code directly invokes schedule_jobs without the token and how did the emulator this? Does it actually cause the sched not to exercise this code path?

SteVwonder:
@dongahn: thanks for the kick to reinvestigate this.  It turns out there is another layer to this onion.

*Background:* When Flux is running in simulation mode, each module keeps track of a `sim_state` which contains, among other things, the current simulation time.  Since the assumption is that while simulating, no module will run unless it is triggered by the simulator, the `sim_state` struct is initialized at the start of `trigger_cb` and free'd at the end of `trigger_cb`.

When the idle watcher kicks in and causes `schedule_jobs` to run "out of turn", the `sim_state` struct has been free'd but not reallocated/reinitialized. So when `schedule_jobs` runs and attempts to read the current time, it realizes it is in a simulation and reads from the `sim_state` struct (which is free'd and the read is technically invalid).  At least on my machine (and I presume on Travis), the invalid read ends up being a 0.  So the `schedule_jobs` will schedule jobs assuming that the current time is 0, allowing any jobs in the queue to be run (up to a limit of course, you will eventually run out of free nodes at time 0, but our test case is too small for that to occur).

I am fairly confident this is truly what is happening because after I removed [this line](https://github.com/SteVwonder/flux-sched/blob/c7497ec0e54b2c6792ecda622d667f1d64482df8/sched/sched.c#L937), the EASY backfilling test case passed (b/c when the idle watcher caused the `schedule_jobs` loop to run, it would read the correct simulation time, as opposed to 0).

To prevent this from going unnoticed in the future, we can add a check in `sim_exec` that verifies that the allocation start time == the simulation time.

SteVwonder:
Another reason we should implement a more exhaustive, but less frequent testing suite.  Invalid/illegal reads like this would be caught by valgrind.

dongahn:
> Another reason we should implement a more exhaustive, but less frequent testing suite. Invalid/illegal reads like this would be caught by valgrind.

I believe this can be done when Sanitizer is enabled within flux-sched. Just a few hours worth of work for me.

dongahn:
@SteVwonder: Just to clarify --

So, if we make scheduling done by the check watcher handler NOOP under emulation, would the emulator be still smart enough to avoid a deadlock in the absence of `sched.res.freed` events? If it intercepts JCS events, in particular job-complete event, perhaps this condition can be handled without a deadlock? 

BTW, does the emulator make use of JSC events other than job creation and complete? 

Just wondering if we can make the whole scheme a bit simpler by changing the emulator such that it queues up the `schedule_jobs` invocation events, which can be generated by the check watcher handler, and plays them on its emulation time line.

If the emulator doesn't make use of JSC events much, this can wean the emulator off job events. Then, the emulator can only depends on `schedule_job` events from the check watcher handler and future `resource events` (i.e., a node died and new nodes added...)

SteVwonder:
Currently, the `sched.res.freed` is used to queue up the running of the `schedule_jobs` function the next time that the scheduler is triggered (following a job completion).  If some other mechanism/flag is used to run the `schedule_jobs` function at the next trigger, then that is fine.

I believe the emulator workflow only really requires the creation and completion events.  The creation and completion events are queued up and handled once the scheduler is triggered.  So if the submission module (or initial program) gets triggered and then submits multiple jobs, the scheduler will have them all queued up when it finally gets triggered.  Currently (on master), the scheduler under simulation will call `schedule_jobs` after handling each job submission, but that wouldn't be hard to change it to only run one time after all job  submissions have been processed.  If we want to do that through the `ev_check_cb`, I think we can make that work, but I am extremely hesitant to involve the actual idle watcher in the emulator at all.  There is no guarantee that the idle watcher will ever run, which breaks the operation of the simulator.  When the simulator triggers a module, that module needs to process events then and only then.  So instead of waiting for the idle watcher to call `ev_check_cb`, `schedule_jobs`, etc., we should just call that function directly when the module is triggered.

Am I making any sense?  If not, it might help if I draw up a diagram showing function calls/messages under the existing scheduler+emulation and under delayed scheduling+emulation.

SteVwonder:
I should also mention that multiple jobs being submitted back-to-back is very rare in the simulator.  The jobs have to have the exact same submission time in the job traces (which have 1 second granularity) in order to be processed together in the simulator.  Otherwise, 1 job would be submitted at time X, the scheduler would be triggered, process the job and run `schedule_jobs` at time X+0.00001, then the 2nd job would be submitted at time X+1 and the scheduler would be triggered, process the job, and run `schedule_jobs` at time X+1.00001.

dongahn:
> I should also mention that multiple jobs being submitted back-to-back is very rare in the simulator. The jobs have to have the exact same submission time in the job traces (which have 1 second granularity) in order to be processed together in the simulator. Otherwise, 1 job would be submitted at time X, the scheduler would be triggered, process the job and run schedule_jobs at time X+0.00001, then the 2nd job would be submitted at time X+1 and the scheduler would be triggered, process the job, and run schedule_jobs at time X+1.00001.

Completely agreed. 

However, I'm more concerned about code maintenance. I have to believe that it would be best if the main sched comms module code becomes completely agnostic of which execution mode it is running under. To the extent possible, I think we want to be able to  make changes to the scheduler code and have it work both real execution or emulation without much effort...

dongahn:
> If we want to do that through the ev_check_cb, I think we can make that work, but I am extremely hesitant to involve the actual idle watcher in the emulator at all.

> There is no guarantee that the idle watcher will ever run, which breaks the operation of the simulator. 

My understanding is that the role of the idle watcher is exactly to avoid the situation that you are worring about. Even if there is no real external event, the idle watcher will keep the event loop from going into sleep and this then guarantees the check watcher to be invoked.

garlick:
> Even if there is no real external event, the idle watcher will keep the event loop from going into sleep and this then guarantees the check watcher to be invoked.

Right:
* prep watcher executes right before event loop would block; starts idle watcher if work todo
* if idle watcher is started, event loop won't block
* check watcher executes right after event loop would unblock; stops idle watcher, runs work

This comes directly from libev- see [this section of the libev docs](http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod#Abusing_an_code_ev_check_code_watche).

SteVwonder:
@dongahn & @garlick : I think I am understanding this a bit better, thanks for the pointers.

[Here is a commit](https://github.com/SteVwonder/flux-sched/commit/0a4d17cbf0c49a7abf746f0542435baf0c2d7b34) that can be applied directly on top of this PR and makes the emulator compatible with the `ev_*_cb` functions.  It still disables the idle watchers, but it follows the same code path now (and it doesn't rely on the `sched.res.free` event).  I have confirmed that `make check` completes successfully with this commit applied. This is probably still not exactly what you want, @dongahn.

Correct me if I'm wrong, but I think what you are envisioning is for the idle watchers to actually be used (in order to maximize the amount of share code and functionality between the emulation and production versions of the scheduler).  This is definitely possible.  To do so, we need to add a boolean flag that keeps `ev_check_cb` from running `schedule_jobs` when the scheduler doesn't have control of the simulation.  `trigger_cb` can set that boolean flag to true (and then immediately return).  The `ev_check_cb` will also have to call the simulator "cleanup" code that returns control of the simulation to the simulator module, etc.  See [here](https://github.com/SteVwonder/flux-sched/blob/delayed_schedule/sched/sched.c#L944) for the code I'm referring to.  It is three function calls: `handle_timer_queue`, `send_reply_request`, and `free_simstate`.  We could pull all of those into a "bridge" function if you want so that only one extra function call ends up in `ev_check_cb`.  Does this sound like what you were thinking, @dongahn?

SteVwonder:
As I think about it, with either method mentioned above, we can remove a lot of the `sim_*_cb`s that queue up events for later replay when the scheduler has control of the simulation.  I think the only thing that we need to delay until the scheduler has control of the simulation is the execution of `schedule_jobs`.  So `sim_job_status_cb` and `sim_res_event_cb` can both be removed, and the emulator can utilize the existing `job_status_cb` and the existing `res_event_cb`.

dongahn:
> Correct me if I'm wrong, but I think what you are envisioning is for the idle watchers to actually be used (in order to maximize the amount of share code and functionality between the emulation and production versions of the scheduler). This is definitely possible. To do so, we need to add a boolean flag that keeps ev_check_cb from running schedule_jobs when the scheduler doesn't have control of the simulation. trigger_cb can set that boolean flag to true (and then immediately return). The ev_check_cb will also have to call the simulator "cleanup" code that returns control of the simulation to the simulator module, etc. See here for the code I'm referring to. It is three function calls: handle_timer_queue, send_reply_request, and free_simstate. We could pull all of those into a "bridge" function if you want so that only one extra function call ends up in ev_check_cb. Does this sound like what you were thinking, @dongahn?

I think so. `ev_check_cb` will call `bridge_schedule_jobs` which then calls `schedule_jobs` under normal execution or emits a `schedule-job` event under emulation?

dongahn:
@SteVwonder: how do you want to handle this though? I can use your patch as a temporary solution and create an issue to provide "better fuse of normal execution and emulation" and address that more leisurely. Or we can work on it together in a near future together as well.

Either way, I hope I can land this PR soon as my other scalability PR depends on it.

SteVwonder:
@dongahn: if your other scalability PR depends on this one, I would just grab the patch and create an issue.

dongahn:
@lipari: just forced a push that cherry-picked @SteVwonder's emulator fix and also added some basic tests. Now as I'm seeing these commits, I realized it may make better sense to combine the first three commits: ce98690, 0f0c438, and 124289e unless you want to keep those intermediate states for later debugging. Please let me know your preference; I can go either way.

lipari:
Sorry for the delay in reviewing... I can see value in combining the three commits given that they are all pieces of the same new functionality.  I don't think there is much debugging value to keeping them separate.

dongahn:
OK. I squashed those commits. Forced a push.

lipari:
Nice work, @dongahn !

PR #196, opened 2016-09-27
Add functions to iterate over all hierarchies in an RDL conf

grondo:
As requested by @dongahn, this PR introduces `rdl:hierarchy_next()` method for Lua rdl to iterate over all hierarchies registered in a give rdl config.

Additionally a C interface `rdl_next_hierarchy (const char *prev)` is introduced to iterate over the hierarchies from C.

### DIFF
--- a/rdl/RDL/memstore.lua
+++ b/rdl/RDL/memstore.lua
@@ -1108,6 +1108,11 @@ function MemStore:serialize ()
     return serialize (self)
 end
 
+function MemStore:hierarchy_next (last)
+    local x,y,z = next (self.__hierarchy, last)
+    return x
+end
+
 local function bless_memstore (t)
     if not t.__resources or not t.__hierarchy or not t.__types then
         return nil, "Doesn't appear to be a memstore table"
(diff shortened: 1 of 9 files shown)


grondo:
Coverage is failing because I added `list-hierarchies` subcommand to `flux-rdltool`, but there are no tests for `flux-rdltool`.  I can add a test for that command, if it is even used anywhere.

dongahn:
@grondo: Sorry I just saw this PR... must have not been alerted. 

Like we discussed in emails the other day, this PR will soon be needed for me to be able to self-discover all the hierarchies. If adding a test case is not that time consuming, it would be great to add. I can see how `list-hierarchies` subcommand and other utilities within `rdltool` can find good use in a near future as the next thrust area for `sched` is on this topic.

grondo:
Ok, I rebased this on current flux-sched master and added requested basic tests for `flux-rdltool`.

dongahn:
@grondo: Nice work! I couldn't find a fault. Once @lipari okays it, I will merge it.

lipari:
I've looked it over and it looks good to me too.

PR #198, opened 2016-09-30
sched: Fix possible missed reservation

### LINKED ISSUE #197: Scheduler Correctness: Reservations
I believe I have found a correctness issue with the current scheduler (specifically in `sched.c`), that occurs when non-homogenous job requests are issued.

If the call to `find_resources` fails (nfound == 0), then nothing else is called and the scheduler moves onto the next job.  In order for hybrid and conservative backfill to work, the scheduler should call `reserve_resources` so that a reservation is still made for the job even though there aren't any free resources for the job to execute on currently (it might be able to run very soon).

Since the reservation is not made, a job later on in the queue can be allocated a run (since it uses less memory or IO bandwidth, for example) without any consideration to if it affects the start time of the previous job.

This case should be considered in our correctness study (#168).

SteVwonder:
If find_resources returned 0, then the scheduler would not call
reserve_resources, as it should.

Fixes #197

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -1489,9 +1489,21 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
                 if (rc) {
                     resrc_tree_destroy (selected_tree, false);
                     job->resrc_tree = NULL;
-                } else
+                } else {
                     job->resrc_tree = selected_tree;
+                }
             }
+        }
+    } else {
+        rc = plugin->reserve_resources (h, &selected_tree, job->lwj_id,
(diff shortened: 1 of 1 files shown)


dongahn:
Hmmm... odd that the compilers don't like the format issue that is not changed with this PR...

SteVwonder:
Yeah, very odd.  I'll see if I can reproduce the issue locally, if not, I pre-apologize for the multiple commits to try and get Travis to compile the code.

SteVwonder:
@lipari and I talked over the phone and this doesn't seem to be the correct fix.  Closing and returning the discussion to #197

PR #201, opened 2016-10-03
A Fix and an Update

lipari:
Fix a compile error with a call to kvs_get_dir() and grab the latest shortjson.h from flux-core

### DIFF
--- a/simulator/simulator.c
+++ b/simulator/simulator.c
@@ -172,11 +172,8 @@ int put_job_in_kvs (job_t *job)
 
     // TODO: Check to see if this is necessary, i assume the kvsdir becomes
     // stale after a commit
-    char *dir_key;
-    dir_key = xasprintf ("%s", kvsdir_key (job->kvs_dir));
     kvsdir_destroy (job->kvs_dir);
-    kvs_get_dir (h, &job->kvs_dir, dir_key);
-    free (dir_key);
+    kvs_get_dir (h, &job->kvs_dir, "%s", kvsdir_key (job->kvs_dir));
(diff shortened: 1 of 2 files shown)


dongahn:
Thanks @lipari.  I will assume the first commit fixes  the Travis failures seen with #198. This is not your code but I thought I might ask a question to @garlick about the need for updating the job's kvs_dir handle.It seems awkward a directory  handle needs to be destroyed and reread after a commit... (There is a TODO note anyway)

garlick:
The `kvsdir_t` contains a list of keys (at the level of the directory) that will become out of date if keys are added/removed at that level.  However, a `kvsdir_t` does not behave like a snapshot with respect to keys at other levels unless it was created by the new `kvs_get_dirat()`.  I know, I know, this API is awful.

So refresh if you need to update the list of keys at that level.

And you should check the return value of all kvs calls for error too.

lipari:
> And you should check the return value of all kvs calls for error too.

Created [Issue 202](https://github.com/flux-framework/flux-sched/issues/202) to capture that advice.

lipari:
@dongahn or @SteVwonder, I believe this is ready for merging.  I've created a separate issue to decouple the return check of the kvs calls to a follow-on effort.

lipari:
@dongahn or @SteVwonder, I merged @grondo's fix instead.  I"ll push a new PR with just the shortsjson update

grondo:
Sorry Don! I somehow missed this PR :-(

lipari:
Np at all

PR #206, opened 2016-10-04
Upate shortjson.h from flux-core, remove JSON

morrone:
Update shortjson.h from flux-core, which removes the JSON typedef.
Update the rest of the code to use json_object * in place of JSON.


morrone:
Should land after flux-framework/flux-core#832, only so the the "Update shortjson.h from flux-core" is not a lie.

lipari:
Thank you, @morrone!  Looks good.  I'll merge once the flux-core side lands.

PR #207, opened 2016-10-11
Adjust to flux_t no longer containing pointer

morrone:
The flux_t typedef in flux-core no longer hides a pointer.
Change flux-sched's use of flux_t to match the new pointer-less
definition of flux_t.


morrone:
This will be needed after flux-framework/flux-core#841 lands.

lipari:
Thank you, @morrone!

Issue #211, opened 2016-10-12
travis-ci: t0001-basic.t fails


garlick:
The travis-ci builds from pr #209 were not working just before the pr was merged.  [This gcc build](https://travis-ci.org/flux-framework/flux-sched/jobs/167119632) is an example.

The first test that fails is this one:
```
2016-10-12T17:28:17.164396Z sched.err[0]: kvs_get resource.hwloc.xml.0: Protocol error
2016-10-12T17:28:17.164620Z sched.err[0]: failed to load resrc using hwloc
2016-10-12T17:28:17.164711Z sched.err[0]: failed to load resources
2016-10-12T17:28:17.164804Z sched.crit[0]: fatal error: Protocol error
not ok 1 - sched: module load works
#	
#		flux module load sched
#	
ok 2 - sched: module remove works
2016-10-12T17:28:17.196121Z sched.err[0]: kvs_get resource.hwloc.xml.0: Protocol error
2016-10-12T17:28:17.196288Z sched.err[0]: failed to load resrc using hwloc
2016-10-12T17:28:17.196373Z sched.err[0]: failed to load resources
2016-10-12T17:28:17.196455Z sched.crit[0]: fatal error: Protocol error
not ok 3 - sched: flux-module load works after a successful unload
#	
#		flux module load sched &&
#		flux module remove sched
#	
rdl_load: foo: No such file or directory
2016-10-12T17:28:17.215201Z sched.err[0]: failed to load resrc
2016-10-12T17:28:17.215322Z sched.err[0]: failed to load resources
2016-10-12T17:28:17.215429Z sched.crit[0]: fatal error: No such file or directory
ok 4 - sched: module load should fail
2016-10-12T17:28:17.230754Z sched.err[0]: kvs_get resource.hwloc.xml.0: Protocol error
2016-10-12T17:28:17.230922Z sched.err[0]: failed to load resrc using hwloc
2016-10-12T17:28:17.231007Z sched.err[0]: failed to load resources
2016-10-12T17:28:17.231089Z sched.crit[0]: fatal error: Protocol error
not ok 5 - sched: module load works after a load failure
#	
#		flux module load sched
#	
ok 6 - sched: module list works
ok 7 - get_instance_size works
# failed 3 among 7 test(s)
1..7
2016-10-12T17:28:17.276716Z broker.err[0]: Run level 2 Exited with non-zero status (rc=1) 0.1s
flux-start: 0 (pid 40522) exited with rc=1
[...]

PR #212, opened 2016-10-13
plugins: add dlopen RTLD_DEEPBIND flag

garlick:
One step towards fixing #211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,9 @@ script:
  - (find $HOME/local2/lib -name \*.la | xargs rm -f)
  - ($HOME/local2/bin/flux keygen)
 
+ # Set LD_LIBRARY_PATH so that flux-core libs can be found
+ - export LD_LIBRARY_PATH=$HOME/local2/lib:${LD_LIBRARY_PATH}
+
  # Enable coverage for $CC+coverage build
  - if test "$COVERAGE" = "t" ; then ARGS=--enable-code-coverage; MAKECMDS="make && make check-code-coverage && lcov -l flux-*coverage.info" ; fi
  
--- a/sched/plugin.c
+++ b/sched/plugin.c
@@ -152,7 +152,7 @@ int sched_plugin_load (struct sched_plugin_loader *sploader, const char *s)
         if (!(name = flux_modname (path)))
             goto error;
     }
-    if (!(dso = dlopen (path, RTLD_NOW | RTLD_LOCAL))) {
+    if (!(dso = dlopen (path, RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND))) {
         flux_log (sploader->h, LOG_ERR, "failed to open sched plugin: %s",
                   dlerror ());
         goto error;
--- a/t/t2002-easy.t
+++ b/t/t2002-easy.t
@@ -38,8 +38,9 @@ while flux kvs get lwj.12.complete_time 2>&1 | grep -q "No such file"; do sleep 
 sleep 0.5
 for x in $(seq 1 12); do echo "$x $(flux kvs get lwj.$x.starting_time)"; done | sort -k 2n -k 1n | cut -d ' ' -f 1 > actual
 
-test_expect_success 'jobs scheduled in correct order' '
-    diff -u ${expected_order} ./actual
-'
+# XXX disabled temporarily - see #212
+#test_expect_success 'jobs scheduled in correct order' '
+#    diff -u ${expected_order} ./actual
+#'
 
 test_done


grondo:
Restarted the travis build

garlick:
I had just rerun the gcc build and it failed on the t2002-easy test again.
Anybody from sched have a clue how to diagnose that one?
Everything is running but jobs are not running in the expected order.

grondo:
I ran `t2002-easy.t` test and dumped the job completion times it is using for the compare, and does indeed look like the backfill isn't working, all jobs appear to run in fcfs order... I wonder if the backfill plugin isn't loading correctly?
```

1 1.004068
2 19643.003816
3 28147.004182
4 30566.006998
5 30566.006998
6 77371.003694
7 80090.003604
8 87576.004911
9 88169.003928
10 96589.003074
11 103162.004577
12 126070.003682
expecting success: 
    diff -u ${expected_order} ./actual

--- /home/ubuntu/flux-sched/t/data/emulator-data/easy_expected  2016-08-12 17:05:26.000000000 +0000
+++ ./actual    2016-10-13 23:12:06.869725418 +0000
@@ -1,12 +1,12 @@
 1
 2
 3
+4
+5
 6
 7
 8
 9
 10
 11
-4
-5
 12
not ok 5 - jobs scheduled in correct order
#
#           diff -u ${expected_order} ./actual
#
```

garlick:
Good theory - I'll go poke at that.

Aside: any idea why `flux wreck ls` output is empty when I run it in the test, just after job 12 was verified complete?

garlick:
Well, there is this subtest in `t1004-module-load.t`:
```
test_expect_success 'module-load: sched loads the backfill plugin' '
    flux module load sched.backfill &&
    flux module list sched
'
```

dongahn:
I also saw the same issue (backfill producing FCFS) as part of #191 but we put a temporary fix for that problem...

grondo:
full dmesg output from the test should be in this [gist](https://gist.github.com/20ad27751346d15c3b2f9ac72363e74e)

grondo:
> Aside: any idea why flux wreck ls output is empty when I run it in the test, just after job 12 was verified complete?

No ... what does `flux kvs dir lwj` return?

garlick:
> what does flux kvs dir lwj return?

```
lwj.1 -> lwj-active.1
lwj.2 -> lwj-active.2
lwj.3 -> lwj-active.3
lwj.4 -> lwj-active.4
lwj.5 -> lwj-active.5
lwj.6 -> lwj-active.6
lwj.7 -> lwj-active.7
lwj.8 -> lwj-active.8
lwj.9 -> lwj-active.9
lwj.10 -> lwj-active.10
lwj.11 -> lwj-active.11
lwj.12 -> lwj-active.12
```

grondo:
are you running with the test with `-d -v`? when I run I get the following error, because the lwj entries don't appear to be "real" jobs:
```
lua: /tmp/inst/libexec/flux/cmd/flux-wreck:213: attempt to index field 'cmdline' (a nil value)
stack traceback:
        /tmp/inst/libexec/flux/cmd/flux-wreck:213: in function </tmp/inst/libexec/flux/cmd/flux-wreck:185>
        /tmp/inst/libexec/flux/cmd/flux-wreck:248: in function 'handler'
        /tmp/inst/share/lua/5.1/flux/Subcommander.lua:146: in function </tmp/inst/share/lua/5.1/flux/Subcommander.lua:127>
        (tail call): ?
        /tmp/inst/libexec/flux/cmd/flux-wreck:397: in main chunk
        [C]: ?
```

garlick:
Oh, I missed that.  Yes I see that too.

garlick:
> I wonder if the backfill plugin isn't loading correctly?

Inserted a `flux-module list sched` and it is indeed loaded:
```
Module               Size    Digest  Idle  S  Nodeset
sched.backfill        258048 2D73E4E    0  R  0
```
The dmesg in the gist shows a successful load also.

garlick:
It sounds like we might need @SteVwonder to have a look here, if the problem is similar to #191.

@grondo had a good idea.  Since "sched's dead" because of the json problems I introduced, maybe we should temporarily disable this test in this PR, get sched passing travis again, then come back to this?

garlick:
OK, I propose we merge this.  Sorry everyone for the trouble that json change caused!

dongahn:
From [gist](https://gist.github.com/anonymous/20ad27751346d15c3b2f9ac72363e74e), there seems to be something else going on than backfill scheduling FCFS. The trace is at least missing some jobs like job (2) etc. Don't know how this could affect that behavior, though.

dongahn:
OK. Merging this and let's have @SteVwonder take a look at the emulator test cases.

PR #213, opened 2016-10-13
resrc:  modify the creation of the properties and tags

lipari:
Check for null when reading the property and tag values from the json
objects.  This check is an improvement independent from the work to
disambiguate json-c and jansson symbols.


dongahn:
Could you please rebase?

lipari:
Done.

PR #215, opened 2016-10-15
Convert JSON handling from libjson-c to libjansson

grondo:
Here's the first stab at removing json-c usage from flux-sched. Sorry about this, but the conversion is all in one big commit. I could try splitting it up, but it might be a bit of a chore to keep the intermediate commits from breaking the build along the way.

As part of the conversion, `shortjson.h` was converted to Jansson, then renamed to `shortjansson.h` to avoid any kind of confusion. I also removed some functions that were not even being used in flux-sched.

The Lua bindings, `json-lua.[ch]` were also converted and renamed to `jansson-lua.[ch]` -- also to avoid any confusion (`jansson-lua.[ch]` could end up being useful in flux-core later perhaps).

The rest of the conversion should be pretty straightforward. The code in `resrc_new_from_json` should be carefully reviewed by @lipari, because I wasn't exactly sure what was being done with the JSON objects there and I hope I got it right (tests are passing at least).

The last few commits on top are for compatibility with Jansson < 2.3 as is used by default in Travis. There is one really big kludge here though: Versions < 2.3 refuse to decode strings to JSON that are not true JSON objects or arrays (Jansson 2.3.0 added `JSON_DECODE_ANY` flag to support this). This caused a failure when decoding JSON objects returned by the kvs (specifically the hwloc xml value, as is used during shed module load). 

For now there is a bad kludge that wraps strings in `{ "foo": %s }` when `json_loads()` fails for Jansson < 2.3 and tries again. This seems to work in Travis at least.

After all this, flux-sched does not need to link with libjson-c, and the segfaults I was seeing on TOSS2 systems are gone.


garlick:
Nice work @grondo.

I should have done what you did in core with the foreach macro.  Instead I implemented two ifdef code paths at the call site, which of course leads to one path not being tested much.

I didn't check the calling context, but could the workaround for parsing bare JSON strings possibly be avoided by calling `kvs_get_string()` instead of `kvs_get()`?

grondo:
> I didn't check the calling context, but could the workaround for parsing bare JSON strings possibly be avoided by calling kvs_get_string() instead of kvs_get()?

Good idea. However, I think the failing call I was seeing is using `kvs_watch_once`, and I don't see a string variant. (see `sched/sched.c: get_string_blocking()` -- I'm sure there is a better way to do this...)

grondo:
> The code in resrc_new_from_json should be carefully reviewed by @lipari, 

Sorry @lipari, I should have said `resrc_flow_from_json`. The change is this:
```diff
@@ -194,13 +194,11 @@ resrc_flow_t *resrc_flow_new_from_json (json_object *o, resrc_flow_t *parent)
     if (Jget_int64 (o, "size", &ssize))
         size = (size_t) ssize;
     if ((jhierarchyo = Jobj_get (o, "hierarchy"))) {
-        struct json_object *val = NULL;
-        struct lh_entry *entry = json_object_get_object (jhierarchyo)->head;
-
-        if (entry) {
-            hierarchy = (char*)entry->k;
-            val = (struct json_object*)entry->v;
-            hierarchy_path = json_object_get_string (val);
+        /* Get first key and value from hierarchy object */
+        const char *key = json_object_iter_key (json_object_iter (jhierarchyo));
+        if (key) {
+            hierarchy = key;
+            Jget_str (jhierarchyo, key, &hierarchy_path);
         }
     }
     if (!Jget_str (o, "path", &path)) {
```

grondo:
It occurs to me I should probably check the `json_object_foreach` macro I implemented here. I just grabbed one from a jansson issue report and modified it for inclusion in flux-sched, made sure it compiled and tests passed, but I didn't check beyond that..

lipari:
> The code in resrc_new_from_json should be carefully reviewed by @lipari,

Looks good, @grondo! And thank you for doing this!

I confirmed the mechanism you used to retrieve the `hierarchy_path` in resrc_flow.c is correct.  In fact, my testing shows that, unlike the resrc.c code, the `path` key is never present when loading a flow hierarchy from an RDL conf file.  So, the check for `path` being present is a performance hit, and I may opt to eliminate it and unconditionally use the value of `hierarchy_path` as the `path` on the resrc_flow side.

I do see a typo in rdl/Makefile.am  I believe the fix is this:
```
@@ -43,7 +43,7 @@ flux_cpuset_la_LDFLAGS = \
  libjansson_lua_la_SOURCES = jansson-lua.c jansson-lua.h
-libflux_rdl_la_LIBADD = $(LUA_LIB) $(JANSSON_LIBS)
+libjansson_lua_la_LIBADD = $(LUA_LIB) $(JANSSON_LIBS)
 
 libflux_rdl_la_SOURCES = rdl.c
 libflux_rdl_la_LIBADD = $(FLUX_CORE_LIBS) $(LUA_LIB) $(CZMQ_LIBS) $(JANSSON_LIBS) \
```

grondo:
Thanks for checking it out @lipari, and thanks for the catch above - I'll include that fix momentarily.

FYI -- the top few commits in this branch are for compatibility with the lwj directory hierarchy changes in flux-framework/flux-core#811. I will move these to another PR, but I couldn't test them without the changes to fix segfaults due to libjansson libjson-c incompatibilities first.

grondo:
OK, @lipari, added your fix to `rdl/Makefile.am`. Give this branch your ok, and if it is all good, I'll strip off lwj-heirarchy fixes and we can perhaps merge this one?

lipari:
@grondo, I reviewed the PR and did not have any other suggestions.  Go ahead and take away the lwj hierarchy commits.

dongahn:
For my PerfExplore testing, should I use this sched PR?

grondo:
@dongahn, I was going to move the lwj.%d adaptations onto their own PR, which will be based off this branch. Once I have that done I'll ping you to use that branch for PerfExplore.  I apologize for the inconvenience

dongahn:
Not a problem! Thank you @grondo.

grondo:
Ok, just the jansson conversion commits are left here, along with new tests for jansson-lua binding. Let's see if this passes Travis.

lipari:
Looks good!  Merging...

PR #217, opened 2016-10-20
use new names for JSON accessors

garlick:
This should not be merged until after flux-framework/flux-core#857 is merged.

Also, before merging, we should restart travis and verify tests pass.  I was having some test problems in my own environment that I don't see how could be related to this PR, but best to be safe.


grondo:
Github is complaining that this branch is out of date, do you need to rebase on current master?

garlick:
Oh indeed.  Sorry a bout that.  Just forced an update.

grondo:
restarted travis after merge of flux-framework/flux-core#857

PR #219, opened 2016-10-26
sim: testing and kvs operation cleanup

SteVwonder:
Restores the `kvs_get_dir` that was removed in #281
Cleans up the sim testing scripts so that they are consistent, use flux-waitjob, and have debug printing.


lipari:
Looks good, @SteVwonder!  Merging...

PR #226, opened 2016-11-22
count cores per node when building JSC request

trws:
This addresses the issue of jobs that don't divide evenly between nodes, counting the cores per node from the resource tree.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -1203,6 +1203,32 @@ static void inline build_contain_1node_req (int64_t nc, int64_t rank,
     json_array_append_new (rarr, e);
 }
 
+static int n_resources_of_type (resrc_tree_t *rt, const char *type)
+{
+    int n = 0;
+    resrc_t *r = NULL;
+
+    if (rt) {
+        r = resrc_tree_resrc (rt);
+        if (! strcmp (resrc_type (r), type)) {
+            return 1;
+        } else {
+            if (resrc_tree_num_children (rt)) {
+                resrc_tree_list_t *children = resrc_tree_children (rt);
+                if (children) {
+                    resrc_tree_t *child = resrc_tree_list_first (children);
+                    while (child) {
+                        n += n_resources_of_type(child, type);
+                        child = resrc_tree_list_next (children);
+                    }
+                }
+            }
+        }
+    }
+    return n;
+}
+
+
 /*
  * Because the job's rdl should only contain what's allocated to the job,
  * we traverse the entire tree in the post-order walk fashion
@@ -1230,8 +1256,11 @@ static int build_contain_req (ssrvctx_t *ctx, flux_lwj_t *job, resrc_tree_t *rt,
         } else {
             if (bridge_rs2rank_tab_query (ctx, r, &rank))
                 goto done;
-            else
-                build_contain_1node_req (job->req->corespernode, rank, arr);
+            else {
+                int cores = job->req->corespernode ? job->req->corespernode :
+                    n_resources_of_type(rt, "core");
+                build_contain_1node_req (cores, rank, arr);
+            }
         }
     }
     rc = 0;


lipari:
LGTM.  Will await @dongahn's opinion as it touches code he wrote.

trws:
Thanks Don.

Would you be interested in having a "greedy" variant of the FCFS scheduler as well? It looks like it might make a noticeable performance improvement for ATS to be able to pack in post-processing alongside the larger, and earlier submitted, jobs.

dongahn:
Interesting. Can the variant be enabled as a sched-param option?

trws:
I don't see why not. The only difference between FCFS and greedy is that greedy would check jobs past the start of the list to look for something that fits. It's really a very small change, so that should be pretty easy.




On November 23, 2016 at 12:42:56 PM PST, Dong H. Ahn <notifications@github.com> wrote:

Interesting. Can the variant be enabled as a sched-param option?

-

lipari:
@trws , I'm fine with the proposed greedy option.  FWIW, I believe the greedy behavior will match the backfill plugin's behavior when the reserve-depth option is zero, once [PR 227](https://github.com/flux-framework/flux-sched/pull/227) is merged.  It basically turns off making any reservations and allows the scheduler to visit every job in the queue, running whatever it can.

trws:
That's probably right, as long as the priority is given to jobs ordered 
by size.  If you're already putting that in, I'll wait to test it before 
working up a separate greedy option.

dongahn:
@trws: In that case, FCFS would be a misnomer? 

OK. I now see @lipari's comment. I think our backfill scheduling can be tailored to the workload using at least two policy parameters: 1) reservation depth and 2) backfilling policy. If one sets 1) to 0 and 2) to best-fit or first-fit, you may get what you need, like @lipari said. 

Reservation-depth=0 for backfill is a new concept to me though...

lipari:
The sched_fcfs plugin is actually a misnomer.  It really deals with now-based scheduling, and does not support scheduling into the future, like the backfill plugin does.  And, both plugins visit jobs in the order in which they were submitted.
My intent was to relegate re-ordering the queue - or assigning job priorities - to an external module or service.  This way the sched plugin could continue to be solely concerned with scheduling jobs and some other agent would be continually re-ordering the pending job queue, based on a multitude of factors like job size.

trws:
Hence calling it greedy, but it depends on how you look at it.  The 
priority could be either first-come first-served or in order of size or 
whatever independently of what the packing policy is.

lipari:
I've been talking about a job priority service for quite a while.  Maybe it's time I actually create one that does something simple like job size ordering.  At that point, I'll rename the sched_fcfs plugin to sched_now and eliminate the confusion.

trws:
Maybe just another plugin to sched?  I'm not sure it warrants being the level of a full service.

lipari:
The job priority service will start off simple, but could ultimately grow to include all the factors available in other schedulers:  fair-share, QoS, job size, other job resources, power, bandwidth ...
I _really_ want to keep the sched orthogonal to any prioritization activities.  This allows parallel and independent activities to proceed along a natural distinction.

trws:
I'm not really disagreeing with that, but currently the list of jobs available to be scheduled is a list local to sched that doesn't have to deal with race conditions.  Making a prioritization system run in a separate thread, which it would have to as a full service, would mean either requesting each job by message or sharing that across threads.  Having it be a pluggable component that is used by sched's thread to add jobs to, or remove jobs from, that list would seem to make those problems go away.

dongahn:
@lipari: yes this makes sense. 

@trws: i guess it depends on who are the user(s) of this service. Whether submit wants to get a priority number from this service and then enqueued into ached's pending queue, in which case a full service can make sense. If a job is simply sent to sched's pending queue first and ordered during the enqueue operation then a sched plugin makes sense.

lipari:
The vision I am promoting is consistent with using a job queue that lives in the kvs and is referenced, but not owned by the scheduling module.  Hence, an external mechanism can be created to cancel a job (remove it from the kvs) or modify it's attributes (including job priority).  The scheduler would be oblivious to all this until it grabbed the current queue to start scheduling jobs.

trws:
That sounds good from a dynamic-update perspective, but the cost of 
updating and accessing the KVS, as quick as it is in many cases, is not 
negligible.  It also implies there will have to be a structure in the 
KVS that stores the jobs in a sorted manner, and I'm not sure there's a 
cheap way to do that.  @garlick, thoughts?

dongahn:
@trws and @lipari: I am really not sure if job management service including queuing belong to kvs. I can see why we might want to have a queuing service but not sure if kvs should be the home for that. (I rather see sched as a home to provide that job management.) Besides, I am thinking to move resource selection as a full service using walker/matcher, at that point what would be the role of scheduler if it doesn't manage the job queues?

lipari:
Non-Flux schedulers do it all.  They respond to queries to display the queue, respond to job submit, modify, and cancel commands, they prioritize the queue based on those multitude factors, and respond to node status requests.  I am looking for every opportunity to avoid that architecture and disperse such services to dedicated agents.

dongahn:
@lipari: Yes I fully agree. Distributed services have lots of merits and that's one of the reasons I wanted to have resource state and selection as a service, and then dynamic scheduler as a service that builds atop of base scheduling and resource services. I was just asking where the job management services and in particular queuing services should belong to if that's not the scheduler service itself. Maybe the scheduler service can be grown into basic job queue services and have other services built around them...

lipari:
Created [Issue 228](https://github.com/flux-framework/flux-sched/issues/228) to capture the job prioritization need.

PR #231, opened 2016-11-30
Ats fixes, performance and memory leaks

trws:
A test of performance with the ATS backend found that more than 55% of the overall samples taken in the broker were in json_loads, in sched_backfill, almost all of that time inside the encoding and decoding of twindow entries.  There may be a couple more tweaks here, and the zhashx is not terribly fast and it still sorts and so-forth, but this should cut out almost all of that overhead.

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -41,6 +41,34 @@
 #include "resrc_reqst.h"
 #include "src/common/libutil/xzmalloc.h"
 
+
+
+typedef struct window {
+    int64_t starttime;
+    int64_t endtime;
+    const char *job_id;
+} window_t;
+
+/* static window_t * window_new (int64_t starttime, int64_t endtime) { */
+/*     window_t *ret = malloc (sizeof *ret); */
+/*     ret->starttime = starttime; */
+/*     ret->endtime = endtime; */
+/*     return ret; */
+/* } */
+
+static void window_destructor (void **window_v) {
+    if (window_v) {
+        free(*window_v);
+        *window_v = NULL;
+    }
+}
+
+static void *window_dup (const void *window) {
+    window_t * ret = malloc(sizeof *ret);
+    memcpy(ret, window, sizeof *ret);
+    return ret;
+}
+
 struct resrc {
     char *type;
     char *path;
@@ -59,7 +87,7 @@ struct resrc {
     zhash_t *tags;
     zhash_t *allocs;
     zhash_t *reservtns;
-    zhash_t *twindow;
+    zhashx_t *twindow;
 };
 
 static zhash_t *resrc_hash = NULL;
@@ -148,13 +176,8 @@ size_t resrc_available (resrc_t *resrc)
 
 size_t resrc_available_at_time (resrc_t *resrc, int64_t time)
 {
-    int64_t starttime = 0;
-    int64_t endtime = 0;
-
     const char *id_ptr = NULL;
-    const char *window_json_str = NULL;
-    json_t *window_json = NULL;
-    zlist_t *window_keys = NULL;
+    window_t *window = NULL;
     size_t *size_ptr = NULL;
 
     size_t available = resrc->size;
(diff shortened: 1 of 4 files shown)


resrc/resrc.c
```diff
@@ -337,18 +333,16 @@ size_t resrc_available_during_range (resrc_t *resrc, int64_t range_starttime,
     zlistx_t *start_windows = matching_windows;
     zlistx_set_comparator(start_windows, compare_windows_starttime);
     zlistx_t *end_windows = zlistx_dup (start_windows);
-    fprintf (stderr, "ZLISTX duplicated %zd\n", zlistx_size(end_windows));
+    // Do not free items in this list, they are owned by the start_windows
+    // list
+    zlistx_set_destructor(matching_windows, NULL);
```
lipari: Don't we need these statements at this point?
zlistx_set_destructor (start_windows, window_destructor);
zlistx_set_destructor (end_windows, window_destructor);

resrc/resrc.c
```diff
@@ -437,10 +392,9 @@ size_t resrc_available_during_range (resrc_t *resrc, int64_t range_starttime,
             min_available;
     }
 
-    zlist_destroy (&end_windows);
+    zlistx_destroy (&end_windows);
```
lipari: Now that start_windows is a copy of matching_windows, I think we need:
zlistx_destroy (&start_windows);

resrc/resrc_flow.c
```diff
@@ -240,7 +240,7 @@ resrc_flow_t *resrc_flow_new_from_json (json_t *o, resrc_flow_t *parent)
 
             json_str = xstrdup (Jtostr (w));
             resrc_twindow_insert (resrc_flow->flow_resrc, "0",
-                                  (void *) json_str);
+                                  starttime, endtime);
```
trws: Thanks!  I'll remove those shortly.

resrc/resrc.c
trws: `start_windows` is currently just an alias for `matching_windows`.  That could certainly be changed, but the free is taken care of by the destroy of matching_windows.

resrc/resrc.c
trws: `start_windows` has its set when it's created as `matching_windows`.  This was meant to set `end_windows`'s destructor, which it inherits from matching windows through the duplication, to NULL because I intentionally didn't use a duplicator to duplicate the underlying values to save on copy overhead.  Thanks for catching the typo there.

resrc/resrc.c
lipari: Agreed.  But once you created the alias for matching_windows, you deactivated its destructor:
zlistx_set_destructor(matching_windows, NULL);
So, now the start_windows alias will need the destruction.

resrc/resrc.c
```diff
@@ -32,6 +32,7 @@
 #include <assert.h>
 #include <czmq.h>
 #include <hwloc.h>
+#include <Judy.h>
```
lipari: I don't see where you used this library?

trws: I haven't yet, that line could just come out for now, but using it would make it easy to do this without sorts and save some memory in each resource.

lipari: That's fine.  I wasn't sure if it was from an abandoned approach.

trws:
More it's the next approach, the zhashx/zlistx switch was a quick and 
dirty way to get it done without new deps, judy would be better overall 
but will take a bit more surgery.

lipari:
Looks like it may be better to remove the Judy.h include for now.  It has build issues:
>   CC     libflux_resrc_la-resrc.lo
> ../../resrc/resrc.c:35:10: fatal error: 'Judy.h' file not found
> #include <Judy.h>

trws:
Rebased and judy removed, assuming this passes, it should be in a good place.

lipari:
Thank you, @trws.   Merging...

PR #232, opened 2016-12-28
Planner: API to enable scheduler-driven aggregate updates

dongahn:
@lipari: Before I start to integrate this planner API further into our scheduler to improve its performance, I thought this is a good stopping point to post an experimental PR. Do the planner abstraction and where I place the source files (including linux Red Black tree code) look reasonable to you? 

I plan to see if I can integrate this to improve twindow to begin with. Depending on how straightforward the work will be, I may next actually add a preorder pruning filter directly to the existing resrc trees or start to add matcher/walker codes.

### DIFF
--- a/resrc/Makefile.am
+++ b/resrc/Makefile.am
@@ -8,9 +8,9 @@ SUBDIRS = . test
 
 noinst_LTLIBRARIES = libflux-resrc.la
 
-noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h planner.h
 
-libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c
+libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c planner.c
 libflux_resrc_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 libflux_resrc_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/src/common/libutil/libutil.la \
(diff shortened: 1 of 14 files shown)


dongahn:
Also @tws: do you have test cases I can run to see if planner can improve twindow's performance overheads?

garlick:
I think you meant @trws

dongahn:
Yeah... Sorry about that. 

Oh. @garlick: while I have your attention, what would you say to pull in that same Linux red black tree source files into flux-core's utils as well. If nothing else, this will at least make flux-sched's util a proper subset of that of core. And I think @chu11 mentioned at a meeting that we may find other use cases for the self balancing trees?

garlick:
I would suggest leaving it private to sched for now until we have a use case in core.  At that point we could choose what makes sense: duplicate it privately in core, or factor it out to a library exported by core.

dongahn:
Ok. No problem leaving it private to sched. FWIW, what I find is that this is built for performance and as such interfaces are not the world's easiest.

trws:
On the tree, I recall @grondo mentioning wanting to pull in the kernel list, and figured if @dongahn was doing it, it might not hurt to just pull both of those into util so we'd have them, but if there's no current user it can certainly wait.

As to benchmarks, you can take any large concurrent submit (a minimum of 256 jobs, preferably more like 4+ times that) and configure sched with the following export.  The worst-case scenario is to submit jobs in reverse order of size, so that the smallest jobs are submitted last.

```
export FLUX_SCHED_OPTIONS='plugin=sched.backfill plugin-opts=reserve-depth=0'
```

If you want to try running what I was actually testing, I can give you the ATS backend code, but the production code it was testing might be a bit harder.

dongahn:
@trws: It is great that you came up with this test case!

This will ultimately be a good test for our new aggregate update scheme using planner (though a twindow improvement alone might still help). I have to believe that your perf issue comes from doing a full tree search for each subsequent unsatisfiable job. Planner will help avoid a full search using a cheap satisfiability check on sub-resource aggregates at a higher-level resource vertex. 

Another area I plan to improve using planner is the current reservation logic. We currently perform a full tree search per each running job's complete time (until enough resources can be reserved), and I believe planner should allow us to limit the number of searches as well. 

Both of these will only get worse at larger graph; so this is a good time to tackle them.

dongahn:
OK. I got good enough feedback from this experimental PR: closing.

PR #233, opened 2017-01-03
Experimental planner PR #2: twindow replacement

dongahn:
This second experimental PR is strictly for @lipari. I replaced the linear twindow searching with planner's API which should be more scalable. 

I confirmed all of the unit and integration tests pass with this commit. But I will ultimately want to run a larger emulation (e.g. borrowing from @SteVwonder's work) and make sure the planner-based scheduler produces identical schedules as the original twindow implementation. 

As you and I discussed this morning, I decided to continue with this branch and add an initial version of scheduler-driven update on top of this before requesting an official PR review. This is a bit of a plan chang, as I initially planned to fork off a new matcher/visitor branch and implement the first version of aggregate work in that branch.

In any case, since I will have to work on other projects/tasks for a while, this is to inform you with my progress.

### DIFF
--- a/resrc/Makefile.am
+++ b/resrc/Makefile.am
@@ -8,9 +8,9 @@ SUBDIRS = . test
 
 noinst_LTLIBRARIES = libflux-resrc.la
 
-noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h
+noinst_HEADERS = resrc.h resrc_tree.h resrc_flow.h resrc_reqst.h planner.h
 
-libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c
+libflux_resrc_la_SOURCES = resrc.c resrc_tree.c resrc_flow.c resrc_reqst.c planner.c
 libflux_resrc_la_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/rdl
 libflux_resrc_la_LIBADD = $(top_builddir)/rdl/libflux-rdl.la \
     $(top_builddir)/src/common/libutil/libutil.la \
--- a/resrc/planner.c
+++ b/resrc/planner.c
@@ -1,0 +1,1389 @@
+/*****************************************************************************\
+ *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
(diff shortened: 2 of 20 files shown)


dongahn:
@lipari: I spent sometime today to map out a few initial steps for scheduler-driven aggregate updates. Here is what I sketched -- so that I can get your feedback better if we have a chance to talk next week.

#### resrc
Start simple and place the planner-based aggregate structure only on a small set of resource type:
- At the resrc vertex of cluster type, maintain aggregates on nodes and cores under its subtree;
- At each resrc vertex of node type, maintain aggregates on cores under its subtree;
- Initialize and populate these aggregate planners at resrc generation time;
- Avoid using graph elements (e.g., BW and power), for now. 


#### resrc_reqst
Instead of computing aggregate requirements of the resrc_reqst by walking its hierarchy at each match-check time, compute node/core aggregate requirements upfront and stick them into resrc_reqst at certain levels:
- Stick the total node and core count requirement in aggregate to the top-level request vertex; 
- Stick the core count requirement in aggregate to the node-level request vertex.

#### match_resource
Augment `match_resource` to query the aggregate planner and avoid unnecessary resrc tree decent

#### insertion point at which the scheduler updates the aggregate planners right after a job is successfully scheduled
I still need to do some hacking to find the appropriate place. I initially envisioned I do this at the job tagging time but I didn't have a chance to look at the code today.

#### Performance testers and target
- Introduce large graph stress test -- we should in general benefit from adding another resrc unit test with a large cluster configuration with common walk patterns (e.g., including a large graph with available resources varying from many to small);
- Significantly speed up and scale for scheduling through many unsatisfiable high-priority jobs in the pending queue;
- Significantly speed up and scale for the case where many jobs need to complete and free resources before a large job can be scheduled under the conservative backfill policy

#### Correctness
I will certainly utilize existing tests and make sure all tests pass. But test cases to validate/verify the correctness of our scheduling algorithms are still somewhat limited, and it feels like we need some solutions there.

lipari:
@dongahn, the resource matching implementation that currently exists is a brute-force approach.  Its advantage is that is very versatile in the complex resource requests that can be matched.  The drawbacks are the performance penalty for visiting each child resource individually.
It is therefore ripe for optimizing and your efforts are welcome.  My only request at this point is that you think of a way to maintain the old brute force approach as a fall-back option.  If we dig too deeply to convert to an aggregate solution, we may discover someday a resource request that is complex enough to be un-satisfiable using the optimized code, but possible using the original, brute force method.

dongahn:
@lipari: Thanks. Yes, this is a great idea for debugging and verification!

dongahn:
As discussed with @morrone and @tpatki today, I'm closing this in preparation for a new planner-only PR.

PR #244, opened 2017-04-03
avoid flux-core bug 1027

garlick:
The main point of this PR is to add a workaround for flux-framework/flux-core#1027 so we don't have to push out another tag immediately, but the workaround turned out to be a bug fix in `flux-waitjob`, which was not exiting when a job had already run.

The "remove old sentinel files" fix was not based on any observed failure.  Happy to rebase without that one if it's not correct, but it seems right based on my read of hte code.

### DIFF
--- a/sched/flux-waitjob.c
+++ b/sched/flux-waitjob.c
@@ -173,6 +173,7 @@ static int wait_job_complete (flux_t *h)
 
     if (jsc_notify_status (h, waitjob_cb, (void *)h) != 0) {
         flux_log (h, LOG_ERR, "failed to register a waitjob CB");
+        goto done;
     }
     /* once jsc_notify_status is returned, all of JSC events
      * will be queued and delivered. It is safe to signal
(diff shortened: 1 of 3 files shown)


garlick:
Oh sorry, the calls to `adjust_sched_session()` do change the filenames used for each synchronization, so that last commit doesn't really fix any problems with the tests, as long as that's being called between every use.  It might be a good hygienic thing to do though.  Either way - I can rebase without it if you like.

lipari:
Thank you, @garlick.  I looked over the changes and am happy to merge with the hygienic measures included.  I wonder though, should we look to replace all calls to flux_log (h, LOG_... calls with the equivalent flux_msg(), flux_err(), and flux_errn()?

garlick:
Thanks.

Generally modules log to the broker ring buffer and commands log to stderr.  This was a case of a command logging to the broker ring buffer.

PR #245, opened 2017-04-12
build: enable -Werror=missing-field-initializers

garlick:
This PR should be merged after flux-framework/flux-core#1032 is merged.

It adds `-Werror=missing-field-initializers` to CFLAGS and adds missing initializers to the various statically defined `struct flux_msg_handler_spec` arrays.  In flux-core, a "rolemask" field has been added to this struct, and due to ambiguities about how these arrays might be declared and implicitly initialized, it is probably best to make sure rolemask isn't ever registered without explicit initialization.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -24,7 +24,7 @@ AC_PROG_MAKE_SET
 AM_PROG_CC_C_O
 AX_CODE_COVERAGE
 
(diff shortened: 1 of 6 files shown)


lipari:
Looks good!  Thanks, @garlick!

Issue #256, opened 2017-07-20
make check fails at put_job_in_kvs


sampollard:
I installed flux-core as usual, trying both v0.7.0 and master (the log I showed is for master). flux-sched's make check fails with issues like this:
```
flux-broker: src/zsock_option.c:1347: zsock_set_sndtimeo: Assertion `rc == 0 || zmq_errno () == (156384712 + 53)' failed.
flux-start: 0 (pid 128482) Aborted
FAIL: t2000-fcfs.t
flux-start: warning: setting --bootstrap=selfpmi due to --size option
not ok 2 - sim: scheduled and ran all jobs
#
#    timed_sync_wait_job 60
#
```

and some more stuff like

```
flux: put_job_in_kvs: Invalid argument
2017-07-20T00:13:12.676051Z submit.err[0]: put_job_in_kvs: failed to put 'submit_time'
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at src/zsys.c:379
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at src/zsys.c:380
E: (flux-broker) 17-07-20 00:13:12 dangling 'REP' socket created at src/zauth.c:77
E: (flux-broker) 17-07-20 00:13:12 dangling 'ROUTER' socket created at overlay.c:455
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at module.c:571
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at shmem.c:259
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at module.c:571
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at shmem.c:259
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at module.c:571
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at shmem.c:259
E: (flux-broker) 17-07-20 00:13:12 dangling 'PAIR' socket created at module.c:571
```

Here are the commit ids I'm using:

flux-sched: a30e2f079d28e35199e32bb19e0c70b789ac0b95
flux-core: 500e3661b3c29b98331b0dab84f603de9901eebc

I have uploaded the output file from make check here
[make_check_flux_sched.txt](https://github.com/flux-framework/flux-sched/files/1160853/make_check_flux_sched.txt)

[...]

PR #257, opened 2017-07-20
sim: fix kvsdir_get_double

SteVwonder:
jansson string format for doubles is "f" not "F"

"starting_time" and "io_time" keys might not always exist in the
KVS, check for them before calling `kvsdir_get_double`

combined with https://github.com/flux-framework/flux-core/pull/1114, this fixes https://github.com/flux-framework/flux-sched/issues/256

### DIFF
--- a/simulator/simulator.c
+++ b/simulator/simulator.c
@@ -206,7 +206,7 @@ int kvsdir_get_double (kvsdir_t *dir, const char *name, double *valp)
         goto done;
     if (!(f = flux_kvs_lookup (h, 0, key)))
         goto done;
-    if (flux_kvs_lookup_get_unpack (f, "F", valp) < 0)
+    if (flux_kvs_lookup_get_unpack (f, "f", valp) < 0)
         goto done;
     rc = 0;
 done:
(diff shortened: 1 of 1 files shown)


garlick:
@SteVwonder: "F" should work in unpack[1]:
```
f (real) [double]
Convert a JSON real to C double.

F (integer or real) [double]
Convert a JSON number (integer or real) to C double.
```
The distinction is a bit subtle and jansson dependent since JSON only supports "numbers"[2]:
<Hr>
[1]  http://jansson.readthedocs.io/en/2.8/apiref.html#c.json_unpack
<br>
[2] http://jansson.readthedocs.io/en/2.8/conformance.html#rfc-conformance

garlick:
Careful: the `kvsdir_exists()` call just checks whether the key exists in the already-fetched directory.  So if it appeared after that, it won't be found.  Not sure if that matters in this particular case.

SteVwonder:
@garlick: Ah. I see.  So it works in unpack but not pack, correct?  I am seeing it in the select cases for unpack (https://github.com/akheron/jansson/blob/master/src/pack_unpack.c#L725) but not for pack (https://github.com/akheron/jansson/blob/master/src/pack_unpack.c#L332).

W.r.t the `kvsdir_exists`, those last two keys are only added into the KVS after the jobs begins running; so the first time `pull_job_from_kvs` is called, the keys won't exist.  Maybe the better option is to always populate those values in the KVS with 0 when the job is created, so that those checks aren't necessary.

SteVwonder:
With the merge of https://github.com/flux-framework/flux-core/pull/1114, this PR shouldn't be necessary since "F" is a valid unpacking format char.  Let me restart the Travis build, and if it passes, I think we can close this PR without merging.

garlick:
> So it works in unpack but not pack, correct?

Right.

> Maybe the better option is to always populate those values in the KVS with 0 when the job is created, so that those checks aren't necessary.

Or maybe just add a boolean `required` option to the typed `kvsdir_get_*()` functions you own now?

Agreed this PR can be closed.

PR #260, opened 2017-08-03
Added ISO time parsing to job traces

sampollard:
It would be nice if we could take the output of `sacct` with no (or very little) modification and use it in a simulator run. This is the first step towards this.

You'll notice a bit of weirdness around line 119. This was required to get make check to pass, and I believe it has to do with the semantics when the Submit column is missing or ill-formed. Normally this should be an error but `atof` never returns an error. I tried to just set `stime = 0.0` or something, but that didn't work either. I'm not quite sure what is going on, but this passes `make check` so I am tempted to just leave it.

### DIFF
--- a/simulator/submitsrv.c
+++ b/simulator/submitsrv.c
@@ -90,7 +90,7 @@ double convert_time_to_sec (char *time)
 }
 
 // Populate a field in the job_t based off a value extracted from the csv
-int insert_into_job (job_t *job, char *column_name, char *value)
+int insert_into_job (flux_t *h, job_t *job, char *column_name, char *value)
 {
     if (!strcmp (column_name, "JobID")) {
         job->id = atoi (value);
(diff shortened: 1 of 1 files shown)


simulator/submitsrv.c
```diff
@@ -107,7 +107,21 @@ int insert_into_job (job_t *job, char *column_name, char *value)
...
+        char *endptr;
+        struct tm tm_spec; 
+        double stime = strtod(value, &endptr);
+        // If this fails then try to parse the string
+        if (endptr != NULL && endptr != value) {
+            endptr = strptime(value, "%Y-%m-%dT%H:%M:%S", &tm_spec);
+            if (endptr == NULL) {
+                // flux_log (h, LOG_WARNING, "Incorrect Submit fmt, expects %s",
+                //         "%Y-%m-%dT%H:%M:%S or seconds since epoch");
+                stime = atof(value);
```
lipari: It seems wrong at this point to assign a value to stime.  I would have expected a NULL assignment having failed to parse value two different ways.

sampollard:
Good catch. Now it checks if a partial match occurred for `strtod` (e.g. `2017-080-26T12:20:30` will match `2017.0` If either `strtod` or `strptime` fail then a warning is thrown.

simulator/submitsrv.c
```diff
@@ -107,7 +107,25 @@ int insert_into_job (job_t *job, char *column_name, char *value)
...
+            if (endptr == NULL) {
+                endptr = value;
+            } else {
+                stime = (double) mktime(&tm_spec);
+            }
+        }
+        if (endptr == value) {
+            flux_log (h, LOG_WARNING, "Incorrect Submit fmt, expects %s"
+                        " or seconds since epoch; replacing '%s' with %f",
+                        "%Y-%m-%dT%H:%M:%S", value, stime);
```
lipari: I'd spell it out for the user: "YYYY-MM-DDTHH:MM:SS".  Also, I defer to @SteVwonder to merge into code he originally created.

SteVwonder: @lipari, this is a good suggestion.  I would keep the strptime format and add the YYYY-MM-DDTHH:MM:SS in the message somewhere (since the strptime format is precise thus necessary but the other format is easier to recognize and intuit without looking at system call documentation).

Side note: it was not immediately clear to me what case this conditional is checking for.  After looking, it seems to be checking for a no-op from `strtod` and `strptime` returning `NULL`.  If that is the case, I would throw a short comment above the conditional to make it clear the exact conditions that lead to this being tripped.

SteVwonder:
Thanks @sampollard for this PR.  Once those tweaks are made, I think this is ready for merging. 👍

PR #261, opened 2017-08-14
Update sim-hackathon with unit tests and small fixes

sampollard:
This is the current version I am using which contains some newer improvements from flux-sched too.


sampollard:
I apologize, this was a mistake. I meant to merge this into a different branch. I request that this be deleted.

PR #268, opened 2017-09-26
planner: initial work for planner

dongahn:
Planner provides a simple abstraction and efficient mechanisms to keep track of the resource states for batch-job scheduling.

- `resource/planner/README.md` has a high-level description on the API and its mechanisms which are based on top of Linux OS kernel's red-black binary search tree code.
    
- `planner.h` has in-lined documentation for its API.
    
- Examples can be found in a unit test: `resource/planner/test/planner_test01.cpp`

Add C++ support into our build system in anticipation for future C++-based development.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -105,6 +105,9 @@ AC_CONFIG_FILES([Makefile
   rdl/test/Makefile
   resrc/Makefile
   resrc/test/Makefile
+  resource/Makefile
+  resource/planner/Makefile
+  resource/planner/test/Makefile
   sched/Makefile
   simulator/Makefile
   etc/Makefile
@@ -116,8 +119,10 @@ echo "
   Prefix...........: $prefix
   Debug Build......: $debug
   C Compiler.......: $CC
+  C++ Compiler.....: $CXX
   CFLAGS...........: $CFLAGS
   CPPFLAGS.......... $CPPFLAGS
+  CXXFLAGS.......... $CXXFLAGS
   FLUX.............: $FLUX
   FLUX_CORE_CFLAGS.: $FLUX_CORE_CFLAGS
   FLUX_CORE_LIBS...: $FLUX_CORE_LIBS
--- a/resource/Makefile.am
+++ b/resource/Makefile.am
@@ -1,0 +1,10 @@
+AM_CFLAGS = \
+	$(WARNING_CFLAGS) \
+	$(CODE_COVERAGE_CFLAGS)
+AM_LDFLAGS = $(CODE_COVERAGE_LDFLAGS)
+
+AM_CPPFLAGS = \
+	-I$(top_srcdir) \
+    $(CZMQ_CFLAGS)
+
+SUBDIRS = . planner
--- a/resource/planner/README.md
+++ b/resource/planner/README.md
@@ -1,0 +1,78 @@
+### PLANNER API
+
+Planner provides a simple abstraction and efficient mechanisms
+to keep track of the resource states for batch-job scheduling.
+Its API builds on two simple concepts:
+
+- Planner: just like your calendar planner, a planner object
+allows you to update and query the available resource states over
+time (i.e., integer unit-less time).
+
+- Span: just like you mark your activities over consecutive days
+in your calendar planner with a span, a span
+allows you to update your resource uses using a start time,
+duration and resource requirement.
+
+Planner is designed for speed. Specifically,
+it uses Linux OS kernel's red-black binary search tree code
+to update and query the resource states over time.
+It manages each of the two scheduled points of spans
+using two highly efficient binary search trees:
+
+- Scheduled point red-black binary search tree [1,2]
+- Min-time resource augmented red-black search tree [1,2]
+
+The scheduled-point search tree allows the planner to find the
+accurate resource state of any given instant time with
+(`O(log n)`) complexity.
+On the other hand, the min-time resource tree enables you to find
+the earliest scheduable point where your resource
+requirement can be satisfied also with (`O(log n)`) complexity.
+Say, you have a total of 10 available resource units and have
+allocated 5 units between time t1 and t2,
+[t1, t2), as well as 3 units from time t2 to t3, [t2, t3).
+Subsequently, you want to find about the earliest scheduled point at which you
+can allocate 7 units. The min-time tree will answer t2 as the
+earliest point in `O(log N)`. By contrast, a request for 5 units
+will be resolved to t1 with the same performance guarantee.
+Using both of these search trees, you can efficiently operate
+on the planner over both the time and resource dimensions alike.
+
+Planner was born out of real-world needs in Flux's
+batch-job scheduling infrastructure. As high performance
+computing (HPC) is undergoing significant changes
+in both performance-limiting resource types
+and workloads, simple data models
+(e.g., bitmaps) representing HPC compute and other
+consumable resources present increasingly greater
+challenges for effective scheduling. In response,
+Flux is in the process of modeling complex HPC
+resources using a graph. While this
+allows a highly sophisticated resource selections,
+it can be done at the expense of excessive, highly
+time-consuming searches of a large graph.
+
+Planner's efficient mechanism is specifically designed
+to mitigate the scalability and performance risk
+of such a graph-based scheduling approach. More specifically,
+it enables a novel scheme called scheduler-driven
+aggregate updates.
+
+As part of scheduling update activities (i.e., update
+for an allocation or reservation), the scheduler will
+update all of the planner objects which were involved
+in the allocation/reservation. This then allows each
+of the resource vertices to retain the up-to-date
+summary information on the vertices
+at its sub-tree or through other relevant topological
+connectivity.
+This information then allows the scheduler to prune the
+graph search for the subsequent allocation
+or reservation.
+
+[1] Red-black tree (Visited 9/27/2017), Wikipedia,
+https://en.wikipedia.org/wiki/Red-black_tree
+
+[2] Red-black Trees (rbtree) in Linux (Visited 9/27/2017), Rob Landley,
+https://www.kernel.org/doc/Documentation/rbtree.txt
+
--- a/resource/planner/planner.c
+++ b/resource/planner/planner.c
@@ -1,0 +1,1275 @@
+/*****************************************************************************\
+ *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <string.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <czmq.h>
+
+#include "src/common/libutil/rbtree.h"
+#include "src/common/libutil/rbtree_augmented.h"
+#include "src/common/libutil/xzmalloc.h"
+#include "planner.h"
+
+#define START(node) ((node)->start)
+#define LAST(node)  ((node)->last)
+
+typedef struct span span_t;
+typedef int64_t resource_array_t[PLANNER_NUM_TYPES];
+typedef char *resource_type_array_t[PLANNER_NUM_TYPES];
+
+typedef struct request {
+    int64_t on_or_after;
+    uint64_t duration;
+    resource_array_t resources;
+    size_t dimension;
+} request_t;
+
+/*! Scheduled point: time at which resource state changes.  Each point's resource
+ *  requirements are tracked as a node in a min-time resource (MTR) binary search
+ *  tree.
+ */
+typedef struct scheduled_point {
+    struct rb_node point_rb;     /* BST node for scheduled point tree */
+    struct rb_node resource_rb;  /* BST node for min-time resource tree */
+    int64_t subtree_min;         /* Min time of the subtree of this node */
+    int64_t at;                  /* Resource-state changing time */
+    int in_mt_resource_tree;     /* 1 when inserted in min-time resource tree */
+    int new_point;               /* 1 when this point is newly created */
+    int ref_count;               /* reference counter */
+    resource_array_t scheduled;  /* scheduled resources at this point */
+    resource_array_t remaining;  /* remaining resources (available) */
+} scheduled_point_t;
+
+/*! Node in a span interval tree to enable fast retrieval of intercepting spans.
+ */
+struct span {
+    int64_t start;               /* start time of the span */
+    int64_t last;                /* end time of the span */
+    int64_t span_id;             /* unique span id */
+    resource_array_t planned;    /* required resources */
+    size_t dimension;            /* vector size of required resources */
+    int in_system;               /* 1 when inserted into the system */
+    scheduled_point_t *start_p;  /* scheduled point object at start */
+    scheduled_point_t *last_p;   /* scheduled point object at last */
+};
+
+/*! Planner context
+ */
+struct planner {
+    resource_array_t total_resources;
+    resource_type_array_t resource_types;
+    size_t dimension;            /* size of the above arrays */
+    int64_t plan_start;          /* base time of the planner */
+    int64_t plan_end;            /* end time of the planner */
+    struct rb_root sched_point_tree;  /* scheduled point rb tree */
+    struct rb_root mt_resource_tree;  /* min-time resrouce rb tree */
+    scheduled_point_t *p0;       /* system's scheduled point at base time */
+    zhashx_t *span_lookup;       /* span lookup table by string id */
+    zhash_t *avail_time_iter;    /* tracking nodes temporarily deleted from MTR */
+    request_t *current_request;  /* the req copy for avail time iteration */
+    int avail_time_iter_set;     /* iterator set flag */
+    uint64_t span_counter;       /* current span counter */
+};
+
+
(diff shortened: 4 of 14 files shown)


dongahn:
Travis caught an issue with a Makefile. I am on it.

dongahn:
Sorry for all the noise. This is now ready for your review. Thanks!

resource/planner/test/Makefile.am
```diff
@@ -0,0 +1,23 @@
+AM_CXXFLAGS = \
+	@GCCWARN@ \
```
morrone: The @GCCWARN@ here should be $(WARNING_CXXFLAGS), but that variable needs fixing.

configure.ac
```diff
@@ -18,10 +18,18 @@ LT_INIT([dlopen])
...
 AC_PROG_LN_S
 AC_PROG_MAKE_SET
 AM_PROG_CC_C_O
+
+AC_PROG_CXX
+AX_CXX_COMPILE_STDCXX([11], [noext], [mandatory])
+if test "$GXX" = "yes"; then
+  WARNING_CXXFLAGS=$WARNING_CFLAGS
+  AC_SUBST([WARNING_CXXFLAGS])
+fi
```
morrone: This hunk should probably be in the commit that added the AX_CXX_COMPILE_STDCXX rather than in the planner commit.  Not a big deal though.

resource/planner/planner.c
```diff
@@ -0,0 +1,1271 @@
...
+#include "src/common/libutil/rbtree.h"
+#include "src/common/libutil/rbtree_augmented.h"
+#include "src/common/libutil/xzmalloc.h"
+#include "planner.h"
+
+#define START(node) ((node)->start)
+#define LAST(node)  ((node)->last)
+
+typedef struct rb_root rb_root_t;
+typedef struct rb_node rb_node_t;
```
morrone: I'm really not a fan of the above two local aliases.  Since the library doesn't provide them, I think it would be best not to use aliases here.

resource/planner/planner.c
```diff
@@ -0,0 +1,1271 @@
...
+        } else {
+            return this_data;
+        }
+    }
+    return last_state;
+}
+
+static int scheduled_point_insert (scheduled_point_t *new_data, rb_root_t *root)
+{
+    rb_node_t **link = &(root->rb_node);
```
morrone: Wouldn't this be equivalent, and much more readable, if this was an rb_node_t * instead of an rb_note_t **?

As far as I can tell, "link" is only a ** so that it can be used in rb_link_node().  But that is just the _return value_ from rb_link_node(), which we are throwing away.  So that usage of "link" is unrelated to the usages of "link" earlier in the function.  You might want to add a separate pointer for rb_link_node's return pointer just to make that distinction clear.

It looks like there are other places in this file where the same thing may be happening.

morrone: Errr, nevermind.  I think I see what is going on now. :)

dongahn: Yeah... I am not a big fan of the interface part of the Linux kernel RB tree. Brian B's perspective was "this is designed for speed; not for interface". FWIW, https://lwn.net/Articles/184495/ was pretty helpful.

configure.ac
dongahn: Yes, I was going to add this in.

resource/planner/planner.c
dongahn: :-) One of the reasons for `typedef` for me is reduce typing. In particular w/ C++'s very long template types, use of `typedef` for this purpose is essential. But this is C code and if you think this will increase readability, I am more than happy to use the original types.

dongahn:
> My first request would be to add a reference to red/black trees to make it easier for the reader to understand the foundation of what you're building on. I suggest http://pages.cs.wisc.edu/~paton/readings/Red-Black-Trees/ or https://en.wikipedia.org/wiki/Red%E2%80%93black_tree - whichever you prefer.

Good point. I will add a reference for RB Tree and then also the kernel article about its implementation/API.

dongahn:
> There are two red/black trees attached to every resource. For those resources that are part of a dominant hierarchy, there are two red/black trees assigned to every resource in the hierarchy.

Yes, and these search trees will be hidden under a planner object. (You will see from my upcoming `resource-query` work.)

resource/planner/planner.c
morrone: Yes, I think readability is more important that reduced typing.  The kernel guys don't allow wrapping every struct in a typedef, and I think there is good reason not to do that.  I think the typedef just adds a layer of indirection for the reader.  If the typedef and the struct declaration are in the same place it isn't quite as bad, but when the alias is completely local to one file and separate from the struct declaration, I think that makes things harder to follow.

I think in C, typedefs should be restricted to things like making an opaque handle.

dongahn:
> The second r/b tree assigned to each resource contains objects that contain minimum time. For example the above resource might have the first object of this min-time r/b tree be time c. (and here is where I need clarification.)

Mintime resource tree is much more general than this. Using your example,

> Therefore, each resource is assigned one red/black tree that tracks the allocation of size units over time. For example, the first r/b tree might have objects that contain 5 allocated units from time a to time b, 3 units from time b to c, and zero allocated units from time c to infinity.

Say, the resource has 10 unit total and a request wants to find about the minimum time where it can get 7 units. Then, the mintime search tree will answer `b` being the minimum time because obvious you can't get 7 units at the time a. On the other hand, if you want to find about the minimum time where it can get 5 units, it will answer `a`. 

The mistime tree answers each query in O(log N).

resource/planner/planner.c
dongahn: good to know. thanks.

dongahn:
Mintime tree is essential for reservation operation as we don't really want to do full tree search for each and all of the resource changing state points... Essentially, what the planner allows me to do is to prune both time and resource graph searches...

dongahn:
> So, perhaps you can see my confusion and insert text that clarifies the design description.

Yes, I will add this. In general, given the complexity of all of these, my strategy is to do a conference paper submission to describe all of the design decisions, performance tradeoffs etc.

dongahn:
> Also a question: will this design accommodate a resource allocation that is not monotonically decreasing? For example, in my above example, would the search algorithm support a case where there was a "d" object in the scheduled point tree which starts in 2 days and spans 5 minutes? IOW, would the search avoid this resource if it is being considered for a job that requests the resource for an entire week? The resource is available after object c, but then in 2 days, it's allocated again.

Yes, as you can see from some of the APIs, "availability" check includes the span check. If the count requirement cannot satisfied for the entire span, the resource vertex isn't available.

resource/planner/planner.h
```diff
@@ -0,0 +1,311 @@
...
+typedef struct span span_t;
+typedef struct planner planner_t;
+
+/*! Construct a planner.
+ *
+ *  \param base_time    earliest schedulable point expressed in integer time
+ *                      (i.e., the base time of the planner to be constructed).
+ *  \param duration     time span of this planner (i.e., all planned spans
+ *                      must end before base_time + duration).
+ *  \param resource_totals
```
lipari: is resource_totals the same as resource_counts?

dongahn: Yes, and good catch. I will change the `resource_counts` in the comment to `resource_totals`.

dongahn:
FYI -- I rebased this PR to PR #270 level. But I won't force a push yet since this can confuse reviewers. I will wait until at least one of the reviewers (e.g., @lipari) complete the review and incorporate all the changes requested by then into that push.

resource/planner/planner.h
dongahn: @lipari: Looking through the comments; I found other issues (typos and out-of-sync documents) and fixed some of these in my PR branch. Feel free to surface up issues in the comments as you will surely find. Thanks.

resource/planner/planner.c
```diff
@@ -0,0 +1,1271 @@
...
+static int64_t mintime_resource_subtree_min (scheduled_point_t *point)
+{
+    int64_t min = point->at;
+    scheduled_point_t *p = NULL;
+    if (point->resource_rb.rb_left) {
+        p = rb_entry (point->resource_rb.rb_left, scheduled_point_t, resource_rb);
+        if (min > p->subtree_min)
+            min = p->subtree_min;
+    }
+    if (point->resource_rb.rb_right) {
```
lipari: Without really understanding what is going here on (yet), the logic looks odd.  I would expect an "else if" here.  As written, the min variable can get assigned in the clause above and then overwritten in this clause.  Is that intended?

dongahn: This can be confusing, but this is correct code. 

Essentially, how linux augmented rbtree interface works is that you have to provide a callback function which allows the tree to maintain an accurate subtree quantify (in my case, the minimum time of the subtree). So, this function checks both branches and return the min between two.

What min-time resource tree does *essentially* is

1) descend to a node where a resource count can be satisfied

2) and then use the minimum time of its subtree to find the schedule point that has the minimum time. 

There are some corner cases but this is the main logic.

In a sense, it is similar to how you can use the augmented rbtree to implement an interval tree. Perhaps, this article can help you to understand the log behind this a bit better (look for interval tree sample usage):

https://www.kernel.org/doc/Documentation/rbtree.txt

lipari:
> FYI -- I rebased this PR to PR #270 level. But I won't force a push yet since this can confuse reviewers. I will wait until at least one of the reviewers (e.g., @lipari) complete the review and incorporate all the changes requested by then into that push.

Go ahead and force a push of your latest changes, @dongahn.  I'd like another day to complete the review, but at this point, it would help to review all the latest changes.

dongahn:
@lipari: okay, pushed.

dongahn:
FYI -- today's performance study revealed that there is a bug in the scheduled point reference counting scheme. I will patch this one liner next time I will force a push and add more "span remove" unit tests.

```diff
@@ -1066,6 +1064,7 @@ int64_t planner_add_span (planner_t *ctx, int64_t start_time,
     start_point->new_point = 0;
     span->start_p = start_point;
     last_point->new_point = 0;
+    last_point->ref_count++;
     span->last_p = last_point;

     update_mintime_resource_tree (ctx, list);

```

resource/planner/README.md
```diff
@@ -0,0 +1,78 @@
...
+computing (HPC) is undergoing significant changes
+in both performance-limiting resource types
+and workloads, simple data models
+(e.g., bitmaps) representing HPC compute and other
+consumable resources present increasingly greater
+challenges for effective scheduling. In response,
+Flux is in the process of modeling complex HPC
+resources using a graph. While this
+allows a highly sophisticated resource selections,
+it can be done at the exchange of excessive, highly
```
lipari: shouldn't this read:  "it can be done at the _expense_ of excessive..."

dongahn: That'd be better. I will change it.

resource/planner/planner.h
```diff
@@ -0,0 +1,312 @@
...
+#endif
+
+#define PLANNER_NUM_TYPES 5
+
+typedef struct span span_t;
+typedef struct planner planner_t;
+
+/*! Construct a planner.
+ *
+ *  \param base_time    earliest schedulable point expressed in integer time
```
lipari: "integer time" seems overly non-committal.  I suspect that this will default to seconds.  But then, I wonder is it epoch seconds?  Unless you see a scenario where integer time represented something else (e.g., milliseconds), would it be asking too much to commit to "seconds".  And if you do that, you may want to qualify that as epoch seconds.  Why?  It helps to remove a layer of abstraction which is (I propose) unnecessary.

resource/planner/planner.h
```diff
@@ -0,0 +1,312 @@
...
+
+/*! Find and return the earliest point in integer time when the request can be
+ *  satisfied.
+ *
+ *  Note on semantics: this function returns a time point where the resource state
+ *  changes. If the number of available resources change at time point t1 and
+ *  t2 (assuming t2 is greater than t1+2), the possible schedule points that this
+ *  function can return is either t1 or t2, but not t1+1 nor t1+2 even if these
+ *  points also satisfy the request.
+ *
```
lipari: > but not t1+1 nor t1+2 even if these points also satisfy the request.

I think this is because, in this example, there is no resource state change in t1+1 or t1+2.  If that's true, you might just drop the discussion of t1+1 and t1+2.  Or perhaps I am missing some relevance to the t1+1 and t1+2 times.

lipari:
I've finished my review and recommend merging pending the reviews of the other reviewers.  My only thought comes in the use of this new code.  As I see it, the span objects are generic - they can be used to make allocations or reservations.  Operationally, once the planner code is put to use, the spans representing the allocations will remain resident in the planner object over time - and removed once the job terminates.  The reservations will come and go with each scheduling loop.  I didn't see a mechanism that will allow you to handle allocations different from reservations.  And I wonder whether such a mechanism to lock in an allocation span should be part of the planner rudiments.

resource/planner/planner.h
dongahn: >  I suspect that this will default to seconds.

While I agree our upper level software (scheduling infrastructure) will program this to seconds, this really depends on the users of the planner API. 

If one wants to introduce a coarser grained scheduling, a tick could be programmed to be coarser than a second; if fine-grained scheduling, the unit could be finer than a second. 

Point is, the planner doesn't want to care about these physical time units and want to remain to treat time as unit-less. 1 just represents 1 tick in the upper level software's time granularity.

I believe this is a better separation of concerns.

lipari: That's what I suspected.  No problem.

resource/planner/planner.h
dongahn: This is just giving an example to make the previous sentence a bit more concrete. Neither t1+1 nor t1+2 has a resource state change. Unless this confuses folks too much, I think it makes sense to keep it.

dongahn:
Thank you for your thorough review @lipari!

> I didn't see a mechanism that will allow you to handle allocations different from reservations. And I wonder whether such a mechanism to lock in an allocation span should be part of the planner rudiments.

Good thought. But I need a bit more progress to further determine whether I need other idioms.

dongahn:
OK. I have made all the changes requested so far including @morrone's avoiding typedefs and forced a push.

src/common/libutil/Makefile.am
```diff
@@ -16,4 +16,8 @@ libutil_la_SOURCES = \
 	log.c \
 	log.h \
 	oom.h \
-	shortjansson.h
+	shortjansson.h \
+	compiler.h \
+	rbtree.c \
```
morrone: I think it would probably be best to put rbtree into its own helper library rather than mixing with libutil.  rbtree is third-party GPL code, so anything that uses it will be ineligible for inclusion in LGPL libraries that we might want to export.  Keeping it separate will make it easier to avoid accidental inclusion in the wrong place.

dongahn:
Thanks @morrone for your thorough review! I skimmed through them, and I believe most, if not all, of your suggestions will improve the quality of the planner. I have a visitor tomorrow so i will try to address them this Friday.

resource/Makefile.am
```diff
@@ -0,0 +1,10 @@
+AM_CFLAGS = \
```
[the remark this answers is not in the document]
dongahn: Make sense.

resource/planner/planner.c
```diff
@@ -0,0 +1,1275 @@
...
+{
+    int i = 0;
+    ctx->plan_start = base_time;
+    ctx->plan_end = base_time + (int64_t)duration;
+    ctx->sched_point_tree = RB_ROOT;
+    ctx->mt_resource_tree = RB_ROOT;
+    ctx->p0 = xzmalloc (sizeof (*(ctx->p0)));
+    ctx->p0->at = base_time;
+    ctx->p0->ref_count = 1;
+    memset (ctx->p0->scheduled, '\0', sizeof (ctx->p0->scheduled));
```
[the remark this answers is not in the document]
dongahn: I didn't realize xzmalloc actually zeroes the buffer. Thanks.

resource/planner/planner.c
```diff
@@ -0,0 +1,1275 @@
...
+}
+
+int planner_reset (planner_t *ctx, int64_t base_time, uint64_t duration)
+{
+    if (duration < 1) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    erase (ctx);
```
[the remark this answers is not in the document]
dongahn: Wow. Great catch, @morrone!

resource/planner/planner.c
```diff
@@ -0,0 +1,1275 @@
...
+    initialize (ctx, base_time, duration);
```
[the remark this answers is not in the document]
dongahn: This initial pseudo schedule point should be freed by `scheduled_points_destroy()`

resource/planner/planner.h
```diff
@@ -0,0 +1,311 @@
...
+ *                      are supported.
+ *  \param resource_types
+ *                      string array of size len where each element contains
+ *                      the resource type corresponding to each corresponding
+ *                      element in the resource_totals array.
+ *  \param len          length of the resource_totals and resource_types arrays;
+ *                      must not exceed PLANNER_NUM_TYPES.
+ *  \return             new planner context; NULL on an error with errno set
+ *                      as follows:
+ *                      0 on success; -1 on an error with errno set:
```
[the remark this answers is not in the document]
dongahn: Sure.

resource/planner/planner.h
```diff
@@ -0,0 +1,311 @@
...
+ */
+int planner_reset (planner_t *ctx, int64_t base_time, uint64_t duration);
+
+/*! Destroy the planner.
+ *
+ *  \param ctx_p        pointer to a planner context pointer returned.
+ *                      from planner_new
+ *
+ */
+void planner_destroy (planner_t **ctx_p);
```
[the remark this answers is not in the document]
dongahn: For now, I will just go with as-is as this is a common form of flux interfaces. If we later decide the other form has a clear advantage, we can make this change globally.

src/common/libutil/Makefile.am
dongahn: Great suggestion. This makes perfect sense!

resource/planner/planner.h
```diff
@@ -0,0 +1,311 @@
...
+ *                      the requested resource counts.
+ *  \param len          length of resource_counts and resource_types arrays.
+ *                      must not exceed PLANNER_NUM_TYPES.
+ *  \return             earliest time at which the request can be satisfied;
+ *                      -1 on an error with errno set as follows:
+ *                          EINVAL: invalid argument.
+ *                          ERANGE: resource_counts contain an out-of-range value.
+ *                          ENOTSUP: internal error encountered.
+ */
+int planner_avail_during (planner_t *ctx, int64_t at, uint64_t duration,
```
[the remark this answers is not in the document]
dongahn: Ah. I can see why this could be confusing with "at" in the interface. I will go with your suggestion.

resource/planner/README.md
```diff
@@ -0,0 +1,78 @@
...
+Flux is in the process of modeling complex HPC
+resources using a graph. While this
+allows a highly sophisticated resource selections,
+it can be done at the expense of excessive, highly
+time-consuming searches of a large graph.
+
+Planner's efficient mechanism is specifically designed
+to mitigate the scalability and performance risk
+of such a graph-based scheduling approach. More specifically,
+it enables a novel scheme called scheduler-driven
```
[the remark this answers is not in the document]
dongahn: I think the "scheduler-driven aggregate updates" scheme would need a more in-depth and formal technical writing (probably a section in a conference paper later). I will just genericize and shorten this to one sentence.

dongahn:
Given all the structural changes I have, it makes more sense to do a fresh PR. Closing this and will post a new PR shortly.

PR #275, opened 2017-10-24
    sched:  Add priority plugin

lipari:
The new priority plugin can be loaded in addition to the existing fcfs
    or backfill plugins.

    With the introduction of the priority plugin, the existing "sched_plugin"
    had to be renamed to the "behavior_plugin".  "sched_plugin" now refers to
    any plugin loaded by the sched module.  The two categories of sched
    plugins, with different API's, are the behavior and priority plugins.

### DIFF
--- a/sched/plugin.h
+++ b/sched/plugin.h
@@ -3,7 +3,7 @@
 
 #include "scheduler.h"
 
-struct sched_plugin {
+struct behavior_plugin {
     void         *dso;                /* Scheduler plug-in DSO handle */
     char         *name;               /* Name of plugin */
     char         *path;               /* Path to plugin dso */
@@ -44,6 +44,16 @@ struct sched_plugin {
                                          const sched_params_t *params);
 };
 
+struct priority_plugin {
+    void         *dso;                /* Scheduler plug-in DSO handle */
+    char         *name;               /* Name of plugin */
+    char         *path;               /* Path to plugin dso */
+
+    int         (*priority_setup)(flux_t *h);
+    void        (*prioritize_jobs)(flux_t *h, zlist_t *jobs);
+    int         (*record_job_usage)(flux_t *h, flux_lwj_t *job);
+};
+
 /* Create/destroy the plugin loader apparatus.
  */
 struct sched_plugin_loader;
@@ -63,7 +73,10 @@ void sched_plugin_unload (struct sched_plugin_loader *sploader);
 /* Retrieve the currently loaded plugin.
  * Return NULL if plugin is not loaded.
  */
-struct sched_plugin *sched_plugin_get (struct sched_plugin_loader *sploader);
+struct behavior_plugin *behavior_plugin_get (struct sched_plugin_loader
+                                             *sploader);
+struct priority_plugin *priority_plugin_get (struct sched_plugin_loader
+                                             *sploader);
 
 #endif /* !_FLUX_SCHED_PLUGIN_H */
 
--- a/sched/sched_priority.c
+++ b/sched/sched_priority.c
@@ -1,0 +1,478 @@
+#if HAVE_CONFIG_H
+# include <config.h>
+#endif
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <argz.h>
+#include <libgen.h>
+#include <errno.h>
+#include <libgen.h>
+#include <czmq.h>
+
+#include <flux/core.h>
+
+#include "src/common/libutil/log.h"
+#include "src/common/libutil/shortjansson.h"
+#include "src/common/libutil/xzmalloc.h"
+#include "scheduler.h"
+
+typedef struct {
+    char   *name;
+    double weight;
+    double (*fn)(flux_lwj_t *job);
+} job_priority_factor_t;
+
+static zlist_t *prio_factors;          /* Job priority factors */
+static zhash_t *associations;          /* charge account / user associations */
+
+/*
+ * An association contains the shares of computing resources assigned
+ * to a charge account and optionally the user permitted to charge
+ * that account.  There can be a hierarchy of associations with the
+ * "root" account at the top.  Each level in the hierarchy contains
+ * associations of associations with the leaves being associations of
+ * users.
+ *
+ * For example, if a user is permitted to charge the "physics"
+ * account, there will be an association of that user and account
+ * "physics".  A different association will exist for every other user
+ * permitted to charge the physics account.
+ *
+ * If the physics account is a child of the "science" account, then
+ * both the physics and science associations will have their user
+ * member set to NULL.
+ *
+ * In addition to shares, an association records the usage of
+ * computing resource, which is nominally in units of core * seconds.
+ *
+ * The fair-share factor for the association represents the difference
+ * between the assigned shares and accrued usage.  The fair-share
+ * factor ranges from 0.0 to 1.0 where:
+ *   0.0 represents an over-serviced association
+ *   0.5 usage is commensurate with assigned shares
+ *   1.0 association has no accrued usage
+ */
+typedef struct association association_t;
+struct association {
+    char   *account;
+    char   *user;
+    double shares;
+    double sibling_shares;      /* sum of all children's shares */
+    double usage;               /* historical cpu-second sum */
+    double sibling_usage;       /* sum of all children's usage */
+    double low_factor;          /* fair-share factor of next lowest sibling */
+    double fs_factor;           /* normalized fair-share factor */
+    association_t *parent;
+    zlist_t *children;
+};
+
+
+static association_t *create_association (char* account, char *user, double shares,
+                                          association_t *parent)
+{
+    association_t *assoc = xzmalloc (sizeof (association_t));
+
+    assoc->account = account;
+    assoc->user = user;
+    assoc->shares = shares;
+    assoc->usage = 0.0;
+    assoc->low_factor = 0.0;
+    assoc->fs_factor = 1.0;
+    assoc->parent = parent;
+
+    return assoc;
+}
+
+/*
+ * Create the associations file that this plugin requires by invoking:
+ * sacctmgr -n -p show assoc cluster=$LCSCHEDCLUSTER
+ *   format=account,share,parentn,user > associations
+ */
+static int read_association_file (flux_t *h)
+{
+    FILE        *fp;
+    association_t *assoc;
+    association_t *parent_assoc;
+    char        line[1028];
+    char        *account = NULL;
+    char        *field = NULL;
+    char        *filename = "associations";
+    char        *key;
+    char        *parent_name = NULL;
+    char        *ptr;
+    char        *user = NULL;
+    double      shares = 1.0;
+    int         n_assoc = 0;
+    int         rc = 0;
+
+    if ((fp = fopen (filename, "r"))) {
+        while (fgets (line, sizeof(line), fp) != NULL) {
+            if ((ptr = strchr (line, '|'))) {
+                *ptr = '\0';
+                account = line;
+                field = ptr + 1;
+            }
+            if ((ptr = strchr (field, '|'))) {
+                *ptr = '\0';
+                shares = atof (field);
+                field = ptr + 1;
+            }
+            if ((ptr = strchr (field, '|'))) {
+                *ptr = '\0';
+                parent_name = field;
+                field = ptr + 1;
+            }
+            if ((ptr = strchr (field, '|'))) {
+                *ptr = '\0';
+                user = field;
+            }
+            /*
+             * In the ordering accounts in the sacctmgr output, parent
+             * accounts appear before child accounts.  Hence
+             * parent_assoc should always found and be non-NULL.  The
+             * only exception is the root account - it has no parent.
+             */
+            parent_assoc = zhash_lookup (associations, parent_name);
+
+            if ((assoc = create_association (account, user, shares, parent_assoc))) {
+                if (user)
+                    /* the key will be a fusing of the account and user name */
+                    key = xasprintf ("%s-%s", account, user);
+                else
+                    /* this is an account, so the key is just the account */
+                    key = xasprintf ("%s", account);
+                zhash_insert (associations, key, assoc);
+                zhash_freefn (associations, key, free);
+                n_assoc++;
+            } else {
+                flux_log (h, LOG_ERR, "%s: failed to create association %s-%s", __FUNCTION__,
+                          account, user);
+                rc = -1;
+            }
+        }
+        fclose(fp);
+        flux_log (h, LOG_INFO, "priority plugin loaded %d associations", n_assoc);
+    } else {
+        flux_log (h, LOG_ERR, "%s: failed to open %s", __FUNCTION__, filename);
+        rc = -1;
+    }
+
+    return rc;
+}
+
+/* Computes the sum of raw shares and used resources across all
+ * siblings */
+static void sum_sibling_shares_usage (association_t *a)
+{
+    association_t *sib;
+    zlist_t *sibs = a->children;
+
+    a->sibling_shares = 0.0;
+    a->sibling_usage = 0.0;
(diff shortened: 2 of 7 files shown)


lipari:
This is an early look into the proposed priority plugin.  Details can be found [here](https://github.com/lipari/flux-sched/wiki/Job-Priority-Plugin-Implementation-Notes).  This is more for review now than actual merging.  There is more to do, time permitting.

SteVwonder:
I think this is a solid start to the priority plugin.  Before merging, I would like to see some coverage by the CI testing.  At minimum, I think an end-to-end simulation with the priority plugin loaded and changing the order of job scheduling/completion is required.  It might be nice to have a few unit tests for some of the idempotent functions (e.g., `calc_fs_factor`).  An example association file should also be included. Is the specification for this file documented anywhere?

sched/plugin.h
```diff
@@ -44,6 +44,16 @@ struct sched_plugin {
                                          const sched_params_t *params);
 };
 
+struct priority_plugin {
```
morrone: This interface probably warrants its own header, with clear documentation of where the hooks are called and what the expectations are from the plugin.

sched/plugin.h
```diff
@@ -44,6 +44,16 @@ struct sched_plugin {
...
+    void         *dso;                /* Scheduler plug-in DSO handle */
+    char         *name;               /* Name of plugin */
+    char         *path;               /* Path to plugin dso */
+
```
morrone: I think that the hooks should be less specific to the current implementation.  One way to do that is to name them more after where they are called, rather than a name that assumes what the plugin does at that call point.

sched/plugin.h
```diff
@@ -44,6 +44,16 @@ struct sched_plugin {
...
+    int         (*priority_setup)(flux_t *h);
+    void        (*prioritize_jobs)(flux_t *h, zlist_t *jobs);
```
morrone: The type being pointed to in the zlist needs to be documented.  This code would probably be well served if we uplifted it to C++.  A std::vector declaration includes the type that it contains, so it is self-documenting in that one respect.

Maybe this should sort the list rather than the caller?  Potentially that lets us do away with the numeric job priority in the job structure, and instead let the plugin decided the factors on which it wishes to prioritize the jobs.  A FIFO priority plugin has no real use for a numeric priority, for instance.  It could directly sort n submission times.

So perhaps instead of a "priority" integer in the jobs stuct, we should have an opaque "void *priority_data" in case the plugin wishes to attach data to the struct.  That implies that we would want to a "job_submission" hook to allow the priority plugin to initialize and attach its data to the job struct.  The job_submission() hook might also be responsible to inserting the job into the correct initial location in a job list.  That would make a FIFO priority implementation even easier, because the order never changes and the sort hook becomes a no-op.

sched/plugin.h
```diff
@@ -44,6 +44,16 @@ struct sched_plugin {
...
+    int         (*record_job_usage)(flux_t *h, flux_lwj_t *job);
```
morrone: A better name for this is probably "job_completion".

sched/scheduler.h
```diff
@@ -55,10 +55,16 @@ typedef struct flux_resources {
 typedef struct {
     int64_t     lwj_id;  /*!< LWJ id */
     job_state_t state;   /*!< current job state */
+    char    *user;       /*!< user name of person who submitted job */
```
morrone: This should probably be the same user ID type as used by flux core (probably unint32_t).  If the priority plugin needs to look up the username string, it can do that and store it in its opaque data struct

sched/scheduler.h
```diff
@@ -55,10 +55,16 @@ typedef struct flux_resources {
...
+    char    *account;    /*!< account to charge for resource usage */
```
morrone: I think this should probably be in the plugin's opaque data in the struct rather than in the job struct.

sched/scheduler.h
```diff
@@ -55,10 +55,16 @@ typedef struct flux_resources {
...
     flux_res_t *req;     /*!< resources requested by this LWJ */
     resrc_tree_t *resrc_tree; /*!< resources allocated to this LWJ */
+    int64_t submittime;
     int64_t starttime;
+    int64_t endtime;
     int64_t enqueue_pos; /*!< the initial enqueue position */
+    double user_prio;    /*!< user-requested priority */
```
morrone: I'm not sure how to handle this, because I'm not sure of its semantics.  Probably this should also be in the opaque data.  When the job is created, the priority plugin's "job_submission" hook either be passed the entire job spec, or maybe some subset like the "system" attributes hash.  Some priority plugins might understand some options, others might understand some other set of options.

sched/scheduler.h
```diff
@@ -55,10 +55,16 @@ typedef struct flux_resources {
...
+    double priority;     /*!< scheduling priority */
```
morrone: Perhaps move to opaque data as mentioned earlier.

sched/plugin.h
```diff
@@ -44,6 +44,16 @@ struct sched_plugin {
                                          const sched_params_t *params);
 };
 
+struct priority_plugin {
+    void         *dso;                /* Scheduler plug-in DSO handle */
+    char         *name;               /* Name of plugin */
+    char         *path;               /* Path to plugin dso */
+
+    int         (*priority_setup)(flux_t *h);
```
morrone: Perhaps this should be the only function that is passed the flux_t handle, and it should just remember it?
We should have a matching shutdown function too.

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+    int rc = -1;
+    job_priority_factor_t* jpf = NULL;
+
+    if (!(prio_factors = zlist_new ()))
+        oom ();
+
+    if (!(associations = zhash_new ()))
+        oom ();
+
+    jpf = &jpfs[0];
```
morrone: Should just be "jpf = jpfs".

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+
+static double user (flux_lwj_t *job)
+{
+    /* User allowed to decrease their job's priority only */
+    if (job->user_prio < 0.0)
+        return (job->user_prio);
+    return 0.0;
+}
+
+job_priority_factor_t jpfs[] = {
```
morrone: Should be static?  Although I am guessing that this is a work in progress.  I imagine these weights will be configurable.

I don't see any value in putting the function pointers in every single zlist on every single job.  Instead there should be a separate name-to-function lookup table, which can always be static.  The name-to-weight able would be the one that is configurable.

morrone:
Note that I haven't scrutinized the fare-share math yet.

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+ * The fair-share factor ranges from 0.0 to 1.0:
+ *  0.0 - represents an over-serviced association
+ *  0.5 - usage is commensurate with assigned shares
+ *  1.0 - association has no accrued usage
+ *
+ * Plotted, it looks like this:
+ * http://www.wolframalpha.com/input/?i=Plot%5B2**(-u%2Fs)%5D,+%7Bu,+0.,+1.%7D,+%7Bs,+0.,+1.%7D
+ *
+ * While this gives us our fs factor relative to our siblings, we need
+ * to "inherit" the fair-share of our parent's account.  Our parent's
```
morrone: I'm confused by that assertion.  "fs" already incorporates parentage information through "S".  So fs would already appear to give us a _global_ factor, not one only relative to our siblings.

Also, the following formula for combinine multiple fs (from the formula above) should really be presented in formula form rather than only be example.

I'd also love to see an explantion of _why_.  I'm really puzzled by the part below where the parent's fair-share value is reduced by the fair-share value of its "next lower sibling".

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+        a->sibling_usage += sib->usage;
+        sib = zlist_next (sibs);
+    }
+}
+
+/*
+ * The formula for the fair-share factor is:
+ *   fs = 2**(-U/S)
+ * where:
+ * U = my_usage / clusters_usage
```
morrone: I think that "clusters_usage" may be incorrect here. Or perhaps the name is just ambiguous without definition.  But I would think that we want the value in U's denominator to be total cpu time over the same time period as "my_usage".  

If a user were to use 1% of a cluster for an hour, and the other 99% was idle, I would think that "clusters_usage" would be 1% of the cpu time in the cluster.  This would result in a U of 1, which would push fs towards 0.  But really U should be 0.01.

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+ * where A, B, and C are three accounts plotted at their fair-share
+ * value.  U3 is a user in account B who has no accrued usage.  Hence,
+ * U3's fair-share on its own is 1.0.  However, it inherits its parent
+ * (B)'s fair-share value.  U1 and U2 have their own fair-share values
+ * of 0.3 and 0.7.  However, these also inherit parent B's fairshare.
+ * This is done by multiplying (one minus their local values) by the
+ * the difference between their parent (B)'s fair-share value and the
+ * fair-share value of the next lower sibling of their parent: A in
+ * this case; and then subtracting this product from their parent B's
+ * fair-share value.
```
morrone: As far as I can tell, this is not how we currently do it in slurm.  So this part probably needs work if I am understanding it correctly.

garlick:
> If this PR does no harm (passes tests, is not egregiously off course), should we try to merge it with a minimum of changes and defer more substantive changes to another PR and/or open some issues for later?

Just realized I left off a reason for this comment.   I just meant it may be useful to have this snapshot of @lipari's thinking in the git history to refer back to, since he brought a lot of experience to the table.  That could be accomplished also by adding to this PR without squashing the original I guess.

morrone:
There is a bug in here too in that the "children" zlist of the associations_t is not initialized nor has children appended.

sched/sched_priority.c
```diff
@@ -0,0 +1,478 @@
...
+                field = ptr + 1;
+            }
+            if ((ptr = strchr (field, '|'))) {
+                *ptr = '\0';
+                user = field;
+            }
+            /*
+             * In the ordering accounts in the sacctmgr output, parent
+             * accounts appear before child accounts.  Hence
+             * parent_assoc should always found and be non-NULL.  The
```
morrone: This does not appear to be a safe assumption.  On one of our clusters, the sacctmgr output begins like this:

```
   Account     Share   Par Name       User 
---------- --------- ---------- ---------- 
    guests         1                   username 
      root         1                       
      root         1                  root 

```

(where "username" is a real user login name on the system)

lipari:
I just added a [Note About The Flux Fair-Share Algorithm](https://github.com/flux-framework/flux-sched/wiki/Note-About-The-Flux-Fair-Share-Algorithm) to describe the fair-share algorithm that appears in the PR.  The implementation in this PR is indeed different from Slurm - both its original multi-factor plugin as well as the Fair-Tree enhancement.

I appreciate @morrone 's review.  I can't commit at this time to submitting a revised PR that address his remarks.

garlick:
Excellent overview @lipari!  Thanks so much for posting that.

dongahn:
The same here, @lipari!

morrone:
HI @lipari, thanks for the information!  I am not expecting you to make any updates on this PR; I'm happy t pick up the baton.

If you have time though, it would be helpful if you could add some detail to your [Note About The Flux Fair-Share Algorithm](https://github.com/flux-framework/flux-sched/wiki/Note-About-The-Flux-Fair-Share-Algorithm) page.  Some questions that I have:

In what way did fair-tree break the sshare's output?  Why does a useful output require an algorithmic change?

It says: "LLNL saw the value of the fair-tree solution and activated it across all of our clusters where it remains to this day."  Can you explain how that is activated?  I find that "PriorityFlags=FAIR_TREE" can be set newer releases of slurm, but our production slurm does not have that option.  It looks like we are using priority/multifactor, and I can't find any obvious changes in the history of that plugin that would have implemented fair-tree.  As far as I can tell, we are using fair-share in production, not fari-tree.  Am I missing anything?

morrone:
I talked to some folks and figured out what is going on with fair-share/fair-tree.  The only places where fair-tree is enabled is on bluegene systems.  Other systems are going to be moving from normal fair-share to fair-tree in the future.

lipari:
Indeed, the Fair_Tree parameter is only available on LC systems running Slurm v17.02.  But this provides a good opportunity to compare the older sshare output which I believe is running on quartz against a BG/Q system or catalyst.

It looks like you've identified the problem with sshare output when Fair_Tree is enabled.  But if you haven't, I can add a description (from memory) or you can talk with Greg T.  He ran first hand into the problem from an inquiring user.

lipari:
> It looks like we are using priority/multifactor, and I can't find any obvious changes in the history of that plugin that would have implemented fair-tree. As far as I can tell, we are using fair-share in production, not fair-tree. Am I missing anything?

Just to be clear, all LC systems have the multi-factor priority plugin loaded.  But only the v17.02 systems, in addition, have the Fair_Tree flag set - which changes the behavior of the multi-factor plugin.

morrone:
There is a bug in the algorithm implementation.  "norm_usage" is calculated like this:

```   norm_usage = a->usage / a->sibling_usage;```

But usage should be divided by the _entire_ system usage.  And really, I think it turns out that "sibling_usage" is not something useful at all.  An account's usage is always going to be the sum of its children's usage, and that is also what sibling_usage is.  So no reason to have sibling_usage at all.

Probably a first approach to fixing norm_usage is to do something like this, since I think it is how slurm does it:

``` norm_usage = a->usage / root_node->usage;```

Better still (as I think I mentioned elsewhere) would be to divide by _available_ usage, not just actually usage.  There is probably no practical difference when a cluster is 100% allocated, but when the cluster usage is less than 100%, we will, I think, be overcharging accounts.

I'll fix norm_usage and get rid of sibling_usage on my development branch.

**Comment by morrone** [comment]
I take back my previous comment!  I was trying to make the code match the algorithm stated in the code comment:

```
 * The formula for the fair-share factor is:
 *   fs = 2**(-U/S)
 * where:
 * U = my_usage / clusters_usage
 * S = ((my_shares / all_siblings_shares) *
 *      (my_parents_shares / all_parents_siblings_shares) *
 *      (grandparents_shares / all_grandparents_siblings_shares) *
 *      ...
```

but now I think I understand that it is the _comment_ that is more incorrect than the code.  The code seems to be trying to implement something a little more like fair-tree's implementation where:

```
LF = S / U
S = Srawself / Srawself+siblings
U = Urawself / Urawself+siblings
```
(See https://slurm.schedmd.com/fair_tree.html)

At least, S and U come from the fair-tree model, but the code still uses this:

```
  fs = 2 ** (-U/S)
```

To ensure that the ratio is constrained to the range of 0-1.0.

So the code to calculate norm_shares and norm_usage just needs to be fixed to really divide by the total self+siblings values rather than dividing by the children values as it does now.

**Comment by morrone** [comment]
Next oddity is in calc_siblings_fs_factors().  This is the initialization:

```
    last_low_factor = p->low_factor;
    sav_low_factor = last_low_factor;
```

But that can't be right, can it?  "a" is the parent of its children, so "p" is the _grandparent_ of the children.  We wouldn't want to use grandparents's next lowest sibling's fs factor as the default...we want to use the _parent's_ next lowest sibling's fs factor.  So that should be "a->low_factor", I think.

lipari:
> but now I think I understand that it is the comment that is more incorrect than the code.

Sorry you have to slog through this, @morrone.  I wish I had the time to have tested everything so you could see how the theory works in practice.  But, w/r/t your comment, yes, the comment is misleading w/r/t the fair-tree-like code which follows.  The comment refers to the original multi-factor algorithm which the code in this PR attempts to improve.

> So that should be "a->low_factor", I think.

I'm afraid too much time has elapsed to be absolutely sure.  But yes, I believe you've spotted a bug.  last_low_factor should get initialized to a->low_factor.

PR #278, opened 2017-11-11
track bulk message handler registration API change

garlick:
This PR will allow sched to compile after flux-framework/flux-core#1277 is merged.

I was able to compile test this but for some reason could not get `make check` to pass on my desktop.  I reverted back to master on both flux-core and flux-sched and couldn't get that to work either so I assume it's environmental and unrelated to this PR, so submitted it anyway.

After the above gets merged, travis will need to be manually restarted here.

### DIFF
--- a/rdl/test/Makefile.am
+++ b/rdl/test/Makefile.am
@@ -5,7 +5,7 @@ AM_CPPFLAGS = -I$(top_srcdir) $(JANSSON_CFLAGS) $(LUA_INCLUDE)
 
 TESTS_ENVIRONMENT = \
     LUA_PATH="$(abs_top_srcdir)/rdl/?.lua;$(FLUX_PREFIX)/share/lua/5.1/?.lua;$(FLUX_PREFIX)/share/lua/5.1/fluxometer/?.lua;$(LUA_PATH);;" \
-    LUA_CPATH="$(abs_top_builddir)/rdl/?.so;$(abs_top_builddir)/rdl/test/.libs/?.so;$(FLUX_PREFIX)/lib64/lua/5.1/?.so;$(LUA_CPATH);;" \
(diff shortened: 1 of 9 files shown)


morrone:
@garlick, forget about the problem I mentioned in the hallway.  I cleaned out my /usr/local, reinstalled flux-core, and all the flux-sched tests pass now for me.

I don't know why we keep getting "FAIL: trdl" in travis.

garlick:
Thanks @grondo, hmm, maybe fallout from

flux-framework/flux-core#1275 Update to the latest ax_lua.m4

?  I'll try adding something to this PR to that effect.

I found out my problems on my desktop were that I had some parts of flux-core installed in /usr/local in unreadable directories (due to restrictive umask setting during install).  After that I get further in the tests but I still get stuck here:
```
FAIL: t1004-module-load.t 11 - module-load: no jobs are lost
```
looking a little closer, 
```
not ok 11 - module-load: no jobs are lost
#	
#	    for i in `seq $sched_start_jobid $sched_end_jobid`
#	    do
#	        state=$(flux kvs get $(job_kvs_path $i).state)
#	        if test $state != "submitted"; then
#	            return 48
#	        fi
#	    done &&
#	    flux module remove sched
#	
```
adding a little echo in there to see what the `state` key is set to, I get:
```
"submitted" != submitted
```
which sounds like fallout from earlier KVS changes, (json string vs raw data).  So maybe that will be the next problem we hit here (though @morrone claims he is passing all the tests on his desktop and hit this along the way before cleaning out _his_ /usr/local.).

grondo:
I forgot to mention that I was also seeing `not ok 11 - module-load: no jobs are lost` after fixing the Lua path issue.

grondo:
> Thanks @grondo, hmm, maybe fallout from
>
> flux-framework/flux-core#1275 Update to the latest ax_lua.m4

Yes, that was it. There was a single modification we'd added to `ax_lua.m4` which was dropped in the update:


https://github.com/flux-framework/flux-core/commit/90e333fce0ce156d6c04cb8c784067b4a80aee40#diff-5c5d94ac801dec27a4153e6500233409

Unfortunately, I'm not sure why we added that change, so I'd be hesitant to put it back into flux-core, since it will certainly bite us again in the future.

garlick:
OK,passing travis now.  I'd say let's merge this despite the poor test coverage results, since sched won't compile against the newer core without it.

PR #281, opened 2017-11-19
planner: fix mintime resource tree corruption bug

dongahn:
- Add a bug fix for a mintime resource tree corruption problem.
- Add a test case.
- This will be needed for upcoming `resource-query` test cases.


dongahn:
Thanks!

PR #289, opened 2018-01-31
Memleak fix

nikhil-jain:
Different sources of memory leak discovered via Valgrind.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -861,7 +861,7 @@ static int sim_job_status_cb (const char *jcbstr, void *arg, int errnum)
         return -1;
     }
 
-    event->jcb = Jget (jcb);
+    event->jcb = jcb;
     event->arg = arg;
     event->errnum = errnum;
 
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -420,6 +420,8 @@ static inline int fill_resource_req (flux_t *h, flux_lwj_t *j)
 done:
     if (jcb)
         Jput (jcb);
+    if(jcbstr)
+        free (jcbstr);
     return rc;
 }
 
@@ -435,6 +437,7 @@ static int update_state (flux_t *h, uint64_t jid, job_state_t os, job_state_t ns
     json_object_set_new (jcb, JSC_STATE_PAIR, o);
     jcbstr = Jtostr (jcb);
     rc = jsc_update_jcb (h, jid, JSC_STATE_PAIR, jcbstr);
+    free (jcbstr);
     Jput (jcb);
     return rc;
 }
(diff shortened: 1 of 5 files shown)


SteVwonder:
Thanks @nikhil-jain for finding and fixing these leaks.

Great catch on the `Jtostr` leak.  Assuming the string is still owned by the JSON library seems to also be made in [several places in flux-core](https://github.com/flux-framework/flux-core/search?utf8=%E2%9C%93&q=Jtostr&type=). This probably happened when flux switched from json-c to jansson.

For the `job->resrc_tree` leak, I am thinking it might make more sense to destroy the resrc_tree of all queued jobs [within schedule_jobs](https://github.com/nikhil-jain/flux-sched/blob/2687e2d945e0fc149d9ac7f4b2a3b08bffc2a1ed/sched/sched.c#L1576), right after all of the reservations are cleared from the main resrc_tree.  If you want to make that change now, feel free, otherwise, I'll open a ticket in flux-sched.

SteVwonder:
Looks like the travis-ci build failed because the `jcbstr`s are marked as `const` and then `free`d.

```
../../../simulator/simulator.c: In function ‘send_reply_request’:
../../../simulator/simulator.c:407:5: error: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [-Werror]
     free (jcbstr);
     ^
In file included from ../../../simulator/simulator.c:29:0:
/usr/include/stdlib.h:483:13: note: expected ‘void *’ but argument is of type ‘const char *’
 extern void free (void *__ptr) __THROW;
             ^
../../../simulator/simulator.c:411:5: error: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [-Werror]
     free (jcbstr);
     ^
In file included from ../../../simulator/simulator.c:29:0:
/usr/include/stdlib.h:483:13: note: expected ‘void *’ but argument is of type ‘const char *’
 extern void free (void *__ptr) __THROW;
```

simulator/simulator.c
```diff
@@ -120,6 +121,8 @@ void free_job (job_t *job)
     free (job->user);
     free (job->jobname);
     free (job->account);
+    if(job->kvs_dir)
+      kvsdir_destroy (job->kvs_dir);
```
SteVwonder: I believe this should be `flux_kvsdir_destroy`

```
../../../simulator/simulator.c: In function ‘free_job’:
../../../simulator/simulator.c:125:7: error: implicit declaration of function ‘kvsdir_destroy’ [-Werror=implicit-function-declaration]
       kvsdir_destroy (job->kvs_dir);
       ^
```

nikhil-jain: Thanks Stephen. I will fix it.

nikhil-jain: Yes, I noticed the travis failures, will update the patch soon.

dongahn:
These have been fixed with our latest valgrind efforts?

sched/sched.c
```diff
@@ -1526,7 +1539,10 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
                 if (rc) {
                     resrc_tree_destroy (ctx->rsapi, selected_tree, false, false);
                     job->resrc_tree = NULL;
-                } else
+                } else {
+                    if(job->resrc_tree != NULL) {
```
SteVwonder: @dongahn: I believe we still have this leak in the scheduler.

dongahn: OK. I just posted a separate PR with this one. I am closing this PR.

sched/sched.c
```diff
@@ -1504,6 +1514,9 @@ int schedule_job (ssrvctx_t *ctx, flux_lwj_t *job, int64_t starttime)
                 // TODO: handle this some other way (JSC?)
                 job->starttime = starttime;
                 job->state = J_SELECTED;
+                if(job->resrc_tree) {
+                    resrc_tree_destroy (ctx->rsapi, job->resrc_tree, false, false);
```
SteVwonder: @dongahn: this is the other leak still in master

PR #291, opened 2018-03-20
sched: Add pending job cancellation support

dongahn:
This adds support for cancelling pending jobs in response to the `flux wreck cancel` command. (Please see [flux-core Issue #1359](https://github.com/flux-framework/flux-core/issues/1359)).

Note that this PR requires [flux-core PR #1374](https://github.com/flux-framework/flux-core/pull/1374) to be merged into flux-core first. 

Two more items:

- add `errno` support when resource loading fails. [flux-core Issue #1361](https://github.com/flux-framework/flux-core/issues/1361).

- add some log print statements which can be used to help identify the performance/scalability bottleneck of resource loading. [flux-core Issue #1363](https://github.com/flux-framework/flux-core/issues/1363).

Maybe, @SteVwonder and/or @grondo can review this?

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -68,6 +68,8 @@ static void ev_check_cb (flux_reactor_t *r, flux_watcher_t *w,
                          int revents, void *arg);
 static void res_event_cb (flux_t *h, flux_msg_handler_t *w,
                           const flux_msg_t *msg, void *arg);
+static void cancel_request_cb (flux_t *h, flux_msg_handler_t *w,
+                               const flux_msg_t *msg, void *arg);
 static int job_status_cb (const char *jcbstr, void *arg, int errnum);
 
 
@@ -102,6 +104,7 @@ typedef struct {
     char         *userplugin;
     char         *userplugin_opts;
     char         *prio_plugin;
+    bool          reap;               /* Enable job reap support */
     bool          sim;
     bool          schedonce;          /* Use resources only once */
     bool          fail_on_error;      /* Fail immediately on error */
@@ -193,6 +196,7 @@ static inline void ssrvarg_init (ssrvarg_t *arg)
     arg->userplugin = NULL;
     arg->userplugin_opts = NULL;
     arg->prio_plugin = NULL;
+    arg->reap = false;
     arg->sim = false;
     arg->schedonce = false;
     arg->fail_on_error = false;
@@ -217,6 +221,7 @@ static inline void ssrvarg_free (ssrvarg_t *arg)
 static inline int ssrvarg_process_args (int argc, char **argv, ssrvarg_t *a)
 {
     int i = 0, rc = 0;
+    char *reap = NULL;
     char *schedonce = NULL;
     char *immediate = NULL;
     char *vlevel= NULL;
@@ -225,6 +230,8 @@ static inline int ssrvarg_process_args (int argc, char **argv, ssrvarg_t *a)
     for (i = 0; i < argc; i++) {
         if (!strncmp ("rdl-conf=", argv[i], sizeof ("rdl-conf"))) {
             a->path = xstrdup (strstr (argv[i], "=") + 1);
+        } else if (!strncmp ("reap=", argv[i], sizeof ("reap"))) {
+            a->reap = xstrdup (strstr (argv[i], "=") + 1);
         } else if (!strncmp ("sched-once=", argv[i], sizeof ("sched-once"))) {
             schedonce = xstrdup (strstr (argv[i], "=") + 1);
         } else if (!strncmp ("fail-on-error=", argv[i],
@@ -255,6 +262,10 @@ static inline int ssrvarg_process_args (int argc, char **argv, ssrvarg_t *a)
     if (!(a->userplugin))
         a->userplugin = xstrdup ("sched.fcfs");
 
+    if (reap && !strncmp (reap, "true", sizeof ("true"))) {
+        a->reap = true;
+        free (reap);
+    }
     if (sim && !strncmp (sim, "true", sizeof ("true"))) {
         a->sim = true;
         free (sim);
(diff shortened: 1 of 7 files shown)


grondo:
@dongahn, looks like this PR needs to be rebased on flux-sched/master. I'll try to review the PR soon (and give it a spin on my desktop or a production machine)

dongahn:
@grondo: Thanks. Odd, though. I actually freshly created a from from the flux-sched/master but still,
the latest pull commit ([a4dae4](https://github.com/flux-framework/flux-sched/commit/a4dae4277ed4a8ee64f177075cf5ff375e4bf385)) is missing from this PR. 

Rebased and forced a push.

dongahn:
Hmmm I think the new reap support in sched is causing a slight coverage difference. But there isn't an easy way to test this... (until reap is implemented in flux-core).

I can remove the code to increase the coverage but I am not convinved that it is a good reason to remove code.

sched/sched.c
```diff
@@ -1743,6 +1794,39 @@ static void res_event_cb (flux_t *h, flux_msg_handler_t *w,
...
+    } else if (!(job = q_find_job (ctx, jobid))) {
+        errno = ENOENT;
+        flux_log (h, LOG_ERR, "couldn't find job (%"PRId64").", jobid);
+        goto error;
+    } else if (job->state != J_SCHEDREQ) {
+        errno = EINVAL;
+        flux_log (h, LOG_ERR, "job state is %d.", job->state);
+        goto error;
+    }
+    zlist_remove (ctx->p_queue, job);
```
SteVwonder: I recommend replacing this with a call to `q_rm_from_pqueue`, which will also set the `pq_state` to `true` automatically.

SteVwonder:
My other recommendation is to take the chunk of code below that is copied and pasted several times in the scheduler state machine (whenever a job is reaped, or if reaping is disabled, when the job is cancelled) and replace it with a `flux_lwj_t` deconstructor, which can later be expanded to include other fields in the `flux_lwj_t` struct like user, account, and resrc_tree.
```
            if (job->req)
                free (job->req);
            key = xasprintf ("%"PRId64"", job->lwj_id);
            zhash_delete (ctx->job_index, key);
            free (key);
            free (job);
```

sched/sched.c
```diff
@@ -1695,31 +1724,53 @@ static int action (ssrvctx_t *ctx, flux_lwj_t *job, job_state_t newstate)
                           job->lwj_id);
             }
         }
+        if (ctx->arg.reap) {
+            /* move the job to complete queue when reap should be supported */
+            q_move_to_cqueue (ctx, job);
+        } else {
+            /* free resource here if reap is not enabled */
+            q_rm_from_pqueue (ctx, job);
```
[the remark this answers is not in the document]
dongahn: Great catch! Will fix it.

sched/sched.c
dongahn: Yes another good catch! Removing a job from a pending queue should be a scheduleable event. Will fix.

dongahn:
Yes, thought was to do it this way and refactor later. But I will have to fix other issues, I'd be happy to look into this as well.

Thanks @StevWonder.

sched/sched.c
```diff
@@ -1743,6 +1807,39 @@ static void res_event_cb (flux_t *h, flux_msg_handler_t *w,
...
+    flux_log (h, LOG_INFO, "cancel request callback invoked.");
+    if (flux_request_unpack (msg, NULL, "{ s:I }", "jobid", &jobid) < 0) {
+        goto error;
+    } else if (!(job = q_find_job (ctx, jobid))) {
+        errno = ENOENT;
+        flux_log (h, LOG_ERR, "couldn't find job (%"PRId64").", jobid);
+        goto error;
+    } else if (job->state != J_SCHEDREQ) {
+        errno = EINVAL;
+        flux_log (h, LOG_ERR, "job state is %d.", job->state);
```
grondo: I'm not sure these calls to `flux_log (LOG_ERR)` belong here. The real error is being returned to the caller via the appropriate errno, and it is not a "system error" so probably shouldn't go the system error log (if that makes sense). If you want these events to be logged, I'd suggest maybe LOG_DEBUG level messages like `"Attempt to cancel [nonexistent|state=STATE] job"`, perhaps including the uid that made the request.

sched/sched.c
```diff
@@ -1743,6 +1807,39 @@ static void res_event_cb (flux_t *h, flux_msg_handler_t *w,
...
+    }
+    zlist_remove (ctx->p_queue, job);
+    if ((update_state (h, jobid, job->state, J_CANCELLED)) != 0) {
+        flux_log (h, LOG_ERR, "couldn't update the job state.");
+        goto error;
+    }
+    flux_log (ctx->h, LOG_INFO, "pending job (%"PRId64") removed.", jobid);
+    if (flux_respond_pack (h, msg, "{ }") < 0)
+        flux_log_error (h, "%s", __FUNCTION__);
+    return;
```
grondo: Since you are already responding with an empty message here, it might be useful to include the jobid in the response, which might make handling asynchronous responses easier on the `sched.cancel` caller, e.g.
```C
  if (flux_respond_pack (h, msg, "{s:I}", "jobid", jobid) < 0)
```

t/t1006-cancel.t
```diff
@@ -0,0 +1,53 @@
...
+test_expect_success 'cancel: cancelling completed jobs must fail' '
+    test_must_fail flux wreck cancel 1 &&
+    state=$(flux kvs get -j $(job_kvs_path 1).state) &&
+    test ${state} = "complete" &&
+    state=$(flux kvs get -j $(job_kvs_path 2).state) &&
+    test_must_fail flux wreck cancel 2 &&
+    test ${state} = "complete" &&
+    flux module remove sched
+'
+
```
grondo: Should a test for cancel of non-existent jobid be added here as well  to test the `ENOENT` case above?

dongahn:
@grondo: Sound good across the board!. I will address those.

sched/sched.c
dongahn: How do you get the uid of the request?

grondo: `int flux_msg_get_userid (const flux_msg_t *msg, uint32_t *userid);`

dongahn:
OK. Forced a push, I think this is ready to go in. @SteVwonder, I decided not to do `flux_lwj_t` refactoring prematurely at this round. I've created a new issue (#293) so we can revisit it later.

SteVwonder:
:+1: LGTM.  Thanks Dong!

PR #295, opened 2018-03-27
Add support for the new wreck.state.submitted event

dongahn:
This requires the pending [flux-core PR #1389](https://github.com/flux-framework/flux-core/pull/1389).

Add support within the scheduler and emulator for the new `wreck.state.submitted` event with which job request info such as nnodes and walltime is piggybacked.

Use the event's payload and other changes within jsc to optimize job submission handling performance -- i.e., reducing KVS accesses and the number of state transitions.

This should improve the job throughput when the submission rate is super high although I didn't measure the improvements.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -404,33 +404,26 @@ static inline void get_states (json_t *jcb, int64_t *os, int64_t *ns)
     Jget_int64 (o, JSC_STATE_PAIR_NSTATE, ns);
 }
 
-static inline int fill_resource_req (flux_t *h, flux_lwj_t *j)
+static inline int fill_resource_req (flux_t *h, flux_lwj_t *j, json_t *jcb)
 {
-    char *jcbstr = NULL;
     int rc = -1;
     int64_t nn = 0;
     int64_t nc = 0;
     int64_t walltime = 0;
-    json_t *jcb = NULL;
     json_t *o = NULL;
 
-    if (!j) goto done;
-
     j->req = (flux_res_t *) xzmalloc (sizeof (flux_res_t));
-    if ((rc = jsc_query_jcb (h, j->lwj_id, JSC_RDESC, &jcbstr)) != 0) {
-        flux_log (h, LOG_ERR, "error in jsc_query_jcb.");
-        goto done;
-    }
-    if (!(jcb = Jfromstr (jcbstr))) {
-        flux_log (h, LOG_ERR, "fill_resource_req: error parsing JSON string");
-        goto done;
-    }
     /* TODO:  add user name and charge account to info the JSC provides */
     j->user = NULL;
     j->account = NULL;
-    if (!Jget_obj (jcb, JSC_RDESC, &o)) goto done;
-    if (!Jget_int64 (o, JSC_RDESC_NNODES, &nn)) goto done;
-    if (!Jget_int64 (o, JSC_RDESC_NTASKS, &nc)) goto done;
+
+    if (!Jget_obj (jcb, JSC_RDESC, &o))
+        goto done;
+    if (!Jget_int64 (o, JSC_RDESC_NNODES, &nn))
+        goto done;
+    if (!Jget_int64 (o, JSC_RDESC_NTASKS, &nc))
+        goto done;
+
     j->req->nnodes = (uint64_t) nn;
     j->req->ncores = (uint64_t) nc;
     if (!Jget_int64 (o, JSC_RDESC_WALLTIME, &walltime) || !walltime) {
@@ -441,8 +434,6 @@ static inline int fill_resource_req (flux_t *h, flux_lwj_t *j)
     j->req->node_exclusive = false;
     rc = 0;
 done:
-    if (jcb)
-        Jput (jcb);
     return rc;
 }
 
@@ -466,7 +457,7 @@ static inline bool is_newjob (json_t *jcb)
 {
     int64_t os = J_NULL, ns = J_NULL;
     get_states (jcb, &os, &ns);
-    return ((os == J_NULL) && (ns == J_NULL))? true : false;
+    return ((os == J_NULL) && (ns == J_SUBMITTED))? true : false;
 }
 
 static int plugin_process_args (ssrvctx_t *ctx, char *userplugin_opts)
@@ -903,8 +894,14 @@ static int sim_job_status_cb (const char *jcbstr, void *arg, int errnum)
     ssrvctx_t *ctx = getctx ((flux_t *)arg);
     jsc_event_t *event = (jsc_event_t*) xzmalloc (sizeof (jsc_event_t));
 
+    if (errnum > 0) {
+        flux_log (ctx->h, LOG_ERR, "%s: errnum passed in", __FUNCTION__);
+        return -1;
+    }
+
     if (!(jcb = Jfromstr (jcbstr))) {
-        flux_log (ctx->h, LOG_ERR, "sim_job_status_cb: error parsing JSON string");
+        flux_log (ctx->h, LOG_ERR, "%s: error parsing JSON string",
+                  __FUNCTION__);
         return -1;
     }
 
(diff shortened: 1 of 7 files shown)


dongahn:
@SteVwonder: this has a change to your emulator code. Please review if you can. Thanks.

sched/sched.c
```diff
@@ -462,11 +453,11 @@ static int update_state (flux_t *h, uint64_t jid, job_state_t os, job_state_t ns
     return rc;
 }
 
-static inline bool is_newjob (json_t *jcb)
+static inline bool is_newjob (flux_t *h, json_t *jcb)
```
SteVwonder: Can the `flux_t *h` argument be removed?  It doesn't appear to be used.

simulator/submitsrv.c
```diff
@@ -239,7 +238,7 @@ int schedule_next_job (flux_t *h, sim_state_t *sim_state)
         return -1;
     }
 
-    f = flux_rpc_pack (h, "job.create", FLUX_NODEID_ANY, 0,
+    f = flux_rpc_pack (h, "job.submit", FLUX_NODEID_ANY, 0,
```
SteVwonder: I think this will cause a race in the KVS for simulator-specific job information.  Previously, the simulator submit module (in an admittedly hacky way) worked by creating the job, then populating the KVS directory with simulator specific information (via the `put_job_in_kvs` function), and then setting the job to `submitted`.  At that point, the scheduler processed the job, scheduled it, and the `sim_execsrv` module would simulate the execution.  In the process of simulating the job execution, `sim_execsrv` [would pull the simulator specific information out of the kvs](https://github.com/flux-framework/flux-sched/blob/e068c924f05e0475ebeb9bbecfa88c3ad167919e/simulator/sim_execsrv.c#L531).  So in the previous version, the simulator specific information was guaranteed to exist in the KVS before the `sim_execsrv` module tried to pull it, because the `submit` module wouldn't permit the scheduling of the job until the custom simulator information was placed into the KVS.

After a quick glance at https://github.com/flux-framework/flux-core/pull/1389, it seems the new wreck still supports `job.create`.  Is there a way we can keep that flow?  Maybe when the `submit` module sends the `wreck.state.submitted` event, it can include the new `ntasks`, `nnodes`, and `walltime` data in the json payload?

I'm not married to that control flow though.  I'm sure there is a cleaner way to solve this race-condition that I'm not seeing....

grondo: > After a quick glance at flux-framework/flux-core#1389, it seems the new wreck still supports job.create. Is there a way we can keep that flow? 

Sorry we didn't check with you first about these changes @SteVwonder!

Since the scheduler should now ignore the `reserved` state, you could still use `job.create` from the simulator to "reserve" a job entry in the kvs, fill in values specific to the scheduler, and once those have been committed to the KVS issue the `job.sumbit` event yourself as you suggested. I don't think the scheduler could tell the difference between the simulator and the `flux-submit` program. (This is exactly the purpose for which the "reserved" state was designed)

That is my $0.02

sched/sched.c
dongahn: It can. It was there to print some debug msgs.

simulator/submitsrv.c
SteVwonder: No need for apologies.  I saw the discussion about the state changes in the github issue, but I forgot this is how the simulator `submit` module works. So I'll take the blame for not catching this earlier.

> you could still use job.create from the simulator to "reserve" a job entry in the kvs, fill in values specific to the scheduler, and once those have been committed to the KVS issue the job.sumbit event yourself as you suggested.

That works for me.  @dongahn, are you ok with the simulator still using `job.create` and the `reserved` state for now?  For the splash target, I think we get this PR in sooner rather than later.  In the future, we could refactor to include the simulator specific information in the `job.submit` message (when in sim mode) to avoid the race condition.

dongahn: Well, it won't work, at least without changes to [flux-core#1398](https://github.com/flux-framework/flux-core/pull/1389).

Right now, when the `reserved` state is seen, jsc puts the new job into the active job hash. So when `submitted` state of this job is seen afterwards, jsc won't be able to insert this job into the active job hash, leading to an error.

The way the scheduler ignores the state change of the job starting with`reserved` state is that, it just doesn't recognize null->reserved as the new job event. So the scheduler won't put the job into its job index at all. Thus, any subsequent event of that job will be ignored.

dongahn: I will try to make a minor modification to the JSC code in [flux-core#1398](https://github.com/flux-framework/flux-core/pull/1389) to allow your original tool flow. Stay tuned.

grondo: Thanks @dongahn. IMO the "reserved" state should be ignored by the scheduler -- it was originally just a state that marked a kvs directory as reserved for a writer. In the wreck changes I considered not emitting an event for it, but I knew it might break the scheduler. (Of course, we can revisit this once we start work on job-ingest module there might not be a use for a reserved state)

dongahn: @grondo: effectively the scheduler is ignoring it. The reason JSC needs to keep track of this event is to allow the scheduler to ignore all of the state transitions that started from `reserved` state which I initially thought was just for the wreckrun use case. (If it makes sense). But now the emulator does the same, and I think I know how to accommodate it.

dongahn: @SteVwonder or @grondo:  `wreck.state.submitted` will have to change something like the following to emulate the new wreck behavior. What kvs_path should I use with the `wreck.state.submitted` event emitted from the submit emulator module?  

```
    if (!(msg = flux_event_pack ("wreck.state.submitted",
                                 "{ s:I s:s s:i s:i s:i}",
                                 "lwj", new_jobid,
                                 "kvs_path", path,
                                 "nnodes", job->nnodes,
                                 "ntasks", job->ncpus,
                                 "walltime", (int64_t)job->time_limit))
        || flux_send (h, msg, 0) < 0) {
        return -1;
    }
```

grondo: @dongahn, I will have to look at the simulator, but it should be the same path returned in the response to the `job.create` call(?)

dongahn:
@SteVwonder: I added this support for both in this PR and https://github.com/flux-framework/flux-core/pull/1389. But somehow the emulator-based tests all failed with "no event remaining message" after scheduling the first job. 

```
2018-03-27T23:44:11.925585Z submit.debug[0]: sent a reply request: {"sim_time": 19643.0, "event_timers": {"sched": -1.0, "sim_exec": -1.0, "submit": -1.0}, "mod_name": "submit"}
2018-03-27T23:44:11.925709Z sim.debug[0]: No events remaining
```

I'm not familiar enough w/ the emulator code that I feel I am not effectively debugging this. Could you please take a look at the broker log and see if you can spot anything? Log is here:
[t2000-fcfs.broker.log](https://github.com/flux-framework/flux-sched/files/1854328/t2000-fcfs.broker.log).

grondo:
@dongahn, for some reason GitHub also thinks this is out-of-date with master, you may need to rebase and push again

grondo:
I restarted the build, though I guess we know it will fail. This does need to be rebased on the current master apparently.

grondo:
@dongahn, I would merge this branch but it does still need to be rebased on master for some reason. Also quickly disable the failing tests so future PRs don't appear to fail known failing tests.

grondo:
@dongahn, I went ahead and rebased this branch on current master, and disabled all the simulator driven tests. Just a warning that I force-pushed over your branch, so if you have local work be sure to save it on a different branch or at least don't reset against your upstream branch.

All simulator tests are disabled for now, to re-enable temporarily set `FLUX_SCHED_ENABLE_SIM_TESTS` in the environment

dongahn:
@grondo: thank you! Sorry for the late response. Had a dentist appointment. 

So if Traverse turns green, this should go in?

grondo:
Yes. I've tested it against my pending flux-core branch and seems to work well. (Note: that PR adds a new `ncores` attribute to job submission, if you have a chance you might want to take a look)

dongahn:
@grondo's change looks great. @garlick, @trws or @SteVwonder, could you push the button? It seems not having this in the master is confusing many at this point.

garlick:
Agreed!

PR #296, opened 2018-03-28
[WIP] Splash

trws:
as with core, this is an indifferentiated mass of changes I'm using in splash right now.


dongahn:
Same here @trws. Anything else I should pick from this branch?

trws:
Not that I know of.   I'm going to keep the branch around just in case, but the PR can go for sure.  Thanks!

Issue #297, opened 2018-03-29
build:  configure should detect unsupported boost versions


garlick:
From closed PR #283:

> A note on boost versions: Probably this will only work with boost 1.53 through, maybe, 1.55. At 1.56 through at least 1.58 the BGL is broken an won't compile with C++11. Unfortunately 1.58 is the standard version supplied in Ubuntu 16.04.3. I tried the latest stable boost 1.66 and that didn't go well either. Our m4 libraries don't detect 1.66 correctly.
> 
> For now I've downgraded to boost 1.53 and that is letting me at compile and run.

Currently configure requires boost >= 1.53, but the version should be further constrained to fail if too new.

garlick:
FWIW, current master of 09ad395b9f13fd2414be413c2f60244848a6a2ea fails on Ubuntu 16.04.4 LTS with default boost version of 1.58.0.1ubuntu as follows
```
  CXX      utilities_resource_query-resource_gen.o
In file included from /usr/include/c++/5/deque:62:0,
                 from resource_gen.cpp:26:
/usr/include/c++/5/bits/stl_construct.h: In instantiation of ‘void std::_Construct(_T1*, _Args&& ...) [with _T1 = boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t>; _Args = {const boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t>&}]’:
/usr/include/c++/5/bits/stl_uninitialized.h:75:18:   required from ‘static _ForwardIterator std::__uninitialized_copy<_TrivialValueTypes>::__uninit_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = __gnu_cxx::__normal_iterator<const boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t>*, std::vector<boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t>, std::allocator<boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t> > > >; _ForwardIterator = boost::detail::stored_edge_property<long unsigned int, Flux::resource_model::resource_relation_t>*; bool _TrivialValueTypes = false]’
[...]

PR #298, opened 2018-03-29
solve portability issues for Ubuntu 16.04.4 LTS

garlick:
This PR contains the changes I needed in order to bulid flux-sched on an Ubuntu 16.04.4 LTS system using a side-installed version of boost-1.53.  The problems solved are:
* fix the build system so that `configure --with-boost=path` works
* deal with new warnings with g++ version 5.4.0 
* avoid namespace collisions between boost and system (possibly also a g++ version issue, since this was not seen earlier)

I tested against the last version of flux-core known to work with flux-sched (two PR's ago, flux-framework/flux-core@cdc9e9b3f41104066d6c0e2d383b6eabb3fe7c45) and it works.

The only thing that's still not quite right is I needed to set LD_LIBRARY_PATH on the "make check" command line in order for in-tree test executables to find the boost libraries.  I guess usually libtool magically takes care of this, but flux-sched doesn't use libtool.

Some of this was discussed in #297


dongahn:
Great cleanup across the board! Sorry about the collision issue due to using the `using` keyword. All changes look good to me.

garlick:
No prob!  I will see if I can add a commit to skip the failing simulator testing, as discussed.

dongahn:
Once this is rebased, this can go in.

garlick:
It's been rebased.  Hopefully travis likes it.

garlick:
Uh oh, stuck here
```
  CXXLD    utilities/grug2dot
utilities_grug2dot-resource_gen_spec.o: In function `read_graphml<boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, Flux::resource_model::resource_pool_gen_t, Flux::resource_model::relation_gen_t> >':
/usr/include/boost/graph/graphml.hpp:214: undefined reference to `boost::read_graphml(std::istream&, boost::mutate_graph&, unsigned long)'
```

dongahn:
> Uh oh, stuck here

Ah... I remember now. I think the reason I switched to `AC_CHECK_LIB` was `AX_BOOST_GRAPH` didn't add the boost graph library to the link line.  Maybe we need to extend `AX_BOOST_GRAPH` further...

garlick:
grug2dot successfully linked on a debian system with packaged boost 1.55, and on my desktop Ubuntu system with 1.53.  Not sure why travis with packaged boost 1.54 cannot be linked or why that function is missing/not matched by this call.  

I guess I'm assuming the link line is fine, and this has to do with the proposed namespace changes.

The prototype in `/usr/include/boost/graph/graphml.hpp` is
```c
namespace boost {
...
void BOOST_GRAPH_DECL
read_graphml(std::istream& in, mutate_graph& g, size_t desired_idx);
...
} // namespace boost
```
I can't see any _obvious_ namespace issues, though when I look at how boost::read_graphml() is called in our code, I find I need to review/learn more C++ to understand how the types manage to match, or what if anything the "using" change could do to impact that.

```c
// resource_gen_spec.hpp
class resource_gen_spec_t {
public:
    resource_gen_spec_t ();
    resource_gen_spec_t (const resource_gen_spec_t &o);
    const gg_t &gen_graph ();
    const gen_meth_t to_gen_method_t (const std::string &s) const;
    int read_graphml (const std::string &ifn);
    int write_graphviz (const std::string &ofn, bool simple=false);

private:
    void setup_dynamic_property (boost::dynamic_properties &dp, gg_t &g);
    gg_t g;
    boost::dynamic_properties dp;
};

// resource_gen_spec.cpp
/*! Load resource generator recipe graph from a graphml file
 *
 *  \param ifn           resource generator recipe file in graphml
 *  \return              0 on success; -1 otherwise
 */
int resource_gen_spec_t::read_graphml (const string &ifn)
{
    int rc = 0;
    ifstream in_file (ifn.c_str ());
    if (!in_file.good ())
        return -1;

    try {
        boost::read_graphml (in_file, g, dp);
    } catch (boost::graph_exception &e) {
        cerr << e.what () << endl;
        rc = -1;
    }

    in_file.close ();
    return rc;
}
```

garlick:
Never mind, I tried dropping the namespace commit on a test branch, and that one fails in travis in the same place, so your lead is probably the right one @dongahn.

garlick:
The problem was AX_BOOST_GRAPH was picking up `boost_graph_parallel` instead of `boost_graph`.

Fixed the glob in there, and also tweaked warning options so that older compilers don't choke on the newer `maybe-uninitialized` warning option.

Forced a push, fingers crossed.

dongahn:
Excellent! 

`boost_graph_parallel` might later provide us w/ further scheduling parallelization. But for now, we want to stick w/ serial version!

garlick:
Finally passed travis by finding the right combination of warning options across all compilers,
dropping `-Wno-maybe-uninitialized` and adding `-Wno-error` (for resource directory only).

I left this "band-aid" as a separate commit so the original set of warning suppressions is there, and the rationale for the band-aid is too.  Ideally we'll address the root cause of those "maybe-uninitialized" warnings and then be able to drop `-Wno-error` so new warnings don't creep in later.

dongahn:
> I left this "band-aid" as a separate commit so the original set of warning suppressions is there, and the rationale for the band-aid is too. Ideally we'll address the root cause of those "maybe-uninitialized" warnings and then be able to drop -Wno-error so new warnings don't creep in later.

Yes, we should look into this. Could you kindly open up an issue for this?

dongahn:
OK. LGTM! Thank you @garlick.

PR #301, opened 2018-03-31
solve portability issues for Ubuntu 17.10 

garlick:
Here are a few minor fixes for issues that came up building sched on my i386 Ubuntu 17.10 test VM, which has gcc-7.2.0.

No issues with the native boost 1.62 on this system.


dongahn:
Nice cleanup @garlick! I will merge it once Travis turns green. 

BTW, this reminds me, at some point, I need to change the `size_t` used for real physical quantities in `resrc` either `uint64_t` or `unsigned int`  since `size_t` shouldn't be used for this kind.

Issue #308, opened 2018-04-09
sched fails to load hwloc data from KVS on sierra


garlick:
(Originally opened as flux-framework/flux-core#1361 but solution turned out to be on the sched side)

On sierra system:

```
2018-03-15T20:43:44.086519Z sched.err[0]: can't load hwloc data: resrc_generate_hwloc_resources: Failed to create resrc from hwloc depth 1: Success
2018-03-15T20:43:44.086548Z sched.err[0]: failed to load resrc using hwloc
2018-03-15T20:43:44.086566Z sched.err[0]: failed to load resources
```

@trws followed up wtih
> I found it, and fixed it, the reader is hard-coded to the resources it will take, and sierra has a new one. WIll upload patch when time.

Then
> It's right [here](https://github.com/flux-framework/flux-sched/pull/296/files#diff-7cc8c5f0cadb0013dfa0589d1e78c50bR801) over in sched.  The "group" type was unhandled.  If possible it would be really good to switch this over to something more generic like what resource.c uses, but this gets it done for sierra.

garlick:
@dongahn says
> It seems this should go into resrc. As part of this, it will be good to take a representative hwloc xml file from sierra and add to our tests. I will see of I can do a quick PR before moving to another project.

dongahn:
@trws:
I gathered hwloc xml files from some sierra compute nodes to use them for testing. For example

```
sierra4371{dahn}23: hwloc-info
depth 0:	1 Machine (type #1)
 depth 1:	1 Group0 (type #7)
  depth 2:	6 NUMANode (type #2)
   depth 3:	2 Package (type #3)
    depth 4:	16 L3Cache (type #4)
     depth 5:	16 L2Cache (type #4)
      depth 6:	32 L1dCache (type #4)
       depth 7:	32 L1iCache (type #4)
        depth 8:	32 Core (type #5)
         depth 9:	128 PU (type #6)
Special depth -3:	24 Bridge (type #9)
Special depth -4:	17 PCI Device (type #10)
Special depth -5:	26 OS Device (type #11)
```

Somehow it shows 32 cores instead of 44 cores. Maybe I'm misinterpreting this... I used `lstopo 0.xml` to generate the xml version of this. Did I do something wrong?

trws:
You ran it on a launch or login node.  Those have 32 cores, it’s the 
compute nodes that have 44.

dongahn:
[...]

PR #305, opened 2018-04-03
Add node exclusion support

dongahn:
This is an experimental PR. Please don't merge yet.

As requested in https://github.com/flux-framework/flux-core/issues/1415#issuecomment-378117471, I added rudimentary support for excluding or including a node resource by the node's name.

To facilitate testing, this PR also adds support for node-exclusion support. 

This PR has not been fully validated for backfill scheduling. Also, only manually tested with no addition test cases yet.

An node-exclusion command is valid only when the target node has nothing allocated or reserved.

This requires another experimental PR I will post to flux-core that includes `flux wreck exclude` and `flux wreck include`

PR #310, opened 2018-04-10
Fix for reading hwloc xml files generated from Sierra compute nodes

dongahn:
Per discussion at https://github.com/flux-framework/flux-sched/issues/308:

Incorporate @trws' patch to handle group type.

Add a test case using hwloc xml files collected from Sierra nodes.


SteVwonder:
@dongahn, do you have a preference for what gets merged first, between this PR and #305?

dongahn:
No I don't. But I think this can go in first now that it looks like i will have to fix a CI failure on the other PR...

PR #318, opened 2018-04-19
test: Fix a race condition bug in a node exclusion test

dongahn:
wait-job now waits until the last job gets backfilled and running.
This way, "flux wreck exclude" is guaranteed to happen after
cab1235 resource gets two reservations.


grondo:
Thanks, @dongahn!

dongahn:
Thanks @grondo and sorry for the race bug.

PR #321, opened 2018-04-24
Add R_lite support

dongahn:
Please note that this PR will fail in Travis CI until PR https://github.com/flux-framework/flux-core/pull/1485 will be merged into flux-core.

- Add simple support APIs (e.g., `resrc_api_map_put`) to the `resrc_api` layer  that allows the user to serialize only certain resource types in two different forms (i.e., gather form and reduce form).

- Introduce `resrc_tree_serialize_lite` and `resrc_to_json_lite` to serialize an individual `resrc` object in these forms.

- Then, in `sched.c`, use the above APIs to only serialize `node` and `core` resource types into the agree-upon `R_lite` format.

- Finally, because `resrc` does not have `rank` information,  introduce`resolve_rank ()` to resolve the node/hostname into the corresponding rank as a post processing step. This was discussed with @grondo and @garlick offline some time back.

- Add some test cases.

References:
We discussed `R_lite` support in https://github.com/flux-framework/flux-core/pull/1485 and https://github.com/flux-framework/flux-core/issues/1378#issuecomment-376610545.

`R_lite` is already understood by wreck per https://github.com/flux-framework/flux-core/pull/1399

This should also resolve https://github.com/flux-framework/flux-core/issues/1439

It should also significantly reduce the memory growth issues with massive large numbers of jobs.


dongahn:
@grondo: once thing I couldn't test was whether the scheduler-generated `R_lite` has led to correct tasks affinity. If you know how to test this easily, please let me know.

dongahn:
Restarted the build as https://github.com/flux-framework/flux-core/pull/1485#event-1592896793 was just meged.

grondo:
I can check this out tomorrow, but can you explain in more detail what you
mean by R_lite resulting in correct affinity?

As of now, wreck is not using R_lite for any affinity. The plan iirc was to
add it only if necessary.

dongahn:
> As of now, wreck is not using R_lite for any affinity. The plan iirc was to
add it only if necessary

Oh sorry I didn't know that. I thought I saw cpuset etc from the new code and mistakenly thought it does do affinity.  I certaintly didn't take a deep dive. Just so that I understand, `R_lite` has info on which particular core or set of cores are allocated to the job. But wreck *currently* only retrieves the count from it? So essecially same as the old `rank.core` scheme? (Still good to have some exercise in preparation for the new execution engine).

grondo:
On Tue, Apr 24, 2018, 6:58 PM Dong H. Ahn <notifications@github.com> wrote:

> As of now, wreck is not using R_lite for any affinity. The plan iirc was to
> add it only if necessary
>
> Oh sorry I didn't know that. I thought I saw cpuset etc from the new code
> and mistakenly thought it does do affinity. I certaintly didn't take a deep
> dive. Just so that I understand, R_lite has info on which particular core
> or set of cores are allocated to the job. But wreck *currently* only
> retrieves the count from it? So essecially same as the old rank.core
> scheme? (Still good to have some exercise in preparation for the new
> execution engine)
>

Yes, that's right. Though it would be trivial for wrexecd to at least set
CPU affinity for all tasks on a node/rank to the set supplied by the
scheduler in rank:N core of R_lite (so at least it is possible, unlike only
a core count)

>

sched/sched.c
```diff
@@ -1344,22 +1371,32 @@ static int req_tpexec_allocate (ssrvctx_t *ctx, flux_lwj_t *job)
...
-
-    if (resrc_tree_serialize (ro, job->resrc_tree)) {
+    json_t *gat = Jnew_ar ();
+    json_t *red = Jnew ();
+    resrc_api_map_t *gmap = resrc_api_map_new ();
+    resrc_api_map_t *rmap = resrc_api_map_new ();
+
+    resrc_api_map_put (gmap, "node", (void *)(intptr_t)REDUCE_UNDER_ME);
+    resrc_api_map_put (rmap, "core", (void *)(intptr_t)NONE_UNDER_ME);
+    if (resrc_tree_serialize_lite (gat, red, job->resrc_tree, gmap, rmap)) {
```
SteVwonder: I want to confirm that I'm reading this correctly: the `red` JSON object is not used directly by the scheduler, correct?

dongahn: Yes, `red` (reduction) is only used by the serializer function, an unfortunate side effect of a recursive API. I could hide it by introducing a wrapper around resrc_tree_serialize_lite, if you want.

dongahn:
Thanks @SteVwonder. I don't think I saw thr multinode hang... Can you consistently reproduve it?

dongahn:
Well on a second thought, if he user only want one resource type to be serialized in the reduced form, red should be used instead. So I think it is better to keep it this way...

dongahn:
@SteVwonder: I think this can go in unless @grondo disagrees. Once this gets merged, I will rebase my other PR: #323. Thanks.

Issue #322, opened 2018-04-26
[resource-query] Incorporate @morrone's suggestions


dongahn:
There are a few nice suggestions made as part of https://github.com/flux-framework/flux-sched/pull/283, which I've never gotten around to address.

Create this ticket to track them:

1. scoring API class: Since this class takes several pages of screen space, only declare the functions in the class definition, and use external method definitions.

2. Use C++ type aliases instead of typedef

3. Move some of the classes that don't have to be a header file only solution to their own source files for better readability (e.g., class color_t).

Issue #319, opened 2018-04-21
Planner: Need an algorithmic fix in mintime resource tree for multiple resource types


dongahn:
Planner maintains two *essentially* binary search-tree to support `O(log N)` retrievals of matching scheduled points. But currently there is a problem when we use a planner object to track the state changes of *multiple* resource types over time.

[The scheduled-point tree](https://github.com/flux-framework/flux-sched/blob/master/resource/planner/planner.c#L89) is a self-balancing binary search tree implemented as Linux kernel's red black tree. This allows for `O(log N)` retrievals of scheduled points given a time requirement as the input. This works great for multiple resource types.

[The min-time-resource tree](https://github.com/flux-framework/flux-sched/blob/master/resource/planner/planner.c#L90tree) is an augmented red-black tree. It allows for `O(log N)` retrievals of minimum-time scheduled points given a resource quantity requirement as the input. That is, given a request of `k` units of a resource type, this answers what is the minimum time when this request can be satisfied in `O(log N)` time. The min-time resource tree works well when we use the planner object to track one resource type, but doesn't work well when we it to track multiple resource type. 

I have a reasonable algorithmic change to extend the min-time-resource tree to work for multiple resource types, and I create this ticket to track that work. 

Essentially, I want to maintain a normal binary search tree (RB) tree on the most significant resource type. And when a multi-dimensional resource quantity requirement query comes, I will retrieve those scheduled points that satisfied each resource type from the most to least significant resource type.
And when I reach the least significant resource type, I will build the minimum resource tree out of only those scheduled points that satisfies all of the more significant resource types and retrieve the minimum time that way. This way, as an optimization, I can still use the current implementation when the number of resource types to track is 1. 

I will need to make progress on the ECP year-end milestone (performance/scalability improvement for our new scheduling infrastructure), and this is one of the items that fits that portfolio.

dongahn:
The more I think about this, the less I am sure if it is a good idea to extend the min-time resource tree to support multiple resource types. I can think of a few different ways to implement this like using kd tree or making each scheduled point node in the resource binary search tree to maintain another sub resource BST for the next significant resource type amd so on. But this can easily lead to much higher software conplexity for one layer.

Maybe an alternative approach would be to limit the planner layer to concern only one resource type. And when the upper layer (resource-query etc) needs to keep tract of multiple resource types, it can instantiate multiple planner objects.

dongahn:
PR #349 fixes this issue.

PR #325, opened 2018-04-27
resource: Refactor policy code + other cleanup

dongahn:
This PR further refactors the `resource` code beyond Issue #322 with a goal to position the code for future work (e.g., https://github.com/flux-framework/flux-sched/issues/319). Please don't merge this yet. I will keep refactoring and cleaning up other parts of the resource layer.


dongahn:
@tpatki and @SteVwonder: This should be ready for your review. Again, these are mostly refactoring for future work and later extendibility though there are some minor code changes.

I expect your advanced scheduling policies to land `resource/policies` folder, so please let me know what you think. Also, for other developers who may want to extend this infrastructure w/ other types of traverses and etc, I've added a doxygen config file, which generally generates pretty neat pages for C++ code.

resource/matcher.hpp
```diff
@@ -110,11 +109,33 @@ class matcher_util_api_t
...
+    bool is_my_pruning_type (const std::string &subsystem,
+                             const std::string &anchor_type,
+                             const std::string &prune_type);
+    bool is_pruning_type (const std::string &subsystem,
+                          const std::string &prune_type);
+
+private:
+    // resource types that will be used for scheduler driven aggregate updates
+    // Examples:
+    // m_pruning_type["containment"]["rack"] -> {"node"}
```
SteVwonder: Just to clarify, this automatic aggregation at a given level has not been implemented yet, correct?  Only the "any level" aggregation, as in the example below, has been implemented?

If so, should we make calling `set_pruning_type` with `anchor_type != ANY_RESOURCE_TYPE` log a warning?

SteVwonder: @dongahn: sorry about this.  I commented directly on an individual commit's code, rather than commenting on the overall view.  So GitHub thinks the comment is out-of-date because there commits that occur after, but it does appear that this code exists in the final set of changes.  I'll keep this UI quirk in mind for the future.

dongahn:
@StevWonder, thanks! IO aware policy implementation isn't a priority. Primarily I wanted to have a review on this PR from someone who will add policies.

resource/matcher.hpp
dongahn: No problem.

> Just to clarify, this automatic aggregation at a given level has not been implemented yet, correct? Only the "any level" aggregation, as in the example below, has been implemented?

This is my next step. So a log warning shouldn't be necessary.

dongahn:
Ok. Rebased. This can go in as is as @SteVwonder already approved it. @tpatki and I can revisit the changes later if needed.

dongahn:
Thanks.

PR #330, opened 2018-05-01
Condense the core id list in R_lite

dongahn:
This is somewhat experimental. I at least wanted to do this to see the valgrind results on Travis. So don't merge yet.

Pull in nodeset class and friends from `flux-core` and add those files to flux-ached's libutil.

Incorporate nodeset to`R_lite` generator in `resrc_tree.c` to condense the list.

### DIFF
--- a/resrc/resrc_tree.c
+++ b/resrc/resrc_tree.c
@@ -33,6 +33,7 @@
 #include <czmq.h>
 
 #include "src/common/libutil/shortjansson.h"
+#include "src/common/libutil/nodeset.h"
 #include "rdl.h"
 #include "resrc_tree.h"
 #include "resrc_api_internal.h"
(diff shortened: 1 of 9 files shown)


dongahn:
That be awesome for various purposes! Could the name also change to something more generic, if that's the case?

garlick:
Sure, I guess - were you thinking of packaging in a separate library similar to libflux-optparse.so?

Suggestions on a new name?  It's implementation is based on the [vEB trie](https://en.wikipedia.org/wiki/Van_Emde_Boas_tree).

Maybe it would be worth making a cleanup pass through the API and creating a man(3) page also.

garlick:
> Suggestions on a new name?

rankset?

grondo:
I'm fine with `flux_nodeset_t`, but if the name must be changed, I would suggest `idset` or `intset` since it can represent any set of integers or identifiers. (e.g. here it is being used for core ids.

dongahn:
I like `idset`

dongahn:
Will rework this against the new `flux_dset` API.

PR #331, opened 2018-05-01
Add valgrind wrap specification for dlclose in sched/plugin.c

grondo:
@dongahn, can you try these two commits and see if they have any effect on the results you are seeing with `t5000-valgrind.t`?

IIUC, this should allow flux-sched valgrind test to work with even a flux-core that was not configured with `valgrind/valgrind.h`. Once `sched.so` is dlopened by `flux-broker`, the valgrind spec added to `sched/plugin.c` should become active and reroute future calls to `dlclose()` to the noop wrapped version.

You will have to apply these two commits and rerun `./autogen` and `./configure`. Then ensure that valgrind/valgrind.h was found with 
```
$ grep VALGRIND config.h
#define HAVE_VALGRIND_VALGRIND_H 1
```
Then retry the valgrind tests and ensure there are no false positives/missing symbols.

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,7 @@ AC_CHECK_LIB([dl], [dlerror],
 PKG_CHECK_MODULES([HWLOC], [hwloc >= 1.11.1], [], [])
 PKG_CHECK_MODULES([JANSSON], [jansson >= 2.6], [], [])
 PKG_CHECK_MODULES([XML2], [libxml-2.0])
+AX_VALGRIND_H
 AX_PROG_LUA([5.1],[5.3])
 AX_LUA_HEADERS
 AX_LUA_LIBS
@@ -63,7 +64,17 @@ AC_CHECK_LIB([readline], [readline],
              [AC_MSG_ERROR([Please install GNU readline])])
 
 # Checks for header files.
-AC_CHECK_HEADERS([inttypes.h limits.h stdint.h stdlib.h string.h sys/time.h unistd.h readline/readline.h readline/history.h])
+AC_CHECK_HEADERS([\
+  inttypes.h \
+  limits.h \
+  stdint.h \
+  stdlib.h \
+  string.h \
+  sys/time.h \
+  unistd.h \
+  readline/readline.h \
+  readline/history.h \
+])
 
 # Checks for typedefs, structures, and compiler characteristics.
 AC_HEADER_STDBOOL # in newer ac version:  AC_CHECK_HEADER_STDBOOL
--- a/m4/ax_valgrind.m4
+++ b/m4/ax_valgrind.m4
@@ -1,0 +1,59 @@
+# ===========================================================================
+#      https://github.com/flux-framework/flux-sched/
+# ===========================================================================
+# SYNOPSIS
+#
+#   AX_VALGRIND_H
+#
+# DESCRIPTION
+#
+#   Find valgrind.h first trying pkg-config, fallback to valgrind.h
+#    or valgrind/valgrind.h
+#
+#  This macro will set
+#  HAVE_VALGRIND if valgrind.h support was found
+#  HAVE_VALGRIND_H if #include <valgrind.h> works
+#  HAVE_VALGRIND_VALGRIND_H if #include <valgrind/valgrind.h> works
+#
+# LICENSE
+#
+#   Copyright (c) 2016 Lawrence Livermore National Security, LLC.  Produced at
+#   the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+#   LLNL-CODE-658032 All rights reserved.
+#
+#   This file is part of the Flux resource manager framework.
+#   For details, see https://github.com/flux-framework.
+#
+#   This program is free software; you can redistribute it and/or modify it
+#   under the terms of the GNU General Public License as published by the Free
+#   Software Foundation; either version 2 of the license, or (at your option)
+#   any later version.
+#
+#   Flux is distributed in the hope that it will be useful, but WITHOUT
+#   ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+#   FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License along
+#   with this program; if not, write to the Free Software Foundation, Inc.,
+#   59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+#   See also:  http://www.gnu.org/licenses/
+#
+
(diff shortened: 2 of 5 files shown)


dongahn:
Thanks @grondo, i will take a look.

dongahn:
On quartz, module load doesn't seem to add the valgrind header directory. 


```
quartz188{dahn}41: ml valgrind/3.13.0
quartz188{dahn}49: ./configure --prefix=/g/g0/dahn/workspace/planner_correction/inst

<CUT>

checking valgrind/valgrind.h usability... no
checking valgrind/valgrind.h presence... no
```

So as a sanity check, I did:

```
quartz188{dahn}50: ./configure CFLAGS="-I/usr/tce/packages/valgrind/valgrind-3.13.0/include" --prefix=/g/g0/dahn/workspace/planner_correction/inst

<CUT>

checking valgrind/valgrind.h usability... yes
checking valgrind/valgrind.h presence... no
configure: WARNING: valgrind/valgrind.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: valgrind/valgrind.h: proceeding with the compiler's result
```

And valgrind test is okay now. 

```
PASS: t5000-valgrind.t 1 - found executable flux-broker
PASS: t5000-valgrind.t 2 - valgrind reports no new errors on single broker run
```

But w/ the above method, some regression tests failed on quartz:
```
FAIL: t1007-exclude.t 1 - excluding a node with no job allocated or reserved works
FAIL: t1007-exclude.t 2 - including an excluded node works
FAIL: t1007-exclude.t 3 - excluding a node does not kill the jobs
FAIL: t1007-exclude.t 4 - -k option kills the jobs using the node
FAIL: t1007-exclude.t 5 - excluding a node with reservations works
```

This could be a side-effect of `CFLAGS="-I/usr/tce/packages/valgrind/valgrind-3.13.0/include"`, though.


Maybe we should add `--with-valgrind-header=<dir>` option? I'm wondering if there is a valgrind m4 script out there that has this option + other options...

grondo:
Does `module load` appropriately set `PKG_CONFIG_PATH`? I notice there is a pkg-config file for valgrind, so maybe we could just add a PKG_CHECK_MODULES stanza for valgrind.

dongahn:
It doesn't look like it is `ml` is setting it:

```
quartz188{dahn}68: echo $PKG_CONFIG_PATH
/g/g0/dahn/workspace/planner_correction/inst/lib/pkgconfig
```

I will manually set it to see if my build will like it.

grondo:
Well we don't check for `PKG_CHECK_MODULES (valgrind, ...)` (yet), but if PKC_CONFIG_PATH was set properly then you could at least do `CPPFLAGS=$(pkg-config --cflags valgrind)`

I wonder how we open issue to get these packages fixed?

dongahn:
@lee218llnl: we see an issue with valgrind TCE package on quartz. How do we open an issue for TCE packages?

dongahn:
@grondo:

```
quartz188{dahn}174: pkg-config --cflags valgrind
-I/usr/tce/packages/valgrind/valgrind-3.13.0/include/valgrind
```

So I made mods like this:

```diff
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,8 @@ AC_CHECK_LIB([dl], [dlerror],
 PKG_CHECK_MODULES([HWLOC], [hwloc >= 1.11.1], [], [])
 PKG_CHECK_MODULES([JANSSON], [jansson >= 2.6], [], [])
 PKG_CHECK_MODULES([XML2], [libxml-2.0])
+PKG_CHECK_MODULES([VALGRIND], [valgrind])
+CPPFLAGS=$(pkg-config --cflags valgrind)
 AX_PROG_LUA([5.1],[5.3])
 AX_LUA_HEADERS
 AX_LUA_LIBS
@@ -73,7 +75,7 @@ AC_CHECK_HEADERS([\
   unistd.h \
   readline/readline.h \
   readline/history.h \
-  valgrind/valgrind.h \
+  valgrind.h \
 ]) 
```


```diff
--- a/sched/plugin.c
+++ b/sched/plugin.c
@@ -29,8 +29,8 @@
 #include <flux/core.h>
 #include <czmq.h>

-#if HAVE_VALGRIND_VALGRIND_H
-#include <valgrind/valgrind.h>
+#if HAVE_VALGRIND_H
+#include <valgrind.h>

@@ -437,7 +437,7 @@ void sched_plugin_loader_destroy (struct sched_plugin_loader *sploader)
     }
 }

-#if HAVE_VALGRIND_VALGRIND_H
+#if HAVE_VALGRIND_H
 /* Disable dlclose() during valgrind operation
  */
 void I_WRAP_SONAME_FNNAME_ZZ(Za,dlclose)(void *dso) {}
```


```diff
--- a/t/t5000-valgrind.t
+++ b/t/t5000-valgrind.t
@@ -16,7 +16,7 @@ fi
 #  However, allow run to be forced on the cmdline with -d, --debug.
 #
 have_valgrind_h() {
-    grep -q "^#define HAVE_VALGRIND_VALGRIND_H" ${SHARNESS_BUILD_DIRECTORY}/config.h
+    grep -q "^#define HAVE_VALGRIND_H" ${SHARNESS_BUILD_DIRECTORY}/config.h
 }
```

With this, the new valgrind test succeeds on quartz:
```
PASS: t5000-valgrind.t 1 - found executable flux-broker
PASS: t5000-valgrind.t 2 - valgrind reports no new errors on single broker run
```

Do you want me to push this somewhere you can pick?

dongahn:
It's [here](https://github.com/dongahn/flux-sched/tree/valgrind-pkg-check).

dongahn:
FYI TCE valgrid package issue has been fixed on Quartz and will be propagated to the rest of TOSS3 system in a month.

grondo:
I also got the `valgrind-devel` package added to TOSS so at least on TOSS3 valgrind.h will be found by default without mucking with the environment.

grondo:
@dongahn, sorry this took me awhile to fix up.

I ran into a couple problems with pkg-config. First, by default `PKG_CHECK_MODULES` makes packages a build *requirement*, so if valgrind.pc was not found the build would fail. Second, on some platforms `valgrind.h` *or* `valgrind/valgrind.h` may be present *without* a valgrind.pc in `PKG_CONFIG_PATH`, so valgrind would not be properly detected.

I tried to come up with a solution that would work for most cases, and placed it into its own macro file as `AX_VALGRIND`. I'll copy this solution back to flux-core also if you can verify it here before merging.

dongahn:
@grondo, looks good to me!

In terms of the commit sequence, though maybe the m4 commit should come first?

grondo:
If you check out the branch that commit should be first. Sometimes it seems
like GitHub orders the commits by timestamp or something. I'll double-check
to be sure though

grondo:
Thanks!

PR #335, opened 2018-05-03
sched: emit wreck.state.cancelled event for cancelled jobs

### LINKED ISSUE #334: issue wreck.state.cancelled event after removal of submitted job
This will allow wreck/job module to remove cancelled jobs from active_jobs hash.

grondo:
Send a wreck.state.submitted event on successful job cancellation.

Fixes #334


dongahn:
Thanks @grondo.

Minor knit in the commit message: 

`wreck.state.submitted` --> `wreck.state.cancelled`

dongahn:
Other than that LGTM!

I was initially pondering if it would be better to emit this event from within `flux-wreck` when it gets the successful response from the sched's request handler. But then, I found other places in `sched` which emit `wreck` events (`wreck.<jobid>.kill`). So, this change is consistent with a slight performance advantage.

grondo:
> I was initially pondering if it would be better to emit this event from within flux-wreck when it gets the successful response from the sched's request handler.

I was wondering the same thing, but the current idiom is to generate the event at the site where the state is changed. Also if flux-wreck command was interrupted before it got the response, this would leave the system in an inconsistent state.

I fixed up the commit message and re-pushed.

dongahn:
Sounds all good. I will merge when Travis reports.

grondo:
Thanks!

PR #339, opened 2018-05-05
Drop rank.N support in favor of R_lite

dongahn:
This PR drops rank.N and introduces GPU support into `R_lite`. If GPUs are involved, `R_lite` will look like:

`[ { "node": "cab1234", "children": { "core": "0-15", "gpu": "0-1" }, "rank": 0 } ]`

@trws and @grondo: will this representation work? I think `wreck` is okay with the introduction of the `gpu` field.

Now, a bunch of scheduler tests will fail with these changes because of this [sharness script code](https://github.com/flux-framework/flux-sched/blob/master/t/sharness.d/sched-sharness.sh#L127).

I will have to adjust this script code to make use of `R_lite` instead of `rank.N.cores`. I can write a simple C wrapper that makes use of jansson and idset for this purpose, but I'm wondering if flux-core already has some utilities that I can use for this purpose. @grondo?

### DIFF
--- a/resrc/test/tresrc.c
+++ b/resrc/test/tresrc.c
@@ -361,6 +361,7 @@ static int test_a_resrc (resrc_api_ctx_t *rsapi, resrc_t *resrc, bool rdl)
     resrc_api_map_put (gather_map, "node", (void *)(intptr_t)REDUCE_UNDER_ME);
     resrc_api_map_t *reduce_map = resrc_api_map_new ();
     resrc_api_map_put (reduce_map, "core", (void *)(intptr_t)NONE_UNDER_ME);
+    resrc_api_map_put (reduce_map, "gpu", (void *)(intptr_t)NONE_UNDER_ME);
 
     init_time ();
     rc = resrc_tree_serialize_lite (gather, reduce, found_tree,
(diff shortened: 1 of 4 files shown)


grondo:
@dongahn, fyi, we do not have Lua bindings for nodeset yet, but there are bindings for `cpu_set_t` as part of flux-core you can use to parse and count these cpu sets, e.g.:
```lua
#!/usr/bin/env lua
local cpuset = require 'flux.cpuset'
local cset = cpuset.new (arg[1])
print ("Cpuset has "..#cset.." members")
```

```
$ lua t.lua 0-1,10
Cpuset has 3 members
$ lua t.lua 0-1   
Cpuset has 2 members
```

trws:
As long as it works for wreck I don't see any issue with it.

dongahn:
Thanks @grondo. I just saw two minor knits. I will force a push at which point this can go in. Also as part of this, I will drop some keys in JSC. So, I will post a PR on flux-core as well.

dongahn:
BTW, Lua is pretty easy to learn and simple! It seems having some hands-on for a simple C binding and module and OO package/module development at some point will be helpful for me.

dongahn:
Ok. Forced a push. If Travis turns green, this can go in.

dongahn:
Thanks.

Issue #350, opened 2018-06-19
Scheduler doesn't build against the current flux-core master


dongahn:
```
make[2]: Entering directory `/collab/usr/global/tools/flux/toss_3_x86_64_ib/build/toss_3_x86_64_ib/flux-sched/build-toss3/simulator'
  CC       libflux_sim_la-simulator.lo
  CC       simsrv_la-simsrv.lo
  CC       sim_execsrv_la-sim_execsrv.lo
  CC       submitsrv_la-submitsrv.lo
../../simulator/simsrv.c: In function 'join_cb':
../../simulator/simsrv.c:223:5: error: implicit declaration of function 'flux_msg_get_json' [-Werror=implicit-function-declaration]
     if (flux_msg_get_json (msg, &json_str) < 0 || json_str == NULL
     ^
../../simulator/sim_execsrv.c: In function 'trigger_cb':
../../simulator/sim_execsrv.c:601:5: error: implicit declaration of function 'flux_msg_get_json' [-Werror=implicit-function-declaration]
     if (flux_msg_get_json (msg, &json_str) < 0 || json_str == NULL
     ^
../../simulator/submitsrv.c: In function 'trigger_cb':
../../simulator/submitsrv.c:361:5: error: implicit declaration of function 'flux_msg_get_json' [-Werror=implicit-function-declaration]
     if (flux_msg_get_json (msg, &json_str) < 0 || json_str == NULL
     ^
cc1: all warnings being treated as errors
make[2]: *** [simsrv_la-simsrv.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
cc1: all warnings being treated as errors
cc1: all warnings being treated as errors
make[2]: *** [submitsrv_la-submitsrv.lo] Error 1
make[2]: *** [sim_execsrv_la-sim_execsrv.lo] Error 1
make[2]: Leaving directory `/collab/usr/global/tools/flux/toss_3_x86_64_ib/build/toss_3_x86_64_ib/flux-sched/build-toss3/simulator'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/collab/usr/global/tools/flux/toss_3_x86_64_ib/build/toss_3_x86_64_ib/flux-sched/build-toss3'
make: *** [all] Error 2
```

dongahn:
Looks like changing `flux_msg_get_json` to `flux_msg_get_string` fixes the issue.

PR #351, opened 2018-06-19
simulator: track a flux-core API change

dongahn:
Use flux_msg_get_string () instead of flux_msg_get_json ().

Resolve Issue #350.


dongahn:
Travis CI isn't happy with one of the configurations. Jus kicked it again.

PR #354, opened 2018-07-03
Add hwloc-based GPU scheduling support

dongahn:
This PR requires a flux-core PR, which I will post soon.

- Modify resrc to recognize and use CoProc devices  as gpu resource type. Include CUDA GPU as well as AMD GPU (at least in theory since I couldn't test this on AMD).

- Add a test using hwloc xml data generated from Sierra using a hwloc library configured with CUDA

- Minor cleanups

### DIFF
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -910,18 +910,10 @@ static resrc_t *resrc_new_from_hwloc_obj (resrc_api_ctx_t *ctx, hwloc_obj_t obj,
     } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_GROUP)) {
         type = xstrdup ("group");
         name = xasprintf ("%s%"PRId64"", type, id);
-#if HWLOC_API_VERSION < 0x00010b00
-    } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_NODE)) {
-#else
     } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_NUMANODE)) {
-#endif
         type = xstrdup ("numanode");
         name = xasprintf ("%s%"PRId64"", type, id);
-#if HWLOC_API_VERSION < 0x00010b00
-    } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_SOCKET)) {
-#else
     } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_PACKAGE)) {
-#endif
         type = xstrdup ("socket");
         name = xasprintf ("%s%"PRId64"", type, id);
 #if HWLOC_API_VERSION < 0x00020000
--- a/resrc/resrc.c
+++ b/resrc/resrc.c
@@ -942,6 +942,19 @@ static resrc_t *resrc_new_from_hwloc_obj (resrc_api_ctx_t *ctx, hwloc_obj_t obj,
     } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_PU)) {
         type = xstrdup ("pu");
         name = xasprintf ("%s%"PRId64"", type, id);
+    } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_OS_DEVICE)
+               && obj->attr
+               && obj->attr->osdev.type == HWLOC_OBJ_OSDEV_COPROC) {
+        /* hwloc doesn't provide the logical index only amongst CoProc devices
+           so we parse this info from the name until hwloc provide better
+           support.
+         */
+        if (strncmp(obj->name, "cuda", 4) == 0)
+            id = atoi (obj->name + 4);
+        else if (strncmp(obj->name, "opencl", 6) == 0)
+            id = atoi (obj->name + 6);
+        type = xstrdup ("gpu");
+        name = xasprintf ("%s%"PRId64"", type, id);
     } else {
         /* that's all we're supporting for now... */
         goto ret;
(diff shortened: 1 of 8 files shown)


dongahn:
FYI -- Aat least my local testing showed that a vanilla hwloc library on Sierra (without CUDA support) can still pick up GPUs from the hwloc xml files generated by an hwloc library configured with CUDA. 

This made writing a test easy and hopefully it will be the case with Travis CI as well.

trws:
That should work everywhere, it's only the detection code that generates the tree that cares if cuda is enabled.  It may also cause problems if you set the flag to tell hwloc that the loaded xml is the current machine, but that would be a lie anyway.

resrc/resrc.c
```diff
@@ -950,6 +942,19 @@ static resrc_t *resrc_new_from_hwloc_obj (resrc_api_ctx_t *ctx, hwloc_obj_t obj,
...
+        /* hwloc doesn't provide the logical index only amongst CoProc devices
+           so we parse this info from the name until hwloc provide better
+           support.
+         */
+        if (strncmp(obj->name, "cuda", 4) == 0)
+            id = atoi (obj->name + 4);
+        else if (strncmp(obj->name, "opencl", 6) == 0)
+            id = atoi (obj->name + 6);
+        type = xstrdup ("gpu");
+        name = xasprintf ("%s%"PRId64"", type, id);
```
trws: I realize the rest are doing this, but at least for the new one maybe just do xasprintf("gpu%"PRId64... rather than doing the extra strdup?

resrc/resrc.c
```diff
@@ -1016,20 +1021,34 @@ static resrc_t *resrc_create_cluster (resrc_api_ctx_t *ctx, char *cluster)
     return resrc;
 }
 
+static void resrc_walk_hwloc (resrc_api_ctx_t *ctx, hwloc_topology_t topo,
```
trws: Thank you, thank you, thank you for writing this code to replace the stuff below.  Thank you. 🙇

dongahn: :-)

resrc/resrc.c
SteVwonder: I believe type still needs to be set to "gpu" since it is accessed at [line 992](https://github.com/flux-framework/flux-sched/pull/354/files/7ffb38fade5d3a947ef59d2e06cccb8e87048f30#diff-7cc8c5f0cadb0013dfa0589d1e78c50bR992).

But as @trws mentions, the xstrdup looks unnecessary. You should just be able to set it to the string literal 'gpu'.

resrc/resrc.c
```diff
@@ -910,18 +910,10 @@ static resrc_t *resrc_new_from_hwloc_obj (resrc_api_ctx_t *ctx, hwloc_obj_t obj,
     } else if (!hwloc_compare_types (obj->type, HWLOC_OBJ_GROUP)) {
         type = xstrdup ("group");
         name = xasprintf ("%s%"PRId64"", type, id);
-#if HWLOC_API_VERSION < 0x00010b00
```
SteVwonder: We have documented as a requirement in the repo README.md that `libhwloc-dev >= 1.11.1`.  Does this removal change that requirement at all?

resrc/resrc.c
SteVwonder: WOW! The recursive code is so much cleaner.  Great work!

SteVwonder:
For the Travis failure, the only error I could find was:
```
2018-07-03T19:34:09.810550Z broker.err[0]: Run level 2 Exited with non-zero status (rc=1) 18.0s
```

So maybe the shutdown of the broker just hung for 18 seconds before being forcefully killed, causing the GPU test to "fail" due to a non-zero exit code?  Very odd.

resrc/resrc.c
dongahn: Should be an easy enough change so I will make a separate commit replacing all instance in this function.

resrc/resrc.c
dongahn: Not really. This just drops those codes that are compiled in when a hwloc version lower than the supported is used, which will be caught by config time.

dongahn:
```
flux-hwloc: /home/travis/build/flux-framework/flux-sched/flux-sched-0.5.0-13-g1c48b0e/t/data/hwloc-data/004N/exclusive/04-brokers-sierra-hwloc-with-cuda/0.xml: No such file or directory
not ok 8 - rs2rank: can schedule GPUs using hwloc
```

Somehow Travis doesn't find this file.

dongahn:
Odd, https://github.com/flux-framework/flux-sched/pull/354/commits/6a690701e399d37296c89c6ad68d519f8f0b39d8 commit introduces those test files.

dongahn:
Ah... Now that I think about this, I had issues with long paths. I will shorten the directory name for 04-`brokers-sierra-hwloc-with-cuda` to see if Travis CI (or maybe it was make dist) likes it better.

dongahn:
@SteVwonder and @trws:

> I believe type still needs to be set to "gpu" since it is accessed at line 992.

> But as @trws mentions, the xstrdup looks unnecessary. You should just be able to set it to the string literal 'gpu'.

It turns out this code need `xstrdup` to keep things simple. `type` can be set through two different paths, one for rdl reader and the other for hwloc reader and a string copy is needed for the general rdl case. If the code assigns a constant literal, the scheduler crashes when the `type` is freed in the destroy function.

dongahn:
@SteVwonder: The Travis problem was actually the long directory path... Once Travis turns completely green, this can go in.

SteVwonder:
Oh. Nice catch. Merging now.

dongahn:
Thanks @SteVwonder. Have a great July 4th!

PR #362, opened 2018-07-13
Issue337

### LINKED ISSUE #364: sched's Travis build fails after pmix update
Travis build for sched fails, probably because libevent-dev (required by pmix) is missing. 

```
============================================================================
== Libevent
============================================================================
checking for libevent in... (default search paths)
looking for header in /usr/include
checking event.h usability... no
checking event.h presence... no
checking for event.h... no
looking for header in /usr/include/include
checking event.h usability... no
checking event.h presence... no
checking for event.h... no
configure: WARNING: LIBEVENT SUPPORT NOT FOUND
configure: error: CANNOT CONTINUE
make: *** No targets specified and no makefile found.  Stop.
travis-dep-builder.sh: Failed to build and install pmix-3.0.0
```

tpatki:
Added request/reponse functionality to update scheduler parameters at runtime. Added a test case for the same. Needs the `flux wreck sched-params get/set` subcommand to be present in flux-core (separate PR). 
Also fixes #364.

### DIFF
--- a/sched/sched.c
+++ b/sched/sched.c
@@ -74,7 +74,12 @@ static void exclude_request_cb (flux_t *h, flux_msg_handler_t *w,
                               const flux_msg_t *msg, void *arg);
 static void include_request_cb (flux_t *h, flux_msg_handler_t *w,
                               const flux_msg_t *msg, void *arg);
+static void sched_params_set_request_cb (flux_t *h, flux_msg_handler_t *w,
+                              const flux_msg_t *msg, void *arg);
+static void sched_params_get_request_cb (flux_t *h, flux_msg_handler_t *w,
+                              const flux_msg_t *msg, void *arg);
 static int job_status_cb (const char *jcbstr, void *arg, int errnum);
+
 
 
 /******************************************************************************
@@ -139,6 +144,7 @@ typedef struct {
     flux_msg_handler_t **sim_handlers;
 } ssrvctx_t;
 
+static int schedule_jobs (ssrvctx_t *ctx);  /* Forward declaration */
 
 /******************************************************************************
  *                                                                            *
@@ -305,28 +311,70 @@ static int adjust_for_sched_params (ssrvctx_t *ctx)
 {
     flux_reactor_t *r = NULL;
     int rc = 0;
+    flux_msg_t *msg = NULL; 
+    /* delay_sched = true. 
+    Watchers need to be created the first time delay_sched is set to true. 
+    Otherwise, start the prepare/check watchers. The idle watcher is started by the ev_prep_cb. */ 
     if (ctx->arg.s_params.delay_sched && !ctx->sctx.in_sim) {
-        if (!(r = flux_get_reactor (ctx->h))) {
-            rc = -1;
-            goto done;
+        if (ctx->before && ctx->after) {                    /* Watchers exist */ 
+            flux_watcher_start (ctx->before); 
+            flux_watcher_start (ctx->after);
         }
-        if (!(ctx->before = flux_prepare_watcher_create (r, ev_prep_cb, ctx))) {
-            rc = -1;
-            goto done;
+       
+        /* If one of the watchers is not available, send error */
+        if ((!ctx->before && ctx->after) || (!ctx->after && ctx->before)) {
+            rc = -1; 
+            errno = EINVAL;
+            goto done; 
+        } 
+ 
+        if (!ctx->before && !ctx->after) {      /* Create and start watchers */
+            if (!(r = flux_get_reactor (ctx->h))) {
+                rc = -1;
+                goto done;
+            }
+            if (!(ctx->before = flux_prepare_watcher_create (r, ev_prep_cb, ctx))) {
+                rc = -1;
+                goto done;
+            }
+            if (!(ctx->after = flux_check_watcher_create (r, ev_check_cb, ctx))) {
+                rc = -1;
+                goto done;
+            }
+            /* idle watcher makes sure the check watcher (after) is called
+            even with no external events delivered */
+            if (!(ctx->idle = flux_idle_watcher_create (r, NULL, NULL))) {
+                rc = -1;
+                goto done;
+            }
+            flux_watcher_start (ctx->before);
+            flux_watcher_start (ctx->after);
         }
-        if (!(ctx->after = flux_check_watcher_create (r, ev_check_cb, ctx))) {
-            rc = -1;
-            goto done;
+    } // End if, delay = true
+
+    /* delay_sched = false. 
+    If set at runtime, stop the watchers, call schedule_jobs */
+    if (!ctx->arg.s_params.delay_sched && !ctx->sctx.in_sim) {
+        if (ctx->before && ctx->after) {
+            flux_watcher_stop (ctx->before); 
+            flux_watcher_stop (ctx->after);
+        
+            /* Create an event to schedule_jobs */   
+            flux_log (ctx->h, LOG_DEBUG, "Update delay_sched parameter");
+            msg = flux_event_encode ("sched.res.param_update", NULL);
+            if (!msg || flux_send (ctx->h, msg, 0) < 0) {
+                flux_log (ctx->h, LOG_ERR, "%s: error sending event: %s",
+                __FUNCTION__, strerror (errno));
+            }
+            flux_msg_destroy (msg);
         }
-        /* idle watcher makes sure the check watcher (after) is called
-           even with no external events delivered */
-        if (!(ctx->idle = flux_idle_watcher_create (r, NULL, NULL))) {
+        /* If one of the watchers is not available, send error */
+        if ((!ctx->before && ctx->after) || (!ctx->after && ctx->before)) {
             rc = -1;
-            goto done;
-        }
-        flux_watcher_start (ctx->before);
-        flux_watcher_start (ctx->after);
-    }
+            errno = EINVAL;
+            goto done; 
+        } 
+    } // End if, delay = false 
 done:
     return rc;
 }
@@ -1077,6 +1125,8 @@ static const struct flux_msg_handler_spec htab[] = {
     { FLUX_MSGTYPE_REQUEST,   "sched.cancel", cancel_request_cb, 0},
     { FLUX_MSGTYPE_REQUEST,   "sched.exclude",  exclude_request_cb, 0},
     { FLUX_MSGTYPE_REQUEST,   "sched.include",  include_request_cb, 0},
+    { FLUX_MSGTYPE_REQUEST,   "sched.params.set", sched_params_set_request_cb, 0},
+    { FLUX_MSGTYPE_REQUEST,   "sched.params.get", sched_params_get_request_cb, 0},
     { FLUX_MSGTYPE_EVENT,     "sched.res.*",  res_event_cb, 0},
     FLUX_MSGHANDLER_TABLE_END
 };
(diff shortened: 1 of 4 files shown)


t/t1008-runtime-sched-params.t
```diff
@@ -0,0 +1,124 @@
...
+test_debug '
+    echo ${basepath} &&
+    echo ${excl_4N4B} &&
+    echo ${excl_4N4B_nc}
+'
+
+ test_expect_success 'sched-params: getting sched params at runtime works' '
+    adjust_session_info 4 &&
+    flux hwloc reload ${excl_4N4B} &&
+    flux module load sched  sched-once=true &&
```
SteVwonder: I recommend explicitly setting the delay-sched and queue-depth arguments on load so that if the defaults ever change, that doesn't affect this test.

t/t1008-runtime-sched-params.t
```diff
@@ -0,0 +1,124 @@
...
+    adjust_session_info 4 &&
+    flux module remove sched &&
+    flux hwloc reload ${excl_4N4B} &&
+    flux module load sched sched-once=true &&
+    timed_wait_job 5 &&
+    submit_1N_nproc_sleep_jobs ${excl_4N4B_nc} 0 &&
+    flux wreck sched-params set queue-depth=4 &&
+    timed_sync_wait_job 10 &&
+    verify_1N_nproc_sleep_jobs ${excl_4N4B_nc} 
+'
```
SteVwonder: I recommend adding a `flux wreck sched-params get queue-depth` here to ensure that the value was actually set correctly.  Right now, I believe this test just ensures that Flux doesn't crash when you call `flux wreck sched-params get`.

SteVwonder: I think you can avoid writing to a file too by just using `test flux wreck sched-params get queue-depth -eq 4`

sched/sched.c
```diff
@@ -2086,6 +2092,81 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_param values requested by user (%u).", userid);
+    if (flux_request_unpack (msg, NULL, "{}") < 0)
+        goto error;
+
+    flux_log (h, LOG_DEBUG, "return values of sched_params ");
+
+    msg2 = flux_event_encode ("sched.params.returned", NULL);
```
SteVwonder: Initially, it seems like the only client that cares about a parameter query being answered is the client that made the original query.  And that client is about to get an RPC response, so this event seems redundant.

Can you elaborate on the use-cases that you had in mind for this event?

dongahn:
@tpatki: great start! 

It seems a few minor changes can help me review your PR. You may want to:

1. Put all the C code changes within `sched.c` into the first commit
2. Put all the C code changes within `sched.c` into the second commit;
3. And remove intermediate commits (you may find git rebase's squash useful)

Also for commit messages. The convention has been:
- Put a particular component name that the commit affects as "component:" in the commit title
- Then, have a short summary. Keep it short so that the fully title will show up full in the PR with no elision. "sched: Add support to update sched params at runtime". "test: Add cases for ..."
- Add some details in the body of the commit. Typically, one uses imperative like "Add request handers for..." instead of "This adds request handers for..." in here. But we can be flexible in the commit body message format.


3.

dongahn:
@tpatki: Sorry I just updated my comments above.

t/t1008-runtime-sched-params.t
tpatki: For now, I only added support for `flux wreck sched-params get`, which returns all the scheduler parameters. 

I didn't add support for  `flux wreck sched-params get queue-depth` or `flux wreck sched-params get queue-depth,delay-sched`. This would require me to introduce redundant code (similar to sched_params_args, but one that "gets" instead of "sets"), and I didn't want to do that. I think this is fine as a first cut, let me know if you think otherwise. 

To support `flux wreck sched-params get queue-depth,delay-sched`, we will 
(1) Either have to change `sched_params_args` to have a flag for set/get and have the latter return the `sched_params_t` object, which can then be packed and sent back to the requester, or,  
(2) Have to add similiar parsing code in either the subcommand in wreck or the callback in sched.c 

I was trying to avoid both of those things.

t/t1008-runtime-sched-params.t
tpatki: @SteVwonder: Will do.

sched/sched.c
tpatki: I was a bit confused about this myself, and was following the convention used by the existing include/exclude/cancel callbacks. 

This event (msg2) is used in the `flux_send` call that goes out to the broker, which is different than the `flux_respond_pack` which contains the data requested by the client. I wasn't sure why the other callbacks had it, and if the response to the client was different than the information sent out to the broker. Maybe @dongahn or @grondo have some clarity here?

tpatki:
@dongahn: Shall fix the PR per your suggestions soon (need to head out for a bit but will revisit later today evening).

sched/sched.c
dongahn: In the case of exclude and include request, an extra event needs to be generated so that the scheduler will get an resource event and invoke a scheduler loop. For this particular case, we probably won' t need an event. Sorry if it wasn't obvious in the code. (Should have had a comment.)

dongahn:
Thanks @tpatki

sched/sched.c
tpatki: Ah, that makes sense. Thanks Dong! Yes it wasn't clear to me, and didn't want to break anything or cause a race condition. I'll remove the event/flux_send from both the callbacks I added (and add a comment to capture the above discussion as well).

tpatki:
@dongahn, @SteVwonder: I submitted a separate PR to flux-core for new subcommand, please take a look at that as well.

tpatki:
@dongahn: updated as per your suggestion. Let me know if I missed anything. Thanks!

sched/sched.c
```diff
@@ -2086,6 +2092,61 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_params change requested by user at runtime (%u).", userid);
+    if (flux_request_unpack (msg, NULL, "{s:s}", "param", &sprms) < 0)
+        goto error;
+   
+    if (sprms)
+        rc = sched_params_args (sprms, &(ctx->arg.s_params));
+
```
dongahn: I don't think calling `sched_params_args ()` alone will lead to scheduling behavior changes for all parameters. `queue-depth` would be okay. But `delay-sched` would not. If the sched load option was `delay-sched=false`, and the user executes `flux wreck sched-param set delay-sched=true` to change it. This alone will not change the scheduler to start delay scheduling. 

Generally speaking, as we will add more parameters, we can't expect only changing a field in the sched context will effect the requested behavior. 

I think this can be addressed if we refactor `adjust_for_sched_params ()` into `effect_sched_params ()` Here, you can have a logic to dynamically adjust the behavior based on the given scheduling parameters. 

I have to believe you can make it so that the function will work for both load time options as well as dynamically requested options.

sched/sched.c
```diff
@@ -2086,6 +2092,61 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+{
+    ssrvctx_t *ctx = getctx ((flux_t *)arg);
+    uint32_t userid = 0;
+    char *sprms = NULL;
+    int rc = 0;
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_params change requested by user at runtime (%u).", userid);
```
dongahn: Minor nit: Maybe `(%u)` can be next to user?

sched/sched.c
```diff
@@ -2086,6 +2092,61 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+                                const flux_msg_t *msg, void *arg)
+{
+    ssrvctx_t *ctx = getctx ((flux_t *)arg);
+    uint32_t userid = 0;
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_param values requested by user (%u).", userid);
+    if (flux_request_unpack (msg, NULL, "{}") < 0)
```
dongahn: Did you want to have bulk get AND individual get?

t/t1008-runtime-sched-params.t
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
+#set -x
+
+test_description='Test dynamic updation of queue depth. 
+First cut. Ensure jobs are correctly scheduled under different values of
```
dongahn: Remove "First cut."

t/t1008-runtime-sched-params.t
```diff
@@ -0,0 +1,124 @@
...
+    flux hwloc reload ${excl_4N4B} &&
+    flux module load sched sched-once=true \
+    sched-params=queue-depth=2048 &&
+    timed_wait_job 5 &&
+    submit_1N_nproc_sleep_jobs ${excl_4N4B_nc} 0 &&
+    flux wreck sched-params set delay-sched=true &&
+    timed_sync_wait_job 10 &&
+    verify_1N_nproc_sleep_jobs ${excl_4N4B_nc} 
+'
+test_expect_success 'sched-params: unloaded sched module' '
```
dongahn: Great you have added many tests! You think it will be possible to test whether "set" has led to scheduler behavior changes as well?

dongahn:
@tpatki: Thank you for putting this together so quickly. I have just added some comments inline.

sched/sched.c
tpatki: Noted, I will work on fixing this and also on testing if the  change took effect. I was thinking about this as well, thanks for the pointers!

sched/sched.c
tpatki: Only bulk get for now, lets discuss individual get in person. Take a look at the comment I added to Stephen's question above regarding that. I'm not sure how to avoid duplicate code there.

dongahn: Ah. OK. Yeah I'm ok with bulk get only since this will have be refactored later anyway within flux sched command. Thanks @tpatki.

tpatki: We should also think about how to refactor `sched_params_args` so we don't write duplicate code for the individual get. I'll add an issue corresponding to this once this PR is done.

t/t1008-runtime-sched-params.t
SteVwonder: > This would require me to introduce redundant code (similar to sched_params_args, but one that "gets" instead of "sets"), and I didn't want to do that. I think this is fine as a first cut, let me know if you think otherwise.

This makes sense.  Thanks @tpatki!

grondo: Could you use `flux wreck sched-params get | grep queue-depth` to simulate a "get" of a single value? It would be nice to verify the set worked in the tests.

tpatki:
@dongahn, @SteVwonder, @grondo:

Please see the modified PR that addresses the changes requested. I have added a few test cases, and have manually verified each of the cases for `adjust_for_sched_params` through logging/stepping, but I can't ensure the correctness of delay_sched (I can only match existing workflow). If you can think of better test cases, please let me know.

sched/sched.c
tpatki: @dongahn, @SteVwonder: PRs now support individual get as well. Please see test cases.

SteVwonder:
LGTM. It just needs a rebase to master. Thanks, @tpatki! This was more work than I thought it would be, but this should come in handy.

tpatki:
@SteVwonder: rebased.

dongahn:
@tpatki and I looked at the code together a bit after our meeting this morning. There will be some minor changes.

tpatki:
@dongahn: updated as per our discussion.

sched/sched.c
```diff
@@ -305,28 +311,68 @@ static int adjust_for_sched_params (ssrvctx_t *ctx)
...
+            flux_watcher_start (ctx->before); 
+            flux_watcher_start (ctx->after);
         }
-        if (!(ctx->after = flux_check_watcher_create (r, ev_check_cb, ctx))) {
-            rc = -1;
-            goto done;
+       
+        /* If one of the watchers is not available, send error */
+        if ((!ctx->before && ctx->after) || (!ctx->after && ctx->before)) {
+            rc = EINVAL;
```
dongahn: Thank you for following up. I would actually go with:

```c
rc = -1;
errno = EINVAL;
```

sched/sched.c
```diff
@@ -305,28 +311,68 @@ static int adjust_for_sched_params (ssrvctx_t *ctx)
...
+                __FUNCTION__, strerror (errno));
+            }
+            flux_msg_destroy (msg);
         }
-        flux_watcher_start (ctx->before);
-        flux_watcher_start (ctx->after);
-    }
+        /* If one of the watchers is not available, send error */
+        if ((!ctx->before && ctx->after) || (!ctx->after && ctx->before)) {
+            rc = EINVAL;
```
dongahn: Thank you for following up. I would actually go with:

```c
rc = -1;
errno = EINVAL;
```

sched/sched.c
```diff
@@ -2086,6 +2134,66 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+{
+    ssrvctx_t *ctx = getctx ((flux_t *)arg);
+    uint32_t userid = 0;
+    char *sprms = NULL;
+    bool prev_delay_sched = ctx->arg.s_params.delay_sched;
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_params change requested by user (%u) at runtime.", userid);
```
dongahn: This line seems long. Does it fit 80 characters? If not, I'd suggest to break the line into two.

sched/sched.c
```diff
@@ -2086,6 +2134,66 @@ static int job_status_cb (const char *jcbstr, void *arg, int errnum)
...
+static void sched_params_get_request_cb (flux_t *h, flux_msg_handler_t *w,
+                                const flux_msg_t *msg, void *arg)
+{
+    ssrvctx_t *ctx = getctx ((flux_t *)arg);
+    uint32_t userid = 0;
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "sched_param values requested by user (%u).", userid);
```
dongahn: Same comment as above.

t/t1008-runtime-sched-params.t
dongahn: Very minor nit. I don't think you need "Frist cut" here.

tpatki:
@dongahn, @SteVwonder: Looks like my testcase fails with certain gcc/clang versions, not sure why. Will debug further in the evening/tmrw.

t/t1008-runtime-sched-params.t
```diff
@@ -0,0 +1,186 @@
...
+#
+# print only with --debug
+#
+test_debug '
+    echo ${basepath} &&
+    echo ${excl_4N4B} &&
+    echo ${excl_4N4B_nc}
+'
+
+ test_expect_success 'sched-params: getting sched params (no args) at runtime works' '
```
dongahn: Any way we can try to shorten the test messages here and below to fit into 80 characters. I don't think it is required but it would be generally good for readability.

dongahn:
>  Looks like my testcase fails with certain gcc/clang versions, not sure why. Will debug further in the evening/tmrw.

Sounds good. I also added some minor comments above. I believe we are close. Thanks!

tpatki:
@dongahn: I think some of the tests might be failing because rc isn't explicitly set to -1.

Regarding, error=EINVAL, should I be setting that to a global error variable somewhere? I don't see one in sched (setting it locally doesn't make sense as it won't propagate through the callstack). Thoughts?

dongahn:
We use `errno` for that purpose. Hopefully it answers your question. http://man7.org/linux/man-pages/man3/errno.3.html

Dong

dongahn:
@tpatki: This looks great. Merging.

PR #376, opened 2018-08-08
sched: remove trailing whitespaces

dongahn:
Just noticed trailing whitespaces that were introduced from the latest changes. 

If vim is used, adding the following to `~/.vimrc` should highlight the trailing whitespaces for C code, which can help one not introduce them:

```
let c_space_errors = 1
```


dongahn:
I think the codecov check is bogus as this PR only contains trailing whitespace removal.

PR #378, opened 2018-08-11
resource-query as resource matching comms module

dongahn:
@SteVwonder and @tpatki: this is just for your two just to show you where I'm headed -- since you are also working on this layer. I don't even have Makefiles for these changes.

`rquery.cpp` is the main file to implement the resource matching service. 

It populates the resource graph database and then handles `match` operations in the request handlers. For now, the `match` request handler will get the serialized jobspec encoded in the message. But in a near future, it will probably get the jobspec from KVS according to the job schema. 

Since we don't have the corresponding scheduling control loop comms module, this handler just returns the `R` in its own native format with the response. Once we have the control loop service around it, and determine what `R` should be and our KVS job schema should be, the `R` format and where it is sent/dumped will change. Changing formats shouldn't be difficult going forward as I plan to make "emitter" as programmable as possible.

Other than this, initially, I probably only want to support a minimal set of other requests -- for now I have `cancel` and `info` requests.

@SteVwonder: I believe your `hwloc` work (Issue #355) should be directly transferrable to this service pattern. 

@tpatki: I believe your locality aware scheduling policy enhancement should also be orthogonal.

If not, let me know.

Any early feedback welcome.

Since this PR won't even compile, feedback related architecture choices and relevance to your other work will be most useful at this point. I know there are lots of typos...

### DIFF
--- a/resource/modules/rquery.cpp
+++ b/resource/modules/rquery.cpp
@@ -1,0 +1,573 @@
+/*****************************************************************************\
+ *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+
+#include <cstdint>
+#include <cstdlib>
+#include <fstream>
+#include <sstream>
+#include <algorithm>
+#include <iterator>
+#include <cstdint>
+#include <cerrno>
+#include <vector>
+#include <map>
+#include <boost/algorithm/string.hpp>
+#include "resource/schema/resource_graph.hpp"
+#include "resource/generators/gen.hpp"
+#include "resource/traversers/dfu.hpp"
+#include "resource/jobinfo/jobinfo.hpp"
+#include "resource/policies/dfu_match_policy_factory.hpp"
+
+extern "C" {
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+}
+
+using namespace std;
+using namespace Flux::resource_model;
+
+/******************************************************************************
+ *                                                                            *
+ *                Resource Matching Service Module Context                    *
+ *                                                                            *
+ ******************************************************************************/
+
+struct rmatch_args_t {
+    string grug;
+    string match_subsystems;
+    string match_policy;
+    string prune_filters;
+    string R_format;
+};
+
+struct rmatch_ctx_t {
+    rmatch_args_t args;          /* Module load options */
+    dfu_match_cb_t *matcher;     /* Match callback object */
+    dfu_traverser_t *traverser;  /* Graph traverser object */
+    resource_graph_db_t db;      /* Resource graph data store */
+    f_resource_graph_t *fgraph;  /* Graph filtered by subsystems to use */
+    map<uint64_t, job_info_t *> jobs;     /* Jobs table */
+    map<uint64_t, uint64_t> allocations;  /* Allocation table */
+    map<uint64_t, uint64_t> reservations; /* Reservation table */
+};
+
+
+/******************************************************************************
+ *                                                                            *
+ *                          Request Handler Prototypes                        *
+ *                                                                            *
+ ******************************************************************************/
+
+static void match_request_cb (flux_t *h, flux_msg_handler_t *w,
+                              const flux_msg_t *msg, void *arg);
+
+
+/******************************************************************************
+ *                                                                            *
+ *                   Module Initialization Routines                           *
+ *                                                                            *
+ ******************************************************************************/
+
+static void freectx (void *arg)
+{
+    if (ctx) {
+        delete ctx->matcher;
+        delete ctx->traverser;
+        delete ctx->fgraph;
+        for (auto &kv : ctx->jobs) {
+            delete kv.second;    /* job_info_t* type */
+            ctx->jobs.erase (kv.first);
+        }
+        ctx->jobs.clear ();
+        ctx->allocations.clear ();
+        ctx->reservations.clear ();
+        delete ctx;
+    }
+}
+
+static void set_default_args (rmatch_args_t &args)
+{
+    args.grug = "none";
+    args.match_subsystems = "CA";
+    args.match_policy = "high";
+    args.prune_filters = "core";
+    args.R_format = "R_lite";
+}
+
+static rmatch_ctx_t *getctx (flux_t *h)
+{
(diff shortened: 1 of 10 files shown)


SteVwonder:
The one thing I thought of while reviewing this PR: what parameters/arguments of this service will we allow/support changing at runtime?

dongahn:
Good question! Currently I am only allowing arguments to be set at module load time.

I had the same question but I wasn't convinced providing runtime option changes would be useful for any of the current args.

One obvious argument to have runtime control would be the selection plugin argument. But only a subset of plugins may make sense to to be changed at runtime.

For example, If the current policy is locality aware selection matcher that uses the containment subsystem and users want to change it to high ID first, it is largely okay. But if they want to change it to power aware policy that uses the power hierarchy, it may not make whole lot of sense. Planner caches and etc state kept at the intermediate vertices (e.g., rack vertex) of the containment hierarchy cannot be used by the power aware policy matcher as its dominant hierarchy is the power hierarchy. Even if we can make this work, given the complexity, I wanted to disallow runtime changes for this round.

Maybe the pruning filter argument is just about the only option where there is a clear benefit providing a runtime change option...

What do you have in mind?

dongahn:
> Given how much would have to change and be rebuilt when changing almost any parameter, it may make sense to just require a complete reload of the entire module to change a parameter.

That's what I was thinking. One orthogonal thing we have to think about is "resilency". If the service is reloaded with the same or compatible argument, we want to restore the existing job state.

SteVwonder:
> If the service is reloaded with the same or compatible argument, we want to restore the existing job state.

Yeah.  Totally agree.  I think we can leverage the KVS for running and allocated jobs.  For reservations though, we may just need to regenerate them from scratch, but that will require some cooperation from the scheduler.  After the resource module is reloaded, we need to make sure that the scheduler restarts it's scheduling loop from the beginning rather than from the middle of the job queue.

dongahn:
I think there are two potentially complementary schemes: "recompute on load" using KVS like you suggested and/or "checkpoint on unload / restart on load" scheme. 

I think the latter will require a little or no support form the scheduler loop service module and maybe easier to implement. We still have to isolate what state needs to be checkpointed and where...

tpatki:
One thing I'm unclear about is how the dependencies between the scheduler module and the resource module will be captured if we have minimal support for runtime parameters (or trigger reloads).  

-- Runtime update for sched module: Will the resource module be responsible for maintaining the current state of nodes (eg node up/down/drain) and communicate that to the sched module? 

-- Runtime update for resource module: When a job terminates prematurely and nodes go down, which entity is responsible for updating resource module? Flux-core?

dongahn:
Capturing an offline discussion with @tpatki:

We probably want to have both the upcoming schedule loop module and resource module to subscribe the node "down
 event. 

    - Upon receiving an event, resource service will mark the node vertex and its subtree resource vertices "unavailable"
    - Upon receiving an event, schedule loop module will go through the job-kill protocol that should include the removal of the job from the queue and send a job-cancel RPC to the resource module.


BTW, are you okay to call this resource matching service module as simply `resource-query` (as coded here) or will this be too confusing?

dongahn:
> Why not just let the job die/exit on its own? What if the job is resilient to a single node failure? For example, what if I am running a sub-instance with a bunch of single core tasks. I don't think we want to tear down the entire sub-instance and sub-jobs just because one node died. If it isn't resilient to node failure (i.e., naive MPI app) then the job should crash/burn and then exit.

Great point!

Maybe this needs to go into a job spec somewhere so that the user can ask "don't kill this job on node failures." And probably have a way to add this to the scheduler property so that a site can configure the default behavior. 

Most traditional MPI jobs won't tolerate a single node failure today. So the worst case would be the remaining MPI tasks are burning cycles at large scale without useful computation. My guess is at least for the system instance, the default would be "kill the job on node failure" unless the job wants the opposite option.

dongahn:
> I like the name resource-query for the command, but I think the module should be named something different. Maybe just resource?

`resource` was actually my first choice. But then in my second thought, a generic name like that can be confusing in particular when we have `resource-hwloc` service. If other folks are okay with `resource`, I'd be happy to go with it.

SteVwonder:
> But then in my second thought, a generic name like that can be confusing in particular when we have resource-hwloc service

Maybe something like `resource-matcher`, `resource-matching`, or `resource-match`.

The main thing that I don't like about "query" is that it makes the module/service sound read-only.

tpatki:
I agree: I like `resource-match`, or even `resmatch` if we want something shorter.

dongahn:
> resource-matcher, resource-matching, or resource-match.

Ha ha. Those were actually my second choices. Then, I realized in writing the request handlers, they become a bit awkward:

```
    { FLUX_MSGTYPE_REQUEST, "resource-match.match",  match_request_cb, 0},
```

etc. 

If `resource` is okay, I think it is actually the best.
```
    { FLUX_MSGTYPE_REQUEST, "resource.match",  match_request_cb, 0},
    { FLUX_MSGTYPE_REQUEST, "resource.cancel",  match_request_cb, 0},
    { FLUX_MSGTYPE_REQUEST, "resource.info",  match_request_cb, 0},
```

SteVwonder:
`resource` works for me.  The RPC topics are definitely cleaner.

grondo:
On Fri, Aug 17, 2018, 3:31 PM Dong H. Ahn <notifications@github.com> wrote:

> Why not just let the job die/exit on its own? What if the job is resilient
> to a single node failure? For example, what if I am running a sub-instance
> with a bunch of single core tasks. I don't think we want to tear down the
> entire sub-instance and sub-jobs just because one node died. If it isn't
> resilient to node failure (i.e., naive MPI app) then the job should
> crash/burn and then exit.
>
> Great point!
>
> Maybe this needs to go into a job spec somewhere so that the user can ask
> "don't kill this job on node failures." And probably have a way to add this
> to the scheduler property so that a site can configure the default behavior.
>
This is traditionally how Slurm behaves. The default behavior of killing a
job on node failure ensures that resources are not wasted by a misbehaving
job (i.e. you can't trust most jobs to clean themselves up, so some sites
might choose to make continue-on-resource-failure opt in only)




Most traditional MPI jobs won't tolerate a single node failure today. So

dongahn:
> This is traditionally how Slurm behaves. The default behavior of killing a
job on node failure ensures that resources are not wasted by a misbehaving
job (i.e. you can't trust most jobs to clean themselves up, so some sites
might choose to make continue-on-resource-failure opt in only)

So we will give a site a way to configure the behavior through the scheduler property config. I have the placeholder [here](https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp), which will later be converted into something like toml.

Then, we give a way for the user to overwrite the behavior using jobspec. @SteVwonder's justification is compelling.

dongahn:
> Just keeping the interesting discussion going -- is this functionality
party of the scheduler, the job manager, or sometimes perhaps even the job
shell?

That is a good question.

Now that I think about this, the job manager needs to know whether to kill the job or not. (even without the involvement of the scheduler).

Then, the scheduler loop service needs to get this job-kill event one way or the other to decide whether to remove the job from its queue or not. 

Next, the resource matching service needs to clear up the job state if the scheduler loop service tells it. 

If you add this property to the system property config, it seems that can take care of this at the job manager level.

But does the job manager necessarily know what's in the jobspec (i.e., `don't kill on node failure`)? Maybe it should?

SteVwonder:
> But does the job manager necessarily know what's in the jobspec (i.e., don't kill on node failure)? Maybe it should?

Maybe it makes sense to give the responsibility of killing/not killing to the job shell in that case, since the job shell already parses the jobspec (am I right on that?). I presume the job shell is also aware of what resources it is running on and can listen for/act on events related to those resources.

dongahn:
Job shell will have to abide by the system config though. (Probably not a problem)

Then the scheduler components can remain simple and reactive (on both node down event and job kill event) without knowing the site configuration or the option in the jobspec.

If this can be done this way, I like the separation of concerns here.

resource/modules/rquery.cpp
```diff
@@ -0,0 +1,573 @@
...
+    const char *js_str = NULL;
+    stringstream R;
+    rmatch_ctx_t *ctx = getctx ((flux_t *)arg);
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        goto error;
+
+    flux_log (h, LOG_INFO, "match requested by user (%u).", userid);
+
+    if (flux_request_unpack (msg, NULL, "{s:s}", "cmd", &cmd) < 0)
```
SteVwonder: These three calls to unpack can be merged into a single call.  [Link to an example](https://github.com/flux-framework/flux-core/blob/3c3890cbc619712409b20bf1fadaabda6451b111/src/broker/sequence.c#L153)

dongahn:
Ok. I believe I have received good enough feedback.

PR #380, opened 2018-08-17
Adding topology-aware scheduler

sampollard:
This is to replace PR #379. Now this has fewer commits, and I addressed all issues added before [this comment](https://github.com/flux-framework/flux-sched/pull/379#issuecomment-413990820)


SteVwonder:
Thanks @sampollard for cleaning up the commits!

My biggest piece of feedback right now is that I am not a huge fan of the `topo_tree_t` data structure.  It doesn't seem like the "right way" to solve the problem of needing plugin-specific data stored at vertices in the resource graph.  It introduces a large runtime computational and memory overhead, as well as a re-implementation of much of the `resrc` library.

I think the "right way" would be to add support for user-defined data storage at `resrc` vertices.  At the same time, this would require modifications to the `resrc`, which we plan on deprecating after we get the `resource` service up and running.  Once the `resrc` is deprecated, this plugin would be deprecated as well.   It would also require a significant re-write of this PR (mainly through the removal of a lot of then unnecessary code). This plugin does seems like the perfect candidate for porting to the new `resource` service.

Given all this, I see two paths forward (maybe others can suggest additional paths):

1. Merge this PR as-is; in the near-future, deprecated along with the `resrc`, port to the `resource` service (which supports a scratchpad at each resource vertex)
2. Invest time into extending `resrc` to support user-defined data at each vertex and then refactor this PR

@dongahn and @trws, any thoughts or ideas?

sched/sched_topo.c
```diff
@@ -0,0 +1,1755 @@
...
+    flux_log (h, LOG_ERR, "This request can never be satisfied");
+    return NULL;
+}
+
+/* Find the nodes (physical compute nodes) of a tree.
+ * Expects an initial argument of an empty list and the root of the tree you
+ * wish to build. Don't forget to call resrc_tree_list_shallow_destroy once
+ * you're done with this.
+ */
+static void find_nodes (resrc_tree_t *r_tree, resrc_tree_list_t *rtl)
```
SteVwonder: I recommend switching this function to producing a `zlist_t`.  Then you can use `zlist_destroy` (which is shallow by default) and remove the need for `resrc_tree_list_shallow_destroy` (which seems weird to have without also having a `resrc_list_shallow_destory`).

sched/sched_topo.c
```diff
@@ -0,0 +1,1755 @@
...
+    // TODO: Update this to go through each level and actually count them
+    int idx;
+    int64_t count = 1;
+    for (idx = 0; idx <= l; idx++) {
+        count *= levels[idx]; 
+    }
+    return count;
+}
+
+// Reservation Depth Guide:
```
SteVwonder: In the comments at the top of the file, it states that this plugin assumes EASY backfilling.  While this function (and other functions copied from the backfilling plugin) assume otherwise.  If EASY backfilling truly is a hardcoded assumption, I would remove this comment and change `reservation-depth` (which is an integer) to `easy-backfilling` (which is a boolean).

dongahn:
I would like to avoid option 2 at all cost. 

Do we have a compelling reason to merge this PR right now though?

Wouldn't it be a better investment if someone can play with `resource` or `resource-query` to see if the logic can be ported as a simple matcher callback class?

sampollard:
My only concern with merging later rather than sooner is about what I claim in this upcoming supercomputing paper. The camera-ready version is due August 28 and I would like to say this plugin is in this repository. Of course, we could say "in the process..."

SteVwonder:
> Wouldn't it be a better investment if someone can play with resource or resource-query to see if the logic can be ported as a simple matcher callback class?

Yeah, that would be the best option for the long-term.  @sampollard are you willing/able to help us with this?  I am less worried about the specifics of how this plugin is implemented if we can reasonably assume it will be replaced by a newer one implemented against the `resource` service.

> The camera-ready version is due August 28 and I would like to say this plugin is in this repository. Of course, we could say "in the process..."

Maybe you can point the SC paper to your fork?  I believe that is what I did for my HPDC paper on IO-aware scheduling.

dongahn:
> Yeah, that would be the best option for the long-term. @sampollard are you willing/able to help us with this? I am less worried about the specifics of how this plugin is implemented if we can reasonably assume it will be replaced by a newer one implemented against the resource service.

Agreed. Don't get me wrong, though. The work is great, and I really do hope to see this work land into our production scheduler land. 

The more advanced matcher plugins we can demonstrate that our resource matching infrastructure can support (through simple matcher callback code), the better.

Depending on the walking pattern of this logic, we may need to extend our infrastructure but that's essentially our growth plan. 

> Maybe you can point the SC paper to your fork? I believe that is what I did for my HPDC paper on IO-aware scheduling.

That sounds right.

sampollard:
I have no problem pointing to my fork of this repo for the paper.

As for a complete refactoring of this code, unfortunately that's not something I have the time for. The most I can reasonably do is provide clearer documentation and reasoning for what I did write, or fix issues like the ones you've mentioned in this PR.

sched/sched_topo.c
sampollard: This is an issue we encountered last summer. We would like to be able to support different reservation depths, but this would require creating copies of the topo_tree, which is already not a great solution to this problem. This is definitely a problem. To at least be safe, I added a check in the process args which ensures you can only enter 0 or 1.

PR #385, opened 2018-08-31
[WIP] resource: support for hwloc ingestion

SteVwonder:
Current state of the code for populating the resource library using HWLOC resource information.

Supports both an HWLOC XML file and the HWLOC in the Flux KVS as input.

Still TODO:

- switch to async KVS operations (only makes sense once the resource matching service is ready since this requires a reactor)
- watch_once on the 'resource.hwloc.loaded' key rather than every rank string
- tests (will be easier to do once it is a service)


dongahn:
Thanks @SteVwonder. I will take a look at this PR as soon as practical. I have just posted a PR to coordinate my work better with yours. Please look at the proposed steps to see if they sound okay with your plan.

resource/utilities/resource-query.cpp
```diff
@@ -465,6 +472,12 @@ int main (int argc, char *argv[])
             case 'e': /* --elapse-time */
                 ctx->params.elapse_time = true;
                 break;
+            case 0:
```
dongahn: Hmmm, 0?

resource/Makefile.am
```diff
@@ -86,5 +86,6 @@ utilities_resource_query_LDADD = \
     $(BOOST_SYSTEM_LIB) \
     $(BOOST_FILESYSTEM_LIB) \
     $(BOOST_GRAPH_LIB) \
-    $(BOOST_REGEX_LIB)
+    $(BOOST_REGEX_LIB) \
+	$(HWLOC_LIBS)
```
dongahn: Use whitespaces.

resource/generators/gen.cpp
```diff
@@ -454,6 +457,166 @@ int resource_generator_t::read_graphml (const string &fn, resource_graph_db_t &d
     return (m_err_msg == "")? rc : -1;
 }
 
+ggv_t add_new_vertex(resource_graph_db_t &db, const ggv_t &parent,
```
dongahn: Wondering if this is a good time to introduce a class hierarchy for the generator class. 

1. Base generator class defines the interfaces
2. Make GRUG-based class as a derived class of that
3. Make Hwloc-based class also as a derived class of that

This way, when we add more reader types, it becomes easier to plug into our system ?

resource/generators/gen.cpp
```diff
@@ -471,6 +471,7 @@ ggv_t add_new_vertex(resource_graph_db_t &db, const ggv_t &parent,
     db.resource_graph[v].type = type;
     db.resource_graph[v].basename = basename;
     db.resource_graph[v].size = size;
+    db.resource_graph[v].rank = rank;
```
dongahn: As discussed, let's go with this method to begin with. If it turns out, introducing a new vertex whose resource type is "rank" makes generating `R` format better, we can change how we handle the rank. 

Without `R` spec, it is not clear which is better for now.

resource/utilities/resource-query.cpp
SteVwonder: Yeah.  It is odd, but AFAIK it is the easy way to handle long arguments that don't have a shorter option (`val` in the option struct is 0).

> getopt_long() and getopt_long_only() also return the option character when a short option is recognized. *For a long option, they return val if flag is NULL, and 0 otherwise.* Error and -1 returns are the same as for getopt(), plus '?' for an ambiguous match or an extraneous parameter. 

Source: https://linux.die.net/man/3/getopt_long

Once #386 is merged, I think this functionality of the resource-query will go away and be moved to the service module (unless we want to use resource-query as the main driver of the tests rather than the service module).

resource/Makefile.am
SteVwonder: Should be fixed with the latest push, and my emacs config is modified so that this doesn't happen again. Thanks!

resource/generators/gen.cpp
SteVwonder: Yeah.  This makes sense.  Off the top of my head, the commonalities of vertex creation (as you have highlighted) can be abstracted into the top class.  I'll have to look deeper at the GRUG code to see what else is shared between the two.

resource/utilities/resource-query.cpp
dongahn: > Once #386 is merged, I think this functionality of the resource-query will go away and be moved to the service module (unless we want to use resource-query as the main driver of the tests rather than the service module).

Actually, I think we want to utilize `resource-query` as much as possible for testing. The CLI should be much faster and easier to handle than resource module.

resource/generators/gen.cpp
dongahn: Great! If we do this right, extending this to `R` reader should also be easily managed. (Once we determine the actual approach).

resource/utilities/resource-query.cpp
SteVwonder: Oh. Sorry. I wasn't intending to suggest that `resource-query` would go away, I meant purely the extraction of hwloc information from a running Flux instance, but I can see how that would be useful to have in `resource-query` for testing.

dongahn:
@SteVwonder: Any progress for this one? I have knocked down most of the cleanup work I needed to do and my next task would be `R` format. I believe landing this PR can significantly help us to make progress on `R`. Thanks!

SteVwonder:
@dongahn: checkpointing before the holiday.

Note: I had to switch the prune filter to `ALL:pu` as we did while debugging in order for matches to occur against the `hwloc` resource graph.  Once #400 is closed, we can go back to using the default prune-filter.

I created a separate issue for the generator class hierarchy (#411).

Still TODO:

1. `watch-once` only on the 'resource.hwloc.loaded' key rather than every rank string
2. switch to async KVS operations

For 1, I think this change would only take an hour or so to implement. 
For 2, this is more involved, and will be much easier once https://github.com/flux-framework/flux-core/issues/1826 is done. Do we want to make a separate issue & PR for that or bundle it in with this PR?  My preference would be to do it in a separate PR.

dongahn:
> Note: I had to switch the prune filter to ALL:pu as we did while debugging in order for matches to occur against the hwloc resource graph.

Yes, this is fine. We can perhaps even pass this as a resource match module load option.

dongahn:
> Do we want to make a separate issue & PR for that or bundle it in with this PR? My preference would be to do it in a separate PR.

We definitely want a separate PR so that we can make progress on other work that depends on hwloc!

dongahn:
> checkpointing before the holiday.

Have a great Thanksgiving, Stephen!

SteVwonder:
> Yes, this is fine. We can perhaps even pass this as a resource match module load option.

Sorry, I should have been more clear.  The code still has `ALL:core` as the default.  I'm passing `prune-filter=ALL:pu` to the resource module when it is being loaded: https://github.com/flux-framework/flux-sched/pull/385/files#diff-18dbb5848dd9395d8821569fdae9a452R47

> Have a great Thanksgiving, Stephen!

You too!

SteVwonder:
Restarted a hung builder.  Travis seems happy now.  Coverage is low due to the light testing of `resource-query` (the testing is currently done through the matching service).  I can add a test that uses `resource-query` to load an hwloc xml file and run a few queries against the resource graph.  Since there is also no testing against an hwloc with GPUs, I'll try and get two birds with one stone.
 
Also, I thought `resource.hwloc.loaded` was a single key, but it is actually a directory of keys, so reducing the total number of kvs watches will require entering the reactor (to watch the entire directory) which is a bigger refactor than I want to do for this PR.  So let's push that off into a separate PR as well.

SteVwonder:
Travis is happy.  Coveralls is happy.  I think this PR is (finally) ready for review. Sorry for the delay on this one.

The only piece of code (outside of error paths) that isn't tested is the opencl co-processor piece of hwloc parsing.  At some point, we should get the hwloc output from a machine with AMD GPUs and add it to the test suite (making an issue about that now).

resource/utilities/Makefile.am
```diff
@@ -33,6 +33,8 @@ resource_query_CXXFLAGS = \
 resource_query_LDADD = \
     ../libresource.la \
     $(READLINE_LIBS) \
+	$(FLUX_CORE_LIBS) \
```
dongahn: There seems to be an extra whitespace or a use of tab in the leading whitespaces... The indentation for $(FLUX_CORE_LIBS) and $(BOOST_LDFLAGS) seems to be a couple of characters too deep...

resource/utilities/Makefile.am
```diff
@@ -20,6 +19,7 @@ grug2dot_CXXFLAGS = \
     $(AM_CXXFLAGS)
 grug2dot_LDADD = \
     ../libresource.la \
+	$(BOOST_LDFLAGS) \
```
dongahn: Seems to have an extra leading whitespace.

resource/utilities/resource-query.cpp
```diff
@@ -33,6 +33,7 @@
 #include <readline/readline.h>
 #include <readline/history.h>
 #include <boost/algorithm/string.hpp>
+#include <flux/core.h>
```
dongahn: I thought we agree that we want to make resource-query not depend on flux-core at all. This way, the `resource-query` utility can continue to serve as a simpler testing and debugging layer without involving other parts of flux.

If this and other relevant changes are interim changes that later removed, could you squash the commits so that the interim commits don't show up in our commit history?

dongahn:
@SteVwonder: I skimmed through the changes. By and large, this looks pretty good. 

Please look at my inline review comments above. It seems there are some interim changes that are later removed and/or fixed. If you can squash those changes appropriately, it will help me to finish up my reviews sooner. 

As part of that, could you also rebase? 

Thanks!

resource/utilities/resource-query.cpp
SteVwonder: We did discuss that in this thread, but I believe the conclusion then swung the other way:

> > SteVwonder: Once #386 is merged, I think this functionality of the resource-query will go away and be moved to the service module (unless we want to use resource-query as the main driver of the tests rather than the service module).

> dongahn: Actually, I think we want to utilize resource-query as much as possible for testing. The CLI should be much faster and easier to handle than resource module.

I'm fine removing the hwloc KVS functionality from resource-query (but leaving hwloc XML features) since there are already tests for it via the resource-match module.  I don't really have a strong opinion one way or the other, so whatever your preference, I say we go with.

SteVwonder:
> If you can squash those changes appropriately, it will help me to finish up my reviews sooner.
> As part of that, could you also rebase?

:+1: Rebased onto master and incremental development has been squashed.

resource/utilities/resource-query.cpp
dongahn: Sorry I probably miscommunicated. I believe it would be best to remove the flux-core kvs dependency while leaving the hwloc XML features. We will have a much better mileage out of `resource-query` that way.

SteVwonder: :+1: Makes sense.  I'll make that change now and push shortly.

SteVwonder:
flux (and jansson) have now been removed from `resource-query` but support for reading from an hwloc xml file was left intact.

resource/modules/resource_match.cpp
```diff
@@ -238,22 +245,103 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
...
+    if (!json_str || !(o = Jfromstr (json_str))
+                  || !json_is_string (o)) {
+        saved_errno = EPROTO;
+        goto error;
+    }
+    free (json_str);
+    return o;
+error:
+    free (json_str);
+    Jput (o);
```
dongahn: Is `Jput (NULL)` valid? It could very well be the case but I thought I should make the comment just in case.

resource/modules/resource_match.cpp
```diff
@@ -238,22 +245,103 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
...
+
+/*
+ * Read the hwloc xml stored in Flux's KVS and populate the resource db
+ *
+ * \param rgen   resource generator
+ * \param h      flux handle
+ * \param db     graph database consisting of resource graph and various indices
+ * \return       0 on success; non-zero integer on an error
+ */
+int read_flux_hwloc (resource_generator_t rgen, flux_t *h, resource_graph_db_t &db)
```
dongahn: You probably want to pass `rgen` by reference. `resource_generator_t &rgen`?

resource/modules/resource_match.cpp
```diff
@@ -238,22 +245,103 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
...
+{
+    int rc = -1;
+    uint32_t rank = 0, size = 0;
+
+    if (flux_get_size (h, &size) == -1) {
+        fprintf(stderr, "%s: error with flux_get_size", __FUNCTION__);
```
dongahn: Since resource-match is a comms module, we will probably want flux log message instead of `fprintf`.

resource/modules/resource_match.cpp
```diff
@@ -238,22 +245,103 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
...
+        return -1;
+    }
+
+    ggv_t cluster_vertex = rgen.create_cluster_vertex (db);
+
+    for (rank=0; rank < size; rank++) {
+        char k[64];
+        int n = snprintf (k, sizeof (k), "resource.hwloc.xml.%" PRIu32 "", rank);
+        assert ((n >= 0) && ((unsigned int) n < sizeof (k)));
```
dongahn: Do we allow `assert` being called from within a comms module? At some point, I vaguely remember this wasn't allowed. Instead set an `errno` and try to return to the infrastructure. @grondo or @garlick?

resource/modules/resource_match.cpp
```diff
@@ -238,22 +245,103 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
...
+
 static int populate_resource_db (resource_ctx_t *ctx)
 {
     int rc = 0;
     resource_generator_t rgen;
-    if (ctx->args.grug != "none") {
+    // TODO: include rgen.err_message()
+    if (ctx->args.grug != "") {
+        if (ctx->args.hwloc_xml != "") {
+            cout << "WARN: multiple resource inputs provided, using grug" << endl;
```
dongahn: Please use `flux_log` instead.

resource/policies/base/matcher.cpp
```diff
@@ -22,6 +22,8 @@
  *  See also:  http://www.gnu.org/licenses/
 \*****************************************************************************/
 
+#include <algorithm>
```
dongahn: This was included in the last PR so needs to be removed. (`git rebase` will probably clear this up...)

dongahn:
@SteVwonder: OK. I finally managed to give my full review. This PR looks great with many needed functionalities! 

There are a few minor knits for which I added in-line comments above. Once those are addressed, I will look at it one last time before this can land. Please rebase though.

@garlick and @grondo: there is one question for you regarding a comms. module calling `assert`

Thanks.

SteVwonder:
Thanks @dongahn.  How would you prefer that I add the changes before your next review?  Is it easier for you to review a single additional commit on top of the existing commits?  Or should I immediately squash the changes into the existing commits?

dongahn:
For this  PR, squashing the changes into the existing commits should work for me. This should minimize your work anyway. For a much more complex PR, I understand why you want to take the former approach. Thanks!

resource/utilities/resource-query.cpp
```diff
@@ -179,7 +182,8 @@ static dfu_match_cb_t *create_match_cb (const string &policy)
 
 static void set_default_params (resource_context_t *ctx)
 {
-    ctx->params.grug = "conf/default";
+    ctx->params.grug = "";
```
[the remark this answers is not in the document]
SteVwonder: The intent was to keep the same behavior as before while supporting the new `hwloc_xml` option which is mutually exclusive with `grug`.  If the user sets both, I currently have it setup to log a warning and then fallback to use whatever is in `grug`.  If the code always sets `grug` to `conf/default`, then when the user sets `hwloc_xml` from the command line, both variables will be set and the code will always log a warning and use `grug`.  Also, by setting it to `""`, it is now possible to disambiguate between the user running `resource-query` and `resource --grug "conf/default"`.

It is good that you called this out though, because if the user sets neither, I currently have that configured as an error.  I will adjust it such that if the users provides neither, it defaults to `grug = conf/default` (restoring the previous behavior).

resource/modules/resource_match.cpp
SteVwonder: That is a great question.  [The docs say](https://jansson.readthedocs.io/en/2.11/apiref.html#c.json_incref) that `json_incref` is safe to call on NULL but do not explicitly address `json_decref`.  Looking directly [at the code](https://github.com/akheron/jansson/blob/71c4e8ec215afa225ac20eed269a14963cd37b50/src/jansson.h#L132) (yay for open source!) it appears to be safe:
```C
static JSON_INLINE
void json_decref(json_t *json)
{
    if(json && json->refcount != (size_t)-1 && JSON_INTERNAL_DECREF(json) == 0)
        json_delete(json);
}
```

resource/modules/resource_match.cpp
SteVwonder: Yeah, that is a great point.  I can switch this to an if statement with a `return -1`. Any suggestion on a relevant errno?  I am almost inclined to set errno to ENOMEM since I am fairly confident this snprintf couldn't fail due to rank being too large, and I think the only way this fails is ENOMEM.

resource/modules/resource_match.cpp
SteVwonder: Sorry for the confusion here.  The switch to flux_log was included in the refactor in https://github.com/flux-framework/flux-sched/pull/385/commits/d5d427debd82cdbf447e3a574bfbe2f030add7ec.

SteVwonder: I probably should have squashed that change down.

resource/policies/base/matcher.cpp
SteVwonder: :+1: Yeah, `git rebase` handled that automatically :tada:

SteVwonder:
@dongahn, I believe I addressed all of your comments.  I pushed them as a single commit on top of the existing commits, thinking that will be easier to review.  I can squash that down before your next pass, if you want, otherwise, I'll squash it down before this gets merged into master.

dongahn:
@SteVwonder: ok. I will get to this soonish...

resource/generators/gen.cpp
```diff
@@ -454,6 +457,243 @@ int resource_generator_t::read_graphml (const string &fn, resource_graph_db_t &d
...
+    db.roots[subsys] = v;
+
+    return v;
+}
+
+int resource_generator_t::read_ranked_hwloc_xml (const char *hwloc_xml, int rank,
+                                                 const ggv_t &root_vertex, resource_graph_db_t &db)
+{
+    if (check_hwloc_version() < 0) {
+        return -1;
```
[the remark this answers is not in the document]
SteVwonder: Added within the `check_hwloc_version` function.

resource/generators/gen.cpp
```diff
@@ -454,6 +457,243 @@ int resource_generator_t::read_graphml (const string &fn, resource_graph_db_t &d
...
+ *
+ * \param ifn    filename of hwloc xml
+ * \param db     graph database consisting of resource graph and various indices
+ * \return       0 on success; non-zero integer on an error
+ */
+int resource_generator_t::read_hwloc_xml_file (const char *ifn,
+                                               resource_graph_db_t &db)
+{
+    if (check_hwloc_version() < 0) {
+        return -1;
```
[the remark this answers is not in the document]
SteVwonder: Same as above.  Added within the function directly.

resource/utilities/resource-query.cpp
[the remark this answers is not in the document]
SteVwonder: I'm totally open to reverting this change if you have a suggestion for a different way to handle this.

resource/generators/gen.cpp
```diff
@@ -454,6 +457,243 @@ int resource_generator_t::read_graphml (const string &fn, resource_graph_db_t &d
...
+
+    hwloc_topology_t topo;
+    if ((hwloc_topology_init (&topo) != 0) ||
+        (hwloc_topology_set_flags (topo, HWLOC_TOPOLOGY_FLAG_IO_DEVICES) != 0) ||
+        (hwloc_topology_set_xmlbuffer (topo, hwloc_xml, len) != 0) ||
+        (hwloc_topology_load (topo) != 0))
+        {
+            hwloc_topology_destroy (topo);
+            // TODO: replace with inserts to m_err_msg
+            fprintf (stderr, "%s: Failed to load hwloc xml from rank %d\n",
```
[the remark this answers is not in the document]
SteVwonder: I opted to just do what the `TODO` says and insert to `m_err_msg`.  Does that work?

resource/generators/gen.cpp
```diff
@@ -454,6 +457,243 @@ int resource_generator_t::read_graphml (const string &fn, resource_graph_db_t &d
...
+    for (unsigned int i = 0; i < obj->arity; i++) {
+        walk_hwloc (obj->children[i], valid_ancestor, rank, db);
+    }
+}
+
+int check_hwloc_version() {
+    unsigned int hwloc_version = hwloc_get_api_version();
+
+    if ((hwloc_version >> 16) != (HWLOC_API_VERSION >> 16)) {
+        fprintf (stderr, "%s: Compiled for hwloc API 0x%x but running on library"
```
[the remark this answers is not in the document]
SteVwonder: Same as above. I just changed this to insert to `m_err_msg`.

SteVwonder:
Sorry! I apparently broke the build.  Probably a silly syntax error.  I'll try and fix later this evening, otherwise I do tomorrow.

resource/utilities/resource-query.cpp
```diff
@@ -489,6 +489,9 @@ int main (int argc, char *argv[])
 
     // Generate a resource graph data store
     resource_generator_t rgen;
+    if (ctx->params.grug == "" and ctx->params.hwloc_xml == "") {
```
dongahn: I didn't know you could you "and"! Had to google this :-). 

This is a valid change but may surprise folks who are not familiar with C++.

Since you will have to make some more changes to this PR. Could you actually change this to `&&`. I think this is less surprising IMHO...

dongahn:
@SteVwonder: OK. Changes look good. You can squash those changes and see if Travis CI will be happy. Once that's done, I will look through the code one last time before merging. Thanks.

resource/utilities/resource-query.cpp
SteVwonder: LOL. Sorry, that was unintended.  My python is leaking out.  I'll be sure to change it. Thanks!

dongahn:
@SteVwonder: LGTM. Thank you for much for finishing up this much anticipated PR! 

If others want to give their reviews, this is the time to tell me. 

Otherwise, I will merge 2PMish this afternoon.

Issue #390, opened 2018-09-20
Need FY-end performance analysis + tuning for resource


dongahn:
This is to finalize my ECP milestone work. 

1. Find the right Linux performance and memory tools to analyze `resource-query` and `resource` module.

2. Gather performance profiles with 4 configurations
      - Tiny machine configuration with 10K `match allocate_orelse_reserve` jobs
      - Medium machine configuration with 10K `match allocate_orelse_reserve` jobs
      - Large machine configuration with 10K `match allocate_orelse_reserve` jobs
      - Largest machine configuration with 10K `match allocate_orelse_reserve` jobs

3. Tuning exploration
     - Explore ways to fully integrate planner-based schemes into both `resource-query` and `resource` 
      - Document ways to reduce the memory footprint and edge traversal performance

dongahn:
This was an initial performance profiling on Mac OS/X, showing that one of the hot spot is edge traversal. 
<img width="1361" alt="28486440-ac973152-6e36-11e7-8453-6938eac10ee3" src="https://user-images.githubusercontent.com/862123/45789153-14049a00-bc33-11e8-9d6b-c24c59400639.png">

dongahn:
Doing such an analysis on Linux with `perf`. I find the output isn't all that intuitive so far. 

I used perf-record as:
$perf record --call-graph dwarf -e cpu-clock resource-query -G tiny.graphml < input
 
I used perf-report as:
perf report --stdio -g graph

![screen shot 2018-09-19 at 4 46 49 pm](https://user-images.githubusercontent.com/862123/45789181-30a0d200-bc33-11e8-85f0-a7765d78d2ce.png)

dongahn:
I gathered some feedback from RedHat and used https://github.com/jrfonseca/gprof2dot.

This tool doesn't seem very helpful. 

![screen shot 2018-09-19 at 4 56 33 pm](https://user-images.githubusercontent.com/862123/45789216-59c16280-bc33-11e8-9e1a-0df4cf97d518.png)

dongahn:
Some other work that falls into this category:

PR https://github.com/flux-framework/flux-sched/pull/349 - This is both correctness and performance, though.

SteVwonder:
[...]

PR #391, opened 2018-09-29
planner: Replace zhash_t to zhashx_t for higher performance

dongahn:
planner_new () can be called from a hot section
of the code and performance profilers discovered that
zhash_new () is way too slow.

With zhash_t, planner_new () accouted for 9.1% of the
graph data store populating. With zhashx_t, it
only accouts for 2.88% for the same operation.

See Issue #390.


SteVwonder:
Thanks @dongahn!  LGTM.  Any idea how much faster zlistx is than zlist?

PR #392, opened 2018-10-05
travis-ci: use docker for test builds

grondo:
I went ahead and copied over the docker scripts from flux-core to get flux-sched building with docker as well. Unfortunately, I wasn't clever enough to write the scripts so they could be exactly shared between different flux-* projects, so the copies here are modified slightly, but work generally the same.

The other difference is that the flux-sched Dockerfiles pull directly from `fluxrm/flux-core:${IMAGE}-latest`, i.e. from the docker images auto-deployed from the flux-core travis builds. If a branch or PR of flux-sched ever requires a specific version of flux-core, the Dockerfiles could be modified to instead pull for `fluxrm/flux-core:${IMAGE}-vx.y.z` to get the tagged flux-core instead.

I also copied over support for TAP_DRIVER_QUIET in `tap-driver.sh`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,98 +1,66 @@
+sudo: required
+services: docker
+dist: trusty
 language: c
-
-sudo: false
 
 matrix:
   include:
-    - compiler: gcc
-      env: LUA_VERSION=5.1
-    - compiler: clang
-      env: LUA_VERSION=5.1
-    - compiler: gcc
-      env: LUA_VERSION=5.1 COVERAGE=t
-    - compiler: gcc
-      env: LUA_VERSION=5.2 TEST_INSTALL=t
-    - compiler: clang
-      env: LUA_VERSION=5.2 CC=clang-3.8 CXX=clang++-3.8
-    - compiler: clang
-      env: LUA_VERSION=5.1 CPPCHECK=t
+    - name: "Ubuntu: gcc-8, distcheck"
+      compiler: gcc-8
+      env:
+       - CC=gcc-8
+       - CXX=g++-8
+       - DISTCHECK=t
+    - name: "Ubuntu: clang-6.0 chain-lint"
+      compiler: clang-6.0
+      env:
+       - CC=clang-6.0
+       - CXX=clang++-6.0
+       - chain_lint=t
+    - name: "Ubuntu: COVERAGE=t"
+      compiler: gcc
+      env:
+       - COVERAGE=t
+    - name: "Ubuntu: TEST_INSTALL docker-deploy"
+      compiler: gcc
+      env:
+       - TEST_INSTALL=t
+       - DOCKER_TAG=t
+    - name: "Centos 7: docker-deploy"
+      compiler: gcc
+      env:
+       - IMG=centos7
+       - DOCKER_TAG=t
+
+env:
+  global:
+  - TAP_DRIVER_QUIET=1
+  - DOCKERREPO=fluxrm/flux-sched
+  - DOCKER_USERNAME=travisflux
+
 cache:
   directories:
-    - $HOME/local
     - $HOME/.ccache
-    - $HOME/.local
-
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-      - llvm-toolchain-precise-3.8
-      - george-edison55-precise-backports # CMake 3.0
-    packages:
-      - cmake
-      - cmake-data
-      - clang-3.8
-      - gcc-8
-      - g++-8
-      - uuid-dev
-      - aspell
-      - libopenmpi-dev
-      - ccache
-      - apport # for coredumps
-      - gdb
-      - libmunge-dev
-      - munge
-      - lcov
-      - libyaml-cpp-dev
-      - libboost-all-dev
-      - libreadline6
-      - valgrind
-      - libevent-dev
 
 before_install:
-  - rm -rf $HOME/.luarocks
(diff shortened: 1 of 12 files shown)


src/test/docker/bionic/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM fluxrm/flux-core:bionic-base-latest
+
+ARG USER=flux
+ARG UID=1000
+
+# Install extra buildrequires for flux-sched:
+RUN sudo apt-get update
+RUN sudo apt-get -qq install -y --no-install-recommends \
+	libboost-graph-dev \
+	libboost-system-dev \
```
SteVwonder: Are all of these libboosts necessary?  I thought it was just libboost-graph that was required by flux-sched?

src/test/docker/centos7/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ARG UID=1000
+
+# Install extra buildrequires for flux-sched:
+RUN sudo yum -y install  \
+	libboost-graph-devel \
+	libboost-system-devel \
+	libboost-filesystem-devel \
+	libboost-regex-devel \
+        libxml2-devel \
+	readline-devel \
```
SteVwonder: I presume this is to get rid of warnings from system tools?  If so, do we want to put this in the flux-core Dockerfile?

t/t2002-easy.t
```diff
@@ -48,7 +48,7 @@ test_expect_success 'jobs scheduled in correct order' '
 '
 
 test_expect_success 'sim: unloaded' '
-    flux module remove sched
+    flux module remove sched &&
```
SteVwonder: Whoops! Great catch!  I guess flux-sched wasn't using `chain-lint`?

src/test/docker/bionic/Dockerfile
grondo: I iterated to get to the smallest set of libboost packages required (libboost-all is just way too much). The Boost checks in flux-sched's configure actually check for a couple of these packages, and I think one might be a dependency of the others (have to check configure again to be certain)

src/test/docker/centos7/Dockerfile
grondo: No, this is a flux-sched's build requirement.. I didn't narrow down what is using it, but libreadline is required by configure.

t/t2002-easy.t
grondo: Yeah I guess it wasn't...

src/test/docker/bionic/Dockerfile
SteVwonder: Gotcha. Makes sense. Thanks!

src/test/docker/centos7/Dockerfile
SteVwonder: Huh.  That's surprising, but I don't think its worth any additional effort to track it down.  Thanks!

src/test/docker/bionic/Dockerfile
grondo: Yeah, configure.ac currently has literally:
```
AX_BOOST_BASE([1.53.0], [], [AC_MSG_ERROR([Please install boost >= 1.53)])])

AX_BOOST_SYSTEM

AX_BOOST_FILESYSTEM

AX_BOOST_GRAPH

AX_BOOST_REGEX

```

I'm not exactly sure why, perhaps it could be cleaned up?

dongahn: As I remember those were needed for resource. I probably need lots of clean up for boost checks including its ridiculously silly messages.

BTW readline is needed for resource-query commandline.

.travis.yml
```diff
@@ -1,98 +1,67 @@
...
+    - name: "Centos 7: docker-deploy"
+      compiler: gcc
+      env:
+       - IMG=centos7
+       - DOCKER_TAG=t
+
+env:
+  global:
+  - TAP_DRIVER_QUIET=1
+  - DOCERREPO=fluxrm/flux-sched
```
SteVwonder: `DOCERREPO` -> `DOCKERREPO`

grondo: Oh, good catch! That would have been unfortunate.

grondo:
I added another couple fixes to this PR:

 - incorporate minor recent updates to `t/sharness.d/flux-sharness.sh` from flux-core
 - add timeout to the valgrind test as was done in flux-core (I think one builder hung in valgrind on my latest push, the timeout will help with debug if it happens again.)

dongahn:
@grondo and @SteVwonder: Is this PR ready to go in. Do you want me to review this also?

SteVwonder:
@grondo, do you mind adding to this PR [the fix for the `--interactive` flag](https://github.com/flux-framework/flux-core/commit/c3b2affe57eb16413bf8856935fdfbeb660d5fa6) that was merged into core?

I'm also fine if we merge this PR as-is and then follow it up with the fix.

grondo:
@SteVwonder, I've pulled in your two commits related to the docker test script into this PR, which includes the --interactive fix and the addition of CAP_SYS_PTRACE.

dongahn:
Once that turns green, I will merge this in and then rebase #394 on top of that.

grondo:
@dongahn, code coverage drop is a known (but not yet investigated) side effect of moving to docker.

dongahn:
Thanks @grondo!

Issue #397, opened 2018-10-19
[armv7l]  error: cast from pointer to integer of different size


garlick:
Encountered this buiuld error on armv7l:
```
  CC       libplanner_la-planner_multi.lo
planner_multi.c: In function ‘planner_multi_rem_span’:
planner_multi.c:414:49: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
         if (planner_rem_span (ctx->planners[i], (int64_t)s) == -1)
                                                 ^
planner_multi.c: In function ‘planner_multi_span_first’:
planner_multi.c:436:10: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
     rc = (int64_t)span;
          ^
planner_multi.c: In function ‘planner_multi_span_next’:
planner_multi.c:453:10: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
     rc = (int64_t)span;
          ^
cc1: all warnings being treated as errors
```

dongahn:
I will have to look at #395 and submit a PR next. So as part of that, I will also check-pick these changes @garlick. Thank you for testing this on armv7l!

garlick:
Sounds good - I'll just leave it there for you to cherry-pick.

dongahn:
@garlick: is armv7l a 32bit system?

Issue #395, opened 2018-10-19
3 resource-matching service test cases fail on LLNL Sierra systems


dongahn:
```
FAIL: t4001-match-allocate.t 2 - match-allocate works with a 1-node, 1-socket jobspec
FAIL: t4001-match-allocate.t 3 - match-allocate fails when all resources are allocated
PASS: t4001-match-allocate.t 4 - detecting of a non-existent jobspec file works
PASS: t4001-match-allocate.t 5 - handling of a malformed jobspec works
FAIL: t4001-match-allocate.t 6 - removing resource works
```

dongahn:
Looks like the `resource` module crashes in the jobspec library with a `free` error on Sierra.

```
Stack Trace
     raise,                                                                                                                                                                      FP=200002238110
     abort,                                                                                                                                                                      FP=200002238260
     __libc_message,                                                                                                                                                             FP=200002238af0
     free,                                                                                                                                                                       FP=200002238c00
     _ZdlPv,                                                                                                                                                                     FP=200002238c20
C++  <nameless>,                                                                                                                                                                 FP=200002238c70
C++  <nameless>,                                                                                                                                                                 FP=200002238cd0
C++  ...td::__detail::_Mod_range_hashing,std::__detail::_Default_ranged_hash,std::__detail::_Prime_rehash_policy,std::__detail::_Hashtable_traits<true,false,true> >::not-in-charge ~_Hashtable, FP=200002238d10
C++  std::unordered_map<std::string,long,std::hash<std::string>,std::equal_to<std::string>,std::allocator<std::pair<const std::string,long> > >::not-in-charge ~unordered_map,   FP=200002238d50
C++  Flux::Jobspec::Resource::not-in-charge ~Resource,                                                                                                                           FP=200002238d90
C++  $unnamed_namespace::parse_yaml_resources,                                                                                                                                   FP=200002238f90
C++  Flux::Jobspec::Resource::Resource,                                                                                                                                          FP=200002239c20
C++  $unnamed_namespace::parse_yaml_resources,                                                                                                                                   FP=200002239e20
C++  Flux::Jobspec::Resource::Resource,                                                                                                                                          FP=20000223aab0
C++  $unnamed_namespace::parse_yaml_resources,                                                                                                                                   FP=20000223acb0
C++  Flux::Jobspec::Resource::Resource,                                                                                                                                          FP=20000223b940
[...]

PR #398, opened 2018-10-24
Integrate jobspec into resource, etc

### LINKED ISSUE #393: libjobspec: temporarily copy from flux-core
As described in flux-framework/flux-core#1714, libjobspec will be rewritten in C.  Meanwhile the C++ interface is being used by sched.  To simplify the transition, it was suggested that libjobspec could be copied into sched.  Then when a new libjobspec from flux-core is available that can provide the interfaces needed by sched, the copy can be dropped and the interfaces changed in one go.

dongahn:
This PR integrates jobspec internally into the resource infrastructure:

1. Tightly managing `libjobspec` from within `flux-sched` solves odd make-check errors that we are seeing on some platforms (e.g., Sierra). The issues seem to arise due to C++ compiler version mismatches.

2. `flux-core` will transition away from `libjobspec` anyway in a near future, and doing this now will help `flux-core` make this transition without having to worry about any breakage in `flux-sched`.

This PR also has a fix for pointer to 64 integer conversion for 32bit systems.

Resolve #393, #395 and #397.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,13 +63,13 @@ AC_DEFINE_UNQUOTED([LIBLUA_SO], ["$LIBLUA_SONAME"],
 PKG_CHECK_MODULES([UUID], [uuid], [], [])
 AX_FLUX_CORE
 
+X_AC_YAMLCPP
 AX_BOOST_BASE([1.53.0], [], [AC_MSG_ERROR([Please use boost == 1.53 or > 1.58])])
 AX_CHECK_BUGGY_BOOST
 AX_BOOST_SYSTEM
 AX_BOOST_FILESYSTEM
 AX_BOOST_GRAPH
 AX_BOOST_REGEX
-PKG_CHECK_MODULES([JOBSPEC], [flux-jobspec], [], [])
 PKG_CHECK_MODULES([IDSET], [flux-idset], [], [])
 AC_CHECK_LIB([readline], [readline],
              [AC_SUBST([READLINE_LIBS], ["-lreadline"])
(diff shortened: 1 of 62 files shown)


garlick:
I think you forgot to add `jobspec.hpp` to `libjobspec_conv_la_SOURCES` and that's why distcheck failed in travis.

Does `configure --disable-jobspec` result in a working sched?  If jobspec is required, that option should go away.

dongahn:
Yes, the jobspec is required. I will look at the new m4 file more closely. Thanks.

garlick:
On the coverage miss, actually you may want to pull in the tests too.

`src/cmd/flux-jobspec-validate` is a jobspec validator program written in c++.

`t/jobspec` contains various inputs including invalid jobspec, and jobspec from our RFC examples.

`t/t0018-jobspec.t` runs the inputs through the validator.

garlick:
(I would probably change `flux-jobspec-validate` to a test program rather than an installed command)

grondo:
>  have to guess that the code coverage issue is the same side effect that @grondo's docker PR had?

No, this is because jobspec.cpp is mostly not covered by tests. We may also need to pull in the `flux jobspec-validate` command and associated sharness test (`t0018-jobspec.t`) to get good coverage.

dongahn:
Ugg... OK that makes sense. I will pull those in as well then.

dongahn:
OK. code coverage looks much better now. Let me know if there is anything else.

garlick:
Looks good.  I'll go ahead and press the button (hope it's OK with sched people!)

Issue #399, opened 2018-10-31
Wire in --prune-filter option for resource-query and resource matching service


dongahn:
Currently, both resource-query and matching service only consider subtree `core` aggregates to prune graph search. We want to allow the users of resource-query and resource matching service to program which subtree resources they want to consider for pruning at what high level resources.

I had a placeholder option called `--prune-filter` in `resource-query` and I need to wire this into both `resource-query` and `resource` matching service. https://github.com/flux-framework/flux-sched/pull/349#issuecomment-402922496

Issue #400, opened 2018-11-01
Ensure correctness without pruning filters


dongahn:
Related #399. I found some of the test cases fails with `make check` fails when there is no pruning filter installed. Because pruning filters should work as a "cache", the scheduling results should be the same with or without the filters.

```
FAIL: t3001-resource-basic.t 3 - match allocate_orelse_reserve 10 jobspecs (pol=hi)
FAIL: t3001-resource-basic.t 5 - match allocate_orelse_reserve 100 jobspecs instead (pol=hi)
FAIL: t3001-resource-basic.t 10 - attempt to match allocate_orelse_reserve 10 jobspecs (pol=low)
FAIL: t3001-resource-basic.t 12 - match allocate_orelse_reserve 100 jobspecs instead (pol=low)
FAIL: t3005-resource-rsv.t 1 - allocate or reserve 17 jobs (pol=hi)
FAIL: t3005-resource-rsv.t 2 - allocate or reserve 17 jobs (pol=low)
FAIL: t3007-resource-iobw.t 1 - match allocate with pfs IO BW (pol=hi)
FAIL: t3007-resource-iobw.t 2 - match allocate with pfs IO BW (pol=low)
FAIL: t3008-resource-cancel.t 1 - allocate or reserve 17 jobs, cancel 3 (pol=hi)
FAIL: t3008-resource-cancel.t 2 - allocate or reserve 17 jobs, cancel all (pol=hi)
FAIL: t3008-resource-cancel.t 3 - allocate or reserve 17 jobs, cancel 3 (pol=low)
FAIL: t3008-resource-cancel.t 4 - allocate or reserve 17 jobs, cancel all (pol=low)
```

dongahn:
As part of hwloc integration, @SteVwonder found that when an intermediate vertex in the resource graph doesn't have a prune filter planner, matching fails at https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl.cpp#L158. This corner case needs to be taken care as part of this issue as well.

Issue #388, opened 2018-09-02
resource module needs load option test


dongahn:
Per discussion at PR #386

dongahn:
At this point we have a plenty of tests that cover various load options. Closing.

PR #401, opened 2018-11-02
resource: wire in --prune-filters option for resource-query and matching module

dongahn:
This PR resolves Issue #399. 

- Clean up how to specify pruning filters through either a command line argument or load option
- Add methods to parse a spec and populate filter maps within `matcher_util_api_t`
- Use these methods within the traverser to implant filters selectively
- Wire in this option within both resource-query and resource matching module
- Default set to `ALL:core` for both CLI and module for now.

Due to time constraints, I plan to address #400 in a separate PR.


dongahn:
Coverage slightly decreased. When I can get to #400 and #388, I plan add more tests to coverage more code.

dongahn:
Yes. Thanks.

Issue #402, opened 2018-11-07
Need to add libtoolize to autogen.sh


SteVwonder:
From @tristanvdb:
> First, I am using:
```
automake (GNU automake) 1.14
autoreconf (GNU Autoconf) 2.69
libtool (GNU libtool) 2.4
```

> The error was: lots of messages like this one:
```
rdl/Makefile.am:30: error: Libtool library used but 'LIBTOOL' is undefined
rdl/Makefile.am:30:   The usual way to define 'LIBTOOL' is to add 'LT_INIT'
rdl/Makefile.am:30:   to 'configure.ac' and run 'aclocal' and 'autoconf' again.
rdl/Makefile.am:30:   If 'LT_INIT' is in 'configure.ac', make sure
rdl/Makefile.am:30:   its definition is in aclocal's search path.
```
> I fixed it by running libtoolize at the root of the repo. You should consider using the autogen script from flux-core which contains a call to libtoolize.

SteVwonder:
Fixed by #406. Closing.

PR #406, opened 2018-11-08
config: Add libtoolize into autogen.sh

dongahn:
This PR uses `libtoolize` to resolve Issue #402. It also pulls in some of the options that `flux-core` is using for consistency. 

The autoreconf verbose option catches `autoreconf: configure.ac: not using Gettext`. But I don't think we are ready to add Gettext support so I will leave it this way.


SteVwonder:
Thanks @dongahn!

PR #419, opened 2018-12-18
Ensure scheduling correctness with different prune filter configurations

dongahn:
This PR add various fixes to ensure correctness around the use of pruning filters.

- Use at least one pruning filter type to ensure the schedule method within dfu.cpp can traverse resource state changing scheduled points for all cases.

- Enable the pruning filter planner, even if the specified filter types are not present at the subtree of the anchor type. This ensures correctness when such a condition ever arises.

- Add test cases using previous failed test codes

- Add test cases to test various module load options for the resource match service

- Other misc. changes

Resolve Issue #400
Resolve Issue #388

### DIFF
--- a/resource/policies/base/matcher.cpp
+++ b/resource/policies/base/matcher.cpp
@@ -220,8 +220,27 @@ void matcher_util_api_t::set_pruning_type (const std::string &subsystem,
                                            const std::string &anchor_type,
                                            const std::string &prune_type)
 {
-    // Note the final container is "set" so it will only allow uniqe prune_types
-    m_pruning_types[subsystem][anchor_type].insert (prune_type);
+    // Use operator[] to create an entry if subsystem key doesn't exist
+    auto &s = m_pruning_types[subsystem];
+    if (anchor_type == ANY_RESOURCE_TYPE) {
+        // Check whether you have already installed prune_type.
+        // If so, remove it as you want to install it against ANY_RESOURCE_TYPE.
+        for (auto &kv : s)
+            kv.second.erase (prune_type);
+        // final container is "set" so it will only allow uniqe prune_types
+        s[anchor_type].insert (prune_type);
+    } else {
+        if ((s.find (ANY_RESOURCE_TYPE) != s.end ())) {
+            auto &prune_set = s[ANY_RESOURCE_TYPE];
+            if (prune_set.find (prune_type) == prune_set.end ()) {
+                // If prune_type does not exist against ANY_RESOURCE_TYPE
+                // Install it against anchor_type, an individual resource type
+                s[anchor_type].insert (prune_type);
+            } // orelse NOOP
+        } else {
+            s[anchor_type].insert (prune_type);
+        }
+    }
     m_total_set[subsystem].insert (prune_type);
 }
 
--- a/resource/policies/base/matcher.cpp
+++ b/resource/policies/base/matcher.cpp
@@ -273,7 +273,8 @@ bool matcher_util_api_t::get_my_pruning_types (const std::string &subsystem,
             for (auto &k : m)
                 out.push_back (k);
         }
-        if (s.find (ANY_RESOURCE_TYPE) != s.end ()) {
+        if (anchor_type != ANY_RESOURCE_TYPE
+            && s.find (ANY_RESOURCE_TYPE) != s.end ()) {
             auto &m = s.at (ANY_RESOURCE_TYPE);
             for (auto &k : m)
                 out.push_back (k);
--- a/resource/policies/base/matcher.cpp
+++ b/resource/policies/base/matcher.cpp
@@ -220,6 +220,7 @@ void matcher_util_api_t::set_pruning_type (const std::string &subsystem,
                                            const std::string &anchor_type,
                                            const std::string &prune_type)
 {
+    // Note the final container is "set" so it will only allow uniqe prune_types
     m_pruning_types[subsystem][anchor_type].insert (prune_type);
     m_total_set[subsystem].insert (prune_type);
 }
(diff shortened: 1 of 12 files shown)


dongahn:
@SteVwonder: I had a few more things I wanted to address. But before I will be out until the end of this year, I thought I wanted to get this much work to be reviewed and possibly merged. Thanks.

dongahn:
@SteVwonder: I have added all of the fixes that I had originally planned, and this PR is ready to be reviewed.

dongahn:
Ok. Rebased and forced a push for PR #427.

resource/policies/base/matcher.cpp
```diff
@@ -257,6 +258,32 @@ bool matcher_util_api_t::is_pruning_type (const std::string &subsystem,
...
+        // Get the value of the subsystem, which is a map
+        // of <string, set> type.
+        auto &s = m_pruning_types.at (subsystem);
+        if (s.find (anchor_type) != s.end ()) {
+            // Get the value of the anchor map, which is a set
+            auto &m = s.at (anchor_type);
+            for (auto &k : m)
+                out.push_back (k);
+        }
+        if (s.find (ANY_RESOURCE_TYPE) != s.end ()) {
```
SteVwonder: If the user specifies `core:pu` on the command line, then the prune-filters list will be `ALL:pu, core:pu`.  In that case, I believe this will produce a second `pu` entry in `out` (the current sets only prevent duplicates with the same prefix, AFAICT).  I don't think the implications of that are terrible (I believe it is just a duplicate entry in the planner), but this could be eliminated by making `out` a set. I don't think will affect the calling function `prime_pruning_filter`, since the values are iterated over and then inserted into a vector there.

resource/traversers/dfu_impl.cpp
```diff
@@ -945,21 +946,26 @@ int dfu_impl_t::prime_pruning_filter (const subsystem_t &s, vtx_t u,
...
+    for (auto &aggr : dfv)
         accum_if (s, aggr.first, aggr.second, to_parent);
-        /* If the aggregate type is "my" filter type, track them in my filter */
-        if (m_match->is_my_pruning_type (s, (*m_graph)[u].type, aggr.first)) {
-            types.push_back (strdup (aggr.first.c_str ()));
-            avail.push_back (aggr.second);
+
+    if (m_match->get_my_pruning_types (s, (*m_graph)[u].type, out_prune_types)) {
+        for (auto &type : out_prune_types) {
+            types.push_back (strdup (type.c_str ()));
```
SteVwonder: Is the `strdup` here necessary?  These values eventually get passed to `planner_multi_new` (through `subtree_plan`) which does an `xstrdup` of its own.

resource/traversers/dfu_impl.cpp
```diff
@@ -945,21 +946,26 @@ int dfu_impl_t::prime_pruning_filter (const subsystem_t &s, vtx_t u,
...
+    std::vector<std::string> out_prune_types;
 
     (*m_graph)[u].idata.colors[s] = m_color.gray ();
     accum_if (s, type, (*m_graph)[u].size, to_parent);
     if (prime_exp (s, u, dfv) != 0)
         goto done;
 
-    for (auto &aggr : dfv) {
-        /* If the aggregate type is any filter type, accume for the parent */
+    for (auto &aggr : dfv)
```
SteVwonder: Is this loop necessary?  If I'm interpreting this correctly, at the end of this loop, `to_parent` should have the exact same contents as `dfv` plus the insertion of this resource at line 952.  In other words, I believe this loop is just copying/merging the contents of `dfv` into `to_parent`. If so, couldn't you just pass `to_parent` into `prime_exp` and eliminate this extra iteration/copying?

dongahn:
@SteVwonder: Thank you for thorough review!

> Related: in #385, I change the default prune filter for resource-match to ALL:pu but didn't change resource-query. So resource-query and resource-match no longer have the same default prune filters (ALL:core and ALL:pu, respectively).

This is actually intended. We use `resource-query` on GRUG whose test cases don't have `pu`, I wanted to use `core` as the filter type. Plus, I thought having `pu` on the resource matching service and `core` on the CLI could increase our testing coverage. 

>  Since resource-query still uses ALL:core, several of the tests in t3011-resource-filt.t are actually testing the same pruning filter set. 

Also intended. Wanted to make sure specifying the same filter type as the default leads to correct scheduling behavior.

resource/traversers/dfu_impl.cpp
dongahn: I suspect this will lead of incorrect behavior. If `to_parent` is directly passed to each subtree walk, the net effect will be that`to_parent` will keep accumulating aggregates. 

Say, you have a tree with `cluster->rack[2]->node[2]->core[2].` Then, by the time you are post-visiting the second rack vertex, `to_parent` will have not only the aggregates of the visiting rack's own subtree but also those aggregates that were accumulated from the visit of the first rack.

I think this will then make rack2's prune filter have more subtree resources than it has under the subtree, leading to incorrect behavior (or at least more walks than necessary)...

resource/policies/base/matcher.cpp
dongahn: This is a good point. I will take a close look at shortly what happens if the same prune filter type is specified with ALL and individual anchor resource. I don't think I have tested enough for individual anchor-based filter implanting.

dongahn:
@SteVwonder: I will try to get to some of your review points maybe tomorrow. Let me know if you disagree with anything I commented above. Thanks.

SteVwonder:
LGTM! Thanks @dongahn!  Let me know whether or not you want to squash any commits.  If not, I'm happy to merge as is, otherwise, I will wait for your signal to merge.

dongahn:
Thanks. @StevWonder. I don't think we need to squash. Once this is merged, we should have all the changes needed for our next release!

dongahn:
Thanks!

Issue #420, opened 2018-12-23
build: should libflux-rdl be installed


garlick:
I noticed flux-sched is still installing `libflux-rdl.so`.  Should that library be made into a noinst to discourage use by, say, a project that builds a sched plugin but is not part of the flux-sched project?  My understanding is that is is being deprecated.

dongahn:
Make sense. I will put this in as part of PR #419 as I will be adding a few more fixes.

dongahn:
Hmmm... I changed `rdl/Makefile.am`:

```diff
diff --git a/rdl/Makefile.am b/rdl/Makefile.am
index 6fa02b7..ebb1ab4 100644
--- a/rdl/Makefile.am
+++ b/rdl/Makefile.am
@@ -26,10 +26,9 @@ nobase_dist_lua_SCRIPTS = \
        RDL/uri.lua \
        RDL/uuid.lua

-noinst_LTLIBRARIES = libjansson-lua.la
-lib_LTLIBRARIES = libflux-rdl.la
-
-fluxschedinclude_HEADERS = rdl.h
+noinst_LTLIBRARIES = libjansson-lua.la \
+                     libflux-rdl.la
+noinst_HEADERS = rdl.h

 libjansson_lua_la_SOURCES = jansson-lua.c jansson-lua.h
 libjansson_lua_la_LIBADD = $(LUA_LIB) $(JANSSON_LIBS)
@@ -43,7 +42,7 @@ libflux_rdl_la_LDFLAGS = $(fluxlib_ldflags) \
     -Wl,--version-script=$(srcdir)/rdl_version.map \
     $(AM_LDFLAGS)

-fluxcmd_PROGRAMS = flux-rdltool
+noinst_PROGRAMS = flux-rdltool
 flux_rdltool_SOURCES = flux-rdltool.c
 flux_rdltool_CFLAGS = $(AM_CFLAGS)
 flux_rdltool_LDADD = libflux-rdl.la $(LUA_LIB) $(JANSSON_LIBS)
```

But somehow I'm getting `multiple definition` errors:
```
make[2]: Entering directory `/usr/WS1/dahn/workspace/flux-sched/build.toss_3_x86_64_ib/simulator'
[...]

PR #421, opened 2019-01-04
build: do not install libflux-rdl.so, fix rebuild of aclocal.m4 on first make

grondo:
This PR turns `libflux-rdl.so` into a convenience library instead of an installed library as discussed in #420. It was simpler to keep the other requisite convenience libs (libutil, liblsd) as part of `libflux-rdl.la` and remove them when duplicated in `LIBADD`, than to leave them out of `libflux-rdl.la` and then add them by hand everywhere `libflux-rdl.la` is used. However, I do have a working implementation of the other way if that is preferred.

I also got annoyed and fixed the issue where `make` thought it had to rebuild `$(srcdir)/aclocal.m4` after every `autogen.sh && ./configure` (this was because `autogen.sh` was moving it to where it couldn't be found)


dongahn:
> This could be because the AUX_DIR and CONFIG_DIR are different in flux-sched (m4 and config
respectively).

Odd... Do you know why sched ends up using both `AUX_DIR` and `CONFIG_DIR`? It seems keeping all m4 scripts in `config` might make thing more readable and consistent with `flux-core.`

dongahn:
@grondo: LGTM! Thanks.

dongahn:
If that is not that time consuming, I think it makes sense just fixing it as part of this PR. Thanks.

grondo:
Ok, I've pushed a commit that moves the contents of `m4` into `config` to more closely match existing practice in `flux-core`

dongahn:
@grondo: Great! I will take a look at the PR after my meeting at 10AM and merge if no other changes are required. Thanks.

dongahn:
@grondo: m4 --> config looks great to me. One thing though, with this move, do we want to revert the https://github.com/flux-framework/flux-sched/pull/421/commits/084b433b4fdb9bee9fcccdbc163d724b58fbdb82 change so that `flux-sched`'s `augogen.sh` becomes more consistent with `flux-core`?

dongahn:
Thanks.

dongahn:
Thanks @grondo. Yes, `AM_MAINTAINER_MODE([enable])` is really a good addition. I vaguely remember I had to add this to avoid some timestamp mishap.

dongahn:
@grondo: Hmmm, for some reason I'm still seeing your `autogen.sh` change in 860b063...

grondo:
> @grondo: Hmmm, for some reason I'm still seeing your autogen.sh change in 860b063...

You mean that commit is still in this branch? I did force push, so you might have an old copy.

If you mean that the move of `aclocal.m4` is still removed from `autogen.sh`, then that is intentional because it was the root cause of the problem where `make` would rerun `configure` the first time after `autogen.sh` (`aclocal.m4` needs to be at the top level)

This will make `autogen.sh` from flux-core and flux-sched match after flux-framework/flux-core#1912 is merged.

dongahn:
> If you mean that the move of aclocal.m4 is still removed from autogen.sh, then that is intentional because it was the root cause of the problem where make would rerun configure the first time after autogen.sh (aclocal.m4 needs to be at the top level)

Ah I see. I somehow mistakenly thought that `aclocal.m4` being moved to `config` directory is okay when you don't have both `AUX_DIR` and `CONFIG_DIR`, not leading to a redundant configure rerun . I will merge as is then.

grondo:
>  I somehow mistakenly thought that aclocal.m4 being moved to config directory is okay when you don't have both AUX_DIR and CONFIG_DIR, not leading to a redundant configure rerun 

I thought that at first as well. However, if you look at `aclocal.m4` location in generated Makefiles, you find it should be in top_srcdir:

```
grondo@peep:~/git/flux-sched$ grep aclocal\\.m4 Makefile
ACLOCAL_M4 = $(top_srcdir)/aclocal.m4
```

I'm not sure why `-I config` doesn't change that, but I didn't want to fight it too much. It was easier to just leave `aclocal.m4` where `./autoreconf` thought it should go...

dongahn:
Agreed.

PR #428, opened 2019-01-15
docker: add yaml-cpp dependency

garlick:
In flux-framework/flux-core#1922, we are dropping yaml-cpp as a dependency, hence dropping it from the flux-core Dockerfiles.

This PR adds it to the sched Dockerfiles.

### DIFF
--- a/src/test/docker/bionic/Dockerfile
+++ b/src/test/docker/bionic/Dockerfile
@@ -10,7 +10,7 @@ RUN sudo apt-get -qq install -y --no-install-recommends \
 	libboost-system-dev \
 	libboost-filesystem-dev \
 	libboost-regex-dev \
-        libxml2-dev \
+	libxml2-dev \
 	python-yaml \
 	libyaml-cpp-dev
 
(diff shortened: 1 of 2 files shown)


src/test/docker/bionic/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN sudo apt-get -qq install -y --no-install-recommends \
 	libboost-filesystem-dev \
 	libboost-regex-dev \
         libxml2-dev \
-	python-yaml
+	python-yaml \
```
dongahn: Looks good to me. 

Using my web browser, I see an extra leading whitespace for the line before this (`libxml2-dev`). While you are on it, can we get rid of it.

Also, I just noticed our main README.md fails to list `python-yaml` and `libyaml-cpp-dev` as our dependencies. Do you think it makes sense to update that REAME.md as part of this PR as well?

garlick: Will address both - thanks!

dongahn:
@garlick: LGTM. Will merge once Travis turns green. Thanks.

PR #429, opened 2019-01-21
build: add release notes for 0.7.0 and update README.md

SteVwonder:


SteVwonder:
Thought just occurred to me.  Should we be listing the version of flux-core that each flux-sched release is compatible with in the README.md? i.e.: `flux-core==0.11.0`.

grondo:
> Thought just occurred to me. Should we be listing the version of flux-core that each flux-sched release is compatible with in the README.md?

That's a good point, though it might get challenging (for now) to keep it up to date on master, for example if flux-sched actually requires a version greater than the latest tagged version of flux-core.

At the very least, we should add a version check to `AX_FLUX_CORE` so that versions of flux-core that are definitely too old are called out during configure.

SteVwonder:
> though it might get challenging (for now) to keep it up to date on master

That's a good point.  If we go the route of making forks for releases, then those READMEs could have the exact versions.

> add a version check to AX_FLUX_CORE so that versions of flux-core that are definitely too old are called out during configure.

That's a good idea.  Should we add that to this release or make an issue and push it off to 0.8.0? (lazy me votes for the latter 😆 )

dongahn:
> That's a good idea. Should we add that to this release or make an issue and push it off to 0.8.0? (lazy me votes for the latter 😆 )

I vote for the next release as well.

dongahn:
And I like @grondo's `AX_FLUX_CORE` idea. 

@SteVwonder: LGTM. I'm merging this in. I really like your script that autogenerate news item! It may be due to my local working copy but I don't understand why we don't see the PR #419 though. 

We can manually add

```
resource: ensure scheduling correctness with different prune filter configurations (#400, #388)
```
though. 

```
Changes since v0.6.0:
 * docker: add yaml-cpp dependency (#428)
 * docker: update repo tag names, ensure flux-sched installed with
   prefix=/usr (#427)
 * travis-ci: fix docker deploy better (#426)
 * travis: fix docker and github release deployment (#424)
 * build: do not install libflux-rdl.so, fix rebuild of aclocal.m4 on
   first make (#421)
 * resource: Add run-level support for resource matching module (#418)
 * resource: support for hwloc ingestion (#385)
 * Compilation fixes (#417)
 * sched/plugin: track flux-core module API changes (#414)
 * config: Add libtoolize into autogen.sh (#406)
 * config: Fix a non-portable use of schell conditionals for automake
   (#405)
 * resource: wire in --prune-filters option for resource-query and
   matching module (#401)
 * Integrate jobspec into resource, etc (#398)
 * Add better autoconf support for libboost (#394)
 * travis-ci: use docker for test builds (#392)
 * planner: Replace zhash_t to zhashx_t for higher performance (#391)
 * Add support for resource matching service (#386)
 * sched: remove trailing whitespaces (#376)
```

grondo:
> It may be due to my local working copy but I don't understand why we don't see the PR #419 though.

The relnotes script processes PRs by listing merges with `git log --merges`, so make sure your working tree is up to date when running the script. FWIW, I see #419 as the first entry output in the script:

```console
grondo@ipa15:~/git/flux-sched.git$ src/test/relnotes.sh  | head -4
Changes since v0.6.0:
 * Ensure scheduling correctness with different prune filter configurations
   (#419)
 * docker: add yaml-cpp dependency (#428)
```

grondo:
> I verified that it was included in this PR here

Great! So is this ready to merge so we can tag flux-sched v0.7.0? :wink:

SteVwonder:
Thanks!

I'll push the annotated tag shortly!

grondo:
> EDIT: Yes, it does. Found it: https://travis-ci.org/flux-framework/flux-sched/builds/483103242

Yeah, hopefully it works! We should hopefully also get `flux-sched:v0.7.0` docker image.

grondo:
Ok, worked: https://github.com/flux-framework/flux-sched/releases

Issue #433, opened 2019-01-30
flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name


grondo:
On a system with the latest flux-core (v0.11.0) and flux-sched (v0.7.0) rpms installed, I'm seeing the following error when using `flux-start`:
```console
$ flux version
commands:    		0.11.0
libflux-core:		0.11.0
$ rpm -q flux-sched
flux-sched-0.7.0-1.ch6.x86_64
$ flux start
2019-01-30T22:07:54.975297Z broker.err[0]: rc1: flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name
2019-01-30T22:07:55.075864Z broker.err[0]: rc1: flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name
2019-01-30T22:07:55.212067Z broker.err[0]: rc1: flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name
```

This is triggered by `flux module load -r 0 resource`, though we're still investigating why:

```
$ flux module load -r 0 resource
flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name
```

grondo:
> This is triggered by flux module load -r 0 resource

That is just due to readdir order, loading any module that comes after `priority_mod_fair_tree.so` in directory traversal will generate this error from the flux-module command.

The problem is that `priority_mod_fair_tree.so` is installed a a flux "extension module" but is not a properly formed extension module. (no `mod_name` symbol)

For now I will generate a `flux-sched-0.7.0-2` RPM that removes this module from the package.

grondo:
Hmm, also see #344. I wonder why this wasn't an issue with previous RPMs?

dongahn:
Hmmm. I don't understand why 

```
$ flux module load -r 0 resource
```

leads to 
```
flux-module: /usr/lib64/flux/modules/sched/priority_mod_fair_tree.so: undefined symbol: mod_name
```

When `resource` shouldn't depend on this module at all...

grondo:
[...]

PR #434, opened 2019-01-31
Fix build issues for priority_mod_fair_tree.so

grondo:
As discussed in #433, this PR fixes a couple build issues for the modified fair tree plugin.


dongahn:
LGTM, @grondo. Thanks!

PR #440, opened 2019-02-06
Sync flux-sched with removed interfaces in flux-core

### LINKED ISSUE #437: remove all code that uses wreck and other removed flux-core interfaces
There is a PR pending in flux-core: flux-framework/flux-core#1988 which will remove the wreck execution system and many other interfaces and APIs that have been deprecated for awhile.

Before this is merged, perhaps, this will require a strategy for flux-sched, a few ideas that come to mind are:

 1. Transition immediately to compatibility with flux-core master by removing all the code and tests that require the old exec system. This would probably mean removal of resrc, RDL, and most of the tests. At some point it will have to be done anyway.
 2. branch off at the v0.7.0 tag for now, and tie testing and configure of that version to v0.11.0 of flux-core.
 3. Back out the most recent change that makes flux-sched incompatible with flux-core-v0.11.0, keep working on master and tie *master* to flux-core-0.11.0.

grondo:
This PR synchronizes flux-sched with the removal of wreck and related interfaces in flux-core. The following were removed in whole:
 - rdl
 - resrc
 - sched
 - libutil/log
 - liblsd

All related testsuite scripts under `t/` were also removed.

The current `t5000-valgrind.t` from flux-core (slightly adapted) was added back to prepare for easily adding new flux-sched valgrind workloads (just add an executable script or scripts under `t/valgrind/workload.d`). I attempted to add a workload that simply loaded/unloaded the `resource` module, but this resulted in valgrind errors so I left that out for now.

Closes #437


grondo:
I pushed some further changes earlier to remove unused code from `xzmalloc.c` (these should be avoided in production code anyway) and updated the codecov.yml (to ignore libtap in code coverage stats).

grondo:
BTW, I don't think there is anything more to do here. There was large changes in the codebase so the code coverage change could perhaps be ignored.

dongahn:
Thanks @grondo!

dongahn:
One of my next priorities: Sorry @grondo, it is taking rather long to get to this.

dongahn:
Thanks @grondo. 

The deletion looks good, but I have a mixed feeling about the removal of the scripts in t/scripts.

As part of our near future work of adding more runtime tests for `resource` module after hooking that up with the rest of the subsystems (execution+schedule loop module), I have to think we will need these scripts to control synchronization needed for testing and to fetch the `R_lite` key from the new `R` format. Any compelling reasons to remove those scripts now?

grondo:
Thanks @dongahn. The scripts as written will no longer work with flux-core after #flux-framework/flux-core#1988 is merged.

Also, removing unused code is good practice in general, because it keeps cruft from accumulating in the project.

It will be simple to go back in history and pull out any scripts (with updates) to use in future testing once they are needed. Similary, you may want to adapt or refer to all the previous sharness tests for the resrc scheduler as we move forward.

However, if you feel strongly that some of the scripts should remain, I'm find we leaving them  for now.

dongahn:
@grondo: I don't feel strongly about it at all in particular since these scripts will stop working with the latest flux-core PR. Just curious, what is the main reason that those scripts will stop working? API changes or Lua? How does flux-core deal with testing synchronization etc going forward?

dongahn:
In any case, from my perspective, this PR looks good me. Will need to go in after flux-framework/flux-core#1988 goes in though.

grondo:
Oh, sorry `waitfile.lua` will still work. However:

 - `R_lite.lua` uses the `flux.affinity` Lua module which will be removed. We'll want to update the script to the Rv1 format anyway.
- `kvs-watch-until.lua` uses the kvs Lua bindings which will be removed.


I guess there were no tests in flux-core that used the `kvs-watch-until.lua` script for synchronization, which is how I was able to remove it. I guess synchronization there depends on the exact use case.

grondo:
> Will need to go in after flux-framework/flux-core#1988 goes in though.

Since flux-sched is dependent on flux-core, I actually think we should merge this PR first, since this PR won't break flux-core, but merging flux-core PR first will break current flux-sched. (Unless I'm missing something)

In the grand scheme of things the order doesn't matter too much though.

dongahn:
When is [flux-framework/flux-core#1988](https://github.com/flux-framework/flux-core/pull/1988) expected to land? I can merge `sched` before that goes in. But some coordination seems a bit beneficial. I do agree this wouldn't matter too much from the grand scheme of things :-)

grondo:
> When is flux-framework/flux-core#1988 expected to land?

I'm not sure. It has been ready for a week or so, but we certainly didn't want to merge it before we had a plan for flux-sched.

I'll force a push there and maybe ask what else reviewers would like.

dongahn:
OK. Merging. Thanks @grondo!

grondo:
Great thanks!

PR #442, opened 2019-02-13
resource: Update for kvs watch API removal

chu11:
Replace call to flux_kvs_watch_once() with flux_kvs_lookup().

Should be merged after https://github.com/flux-framework/flux-core/pull/2011 is accepted.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -252,11 +252,17 @@ error:
  */
 static json_t *get_string_blocking (flux_t *h, const char *key)
 {
-    char *json_str = NULL; /* initial value for watch */
+    flux_future_t *f = NULL;
+    const char *json_str;
     json_t *o = NULL;
     int saved_errno;
 
-    if (flux_kvs_watch_once (h, NULL, key, &json_str) < 0) {
+    if (!(f = flux_kvs_lookup (h, NULL, FLUX_KVS_WAITCREATE, key))) {
+        saved_errno = errno;
+        goto error;
+    }
+
+    if (flux_kvs_lookup_get (f, &json_str) < 0) {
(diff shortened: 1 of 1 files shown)


grondo:
> Should be merged after flux-framework/flux-core#2011 is accepted.

isn't this a change that can be merged in to flux-sched now (moving from deprecated to non-deprecated interface)?

chu11:
> isn't this a change that can be merged in to flux-sched now (moving from deprecated to non-deprecated interface)?

Oh yeah, you're correct!  It doesn't require that the old watch API is removed :-)

resource/modules/resource_match.cpp
```diff
@@ -252,11 +252,18 @@ static resource_ctx_t *init_module (flux_t *h, int argc, char **argv)
  */
 static json_t *get_string_blocking (flux_t *h, const char *key)
 {
-    char *json_str = NULL; /* initial value for watch */
+    flux_future_t *f = NULL;
+    const char *json_str;
     json_t *o = NULL;
+    int flags = FLUX_KVS_WATCH | FLUX_KVS_WAITCREATE;
```
SteVwonder: Since the value normally won't change (and we don't have a mechanism for handling a change in value), is the `FLUX_KVS_WATCH` necessary?  Or can we just do a `WAITCREATE`?  If the `WATCH` is necessary, does there need to be a `lookup_cancel`?

SteVwonder:
Thanks @chu11!  Sorry that I propagated the use of a deprecated function.

resource/modules/resource_match.cpp
chu11: doh!  you're absolutely right.  And I was the one that added that feature.  Re-push coming shortly.

Issue #453, opened 2019-03-21
hwloc reader for resource doesn't fill in rank 


dongahn:
Just ran a simple test of the resource-matching service with RV1 emitter (I used `rv1_nosched` emitter instead of `rv1` more specifically). I will see if I can add rank support. But if rank support is already there but somehow this test doesn't appear to run incorrect, please let me know. 

```
ahn1@c7ec51045e3e:/usr/src/t$ flux resource match allocate data/resource/jobspecs/basics/test008.yaml
JOBID                STATUS               AT                   OVERHEAD (Secs)
0                    ALLOCATED            Now                  0.000555992126465
================================================================================
MATCHED RESOURCES:
{"execution":{"R_lite":[{"rank":-1,"node":"c7ec51045e3e","children":{"core":"3"}}]}}
```

grondo:
>  But if rank support is already there but somehow this test doesn't appear to run incorrect, please let me know.

Sorry, I didn't parse that @dongahn.  How can I help?

The current execution system prototype does require correct `rank` information, and ignores any hostname in `node`.

dongahn:
Sorry @grondo. I should have been clearer. The comment was more for @SteVwonder who added hwloc reader support within resource.  Notice:

```
{"R_lite":[{"rank":-1,
```

The emitter simply emits whatever value we have for the `rank` field within each resource vertex. The test shows that this field just has the initial value which is -1. So it seems I will need to add "rank" support for the hwloc reader. I just wanted to confirm with @SteVwonder that it is not that case that he already added rank support and my test somehow wasn't picking up.

grondo:
Ah, ok, sorry!

dongahn:
The most of rank support was already there. Apparently, some BGL algorithms call either the copy constructor or assignment constructor of a resource vertex, which doesn't copy the rank field:  https://github.com/dongahn/flux-sched/blob/match-format/resource/schema/resource_data.cpp#L42

An easy fix is to add rank support into those constructors, which I will do for now.

Then, still I need to fully understand which BGL algorithms are calling these methods. Currently, the code doesn't make a deep copy of the "schedule" data of a resource vertex, so depending on when and how these methods are called, the copying can having a side effect. I vaguely remember I actually tested this when I first implemented, this but it wouldn't hurt to verify this once again since it has been for a while.

[...]

Issue #451, opened 2019-03-12
uuid_generate adds a huge startup overhead


dongahn:
I tried to add uuid support into `resource` and found that this blows up the graph populating time by multiple factors. 

I don't remember why we decide to use an uuid for each resource. Does anyone remember? 

If there isn't a good reason for this, can we just use a "unique" integer id into each resource? Our graph implementaiton is such as way that `vtx_t` and `edg_t` resolve to integers and uniquely identify each vertex and edge. Can we use this instead of the heavyweight uuid?

grondo:
> If there isn't a good reason for this, can we just use a "unique" integer id into each resource? Our graph implementaiton is such as way that vtx_t and edg_t resolve to integers and uniquely identify each vertex and edge. Can we use this instead of the heavyweight uuid?

Yes, that seems fine to me.

The UUID per resource was just a way to resolve the same resource between any instance running within a datacenter (or I suppose any instance anywhere). This allows resolving an error, data, resource tags, etc from any child instance to the parent easily.

However, a uuid per resource seems like a concept that doesn't really have a good (or possible?) implementation, and besides, we don't really have a use case yet for matching resources across instances (and I bet there is a better way for parent<->child instances anyway)

grondo:
> Our graph implementaiton is such as way that vtx_t and edg_t resolve to integers and uniquely identify each vertex and edge. Can we use this instead of the heavyweight uuid?

Just curious, will the implementation keep the same vtx_t id in all Rs generated from the parent instance R?

dongahn:
> Just curious, will the implementation keep the same vtx_t id in all Rs generated from the parent instance R?

Yes, that is my current plan. 

The top-level instance that will use a grug-based  or hwloc-based generator, which will use the vtx_t id. Then, the nested instances will use an R reader such as RV1 reader which will keep this same unique id instead of regenerating a new one.

Issue #448, opened 2019-03-08
Need a performance benchmark for resource


dongahn:
This was discussed at https://github.com/flux-framework/flux-core/pull/2062#issuecomment-469775105 and I also discussed this with @grondo and @garlick. 

We need a testing environment that configures `resource-query` with a very simple system configuration consisting of only nodes and cores. Comparing its performance against that of the simple scheduler can provide the overheads that we have to pay in exchange for the advanced capabilities.

dongahn:
I added https://github.com/flux-framework/flux-sched/pull/446/commits/3334e56b56d2a5fbfd5eab352e5be8e29774b886 to support this and ran some tests on one of the quartz compute nodes:

### Flat cluster configuration
- Target system: `cluster->node[256]->core[32]`
-  jobspec: `cluster->node[1]->slot[1]->core[1]`
- resource-query: use default parameter set (e.g., high id first match policy etc).

The overhead stats of 4096 jobs of successful `match allocate jobspec`:

```
Num. of Jobs Matched: 4096
Min. Match Time: 0.01494
Max. Match Time: 0.0355251
Avg. Match Time: 0.022324
```
= ~45 jobs per second

The overhead stats of 8192 jobs of successful `match allocate jobspec`:

```
Num. of Jobs Matched: 8192
Min. Match Time: 0.000808001 
Max. Match Time: 0.0336521
Avg. Match Time: 0.0149624
```
= ~67 jobs per second

### Hierarchical cluster configuration
- Target system: `cluster->rack[8]->node[32]->socket[2]->core[16]`
-  jobspec: `cluster->node[1]->rack[1]->node[1]->slot[1]->socket[1]->core[1]`
- resource-query: use default parameter set (e.g., high id first match policy etc).
The overhead stats of 8192 jobs of successful `match allocate jobspec`:

```
Num. of Jobs Matched: 8192
Min. Match Time: 0.000168085
Max. Match Time: 0.0442641
Avg. Match Time: 0.0167113
```
= ~59 jobs per sec

### Overall
As this part of the code has not been fully optimized for throughput, I am not that unhappy with the results and this gives me some ideas about how we can **further specialize** our policies tailored to to the throughput workloads like I used here.  

[...]

Issue #445, opened 2019-02-20
Need an R_Lite emitter for RFC20


dongahn:
Make this as a separate issue because I couldn't yet decide the best way to generate the `R_Lite` segment of RFC20.  This may need some discussions later.

PR #455, opened 2019-03-22
RFC20 support + other prep work for future integration with new execution system

dongahn:
This PR adds a bunch of support needed in preparation for future integration with the new execution systems: 

* Add support for various emit formats including RFC20 RV1 (Issue #436, #443 and #445)
* Add performance improvements such as adding `reserve-vtx-vec` options and remove uuid (Issue #451)
* Add better performance measurement rigs (Issue #448)
* Various bug fixes (Issue #453)

### DIFF
--- a/t/t6000-graph-size.t
+++ b/t/t6000-graph-size.t
@@ -5,15 +5,35 @@ test_description='Test Correctness of Populating Graph Data Store'
 . $(dirname $0)/sharness.sh
 
 tiny_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+large_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/sierra.graphml"
 query="../../resource/utilities/resource-query"
 
 test_expect_success "vertex/edge counts for a tiny machine are correct" '
-    echo "quit" >> input1.txt &&
+    echo "quit" > input1.txt &&
     ${query} -G ${tiny_grug} -e -S CA -P high < input1.txt > out1.txt &&
     vtx=$(grep "Vertex" out1.txt  | sed "s/INFO: Vertex Count: //") &&
     edg=$(grep "Edge" out1.txt  | sed "s/INFO: Edge Count: //") &&
     test ${vtx} -eq 100 &&
     test ${edg} -eq 198
 '
+
+test_expect_success "--reserve-vtx-vec works" '
+    echo "quit" > input2.txt &&
+    ${query} -G ${tiny_grug} -e -r 1024 -S CA -P high < input2.txt > out2.txt &&
+    vtx=$(grep "Vertex" out2.txt  | sed "s/INFO: Vertex Count: //") &&
+    edg=$(grep "Edge" out2.txt  | sed "s/INFO: Edge Count: //") &&
+    test ${vtx} -eq 100 &&
+    test ${edg} -eq 198
+'
+
+test_expect_success LONGTEST "--reserve-vtx-vec improves loading performance" '
+    echo "quit" > input3.txt &&
+    ${query} -G ${large_grug} -e -r 400000 < input3.txt > out3A.txt &&
+    ${query} -G ${large_grug} -e < input3.txt > out3B.txt &&
+    with=$(grep "Graph" out3A.txt  | sed "s/INFO: Graph Load Time: //") &&
+    without=$(grep "Graph" out3B.txt  | sed "s/INFO: Graph Load Time: //") &&
+    test $(awk "BEGIN{ print $with<$without }") -eq 1
+'
+
 
 test_done
--- a/t/t6000-graph-size.t
+++ b/t/t6000-graph-size.t
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+test_description='Test Correctness of Populating Graph Data Store'
+
+. $(dirname $0)/sharness.sh
+
+tiny_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+query="../../resource/utilities/resource-query"
+
+test_expect_success "vertex/edge counts for a tiny machine are correct" '
+    echo "quit" >> input1.txt &&
+    ${query} -G ${tiny_grug} -e -S CA -P high < input1.txt > out1.txt &&
+    vtx=$(grep "Vertex" out1.txt  | sed "s/INFO: Vertex Count: //") &&
+    edg=$(grep "Edge" out1.txt  | sed "s/INFO: Edge Count: //") &&
+    test ${vtx} -eq 100 &&
+    test ${edg} -eq 198
+'
+
+test_done
--- a/configure.ac
+++ b/configure.ac
@@ -83,6 +83,12 @@ AM_CHECK_PYMOD(yaml,
                [StrictVersion(yaml.__version__) >= StrictVersion('3.10')], [],
                [AC_MSG_ERROR([[could not find python module yaml, version 3.10+ required]])])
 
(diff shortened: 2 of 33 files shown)


dongahn:
OK. I closed the old PR and newly posted this PR as I think this is getting closer. Before I will turn into a pumpkin, I will try to add a few test cases that uses JGF validator (and run some valgrind tests) at which point this should be ready for review. Hence I removed WIP from the title. 

There are whole bunch of other things I need to work on to be ready for the integration work with the new execution system. But this would be a good merge point as this PR becomes larger and larger, and I won't be able to get to this work for awhile due to travel+vacation.

SteVwonder:
Hey @dongahn.  Thanks for putting this PR together!

I like the new interface for writers.  It seems very extensible to new formats!  For the JSON-based writers, I think we should seriously consider pulling in a C++ JSON library (crossref #383).   It would remove the need for a lot of the manual string manipulation, easing future extension and maintenance, and the pretty printing could be handled by the JSON library, removing a lot of the duplicate code.

Per the naming, it looks like `writers` is used for the classes/types, but `emitters` is used for the filenames/directory.  I think those should probably be the same (either one works for me).  When looking for the `jgf_match_writers_t` implementation, I was looking for a `writers` directory, but it was actually under `emitters`.

When I tested out the `resource-query` command on the new `sierra.graphml` and the old `t/data/resource/jobspecs/basics/test004.yaml`, I noticed some oddities on the JGF output ([link to the output that I saw](https://gist.github.com/SteVwonder/9045a9f0d718f2420eabdab5c0c7c0c0)).  Specifically, it looks like there are edges to targets whose IDs don't exist in the `nodes` section of the graph.  My gut instinct is that this is related to pruning of resources that aren't a `core`, `node`, `socket`, etc, but I haven't tracked it down.

Adding the json-schema for JGF is a great addition!  Is the `resource-query` JGF output being passed through `validate_schema.py` with the JGF json-schema?  If not, I think that would be a useful test to add.

The `reserve_vtx_vec` is a nice addition.  Noticeable speed-ups on the sierra GRUG!

dongahn:
Great reviews @SteVwonder across the board!

>  C++ JSON library
I haven't looked at this library yet. Maybe this is the time. 

> emitters vs. writers for the directory/file names.
I will change this to `writers` since this will reduce confusion like what you had. 

> When I tested out the resource-query command on the new sierra.graphml and the old t/data/resource/jobspecs/basics/test004.yaml, I noticed some oddities on the JGF output (link to the output that I saw). Specifically, it looks like there are edges to targets whose IDs don't exist in the nodes section of the graph. My gut instinct is that this is related to pruning of resources that aren't a core, node, socket, etc, but I haven't tracked it down.

Hmmm. I can't say why there are disconnected edges like this at this point. I do need to seriously look at this issue before this PR gets merged (we may need a test case for this as well). Thank you for kicking the tire for this!

> Adding the json-schema for JGF is a great addition! Is the resource-query JGF output being passed through validate_schema.py with the JGF json-schema? If not, I think that would be a useful test to add.

Yes, this is the intent. I have not been able to work on this because of other duties... much slower than my expectations.

SteVwonder:
> C++ JSON library
> I haven't looked at this library yet. Maybe this is the time.

To keep this PR from growing too large in scope and to get it merged quicker, we can make the C++ JSON library a separate PR if you want.  Also, if you want, I am willing to pick up that task, so you can focus on other aspects of the PR and integration with the new exec system.

dongahn:
> To keep this PR from growing too large in scope and to get it merged quicker, we can make the C++ JSON library a separate PR if you want. Also, if you want, I am willing to pick up that task, so you can focus on other aspects of the PR and integration with the new exec system.

Sounds good for both counts. Thanks you @SteVwonder!

dongahn:
> I like the new interface for writers. It seems very extensible to new formats! For the JSON-based writers, I think we should seriously consider pulling in a C++ JSON library (crossref #383). It would remove the need for a lot of the manual string manipulation, easing future extension and maintenance, and the pretty printing could be handled by the JSON library, removing a lot of the duplicate code.

@SteVwonder: rereading again: I do believe this IS the right way to go for the future. I found myself worrying about indentations/formats and maintainability. The introduction of this library should solve this headache.

dongahn:
> When I tested out the resource-query command on the new sierra.graphml and the old t/data/resource/jobspecs/basics/test004.yaml, I noticed some oddities on the JGF output (link to the output that I saw). Specifically, it looks like there are edges to targets whose IDs don't exist in the nodes section of the graph. My gut instinct is that this is related to pruning of resources that aren't a core, node, socket, etc, but I haven't tracked it down.

Nice catch! I reproduced this, and it appears that somehow higher-level resources (e.g., rack) that do not contain the allocated resources are also emitted! This is clearly a bug. I will work on it. Thanks.

dongahn:
> Hmmm. I can't say why there are disconnected edges like this at this point. I do need to seriously look at this issue before this PR gets merged (we may need a test case for this as well). Thank you for kicking the tire for this!

@SteVwonder: I found the root cause. It was basically due to a mistake where unqualified edges at a high level resource (e.g., cluster) are emitted along with those qualified edges. Fixed it.

dongahn:
I am getting CI errors for the newly added jsonschema validation test for JGF. 

> FAIL: t6001-match-formats.t 1 - R emitted with -F jgf validates against schema
> FAIL: t6001-match-formats.t 2 - R emitted with -F pretty_jgf validates against schema

Perhaps jsonschma python module is not installed for CI...

dongahn:
Ok. This is ready for another review. Some of the commits need to be squashed and reordered but I'm leaving it this way for quicker review. @SteVwonder: do you want to kick the tire for this once again?

SteVwonder:
> do you want to kick the tire for this once again?

Yep.  Just added to my TODO list for this week.

SteVwonder:
LGTM!  Let me know once you have squashed/reordered any commits that you want to, and then I'll push the button.

dongahn:
> LGTM! Let me know once you have squashed/reordered any commits that you want to, and then I'll push the button.

Thanks @SteVwonder! I will do this either today or tomorrow.

t/t6000-graph-size.t
```diff
@@ -0,0 +1,39 @@
...
+test_description='Test Correctness of Populating Graph Data Store'
+
+. $(dirname $0)/sharness.sh
+
+tiny_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+large_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/sierra.graphml"
+query="../../resource/utilities/resource-query"
+
+test_expect_success "vertex/edge counts for a tiny machine are correct" '
+    echo "quit" >> input1.txt &&
```
SteVwonder: Sorry. I just caught this will checking out #460, but I believe you want this to be:
```suggestion
    echo "quit" > input1.txt &&
```
in the case where `input1.txt` already exists and has other commands in it.  Same goes for the other tests in this file.

dongahn: Good catch! Will address it shortly.

dongahn:
> LGTM! Let me know once you have squashed/reordered any commits that you want to, and then I'll push the button.

I reviewed this again and I actually don't see a compelling reason to squash and reorder much. I will force a push to address your last review point. And this can go in from my perspective.

dongahn:
@SteVwonder: Forgot to drop a line. Forced a push and this is ready to be merged.

SteVwonder:
Thanks! Merged.

dongahn:
Thanks @StevWonder for reviewing this thoroughly. I know this was pretty substantial!

Issue #466, opened 2019-06-11
misc. problems with flux resource match allocate


dongahn:
Encountered a few issues when testing my hwloc whitelist support for #454 

1. When there is no match, `flux resource match allocate` prints out:

```console
ahn1@b71137318c0b:/usr/src$ flux resource match allocate ./test001.yaml
JOBID                STATUS               AT                   OVERHEAD (Secs)
0                    ALLOCATED            Now                  0.0010130405426
================================================================================
MATCHED RESOURCES:
{"execution":}
```

2. with Sierra2 hwloc xml, `flux resource match allocate` prints an error message and hangs

```console
flux start -s 1
flux module remove resource
flux hwloc reload /usr/src/t/data/hwloc-data/004N/exclusive/04-brokers-sierra2/
flux module load resource
flux resource match allocate ./test001.yaml
2019-06-11T00:20:49.490515Z resource.err[0]: match_request_cb: flux_respond_error: Invalid argument
```

dongahn:
I plan to add fixes for those as part of PR #454.

dongahn:
Item #2 above was easy to fix. Just need to add `errno=EBUSY` when resources cannot be allocated. Changes are committed to PR #467.

Iterm #1: I reproduced it using `resource-query`.

test001.yaml:
```yaml
ahn1@b71137318c0b:/usr/src$ cat test001.yaml
version: 1
resources:
    - type: node
      count: 1
      with:
          - type: slot
            count: 1
            label: default
            with:
              - type: socket
                count: 1
                with:
                  - type: core
                    count: 1
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: app
    slot: default
    count:
      per_slot: 1
```

```console
ahn1@b71137318c0b:/usr/src$ resource/utilities/resource-query -X /usr/src/t/data/hwloc-data/001N/amd_gpu/corona11.xml
resource-query> match allocate ./test001.yaml
      ---------socket1[1:x]
      ------corona11[1:s]
      ---cluster0[1:s]
INFO: =============================
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
INFO: =============================
```

Apparently, matcher couldn't match the jobspec at the `core` level but still reported success. Clearly a bug. FWIW, when I normalized the hwloc with whitelist, the matching worked fine.

```console
[...]

Issue #454, opened 2019-03-22
hwloc resource type whitelist and blacklist support


dongahn:
It seems we will soon need a better way to filter unnecessary resource types out in building the resource graph data store within `resource` for hwloc reader. I propose we add whitelist and blacklist support as a module load option to achieve this. 

For example, hwloc-blacklist=numanode,group,pu will make our hwloc reader not to populate resource vertices for numanode, group and pu type. hwloc-whitelist=node,socket,core will only create resource types for the node, socket and core types if they exist.

This will allow us to coarsen the resource model or make it fine grained even with the hwloc reader. If coarsened, this will improve the scheduler performance and reduce graph complexity.

dongahn:
I believe this would be a generalized solution that happens to address Issue #438.

dongahn:
I found writing "normalized" job specs with hwloc reader is pretty difficult because different systems report their topologies differently via hwloc. A jobspec that works perfectly fine on one system (e.g., Sierra) doesn't necessarily match well with another system (Corona). We should give this a high priority to reduce such complexity. I believe this will be absolutely required before deploying graph-based schedulers in production.

dongahn:
WIP PR has been posted PR #467. This turns out to be pretty simple support and I will add one more test case with `resource-match`.

dongahn:
PR #467 resolved this.

PR #467, opened 2019-06-11
Add hwloc whitelist support

dongahn:
This PR adds support for hwloc whitelist, which allows the resource-query and -match to load in hwloc resources that are in the whitelist. 

This alleviate two problems: 1) the graph date store created with hwloc is relatively large with no clear benefit to scheduling; 2) because hwloc topology oftentimes exhibit variations between different platforms, one jobspec that work on one platform doesn't necessarily work for another platform.

This also has a fix for a couple bugs: Issue #466. 

Resolve Issue Issue #[454](https://github.com/flux-framework/flux-sched/issues/454) and #466.


dongahn:
@SteVwonder: OK. Travis is green. This is ready for your review. This is kind of important to support minimal portable jobspec. One caveat is this may still not support the complete portability of a jobspec. Even though you will whitelist and downselect resources to populate the resource graph data store with this, the logic preserves the relative hierarchy of those resources. So if the relative hierarchy of a resource as reported by hwloc is different between two platforms, one jobspec may not work across both platforms. I suspect, GPUs could be in that category. I can imagine in one platform, hwloc may report this as: 

```
node->socket->cpu
    ->gpu
```

And in another platform: 
```
node->socket->cpu
            ->gpu
```

Although I haven't seen this yet.

dongahn:
I don't understand why codecov/patch coverage is reduced significantly...

dongahn:
> I don't understand why codecov/patch coverage is reduced significantly...

@SteVwonder or @grondo: Should this be just okay? I vaguely remember there were problems with code coverage with docker support.

resource/traversers/dfu_impl.cpp
```diff
@@ -253,21 +253,24 @@ bool dfu_impl_t::slot_match (vtx_t u, const Resource *slot_resources)
 
 const vector<Resource> &dfu_impl_t::test (vtx_t u,
                                           const vector<Resource> &resources,
-                                          match_kind_t *spec)
+                                          bool &prestine, match_kind_t &spec)
```
SteVwonder: `s/prestine/pristine/`?  Even with that, I don't immediately grok the naming of this variable.  What was the intended semantics/meaning behind this name?

Can you include the text in your commit message as a comment near this code?  I think that would be very helpful to anyone working on this code in the future.

It is also a little confusing at first to have prestine be an in/out variable.  I wonder if it would make sense to modify prestine based on the returned value of spec (i.e., in `dom_dfv`).

resource/traversers/dfu_impl.hpp
```diff
@@ -43,10 +43,14 @@ namespace detail {
...
 
-enum class match_kind_t { RESOURCE_MATCH, SLOT_MATCH, NONE_MATCH };
+enum class match_kind_t { RESOURCE_MATCH,
+                          SLOT_MATCH,
+                          NONE_MATCH,
+                          PRESTINE_NONE_MATCH };
 
 struct jobmeta_t {
     bool allocate = true;
+    bool never_matched = true;
```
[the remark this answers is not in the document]
dongahn: Ah. Good catch. This is a deadwood! Thanks.

resource/traversers/dfu_impl.cpp
dongahn: I was struggling with the naming myself as well. 

Just a background: I need to differentiate two similar but distinct cases and `pristine` is used to differentiate them. 

1) Jobspec is allowed to omit the prefix so you can specify things like `socket[1]->core[2]` which will still match `cluster[1]->node[1]->socket[2]->core[22]`. For this case, when you are visiting the `node` resource, the next jobspec resource that should be used at the next recursion should still be `socket[1]`.  And the logic does this if `pristine` is true. 

2) But then once the first-level match is made, any mis-mismatch afterwards should result in a match failure. For example, `socket[1]->core[2]` should fail to match `cluster[1]->socket[1]->numanode[1]->core[22]`. And `pristine` is used to detect this case. 

I will take a look at the code once again to see if I can clarify this a bit better.

t/t4004-match-hwloc.t
```diff
@@ -124,6 +136,15 @@ test_expect_success 'match-allocate fails when all resources are allocated' '
     test_expect_code 16 flux resource match allocate ${jobspec_2socket}
 '
 
+test_expect_success 'reloading sierra xml' '
+    flux module remove resource &&
+    flux hwloc reload ${excl_4N4B_sierra2} &&
+    flux module load -r 0 resource subsystems=containment \
+policy=high hwloc-whitelist=node,socket,core,gpu &&
+    flux resource match allocate ${jobspec_1socket_2gpu} &&
+    flux resource match allocate ${jobspec_1socket_2gpu}
```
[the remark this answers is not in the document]
dongahn: Yes good point. I should be able to augment the tests a bit.

resource/traversers/dfu_impl.cpp
dongahn: Now that I think about this, there maybe one case I also overlooked as well. Let me look into this.

SteVwonder: Per the naming, `pristine` is fine with me.  I think a comment in the code to explain the semantics would go a long way to clarifying things.  Your description here makes a lot of sense.  Thanks!

SteVwonder:
> Should this be just okay? I vaguely remember there were problems with code coverage with docker support.

From what I'm seeing, it looks like the overall coverage stayed basically the same, but of the code that changed in this PR, a slightly smaller percentage was covered than the percentage covered in the repo as a whole.  Peeking at the [coverage diff breakdown](https://codecov.io/gh/flux-framework/flux-sched/compare/972065687c0cf24d8ba15c7de2059b92ebb4d81e...5c2e852342df1655614e48a15e65f0ae2f6fb1bc/diff#diff-cmVzb3VyY2UvZ2VuZXJhdG9ycy9nZW4uY3Bw), it looks like the only changes that aren't covered by testing are error paths.  So I'm fine with this coverage as-is.

dongahn:
Will try to get to this tomorrow.

dongahn:
@SteVwonder: I push my changes that address all of your (great) review comments. Can you please quickly take a look the new commits? If all is good, I will squash them and then this should be good to be merged.

SteVwonder:
Thanks @dongahn. LGTM! Ready for squashing and then merging.

dongahn:
@SteVwonder: Thanks. Squashed and pushed. Once Travis is ok, this is good to go.

SteVwonder:
Thanks @dongahn!

PR #474, opened 2019-06-22
WIP: Add high-level resource API skeleton support

dongahn:
This is experimental in case people like @cmisale wants to take a peek at where I'm headed. (So don't even attempt to merge at all!)

For our main work, these high-level resource APIs are required to reduce software complexity for our upcoming queue manager support, the module that will directly interface with `job-manager`. In particular, this helps designing our future queue manager to be layered both with RPC and CLI (i.e., for testing) without having to specialize the main code base for each case.  

As I will start to use these APIs in the upcoming queue manager code, the APIs and implementations will be likely to change, but the structure should remain the same.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -166,6 +166,9 @@ AC_CONFIG_FILES([Makefile
   resource/utilities/Makefile
   resource/utilities/test/Makefile
   resource/modules/Makefile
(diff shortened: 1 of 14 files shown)


dongahn:
This will be subsumed by an upcoming PR.

Issue #480, opened 2019-06-28
The version attribute is missing from the R


dongahn:
What's emitted
```console
2019-06-28T17:57:30.207491Z qmanager.info[0]: jobid (576431587328): {"execution":{"R_lite":[{"rank":0,"node":"59fa0c6cd2b8","children":{"core":"3"}}]}}
```

This led to the new execution system's reader error. 
```console
{"timestamp":1561744650.2166536,"name":"exception","context":{"note":"reading R: Object item not found: version"}}
```

dongahn:
It was easy to fix. But then, I get

```
{"timestamp":1561745825.5159779,"name":"exception","context":{"note":"reading R: R_lite: failed to read target rank list"}}
```

```
{"version": 1,"execution":{"R_lite":[{"rank":-1,"node":"node0","children":{"core":"35"}}]}}
```

Change the issue title to capture all of these RV1 compat issues.

dongahn:
Wrong log:

```diff
> {"version": 1,"execution":{"R_lite":[{"rank":-1,"node":"node0","children":{"core":"35"}}]}}
< {"version": 1,"execution":{"R_lite":[{"rank":0,"node":"59fa0c6cd2b8","children":{"core":"3"}}]}}
```

So it turned out, rank info was correctly there with hwloc reader. But the rv1 writer emits the rank as an integer not as a string. After this schema fix:

```console
2019-06-28T19:01:35.899599Z qmanager.info[0]: jobid (212114341888): {"version":1,"execution":{"R_lite":[{"rank":"0","node":"59fa0c6cd2b8","children":{"core":"3"}}]}}

ahn1@59fa0c6cd2b8:/usr/src/scratch$ flux job id --to=kvs 212114341888
job.0000.0031.6300.0000
ahn1@59fa0c6cd2b8:/usr/src/scratch$ flux kvs get -r job.0000.0031.6300.0000.guest.exec.eventlog
{"timestamp":1561748495.907171,"name":"init"}
{"timestamp":1561748495.9112291,"name":"starting"}
{"timestamp":1561748495.9112711,"name":"running","context":{"timer":"3600.000000s"}}
ahn1@59fa0c6cd2b8:/usr/src/scratch$
```

This was the first working end-to-end run!

dongahn:
PR #481 resolved this.

Issue #477, opened 2019-06-26
The absence of duration in flux-jobspec produced JSONs crashes libjobspec


dongahn:
Not sure if this is a flux-core issue or flux-sched issue so I'm starting this at the core first.

```console
ahn1@8a06403418e4:/usr/src/scratch$ flux jobspec srun -n1 hostname | jq > test.json
```

```json
{
  "tasks": [
    {
      "slot": "task",
      "count": {
        "per_slot": 1
      },
      "command": [
        "hostname"
      ],
      "attributes": {}
    }
  ],
  "attributes": {
    "system": {}
  },
  "version": 1,
  "resources": [
    {
      "count": 1,
      "with": [
        {
          "count": 1,
          "type": "core"
        }
      ],
      "type": "slot",
      "label": "task"
    }
  ]
}
```

```console
ahn1@8a06403418e4:/usr/src/scratch$ ../resource/utilities/resource-query -G ../t/data/resource/grugs/tiny.graphml
INFO: Loading a matcher: CA
resource-query>
resource-query> match allocate ./test.json
Segmentation fault
```

I will try to diagnose the issue. But I think we would need testing coverage for this.

SteVwonder:
> OK. Turns out this is not a huge issue. Looks to me: it is just that libjobspec (or resource) requires the duration field. W/o this, the utility led to SIGSEGV.

Nice catch!

> Ultimately, flux-sched will need a config file that allows resource to fill default values into the jobspec fields that are not specified. I will move this ticket to flux-sched.

I wonder if the scheduler is the right place for this or not.  By the time that the scheduler sees the jobspec, it has already been signed by the user and cannot be modified.  Long term, I wonder if this is something that should be handled by front-end tools like `flux-jobspec` and `flux-submit`.  One benefit of the latter being that the ultimate duration used for the job will be documented in the jobspec stored in the KVS.

> I will try to diagnose the issue. But I think we would need testing coverage for this.

Agreed.  The strawman jobspec v1 in https://github.com/flux-framework/rfc/issues/149 requires that the `duration` key exists.  So at least for the period of time that we only support jobspec V1, the `flux-jobspec` command should be modified to always include the `duration` key and the job validator json-schema needs to be updated to require `duration` for the jobspec to pass validation.

dongahn:
[...]

Issue #471, opened 2019-06-19
Need a high-level, simple resource interface within resource (module)


dongahn:
Two use cases:

- As I'm looking at the queue manager layer (a new service that will interact with job manager), I found a need to call either `match.allocate` rpc (to build a service module) or match_allocate function call (to build a testing utility). 

- This will help @cmisale interface our resource infrastructure with Kubernetes and Go.

cmisale:
I saw that you restructured the code quite recently. The code I used as reference for the Go wrapper is older than those changes. 
I see it is basically the same, the newer version has the code just restructured in functions.
I can confirm that such high-level interface will help also in my case.

Talking specifically about my instance, I made that kind of restructuring by myself time ago, to allow a rpc kind of service.
In my code, I have a `ResQWrapper` class where I have the initialization logic (i.e., create the context, set the match subsystem and policy, grug file..) as well as the match method, which prepares the command using method's argument, and then call `cmd_match(ctx, tokens)` function in `command.hpp` file.
I will be available to help and/or provide feedback.

dongahn:
Thanks @cmisale. 

> I see it is basically the same, the newer version has the code just restructured in functions.

Sorry about this. But this project is under active development and this kind of implementation and internal API refactoring is unavoidable. 

One benefit fo introducing a high level interface is to make it stable so that users like you (and me) don't have to change our code much.

dongahn:
@grondo: A question. It would be best for me to minimize `resource`'s flux-core dependency (other than the module itself). One thing that I noticed is that the new execution system introduces a new jobid type `flux_jobid_t`. How does this type work if being casted to `int64_t`? As you can guess, I am wondering if I can still use a non-flux-specific`int64_t` type as the internal jobid type within `resource` and somehow convert`flux_jobid_t` into it.  

If this is not feasible, I will think about other ways.

grondo:
>  I noticed is that the new execution system introduces a new jobid type flux_jobid_t. How does this type work if being casted to int64_t?

`flux_jobid_t` is just:
```C
typedef uint64_t flux_jobid_t;
```

int64_t should work I guess if the code doesn't assume values > 0 and is just using the jobid as a unique identifier. However, it might be better to switch to `uint64_t`.

dongahn:
[...]

Issue #468, opened 2019-06-11
Integrate resource-match with the new scheduler interface


dongahn:
Had a discussion with @grondo yesterday. Though the scheduler interface isn't where he wants it to be, we thought it would be a good idea to start doing the integration work sooner rather than later. Once #467 is merged, I plan to take a crack at this.

dongahn:
@grondo: could you post some pointers as to what files I should start to take a look? Thanks.

grondo:
The helper functions for scheduler integration were developed by @garlick and can be found in `src/common/libschedutil`.

For example uses check out `src/modules/sched-simple` and `t/job-manager/sched-dummy.c`.

We had planned to abstract `libschedutil` into a scheduler specific module interface to simplify scheduler development, however, that work has been pushed off due to other priorities. For now it might be easiest to just copy libschedutil into `flux-sched`. Once we transition to a more polished scheduler interface we could remove that convenience library from flux-sched.

dongahn:
Thanks @grondo. This will be my next priority.

dongahn:
I looked at some of these suggested files. They look great. My feeling is, though, the scheduler loop service at flux-sched will have to be much more complex, and we will need to manage this complexity very carefully. Note that the original `sched` was pretty complex with various scheduler parameter and queueing policy variations (plus embedded emulator, which isn't an issue for this round) and we should use that experience to design this better. My proposal is to make a subdirectory at the top level called `sched` in which we use a similar strategy as `resource`.

- Build up abstractions using C++ classes like `class sched_t` and etc;
- Build a command-line utility in `sched/utilities` like `% sched>` that uses these abstractions such that we can test and debug using the CLI more comprehensively (like `resource-query` did it for resource matching);
- Build a `sched-loop` module as a thin layer atop of sched classes in `sched/modules`.

I also talked with @garlick and @grondo and I will copy `libschedutil` from flux-core to flux-sched in the corresponding location.

dongahn:
@garlick and @grondo: I assume you won't have queueing policies other than fcfs at the job-manager level, correct? So essentially `alloc` will be issued in the fcfs order from 'job-manager'. (though priority and such can change this order). 

This is fine but I thought I should doublecheck.   

When users want the any out of order policy like backfilling at the flux-sched level, I will assume sched will have to use "unlimited" to replicate the entire job queue.

[...]

PR #481, opened 2019-06-28
qmanager to integrate with the new exec system

dongahn:
This PR has the following:

- Define the high level resource API so that our resource infrastructure can be better used by other module users and CLI users. This will also facilitate other language binding (such as Golang) creations.
- Implement this API for the module users.
- Incorporate schedutil from flux-core into flux-sched.
- Introduce the new queuing policy interface. I expect that many classical queueing polices can be implemented by deriving from this class and overriding its `run_sched_loop` interface. Basic queuing operations are already supported by the base classes using C++ STL containers. To make certain queuing operation efficient, I decided not to use `std::list` but use `std::map` keyed by monotonically increasing queuing time.
- Implement the FCFS policy derived class and also added a skeleton EASY policy class with no implementation as a placeholder. Note that the queuing policy interface is designed to be used by future CLI users as well as module users -- useful for testing.
- Use these primitives to implement the first baseline version of qmanager which provides the schedule loop service integrating both the new execution system within flux-core and the resource match service within flux-sched.
- Fix a few RV1 compatibility issues including libjobspec fix

Resolve Issue #480, #468, and #471, #483, and #477.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -1,0 +1,231 @@
+/*****************************************************************************\
+ *  Copyright (c) 2019 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+
+extern "C" {
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+#include "src/common/libschedutil/schedutil.h"
+}
+
+#include "qmanager/policies/base/queue_policy_base.hpp"
+#include "qmanager/policies/base/queue_policy_base_impl.hpp"
+#include "qmanager/policies/queue_policy_factory_impl.hpp"
+
+
+using namespace Flux;
+using namespace Flux::queue_manager;
+using namespace Flux::queue_manager::detail;
+
+/******************************************************************************
+ *                                                                            *
+ *                 Queue Manager Service Module Context                       *
+ *                                                                            *
+ ******************************************************************************/
+
+struct qmanager_ctx_t {
+    flux_t *h;
+    ops_context *ops;
+    queue_policy_base_t *queue;
+};
+
+
+/******************************************************************************
+ *                                                                            *
+ *                     Internal Queue Manager APIs                            *
+ *                                                                            *
+ ******************************************************************************/
+
+extern "C" int jobmanager_hello_cb (flux_t *h, const char *R, void *arg)
+{
+    return 0;
+}
+
+extern "C" void jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
+                                     const char *jobspec, void *arg)
+{
+    uint32_t userid;
+    qmanager_ctx_t *ctx = (qmanager_ctx_t *)arg;
+    std::shared_ptr<job_t> job = std::make_shared<job_t> ();
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        return;
+
+    flux_log (h, LOG_INFO, "alloc requested by user (%u).", userid);
+
+    if (schedutil_alloc_request_decode (msg, &job->id, &job->priority,
+                                        &job->userid, &job->t_submit) < 0) {
+        flux_log_error (h, "%s: schedutil_alloc_request_decode", __FUNCTION__);
+        return;
+    }
+    job->jobspec = jobspec;
+    job->msg = flux_msg_copy (msg, true);
+    if (ctx->queue->insert (job) < 0) {
+        flux_log_error (h, "%s: queue insert", __FUNCTION__);
+        return;
+    }
+    if (ctx->queue->run_sched_loop ((void *)ctx->h, true) < 0) {
+        flux_log (ctx->h, LOG_DEBUG,
+                  "%s: return code < 0 from schedule loop", __FUNCTION__);
+    }
+    while ((job = ctx->queue->alloced_pop ()) != nullptr) {
+        flux_log (ctx->h, LOG_DEBUG, "jobid (%ju): %s",
+                  (intmax_t)job->id, job->schedule.R.c_str ());
+        if (schedutil_alloc_respond_R (ctx->h, job->msg,
+                                       job->schedule.R.c_str (), NULL) < 0) {
+            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_R",
+                            __FUNCTION__);
+        }
+    }
+}
+
+extern "C" void jobmanager_free_cb (flux_t *h, const flux_msg_t *msg,
+                                    const char *R, void *arg)
+{
+    uint32_t userid;
+    flux_jobid_t id;
+    qmanager_ctx_t *ctx = (qmanager_ctx_t *)arg;
+    std::shared_ptr<job_t> job;
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
+        return;
+
+    flux_log (h, LOG_INFO, "free requested by user (%u).", userid);
+
+    if (schedutil_free_request_decode (msg, &id) < 0) {
+        flux_log_error (h, "%s: schedutil_free_request_decode",
+                        __FUNCTION__);
+        return;
+    }
+    if (ctx->queue->remove (id)) {
+        flux_log_error (h, "%s: remove", __FUNCTION__);
+        return;
+    }
+    if (ctx->queue->run_sched_loop ((void *)ctx->h, true) < 0) {
+        // TODO: Need to tighten up anomalous conditions
+        // returned with a negative return code
+        // (e.g., unsatisfiable jobs).
+        flux_log (ctx->h, LOG_DEBUG,
+                  "%s: return code < 0 from schedule loop", __FUNCTION__);
+    }
+    if (schedutil_free_respond (h, msg) < 0) {
+        flux_log_error (h, "%s: schedutil_free_respond", __FUNCTION__);
+    }
+    while ((job = ctx->queue->alloced_pop ()) != nullptr) {
+        flux_log (ctx->h, LOG_DEBUG, "jobid (%ju): %s",
+                  (intmax_t)job->id, job->schedule.R.c_str ());
+        if (schedutil_alloc_respond_R (ctx->h, job->msg,
+                                       job->schedule.R.c_str (), NULL) < 0) {
+            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_R",
+                            __FUNCTION__);
+        }
+    }
+}
+
+static void jobmanager_exception_cb (flux_t *h, flux_jobid_t id,
+                                     const char *t, int s, void *a)
+{
+    return;
+}
+
+static qmanager_ctx_t *qmanager_new (flux_t *h)
+{
+    int queue_depth = 0;
+    qmanager_ctx_t *ctx = NULL;
+
+    if (!(ctx = new (std::nothrow) qmanager_ctx_t ())) {
+        errno = ENOMEM;
+        goto out;
+    }
+    ctx->h = h;
+    if (!(ctx->queue = create_queue_policy ("fcfs", "module"))) {
+        flux_log_error (h, "%s: create_queue_policy", __FUNCTION__);
+        goto out;
+    }
+    if (schedutil_hello (h, jobmanager_hello_cb, ctx) < 0) {
+        flux_log_error (h, "%s: schedutil_hello", __FUNCTION__);
+        goto out;
+    }
+    if (schedutil_ready (h, "single", &queue_depth)) {
+        flux_log_error (h, "%s: schedutil_ready", __FUNCTION__);
+        goto out;
+    }
+    if (!(ctx->ops = schedutil_ops_register (h,
+                                             jobmanager_alloc_cb,
+                                             jobmanager_free_cb,
+                                             jobmanager_exception_cb, ctx))) {
+        flux_log_error (h, "%s: schedutil_ops_register", __FUNCTION__);
+        goto out;
+    }
+out:
+    return ctx;
+}
+
+static void qmanager_destroy (qmanager_ctx_t *ctx)
+{
+    if (ctx) {
+        int saved_errno = errno;
+        std::shared_ptr<job_t> job;
+        while ((job = ctx->queue->pending_pop ()) != nullptr)
+            flux_respond_error (ctx->h, job->msg, ENOSYS, "unloading");
+        while ((job = ctx->queue->complete_pop ()) != nullptr)
+            flux_respond_error (ctx->h, job->msg, ENOSYS, "unloading");
+        delete ctx->queue;
+        ctx->queue = NULL;
+        schedutil_ops_unregister (ctx->ops);
+        free (ctx);
+        errno = saved_errno;
+    }
+}
+
+
+/******************************************************************************
+ *                                                                            *
+ *                               Module Main                                  *
+ *                                                                            *
+ ******************************************************************************/
+
+extern "C" int mod_main (flux_t *h, int argc, char **argv)
+{
+    int rc = -1;
+    try {
+        qmanager_ctx_t *ctx = NULL;
+        if (!(ctx = qmanager_new (h)))
+            flux_log_error (h, "%s: qmanager_new", __FUNCTION__);
+        if ((rc = flux_reactor_run (flux_get_reactor (h), 0)) < 0)
+            flux_log_error (h, "%s: flux_reactor_run", __FUNCTION__);
+        qmanager_destroy (ctx);
+    }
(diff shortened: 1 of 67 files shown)


dongahn:
@SteVwonder, @garlick and maybe @grondo: This PR reached a reasonable stopping state to review this PR. The testing coverage went down a bit because this is actually an intermediate work before our end-of-July milestone. As such contains fair amounts of placeholder codes.

garlick:
Just some general comments, initially:
* You have two modules named `qmanager` and `resource`.  Should they be named with a common prefix to indicate that they are part of a single scheduler implementation, like `sched-full-qmanager` and `sched-full-resource`?  (substitute something more creative and cool sounding for "full" :-)
* Your qmanager  `jobmanager_hello_cb` needs to be filled in so that resources that are already allocated with the module loads can be marked allocated.
* Your qmanager `jobmanager_exception_cb` needs to be filled in so that `flux job cancel` works for jobs that have an outstanding alloc request.
* functions in qmanager.cpp look a bit chatty with the LOG_INFO messages on every message.  Perhaps LOG_DEBUG?
* Have you tried running this end to end with flux-core?
* I see a ctx->queue->insert (job) call that isn't checked for failure.  What happens to the broker if this hits a c++ exception, or am I missing how exceptions are handled?)

dongahn:
Thank you for the quick review. I note that this is preliminary work so some of the logic only has a place holder. More will come as part of the next sprint.

> You have two modules named qmanager and resource. Should they be named with a common prefix to indicate that they are part of a single scheduler implementation, like sched-full-qmanager and sched-full-resource? (substitute something more creative and cool sounding for "full" :-)

I can certainly do this. I was initially a bit ambivalent because resource can be used as a stand alone service independent of qmanager. I will create an issue to have a bit more discussion. But it doesn't have to be a part of this PR, does it?

> Your qmanager jobmanager_hello_cb needs to be filled in so that resources that are already allocated when the module loads can be marked allocated.

Yes. This will be done as part of the end of July milestone. And this will be further utilized for resilience as logged with a scheduler resiliency ticket even later.


> Your qmanager jobmanager_exception_cb needs to be filled in so that flux job cancel works for jobs that have an outstanding alloc request.

Same as above.

> functions in qmanager.cpp look a bit chatty with the LOG_INFO messages on every message. Perhaps LOG_DEBUG?

Ok. Good feedback. I will change some of the messages to LOG_DEBUG as part of this PR.

> Have you tried running this end to end with flux-core?

Yes. I had some manual testing. And the test case demonstrates this. 


> I see a ctx->queue->insert (job) call that isn't checked for failure. What happens to the broker if this hits a c++ exception, or am I missing how exceptions are handled?

I will double check. I generally don't want to raise an exception unless necessary. And I think qmanager also needs the top level exception catch clause. (Will add that)

garlick:
> I think qmanager also needs the top level exception catch clause. (Will add that)

That sounds like the right thing!

All good on filling in stuff and considering the name later as far as I'm concerned.

dongahn:
Thanks @garlick. I should say that the new rpc based scheduler interface and infrastructure  made my job far easier! Great work.

dongahn:
> I will double check. I generally don't want to raise an exception unless necessary. And I think qmanager also needs the top level exception catch clause. (Will add that)

I am using stl map and a few of its methods: `erase`, `insert`, `find` and `empty`. The only methods that can throw an exception would be `erase` and `find` when used with the key as their method arguments. But even then their exceptions are essentially transparently passing an exception that can be thrown from the comparator.

Now, I just used the default comparator which is `std::less<key>` which doesn't throw exceptions. So all of these should be pretty much exception-free. But there could be `std::bac_alloc` exceptions and etc, which can be thrown from object creation time etc, so I will wrap the entire `mod_main` with

```c++
catch (std::exception &e) {
}
```

dongahn:
OK. Rebased to the upstream/master and then pushed.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -0,0 +1,182 @@
...
+    /*! Append a job into the internal pending-job queue.
+     *
+     *  \param job       a shared pointer pointing to a job_t object.
+     *  \return          0 on success; -1 on error.
+     *                       EINVAL: invalid argument.
+     */
+    int insert (std::shared_ptr<job_t> job);
+
+    /*! Remove a job whose jobid is id from any queue (e.g., pending queue
+     *  or running queue.)
```
SteVwonder:
```suggestion
     *  or alloced queue.)
```

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+    default:
+        break;
+    }
+    rc = 0;
+out:
+    return rc;
+}
+
+
+std::map<uint64_t, flux_jobid_t>::iterator queue_policy_base_impl_t::
```
SteVwonder: Can you add a comment that the return value is the next element in the queue?  It makes sense after looking up the return semantics of `std::map::erase`, but being mostly unfamiliar with the C++ STL, this behavior was initially surprising to me.

dongahn:
+



+std::map<uint64_t, flux_jobid_t>::iterator queue_policy_base_impl_t::



> Can you add a comment that the return value is the next element in the queue? It makes sense after looking up the return semantics of std::map::erase, but being mostly unfamiliar with the C++ STL, this behavior was initially surprising to me.



Good point. I will add a comment there. Incorrectly removing an element while iterating through a STL container is one of the common programming mistakes, and this semantics comes in handy to implement it correctly.

qmanager/modules/qmanager.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+extern "C" void jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
+                                     const char *jobspec, void *arg)
+{
+    uint32_t userid;
+    qmanager_ctx_t *ctx = (qmanager_ctx_t *)arg;
+    std::shared_ptr<job_t> job = std::make_shared<job_t> ();
+
+    if (flux_msg_get_userid (msg, &userid) < 0)
```
SteVwonder: I am probably just misunderstanding how this interface works, but IIUC, the `alloc` RPCs will always come from the `jobmanager`.  Doesn't this mean that the `userid` will always be whichever user flux is running as?

resource/config/system_defaults.hpp
```diff
@@ -30,6 +30,7 @@ namespace Flux {
 namespace resource_model {
 namespace detail {
     const uint64_t SYSTEM_DEFAULT_DURATION = 43200;
+    const uint64_t SYSTEM_MAX_DURATION = 604800;
```
SteVwonder: Can you add a comment next to this along the lines of `// 7 days`?  And a similar one for the above duration `// 12 hours`?   Since they both have two trailing zeros and are offset in space, they appear at a quick glance to be of similar magnitude (not a factor of 10 different).

t/sharness.d/sched-sharness.sh
```diff
@@ -2,20 +2,21 @@
...
+FLUX_QMANAGER_RC_NOOP=1
 if test -n "$FLUX_SCHED_TEST_INSTALLED"; then
   # Test against installed flux-sched, installed under same prefix as
   #   flux-core.
-  # (Assume sched modules installed under PREFIX/lib/flux/modeuls/sched)
+  # (Assume sched modules installed under PREFIX/lib/flux/modules)
   # (We also support testing this when sched is installed
   # another location, but only via make check)
-  FLUX_MODULE_PATH_PREPEND=$(which flux | sed -s 's|/bin/flux|/lib/flux/modules/resource|'):${FLUX_MODULE_PATH_PREPEND}
+  # FLUX_MODULE_PATH_PREPEND=$(which flux | sed -s 's|/bin/flux|/lib/flux/modules/resource|'):${FLUX_MODULE_PATH_PREPEND}
```
SteVwonder: Was it intended for this to remain commented out?

t/t1001-qmanager-basic.t
```diff
@@ -0,0 +1,107 @@
...
+    flux job wait-event -t 2.5 ${jobid} exception > exception.1.out &&
+    test_debug "flux job eventlog ${jobid}" &&
+    grep "type=\"exec\"" exception.1.out &&
+    grep "mock initialization exception generated" exception.1.out &&
+    flux job wait-event -qt 2.5 ${jobid} clean &&
+    flux job eventlog ${jobid} > eventlog.${jobid}.out &&
+    test_must_fail grep "finish" eventlog.${jobid}.out
+'
+
+test_expect_success 'qmanager: exception during initialization is supported' '
```
SteVwonder: this looks to be a duplicate of the previous test

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -0,0 +1,182 @@
...
+     *
+     *  \param h         Opaque handle. How it is used is an implementation
+     *                   detail. However, when it is used within a Flux's
+     *                   service module such as qmanager, it is expected
+     *                   to be a pointer to a flux_t object.
+     *  \param use_alloced_queue
+     *                   Boolean indicating if you want to use the
+     *                   allocated job queue or not. This affects the
+     *                   alloced_pop method.
+     *  \return          0 on success; -1 on error.
```
SteVwonder: In one of the implementations, `run_sched_loop` returns `rc1 + rc2`.  Maybe update this documentation to say `< 0 on error`?

qmanager/modules/qmanager.cpp
```diff
@@ -122,14 +122,14 @@ extern "C" void jobmanager_free_cb (flux_t *h, const flux_msg_t *msg,
         return;
     }
     if (ctx->queue->run_sched_loop ((void *)ctx->h, true) < 0) {
-        flux_log (ctx->h, LOG_INFO,
+        flux_log (ctx->h, LOG_DEBUG,
```
SteVwonder: A `< 0` rc from the scheduler loop is an error, right?  If so, I think this should actually be `LOG_ERR`.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -0,0 +1,182 @@
...
+     *  must determine its queueing policy. For example, a pedantic FCFS
+     *  class should only iterate through the first set of jobs in the
+     *  pending-job queue which it can schedule using the underlying
+     *  resource match infrastructure.
+     *
+     *  \param h         Opaque handle. How it is used is an implementation
+     *                   detail. However, when it is used within a Flux's
+     *                   service module such as qmanager, it is expected
+     *                   to be a pointer to a flux_t object.
+     *  \param use_alloced_queue
```
SteVwonder: What is your plan for this variable going forward?  AFAICT, in this PR, it is only ever set to true.

It seems like, for the sake of generalization, this variable should evolve (in a future PR) to encompass all/most scheduler configuration options (e.g., `queue-depth`, `reservation-depth`).  Does that align with what you were thinking?

SteVwonder:
Oh.  One other question I had while reading through the code: it wasn't clear to me the need for the duplication of all of the interfaces for the CLI support.  It'll probably be clear once there is an implementation filled out in there, but in the meantime, can you briefly explain your plans for the CLI piece?  Will it still use RPCs etc to interface with the `resource` module, or are you planning on running the `resource-query` (or similar) tool(s) as subprocesses of the `qmanager`?

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    std::shared_ptr<job_t> job;
+    flux_jobid_t id;
+    if (m_alloced.empty ())
+        return nullptr;
+    id = m_alloced.begin ()->second;
+    if (m_jobs.find (id) == m_jobs.end ())
+        return nullptr;
+    job = m_jobs[id];
+    m_alloced.erase (job->t_stamps.running_ts);
```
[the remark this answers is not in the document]
dongahn: Actually not. "allocated job queue" serves as a bit like a pseudo queue. The fact that a job is popped from the allocated job queue must not mean the job itself has completely graduated from the main queues that the queue policy layer manages. (Whereas, if you pop a job from the pending or complete queues, the jobs will graduate.  

"allocated job queue" is introduced so that this layer can remain "jobmanager internal" agnostic. For example, I didn't want this queue-policy layer to directly respond to `alloc` RPCs; but rather provide primitives for `qmanager` to do this job. "allocated job queue" supports this separation of concern.

qmanager/policies/base/queue_policy_base_impl.hpp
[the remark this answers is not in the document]
SteVwonder: >  I somehow responded to this via an email. 
That was my fault.  I accidentally made it as a top-level comment, hit delete, and then re-did it as a review comment.  I forgot that comments spawned emails.  Sorry about that.

t/sharness.d/sched-sharness.sh
dongahn: I can remove this of course ;-)

t/t1001-qmanager-basic.t
dongahn: Ugh.. probably copy-and-paste error from a flux-core test case. Will clean up.

qmanager/modules/qmanager.cpp
SteVwonder: > Maybe I can add "TODO" comment to satisfy your review comment?

:+1: works for me

qmanager/modules/qmanager.cpp
SteVwonder: > But for now, what happens if an alloc RPC is directly issued by someone other than the job-manager? We probably at least log it for now?

Yeah, that's a good question.  Ultimately, should we just ignore those requests (or register the handler in such a way that only instance owners can submit them)?

dongahn:
> Oh. One other question I had while reading through the code: it wasn't clear to me the need for the duplication of all of the interfaces for the CLI support. It'll probably be clear once there is an implementation filled out in there, but in the meantime, can you briefly explain your plans for the CLI piece? Will it still use RPCs etc to interface with the resource module, or are you planning on running the resource-query (or similar) tool(s) as subprocesses of the qmanager?

Great question! Yes, ultimately I want to add `queue-query` or similar as a way to more aggressively test our queue policy layer. This will be a CLI utility which isn't hooked up to flux-core components at all. (like `resource-query` did) 

And this will help us easily create more advanced queueing policies later down the road... (including greedy algorithms they we talked about in the past). 

Even more longer-term picture: All of these utilities will ensure that the main infrastructures designed and developed as part of flux-sched have good separations of concerns so much that our core scheduling components can be used by other resource manager (e.g., Kubes) for both HPC and Cloud alike. And this is why I thought it would be a good idea to have cool names for our scheduler project. Issue https://github.com/flux-framework/flux-sched/issues/485

 When it is used by flux, our solutions will be named

Flux yyy (to be named)

If this is used by Kubernetes, it will be 

Kubernetes yyy (to be named)

etc.

qmanager/policies/base/queue_policy_base_impl.hpp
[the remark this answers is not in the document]
SteVwonder: > Actually not. "allocated job queue" serves as a bit like a pseudo queue. The fact that a job is popped from the allocated job queue must not mean the job itself has completely graduated from the main queues that the queue policy layer manages.

Ah. That makes sense.  Thanks!

> (Whereas, if you pop a job from the pending or complete queues, the jobs will graduate.

One quick follow up, it looks like jobs are only added to `m_jobs` in the `insert` method (which is called when an `alloc` request comes in and a job becomes `PENDING`).  If they are `erase`d when `pending_pop` is called, how does the lookup for that job in `m_jobs` that takes place during `to_complete` and `complete_pop` succeed?    No worries if that isn't hooked up yet, just trying to put the pieces together in my mind.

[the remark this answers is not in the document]
dongahn: Hmmm. The jobs erased with `pending_pop ()` shouldn't make into the complete queue so that lookup shouldn't occur within `complete_pop ()` 

Ah. I think the confusion might be whether `pending_pop ()` is used as part of the schedule loop... No, it isn't the interface that the scheduler loop uses to iterate through the pending job queue. The queue itself is directly exposed to `run_sched_loop` (protected class member of `queue_policy_base_t` class.)

`pending_pop` is an external interface that is currently only used by qmanager to remove all of the pending jobs from the pending job queue in its destructor.

If the public interface name is too confusing, I am willing to change the interface name to something else. Let me know.

qmanager/modules/qmanager.cpp
dongahn: > Ultimately, should we just ignore those requests (or register the handler in such a way that only instance owners can submit them)?

Can we actually implement this using the current capabilities within flux-core?

SteVwonder: Someone more knowledgeable than me, please correct me if I'm wrong, but after some digging, it looks like, by default, only messages sent by the instance owner will cause a message to be handled by the callback.  You have to specifically add the `FLUX_ROLE_USER` rolemask when registering the callback to enable messages from other users.

Source: https://github.com/flux-framework/flux-core/issues/1260 & https://github.com/garlick/flux-core/commit/009b7ea9c32bb456e6c85c853ccb8ff11bb52e48

I initially thought the default was the opposite.  So it looks like nothing needs to be done here in terms of security/handling, but the `get_userid` function will only ever return the instance owner's id.

qmanager/policies/base/queue_policy_base_impl.hpp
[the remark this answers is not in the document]
SteVwonder: >  I think the confusion might be whether pending_pop () is used as part of the schedule loop
> pending_pop is an external interface that is currently only used by qmanager to remove all of the pending jobs from the pending job queue in its destructor.

Ah. That makes total sense now.  Thanks!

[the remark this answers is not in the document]
dongahn: Ok. Thanks. You have good points here.  I will add some oxygen comments on those public queue pop interfaces to reduce confusion.

dongahn:
Thank you for the research @SteVWonder. I will create a ticket to add role based control since this will be a part of our multiuser suppprt within sched.

dongahn:
@SteVwonder: I pushed some more commits to address all of your review comments. This revision has changes for both yours and @garlick. So, if Travis turns green and you are okay with the latest changes, I will squash the latest commits and then this PR should be good to go. Thanks!

dongahn:
OK. I squashed those later commits. Once Travis turns green, this is good to be merged @SteVwonder. Thanks.

dongahn:
@grondo: Thanks. Actually, I was concerned about these per-job messages myself as well. And I plan to go over the logging messages in a later PR across both qmanager and resource. Given that this is an intermediate PR, can this go in as is and then a later PR addresses this issue for all?

grondo:
Fine with me, I didn't consider them mandatory changes.

dongahn:
Thanks @grondo and @SteVwonder!

dongahn:
Yeah! Thanks!

PR #490, opened 2019-07-11
resource-query: add support for properties

tpatki:
WIP PR that adds support for set-property and get-property in resource-query. 
TODO: Add similar support to [flux-resource script](https://github.com/flux-framework/flux-sched/blob/master/t/scripts/flux-resource) and `resource-match`.


resource/utilities/command.cpp
```diff
@@ -209,6 +213,75 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+		goto done;
+	}
+
+	if (rc != 0) {
+		cerr << "ERROR: error encountered while getting property. " << endl;
+		cerr << "ERROR: " << strerror (errno) << endl;
+	}
+
+ done:
+	return 0;
```
dongahn: Same comment with `rc`.

resource/utilities/command.cpp
```diff
@@ -209,6 +213,75 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+	std::map<std::string, std::vector <vtx_t>>::const_iterator it = ctx->db.by_path.find(resource_path);
+	if (it == ctx->db.by_path.end())
+		out << "Could not find path in resource graph." << endl;
+	else {
+		vtx_t v = it->second[0];
+		if (ctx->db.resource_graph[v].properties.size() == 0)
+			out << "No properties were found for " << resource_path << ". " << endl;
+		else {
+			std::map<std::string, std::string>::const_iterator p_it;
+			for (p_it=ctx->db.resource_graph[v].properties.begin(); p_it!=ctx->db.resource_graph[v].properties.end(); p_it++)
```
dongahn: Line break; function calling style; and also I will appreciate if you have spaces around the assignment statements.

t/t3012-resource-properties.t
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
```
dongahn: I would suggest to have some mal-formed inputs to ensure your code behaves as expected.

dongahn:
@tpatki: looks like a great start. Please see my in-line comments. I think the `resource` infrastructure overall don't have good support for resource properties yet. (e.g., emitters probably don't write properties -- even if they do, I don't think we have test coverage.) So could you also please open a ticket with "tighten up the property support within resource"? 

Thank you for being willing to share your WIP!

dongahn:
Forget to mention. You probably want to break the commit into at least two commits: one for the code addition and the other for test addition.

tpatki:
Thank you so much for the constructive comments, Dong! Yes, I was wondering about error checks too, so thanks for the feedback there, I'll add those in. And once that's done, I'll add in the script/CB. 
(It has been slower than expected but an excellent lesson for me in terms of time estimation).

dongahn:
> And once that's done, I'll add in the script/CB.

I'd suggest you would make an effort to make this a real PR first. And then, the resource-match part as another PR.

tpatki:
> I'd suggest you would make an effort to make this a real PR first. And then, the resource-match part as another PR.

Sounds good, will do.

tpatki:
@dongahn: Please take a look. 
I'll open a separate issue for the script/resource-match and get started on that soon.

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+		cerr << "ERROR: malformed command" << endl;
+		return 0;
+	}
+
+	string resource_path = args[1];
+	string property_key, property_value;
+	ostream &out = (ctx->params.r_fname != "") ? ctx->params.r_out : cout;
+	int eq_count = count (args[2].begin (), args[2].end (), '=');
+
+	if (eq_count == 1) {		/* Exactly 1 '=' sign, try to extract LHS and RHS */
```
dongahn: Perhaps we want to support a value that may contain `=`? I can imagine a value string containing `=` like 

```
perf-class=mem=1:cpu=2
```

Then, `mem=1:cpu=2` will be further parsed by your match callback to decide which subclass you want to use?

dongahn: A knit: over 80 characters :-)

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+		property_key = args[2].substr (0, args[2].find ("="));
+		property_value = args[2].substr (args[2].find ("=") + 1);
+	}
+
+	if ((eq_count != 1) || property_key.empty () || property_value.empty () ) {
+		out << "Incorrect input format. Please use "
```
dongahn: I slightly prefer breaking the output into two lines.

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+				"`set-property <resource> PROPERTY=VALUE`." << endl;
+		return 0;
+	}
+
+	std::map<std::string, std::vector <vtx_t>>::const_iterator it =
+			ctx->db.by_path.find (resource_path);
+
+	if (it == ctx->db.by_path.end ()) {
+		out << "Couldn't find path" << resource_path <<
```
dongahn: Seems like you will need whitespaces in the surrounding strings.

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+			"in resource graph." << endl;
+	}
+	else {
+		vtx_t v = it->second[0];
+		std::pair<std::map<string,string>::iterator,bool> ret;
```
dongahn: Seems like a deadwood?

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+	}
+
+	string resource_path = args[1];
+	ostream &out = (ctx->params.r_fname != "") ? ctx->params.r_out : cout;
+
+	std::map<std::string, std::vector <vtx_t>>::const_iterator it =
+			ctx->db.by_path.find (resource_path);
+
+	if (it == ctx->db.by_path.end ()) {
+		out << "Could not find path" << resource_path <<
```
dongahn: Also need spaces in the surrounding strings.

resource/utilities/command.cpp
```diff
@@ -209,6 +213,91 @@ int cmd_cancel (resource_context_t *ctx, vector<string> &args)
...
+	else {
+		vtx_t v = it->second[0];
+		if (ctx->db.resource_graph[v].properties.size () == 0) {
+			out << "No properties were found for " << resource_path
+				<< ". " << endl;
+		}
+		else {
+			std::map<std::string, std::string>::const_iterator p_it;
+			for (p_it = ctx->db.resource_graph[v].properties.begin ();
+						p_it != ctx->db.resource_graph[v].properties.end (); p_it++)
```
dongahn: This appears to exceed 80 characters in my web browser. Yours may render this differently...

dongahn:
Tapasya, please close. I made a few minor comments. Please take a look. For the next round, you may want to have your changes as separate commits with a caveat that they will be squashed before this PR lands. That way, reviewers don't have to review the whole things when the PR only requires minor revision. Not a huge issue with this PR since it is pretty small. But this can help quiet bit for large PRs. Thanks!

resource/utilities/command.cpp
tpatki: How about we do this after the first version of variation-scheduler is in?
As in, we get that first ECP milestone out and then add:
(1) support for adding multiple properties at once (egmem/cpu examples or comma-separated as opposed to writing out "set-property" every time), 
(2) support for get-property <resource> PROPERTY

tpatki:
@dongahn: Thanks, Dong, let me make the changes you suggested. 
Looks like 80 chars on my eclipse editor is messed up, let me cap off at 70- or so, that way we have more readable code. Is there a nice automated way to check for this?

resource/utilities/command.cpp
dongahn: Adding multi-subclass capability support within variation scheduler is later okay by me. But maybe this code can be slightly revised to support values that contain `=` to position the future work? (Since you will have more commits anyway?)

dongahn:
I vaguely remember @SteVwonder had clang based style checker... I maybe wrong though.

resource/utilities/command.cpp
tpatki: Sure. 
The question will be, when is having more than one`=` ok, and when it isn't. As in, where do we check and reject the malformed inputs: in the matcher or in the set-property?
`perf-class=mem=1:cpu=2` versus
`perf-class=mem=5==cpu=2:?garbage`

dongahn: I would take everything that comes after the first occurrence of `=` *valid* at this point of execution as far as it is not empty. That is, how to interpret the value object is really up to the user of that value string. If a garbage is there, the user will raise an exception.

tpatki: Got it, and makes sense. The matcher can reject input that it cannot parse, I suppose. 
That's an easy fix (`= 1` to `>= 1`) should do it.

dongahn: > That's an easy fix (= 1 to >= 1) should do it.

Yes. However, a more efficient trick would be just to find the first occurrence instead of count all of them? 

```c++
   string key;
   string val;
   size_t pos = 0;
   std::string delim = "=";

    if ((pos = args[2].find (delim)) == std::string::npos
         || pos == 0 || pos == args[2].length () - 1) {
            // input error 
    } else {
        key = args[2].substr (0, pos);
        val = args[2].substr (pos+1);
    }
```
The code not tested...

tpatki:
@dongahn: Made the changes you requested. 
Let me know what the process to squash is for this to get merged in when it's ready. Thanks!

resource/utilities/command.cpp
```diff
@@ -223,29 +223,27 @@ int cmd_set_property (resource_context_t *ctx, std::vector<std::string> &args)
 	string resource_path = args[1];
 	string property_key, property_value;
 	ostream &out = (ctx->params.r_fname != "") ? ctx->params.r_out : cout;
-	int eq_count = count (args[2].begin (), args[2].end (), '=');
+	size_t pos = args[2].find ('=');
```
dongahn: @tpatki: Thank you. The code looks great. I *think* there are problems with our editor though. I just realize the indentation is much larger than the rest of the code. Could you please align the indentation with the rest of the code, so the code will look as if one person wrote?

resource/utilities/command.cpp
```diff
@@ -223,29 +223,27 @@ int cmd_set_property (resource_context_t *ctx, std::vector<std::string> &args)
...
 
-	if (eq_count == 1) {		/* Exactly 1 '=' sign, try to extract LHS and RHS */
-		property_key = args[2].substr (0, args[2].find ("="));
-		property_value = args[2].substr (args[2].find ("=") + 1);
+	if (pos == 0 || (pos == args[2].size() - 1) || pos == string::npos) {
```
dongahn: A nit. Space after the `size ()` method.

resource/utilities/command.cpp
```diff
@@ -289,7 +287,7 @@ int cmd_get_property (resource_context_t *ctx, std::vector<std::string> &args)
 		else {
 			std::map<std::string, std::string>::const_iterator p_it;
 			for (p_it = ctx->db.resource_graph[v].properties.begin ();
-						p_it != ctx->db.resource_graph[v].properties.end (); p_it++)
+				p_it != ctx->db.resource_graph[v].properties.end (); p_it++)
 							out << p_it->first << "=" << p_it->second << endl;
```
dongahn: The indentation for this statement is way off. Eclipse?

dongahn:
@tpatki: Just a few minor final reviews in-lined. Once those are addressed and the commits are squashed, this PR LGTM.

dongahn:
@tpatki: once the above issues are handled, this PR will need to be rebased with the upstream and pushed.

```console
$ git remote add upstream git@github.com:flux-framework/flux-sched.git
$ git fetch upstream
$ git rebase upstream/master
```

If all is good (no conflict is expected), then force a push.

tpatki:
@dongahn: Yes, will clean this up shortly this afternoon. Sorry, was out of office this morning, just getting back.

tpatki:
@dongahn: I fixed the indentation, and I hope that I did the rebase/squash correctly. Let me know if this was ok.

dongahn:
Thanks. 

- Seems you lost your last commit splits. Now your single commit has both code change and test.

- Please rebase your PR with upstream master. You can find this by looking at the warning  above "This beach is out of sync with the base branch"

Otherwise LGTM.

tpatki:
That's weird, I did rebase (that's how I squashed commits). Trying again.

dongahn:
> That's weird, I did rebase (that's how I squashed commits). Trying again.

Maybe you didn't rebase with upstream but origin? 

``` console
$ git remote add upstream git@github.com:flux-framework/flux-sched.git
$ git fetch upstream
$ git rebase upstream/master
```

Should rebase your PR against the upstream master.

dongahn:
I also saw the name of the PR has [WIP] on it. I guess it's the time to remove that tag.

tpatki:
@dongahn: let me know if all looks good now.

dongahn:
> Seems you lost your last commit splits. Now your single commit has both code change and test.

Hmmm this needs to be addressed?

tpatki:
Sorry I had missed that -- done.

dongahn:
@tpatki: Makefile.am change seems to belong to the test commit.

dongahn:
Thanks!

Issue #470, opened 2019-06-14
Add state reconstruction support to resource module


dongahn:
This will be one of the work items for the next year. Looking at the scheduler interface, it feels like we can get some good mileage by using a scheduler state reconstruction scheme instead of the traditional checkpoint/restart-based resiliency.

At initialization, the resource match service can walk through each and every currently allocated (or reserved) job and then fetch `R` associated with each job. Then, this service can simply update the resource graph data store using this information. 

We don't currently have ways to update the resource data store in this way. I have to believe it is quite doable. We will need some initial feasibility study though. One thing that isn't immediately clear to me is how easy or difficult we can reconstruct some of the metadata like pruning filter states etc in addition to the resource vertex allocation data. 

The traditional technique would be the checkpoint and restart thereby the scheduler periodically dumps the resource data store state somewhere and restarts when we need it. But given that much of the job state is already kep at the job-manager and kvs, it seems the traditional approach would be an overkill in terms of performance. And perhaps more importantly, this can lead to state inconsistency between the scheduler state and job-manager and kvs state. Of course, with this, we will loose the reservation (for backfilling) information that was made at the last scheduler loop as we won't store this into the kvs job schema. But I have to believe this can be accurate recomputed as far as we can accurately reconstruct the currently allocated jobs state.

BTW, this assumes that job-manager and kvs will be resilient. But this is something we must solve in this timeframe anyway.

dongahn:
This also tells me that `R` here must be *full*. Without `JFG` section (IOW, only with `R_lite`), we won't be able to reconstruct the state. Finally, it feels like we really want to put some thoughts for https://github.com/flux-framework/flux-sched/issues/444 so that we can use this JGF reader work not only for nested scheduling support but also this kind of reconstruction-based resiliency scheme.

dongahn:
@garlick or @grondo: It appears that the hello callback invoked by job-manager is currently only being passed an `R` string per each allocated job. Was there any reason that their jobids are not passed as well? The reason I'm asking is that all of the fail-over cases for the scheduler components (`qmanager` and `resource`) would need the jobid per each existing allocated job to be efficient with their state reconstruction schemes.

garlick:
Part of the design for minimizing complexity in the scheduler was to make it a "resource allocator" that is not job aware.  I think we could attach the jobid trivially, if it's needed.

Note: eventually we will return R fragments, not the original full R, when parts of the job terminate before others.   For now it's the full allocated R.

[...]

PR #493, opened 2019-07-12
qmanager: Add hello/exception callback support

### LINKED ISSUE #492: Exception and hello callback implementation within qmanager
From @garlick, as part of PR #481, 

- Your qmanager jobmanager_hello_cb needs to be filled in so that resources that are already allocated when the module loads can be marked allocated.
- Your qmanager jobmanager_exception_cb needs to be filled in so that flux job cancel works for jobs that have an outstanding alloc request.

dongahn:
This PR fills in hello callback and exception callback as required by job-manager. The hello callback allows `qmanager` to fill in its running queue as part of load sequence and the exception callback allows for canceling the pending jobs (i.e., jobs with pending alloc requests).

Note that the hello callback is working in progress towards full resilience. More comprehensive solution with respect to `resource` will require a solution for Issue #470. This isn't planned until later this year.

Resolve #492.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -162,21 +162,22 @@ static qmanager_ctx_t *qmanager_new (flux_t *h)
         flux_log_error (h, "%s: create_queue_policy", __FUNCTION__);
         goto out;
     }
-    if (schedutil_hello (h, jobmanager_hello_cb, ctx) < 0) {
-        flux_log_error (h, "%s: schedutil_hello", __FUNCTION__);
+    if (!(ctx->ops = schedutil_ops_register (ctx->h,
+                                             jobmanager_alloc_cb,
+                                             jobmanager_free_cb,
+                                             jobmanager_exception_cb, ctx))) {
+        flux_log_error (ctx->h, "%s: schedutil_ops_register", __FUNCTION__);
+        goto out;
+    }
+    if (schedutil_hello (ctx->h, jobmanager_hello_cb, ctx) < 0) {
+        flux_log_error (ctx->h, "%s: schedutil_hello", __FUNCTION__);
         goto out;
     }
     if (schedutil_ready (h, "single", &queue_depth)) {
         flux_log_error (h, "%s: schedutil_ready", __FUNCTION__);
         goto out;
     }
-    if (!(ctx->ops = schedutil_ops_register (h,
-                                             jobmanager_alloc_cb,
-                                             jobmanager_free_cb,
-                                             jobmanager_exception_cb, ctx))) {
-        flux_log_error (h, "%s: schedutil_ops_register", __FUNCTION__);
-        goto out;
-    }
+
 out:
     return ctx;
 }
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -145,7 +145,20 @@ extern "C" void jobmanager_free_cb (flux_t *h, const flux_msg_t *msg,
 static void jobmanager_exception_cb (flux_t *h, flux_jobid_t id,
                                      const char *t, int s, void *a)
 {
-    return;
+    std::shared_ptr<job_t> job;
+    qmanager_ctx_t *ctx = (qmanager_ctx_t *)a;
+
+    if (s > 0 || (job = ctx->queue->lookup (id)) == nullptr
+        || !job->is_pending ())
+        return;
+    if (ctx->queue->remove (id) < 0) {
+        flux_log_error (h, "%s: remove job (%ju)", __FUNCTION__, (intmax_t)id);
+        return;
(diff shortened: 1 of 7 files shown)


garlick:
Recovery of `userid`, `t_submit`, `priority` from the job eventlog within the hello phase looks good to me, and so does including the jobid in the free callback.  It might be useful to move the eventlog stuff into libschedutil alongside the code that reads in `R` from the KVS rather than embedding it in qmanager.

19c37e9 is the heart of this PR and doesn't have any commit description, so I would suggest adding a few paragraphs of explanation.  Splitting out the changes to libschedutil to their own commit would be helpful in case we backport to flux-core, or to enable standalone testing.

dongahn:
> It might be useful to move the eventlog stuff into libschedutil alongside the code that reads in R from the KVS rather than embedding it in qmanager.

Excellent suggestion! Other schedulers may need this through `libschedutil`. Will do.

> Splitting out the changes to libschedutil to their own commit would be helpful in case we backport to flux-core, or to enable standalone testing.

I thought about this. But `qmanager` will no longer build at that commit. If having self contained commit is more important than making each commit buildable, I can do this. It seems this is one of those cases.

garlick:
OK by me then.

Incidentally, my comment about standalone testing applies to the first suggestion (move eventlog code from qmanager to libschedutil), not the second one about splitting the commit.  Sorry if that was confusing/silly.

dongahn:
Maybe, I can move the event code stuff into `libschedutil` without the API change -- one commit that solely contains schedutil changes alone. 

And then the next commit can contain the bulk of the hello callback support and two-line API changes in `libschedutil`. This way, at least back porting becomes an inch easier...

grondo:
Should we think about exporting libeventlog from libflux-core.so (or its own dso)? It seems more general purpose and less likely to change than libschedutil.

Also, are the changes here going to be easy to resolve with flux-framework/flux-core#2226?

dongahn:
> Should we think about exporting libeventlog from libflux-core.so (or its own dso)? It seems more general purpose and less likely to change than libschedutil.

I think so -- ultimately. I imagine there would be more external users of libeventlog than schedulers and it does make sense. If this happens, I can latch on to it. (I don't think it is a priority though).

dongahn:
> Also, are the changes here going to be easy to resolve with flux-framework/flux-core#2226?

We will see. But my changes will be pretty isolated with a minimal API change so hopefully this doesn't give too much trouble to @SteVwonder.

garlick:
Just occurred to me - you can request both R and the eventlog through the job-info module with one RPC (instead of going direct to KVS)...

grondo:
But doesn't job-info have to do 2 rpcs, so 3 in total and higher latency?

garlick:
I think job-info fetches keys in parallel so probably the same latency.  Yeah, never mind - he's already got code to fetch and parse the eventlog, so I think I'm not being especially helpful here!

garlick:
Let's just fix the job manager to send the entire queue object in the hello response.  I'll propose something in flux-core.  Stand by...

dongahn:
> Let's just fix the job manager to send the entire queue object in the hello response. I'll propose something in flux-core. Stand by...

OK. I will hold until I see your changes then. Thanks.

dongahn:
> OK. I will hold until I see your changes then. Thanks.

@garlick: Thank you for the hello protocol changes. Just pushed the changes to make use of them. I believe I addressed all the comments that came up so far. Let me know if these are okay I will squash them before merge.

garlick:
I think it's OK to squash.

I'll have one more look after you do that but I think the job-manager interface part looks good.

dongahn:
Thanks @garlick. Squashed and forced a push.

garlick:
315cce4 (add hello/exception callback) seems to include more change than is described in the commit message, e.g. modifications to the free callback error handling, a reordering of qmanager_new(), and queue policy interface changes.  Add to commit message or split out interface changes to their own commit?

LGTM otherwise.

dongahn:
> 315cce4 (add hello/exception callback) seems to include more change than is described in the commit message, e.g. modifications to the free callback error handling, a reordering of qmanager_new(), and queue policy interface changes. Add to commit message or split out interface changes to their own commit?

@garlick: I'd be happy to do that. I just split the commit into multiples. To minimize later squash, I took liberty forcing a push. From my perspective, this is ready to be merged.

dongahn:
Wait... just one more change.

dongahn:
Done.

dongahn:
@garlick: 

Thank you. 

> Did you want me to press the button or did you want someone wither a better understanding of the new architecture to do it?

Please go ahead and press the button. If there are any problems with other parts of the code, we can revisit that later. This will help meet our next milestones!

dongahn:
Thanks!!

Issue #495, opened 2019-07-13
[bug] exclusive req. can allocate a partially allocated resource


dongahn:
Using `tiny.graphml` GRUG file (`cluster[1]->rack[1]->node[2]->socket[2]->cores[18]`):

`jobspec1.yaml`:
```yaml
version: 1
resources:
    - type: cluster
      count: 1
      with:
        - type: rack
          count: 1
          with:
            - type: node
              count: 2
              with:
                - type: socket
                  count: 2
                  with:
                    - type: slot
                      label: default
                      count: 1
                      with:
                        - type: core
                          count: 1
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: app
    slot: default
    count:
      per_slot: 1
```

`jobspec2.yaml`:
```yaml
version: 1
resources:
  - type: slot
    label: default
    count: 1
    with:
      - type: socket
        count: 1
        with:
          - type: core
            count: 1
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: app
    slot: default
    count:
      per_slot: 1
```

```console
ahn1@28fae5c581ab:/usr/src/resource/utilities$ ./resource-query -G ../../t/data/resource/grugs/tiny.graphml -F pretty_simple -P low
INFO: Loading a matcher: CA

resource-query> match allocate jobspec1.yaml
      ---tiny0[1:shared]
      ------rack0[1:shared]
      ---------node1[1:shared]
      ------------socket1[1:shared]
      ---------------core18[1:exclusive]
      ------------socket0[1:shared]
      ---------------core0[1:exclusive]
      ---------node0[1:shared]
      ------------socket1[1:shared]
      ---------------core18[1:exclusive]
      ------------socket0[1:shared]
      ---------------core0[1:exclusive]
INFO: =============================
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
INFO: =============================

resource-query> match allocate jobspec2.yaml
      ---tiny0[1:shared]
      ------rack0[1:shared]
      ---------node0[1:shared]
      ------------socket0[1:exclusive]
      ---------------core1[1:exclusive]
INFO: =============================
INFO: JOBID=2
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
INFO: =============================
```

Unless I'm mistaken, the second match-allocate (on `jobspec2.yaml`) shouldn't be matched.

Issue #478, opened 2019-06-27
Satisfiability check within resource


dongahn:
Creating this issue so that this can be addressed separate from Issue #468. 

`resource` has two match operations: `match allocate` and `match allocate_orelse_reserve`. The former doesn't tell whether the allocation request failed because of the current resource state (i.e., no resource available now) or the jobspec cannot be satisfied at all. By contrast, if the latter fails, that means the jobspec is not satisfiable -- No matter how far in the timeline I move my schedule point (as far as the resource state becomes pristine), the jobspec cannot be reserved.

I propose we add `match allocate_orelse_satisfiability` to overcome this challenge. This is required to complete the qmanager integration with job-manager: if the jobspec is unsatisfiable, qmanager needs to send back a "request denied" to the alloc request; otherwise, it will keep trying to allocate on every job and resource event.

SteVwonder:
Great idea!

It may also be worthwhile to have a `match satisfiability` command that can be run once per job at ingestion, to give users an early feedback about their job (at the expense of extra work at the scheduler).

dongahn:
Conceptually, that's a good idea. But implementation wise, we will have to be very careful. 

In particular, I am worried about the performance implication that this will have on `resource-match`. A satisfiability check can be very expensive (in fact much more expensive then `match_allocate` itself) as it has to walk the entire resource graph. So it would be wise to avoid a full walk unless we have to. 

I was thinking to add this for the `alloc` request handler within `qmanager` for the first cut. It will be only called if the `match allocate` request of a jobspec fails so it shouldn't affect the scheduler performance for those "normal" jobs at all. Users won't get early feedback. But they will eventually get their feedback when the job is at least checked at least once within the schedule loop. 

Now, for early feedback, I can see why we can add a lightweight check (check various limits) though. This won't be a full satisfiability check. But this would be useful as the first filter. Such a lightweight check will probably also belong to `resource match` but I have't given much thought there.

dongahn:
PR #504 resolves this.

PR #500, opened 2019-07-27
EASY/HYBRID/CONSERVATIVE policies + satisfiability checking

dongahn:
This PR adds the following support:

- Detect unsatisfiable jobs
As discussed in Issue #478, the match RPC within `resource` only supports two match operations: `allocate` and `allocate_orelse_reserve`. The former doesn't tell whether the allocation request failed because of the current resource state (i.e., no resource available now) or the jobspec cannot be satisfied at all. By contrast, if the latter fails, that means the jobspec is not satisfiable: no matter how far in the timeline I move my schedule point, the jobspec cannot be matched.
The main feature of this PR is to add  `allocate_with_satisfiability` to overcome the challenge with `allocate`.  It first attempts to allocate. If succeeds, it returns the matching info as before. If fails, however, it sets the scheduled time to a point as late as possible within the time box of the planner
and try the match. Then, if it still can't find matching resources at that scheduled point, the jobspec is deemed unsatisfiable. This add no additional overhead for those "qualified" jobspec. However, for "unqualified" jobspecs, checking satisfiability requires an entire tree walk. Going forward, we will more rigorously validate the input job specs before they come to `resource`, so we will be able to manage this cost better.

- Tighten up `qmanager`'s error handling 
Integrate this feature to the rest of the systems including `qmanager` and its queue policy layer. This tightens up error handling of `qmanager` to become more robust.

- Bug fix for incorrect handling of implicit exclusivity per slot
Include a fix for Issue #495: when we have some sub-resources under the visiting
resource have been allocated to another job, we can prune the further walk even though the visiting resource itself has not been exclusively allocated to another job. But it appears `by_excl` does not factor into account the implicit exclusivity that can arise from slot: e.g., with `slot[2]->socket[1]->core[1]`, `socket`s and `core`s under slots are implicitly exclusive.

- Trim message logging for `resource` and `qmanager`
Significantly reduce the number of per-job messages in particular.

- Add EASY/HYBRID/CONSERVATIVE policies
Introduce `queue_policy_bf_base.hpp` and `queue_policy_bf_base_impl.hpp` that implement core algorithms for a majority of backfillling strategies. Derive from this class to implement EASY/HYBRID/CONSERVATIVE policies.

- Add queue-depth support for backfill policies

- Introduce the new name for our graph based scheduler: **Fluxion**
Prefix the module names with fluxion and adjust RPC names in `resource`, rc scripts for both `resource` and `qmanager` and tests with this new prefix.


dongahn:
@SteVwonder, @grondo and @garlick: this PR should contain all of the tasks I planned for the 7/31 milestone. Please help review this PR. The changes are pretty substantial.

garlick:
@dongahn - it seems like it would be beneficial for reviewability and for a clean history to split this PR into topically distinct chunks.  I'm not sure of the best way to do that but for a start, could you split off the rename to a new PR that we can get merged first?

I did have some quick suggestions on that part:
* The module names should have a `sched-` prefix, e.g. `sched-fluxion-qmanager` so that `flux module list` shows something that looks like a scheduler is loaded and naming is consistent with `sched-simple`.
* Just use the framework names for automake vars for rc dirs, etc, e.g. `fluxrc1dir` not `fluxionresourcerc1` or `fluxionqmanagerrc1`
* Keep rebranding/documentation that is not strictly necessary for functionality in separate commits from the module name change.

dongahn:
@garlick: good suggestions. I will follow up. Others: please let me know if you have other quick suggestions. It would be wise to do the work in bulk as much as possible. BTW one test is failing on one of the travis configuration. I need to look at that too. Thanks

garlick:
BTW I shouldn't have suggested that the rename PR go first, since the order doesn't particularly matter and it's last in your stack.  It's probably a pain to do it first.

dongahn:
@garlick: yes I am actually thinking not to do a PR on the name change for this round. After writing a paper, I will have better proses and at that point I can do a better PR on it.

dongahn:
I think splitting this into atisfiability and queuing policy makes sense to me as originally attempted.

But the queuing PR depends on the changes in the satisfiability. So maybe doing a PR for sat first and review and land it. And then rebase the queuing on that and do a separate PR afterward?

garlick:
Sounds reasonable to me.

If there are some fixes  here that can be split out, those might be easier to get in quickly before the big features PR's.

grondo:
>  Others: please let me know if you have other quick suggestions. It would be wise to do the work in bulk as much as possible.

Sorry I had a family function yesterday and could not take a look until this AM.

I probably don't have much to add, but I did poke at this PR a bit and didn't find any major problems. In fact, the scheduler seems quite performant when run through sched-bench (a simple test, but looks great so far). So nice work!

I only have a couple other comments for now:

 * Now that running a workload is possible again, it might be useful to turn the valgrind tests back on. (Of course, this could be saved for a future PR)
 * A quick run under valgrind showed a mismatched `new`/`free` for the qmanager ctx:
```
==14218== Mismatched free() / delete / delete []
==14218==    at 0x4C30D3B: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14218==    by 0x157E913A: ??? (qmanager.cpp:316)
==14218==    by 0x157EA4B7: mod_main (qmanager.cpp:351)
==14218==    by 0x113CC6: module_thread (module.c:162)
==14218==    by 0x57C16DA: start_thread (pthread_create.c:463)
==14218==    by 0x5F4F88E: clone (clone.S:95)
==14218==  Address 0x10a07be0 is 0 bytes inside a block of size 120 alloc'd
==14218==    at 0x4C3040F: operator new(unsigned long, std::nothrow_t const&) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14218==    by 0x157EA288: qmanager_new (qmanager.cpp:293)
==14218==    by 0x157EA288: mod_main (qmanager.cpp:333)
==14218==    by 0x113CC6: module_thread (module.c:162)
==14218==    by 0x57C16DA: start_thread (pthread_create.c:463)
==14218==    by 0x5F4F88E: clone (clone.S:95)
```

dongahn:
Oh great @grand9. Yeah I should have run this through valgrid. I think I attempted but didn't find how to. Sorry about that.

Fir performance, I have some more ideas to improve it but I plan to get there much later in our 1 year plan. Thanks.

dongahn:
OK. I am closing this PR to split it into multiple,

PR #502, opened 2019-07-30
Bug fix for incorrectly handling implicit exclusivity

dongahn:
This PR fixes a bug where an implicit resource exclusivity created by slot is incorrectly handled.

Problem: When we have some sub-resources under the visiting resource have been allocated to another job, we can prune the further walk even though the visiting resource itself has not been exclusively allocated to another job. But it appears `dfu_impl_t::by_excl ()` does not factor into account the implicit exclusivity that can arise from slot: e.g., with slot[2]->socket[1]->core[1], sockets and cores under slots are implicitly exclusive.

Augment `dfu_impl_t::by_excl ()` to handle this case. 

Augment `t/t3001-resource-basic.t` to cover this case.

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -115,14 +115,27 @@ done:
 }
 
 int dfu_impl_t::by_excl (const jobmeta_t &meta, const std::string &s, vtx_t u,
-                         const Jobspec::Resource &resource)
+                         bool exclusive_in, const Jobspec::Resource &resource)
 {
     int rc = -1;
     planner_t *p = NULL;
     int64_t at = meta.at;
     int64_t njobs = -1;
     uint64_t duration = meta.duration;
-    if (resource.exclusive == Jobspec::tristate_t::TRUE) {
+
+    // If a non-exclusive resource request is explicitly given on a
+    // resource that lies under slot, this spec is invalid.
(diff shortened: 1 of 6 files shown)


dongahn:
@SteVwonder: Ok, this is ready for your review.

resource/traversers/dfu_impl.cpp
```diff
@@ -115,14 +115,27 @@ int dfu_impl_t::by_avail (const jobmeta_t &meta, const std::string &s, vtx_t u,
...
+                         bool exclusive_in, const Jobspec::Resource &resource)
 {
     int rc = -1;
     planner_t *p = NULL;
     int64_t at = meta.at;
     int64_t njobs = -1;
     uint64_t duration = meta.duration;
-    if (resource.exclusive == Jobspec::tristate_t::TRUE) {
+
+    // If a non-exclusive resource request is explicitly given on a
```
SteVwonder: I don't believe that RFC 24 for jobspec V1 specifies this constraint.  So I'll make a note in the outstanding PR to make `exclusive: false` under a `slot` invalid.

dongahn: Yes, this would be good. It would still be good to test condition at the this level too.

dongahn:
@SteVwonder: Yes this should be merged as soon as possible and I believe this is ready to go in. Thanks.

Issue #489, opened 2019-07-10
Trim per-job log msgs from qmanager/resource


dongahn:
Per https://github.com/flux-framework/flux-sched/pull/481#issuecomment-510176192

dongahn:
PR #482 fixes this.

PR #503, opened 2019-07-30
Add support for checking a job's satisfiability 

### LINKED ISSUE #495: [bug] exclusive req. can allocate a partially allocated resource

dongahn:
This PR adds the following support:

- Detect unsatisfiable jobs
As discussed in Issue #478, the match RPC within `resource` only supports two match operations: `allocate` and `allocate_orelse_reserve`. The former doesn't tell whether the allocation request failed because of the current resource state (i.e., no resource available now) or the jobspec cannot be satisfied at all. By contrast, if the latter fails, that means the jobspec is not satisfiable: no matter how far in the timeline I move my schedule point, the jobspec cannot be matched.
The main feature of this PR is to add  `allocate_with_satisfiability` to overcome the challenge with `allocate`.  It first attempts to allocate. If succeeds, it returns the matching info as before. If fails, however, it sets the scheduled time to a point as late as possible within the time box of the planner and try the match. Then, if it still can't find matching resources at that scheduled point, the jobspec is deemed unsatisfiable. This add no additional overhead for those "qualified" jobspec. However, for "unqualified" jobspecs, checking satisfiability requires an entire tree walk. Going forward, we will more rigorously validate the input job specs before they come to `resource`, so we will be able to manage this cost better.

- Tighten up `qmanager`'s error handling 
Integrate this feature to the rest of the systems including `qmanager` and its queue policy layer. This tightens up error handling of `qmanager` to become more robust.

- Trim message logging for `resource` and `qmanager`
Significantly reduce the number of per-job messages in particular.

Resolve #495, #478, and #489.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -58,6 +58,37 @@ struct qmanager_ctx_t {
  *                                                                            *
  ******************************************************************************/
 
+static int post_sched_loop (qmanager_ctx_t *ctx)
+{
+    int rc = -1;
+    std::shared_ptr<job_t> job = nullptr;
+
+    while ((job = ctx->queue->alloced_pop ()) != nullptr) {
+        if (schedutil_alloc_respond_R (ctx->h, job->msg,
+                                       job->schedule.R.c_str (), NULL) < 0) {
+            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_R",
+                            __FUNCTION__);
+            goto out;
+        }
+        flux_log (ctx->h, LOG_DEBUG,
+                  "alloc success (id=%jd)", (intmax_t)job->id);
+    }
+    while ((job = ctx->queue->rejected_pop ()) != nullptr) {
+        std::string note = "alloc denied due to type=\"" + job->note + "\"";
+        if (schedutil_alloc_respond_denied (ctx->h, job->msg, note.c_str ()) < 0) {
+            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_denied",
+                            __FUNCTION__);
+            goto out;
+        }
+        flux_log (ctx->h, LOG_DEBUG,
+                  "%s (id=%jd)", note.c_str (), (intmax_t)job->id);
+    }
+    rc = 0;
+
+out:
+    return rc;
+}
+
 // FIXME: This will be expanded when we implement full scheduler
 // resilency schemes: Issue #470.
 extern "C" int jobmanager_hello_cb (flux_t *h,
@@ -104,18 +135,10 @@ extern "C" void jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
         flux_log_error (h, "%s: queue insert", __FUNCTION__);
         return;
     }
-    if (ctx->queue->run_sched_loop ((void *)ctx->h, true) < 0) {
-        flux_log (ctx->h, LOG_DEBUG,
-                  "%s: return code < 0 from schedule loop", __FUNCTION__);
-    }
-    while ((job = ctx->queue->alloced_pop ()) != nullptr) {
-        flux_log (ctx->h, LOG_DEBUG, "jobid (%ju): %s",
-                  (intmax_t)job->id, job->schedule.R.c_str ());
-        if (schedutil_alloc_respond_R (ctx->h, job->msg,
-                                       job->schedule.R.c_str (), NULL) < 0) {
-            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_R",
-                            __FUNCTION__);
-        }
+    if (ctx->queue->run_sched_loop ((void *)ctx->h, true) < 0
+        || post_sched_loop (ctx) < 0) {
+        flux_log_error (ctx->h, "%s: schedule loop", __FUNCTION__);
+        return;
     }
 }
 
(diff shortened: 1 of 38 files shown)


dongahn:
@StevWonder, @grondo and @garlick: this is also ready to be reviewed again. I had to comment out one test because this builds on the by_excl() bug fix PR. If the bug fix PR lands first, I will turn it back on.

dongahn:
Rebased to the master after PR #502 and forced a push.

resource/modules/resource_match.cpp
```diff
@@ -644,25 +654,18 @@ static void match_request_cb (flux_t *h, flux_msg_handler_t *w,
...
-    flux_log (h, LOG_INFO, "match requested by user (%u).", userid);
-
     if (flux_request_unpack (msg, NULL, "{s:s s:I s:s}", "cmd", &cmd,
                              "jobid", &jobid, "jobspec", &js_str) < 0)
         goto error;
-
     if (is_existent_jobid (ctx, jobid)) {
-        flux_log (h, LOG_INFO, "invalid jobid (%ld).", (intmax_t)jobid);
         errno = EINVAL;
+        flux_log_error (h, "existent job (%jd).", (intmax_t)jobid);
```
SteVwonder:
```suggestion
        flux_log_error (h, "nonexistent job (%jd).", (intmax_t)jobid);
```

Further down in `info_request_cb`, the same nonexistent job error results in an `ENOENT`. I wonder if that fits better here too.

qmanager/modules/qmanager.cpp
```diff
@@ -58,6 +58,37 @@ struct qmanager_ctx_t {
...
+{
+    int rc = -1;
+    std::shared_ptr<job_t> job = nullptr;
+
+    while ((job = ctx->queue->alloced_pop ()) != nullptr) {
+        if (schedutil_alloc_respond_R (ctx->h, job->msg,
+                                       job->schedule.R.c_str (), NULL) < 0) {
+            flux_log_error (ctx->h, "%s: schedutil_alloc_respond_R",
+                            __FUNCTION__);
+            goto out;
```
SteVwonder: This is more thinking out loud than it is giving concrete review feedback, but I wonder what the best practice is for handling an error like this.  Should the first failure cause the `qmanager` to bail out of trying to respond to the remaining messages?  Or should it continue on trying to respond to the other messages, even if they too error out.  I'm not sure.  I'm kind of leaning towards the latter if only because then the `alloced` and `rejected` queues are empty at the end of the scheduling loop and every request has gotten a response (or a response has at least been attempted rather than delayed indeterminately).

resource/modules/resource_match.cpp
dongahn: Ah. Thanks. Yes, `ENOENT` should be used instead unless it has a side effect (I don't believe so). I will follow up.

qmanager/modules/qmanager.cpp
dongahn: Good point! I am not sure either. I think what is best would depend on what is the most failure mode of `schedutil_alloc_respond_R`. If the mode is a network/protocol error or job-manager is in a limbo state, it doesn't seem like issuing more respond would help... Even if we error out right away, the allocated and rejected jobs will still be in the corresponding queues, the responds will be sent at the next invocation of the scheduler loop. So I'm kind of leading toward the former. 

I'd suggest we merge this as-is. And then, have a ticket created so that if this becomes an issue, we can easily go back to this discussion.

resource/modules/resource_match.cpp
dongahn: Looking at this once again, this is actually intended. In this case, the code is checking whether the job id already exists and if so it refuses the match request. So `EINVAL` should correct as well. So I will just push one commit that fixes the `done` label issue.

I was confused myself as well. So thank you for brining this up!

dongahn:
I pushed the commit and Travis is happy now. @SteVwonder: once you look at it briefly, I will squash that. Then, this PR is good to be merged.

SteVwonder:
> once you look at it briefly, I will squash that. Then, this PR is good to be merged.

LGTM!  Squash away!

dongahn:
Ok. Squashed. Good to go. Thanks!

SteVwonder:
Thanks @dongahn!

Issue #498, opened 2019-07-26
Default hwloc-whitelist for our next near-term target?


dongahn:
I think this is a good time to add `hwloc-whitelist=` into `resource`'s `rc1` script so that our graph scheduler can build and operate on a significantly trimmed  graph for scheduling (thus fast). I have two options and I am hoping to get feedback as to the best default given our next production target.

#### 2. flux module load -r 0 resource hwloc-whitelist=node,core,gpu

The advantage of option 1 is that it will provide a better topology for scheduling as it includes socket. But the downside is, your node will be represented as `node[1]->socket[2]->core[8]` and so if your jobspec is

```json
  "resources": [
    {
      "count": 1,
      "with": [
        {
          "count": 16,
          "type": "core"
        }
      ],
      "type": "slot",
      "label": "task"
    }
  ]
```

The scheduler will deem this to be unsatisfiable. From the graph matching point of view, this jobspec will be interpreted as prefix-omitted *"->core[16]"* and clearly there is no match. 

By contract, option 2 will represent each node as `node[1]->core[16]` and this job spec will match just fine (at the expense of lesser topological info). 

My current thinking is to go with option 2 as the default hwloc-whitelist unless there are other compelling reasons not to do that. 

Are we even going to support `socket` in our near-term target? 

Thoughts?

SteVwonder:
> My current thinking is to go with option 2 as the default hwloc-whitelist unless there are other compelling reasons not to do that.

> Are we even going to support socket in our near-term target?

I agree with your plan.  At the top-level, we will be doing node-exclusive allocations anyway, so I don't think that specifying sockets at the top-level scheduler helps a ton (maybe it helps when the user has multiple tasks in a single jobspec, but v1 doesn't support that either).  So for now, I think we allocate gpus/cores/nodes and then let the advanced users that care about sockets take advantage of mpibind or something similar.

For user-level schedulers, which will be doing core-level scheduling, the socket information may be more important.  I think we wait until Flux is deployed on LC systems and someone has that particular use-case, and then we develop a solution in coordination with them (or if we have extra time at the end of the year, we do it then).

dongahn:
Thanks @SteVwonder!

> At the top-level, we will be doing node-exclusive allocations anyway

Our new scheduler is capable of doing node-exclusive and -nonexclusive at the same time.

[...]

Issue #482, opened 2019-06-28
Add EASY and CONSERVATIVE queueing policy into qmanager


dongahn:
It is likely #468 will be demonstrated with FCFS policy only. So, spin off a new task: implement other classical backfill policies into the queue policy interface and use it within qmanager.

dongahn:
PR #504 added this support.

PR #504, opened 2019-07-30
Add EASY/HYBRID/CONSERVATIVE policies

### LINKED ISSUE #476: Add queuing optimization support
So that Issue https://github.com/flux-framework/flux-sched/issues/468 will only add a baseline integration and then this issue will add some traditional queueing optimization such as queue-depth, reservation depth and delay scheduling.

dongahn:
This PR adds various backfill queuing polices and queue-depth support for optimization:

- Add EASY/HYBRID/CONSERVATIVE policies
Introduce queue_policy_bf_base.hpp and queue_policy_bf_base_impl.hpp that implement core algorithms for a majority of backfillling strategies. Derive from this class to implement EASY/HYBRID/CONSERVATIVE policies.

- Add queue-depth support for backfill policies

This PR builds on PR #502 and PR #503 so please don't merge it before those PRs land. Resolve #476, #482, and #498.

### DIFF
--- a/qmanager/config/queue_system_defaults.hpp
+++ b/qmanager/config/queue_system_defaults.hpp
@@ -28,9 +28,9 @@
 // FIXME: These need to be coverted into a scheduler configuration file
 namespace Flux {
 namespace queue_manager {
-    const unsigned int MAX_QUEUE_DEPTH=16384;
-    const unsigned int DEFAULT_QUEUE_DEPTH=8192;
-    const unsigned int MAX_RESERVATION_DEPTH = 4096;
+    const unsigned int MAX_QUEUE_DEPTH = 1000000;
+    const unsigned int DEFAULT_QUEUE_DEPTH = 8192;
+    const unsigned int MAX_RESERVATION_DEPTH = 100000;
     const unsigned int HYBRID_RESERVATION_DEPTH = 64;
 } // namespace resource_model
 } // namespace Flux
--- a/qmanager/config/queue_system_defaults.hpp
+++ b/qmanager/config/queue_system_defaults.hpp
@@ -1,0 +1,42 @@
+/*****************************************************************************\
+ *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+ *  LLNL-CODE-658032 All rights reserved.
+ *
+ *  This file is part of the Flux resource manager framework.
+ *  For details, see https://github.com/flux-framework.
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the license, or (at your option)
+ *  any later version.
+ *
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+
+#ifndef QUEUE_SYSTEM_DEFAULT_HPP
+#define QUEUE_SYSTEM_DEFAULT_HPP
+
+// FIXME: These need to be coverted into a scheduler configuration file
+namespace Flux {
+namespace queue_manager {
+namespace detail {
+    const unsigned int MAX_RESERVATION_DEPTH = 4096;
+    const unsigned int HYBRID_RESERVATION_DEPTH = 64;
+} // namespace detail
+} // namespace resource_model
+} // namespace Flux
+
+#endif // QUEUE_SYSTEM_DEFAULT_HPP
+
+/*
+ * vi:tabstop=4 shiftwidth=4 expandtab
+ */
--- a/qmanager/config/queue_system_defaults.hpp
+++ b/qmanager/config/queue_system_defaults.hpp
@@ -28,10 +28,10 @@
 // FIXME: These need to be coverted into a scheduler configuration file
 namespace Flux {
 namespace queue_manager {
-namespace detail {
+    const unsigned int MAX_QUEUE_DEPTH=16384;
+    const unsigned int DEFAULT_QUEUE_DEPTH=8192;
     const unsigned int MAX_RESERVATION_DEPTH = 4096;
     const unsigned int HYBRID_RESERVATION_DEPTH = 64;
-} // namespace detail
 } // namespace resource_model
 } // namespace Flux
 
--- a/qmanager/policies/base/queue_policy_base_impl.hpp
+++ b/qmanager/policies/base/queue_policy_base_impl.hpp
@@ -27,10 +27,72 @@
 
 #include <iostream>
 #include <cerrno>
+#include <algorithm>
 #include "qmanager/policies/base/queue_policy_base.hpp"
 
 namespace Flux {
 namespace queue_manager {
+
+int queue_policy_base_t::set_param (std::string &p_pair)
+{
+    int rc = -1;
+    size_t pos = 0;
+    std::string k, v;
+    std::string split = "=";
+
+    if ((pos = p_pair.find (split)) == std::string::npos) {
+        errno = EINVAL;
+        goto done;
+    }
+    k = p_pair.substr (0, pos);
+    k.erase (std::remove_if (k.begin (), k.end (), ::isspace), k.end ());
+    if (k.empty ()) {
+        errno = EINVAL;
+        goto done;
+    }
+    v = p_pair.erase (0, pos + split.length ());
+    v.erase (std::remove_if (v.begin (), v.end (), ::isspace), v.end ());
+    if (m_params.find (k) != m_params.end ())
+        m_params.erase (k);
+    m_params.insert (std::pair<std::string, std::string>(k, v));
+    rc = 0;
+done:
+    return rc;
+}
+
+
+int queue_policy_base_t::set_params (const std::string &params)
+{
+    int rc = -1;
+    size_t pos = 0;
+    std::string p_copy = params;
+    std::string delim = ",";
+
+    try {
+        while ((pos = p_copy.find (delim)) != std::string::npos) {
+            std::string p_pair = p_copy.substr (0, pos);
+            if (set_param (p_pair) < 0)
+                goto done;
+            p_copy.erase (0, pos + delim.length ());
+        }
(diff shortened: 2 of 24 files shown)


dongahn:
It is likely a test case will fail on one of the travis config. So I will work on that before our review. It would be best if this PR is reviewed last anyway.

dongahn:
OK. This PR fixed all the issues including CI test failures and new/free mismatch problem that @grondo reported. This should be ready for your reviews -- But still preferably after PR #503 and #502.

dongahn:
Rebased to the master after PR #502 and forced a push.

resource/modules/resource_match.cpp
```diff
@@ -614,59 +629,56 @@ static inline bool is_existent_jobid (const resource_ctx_t *ctx, uint64_t jobid)
...
-    if ((rc = tr.remove (jobid)) == 0) {
-        if (is_existent_jobid (ctx, jobid)) {
-           job_info_t *info = ctx->jobs[jobid];
-           info->state = job_lifecycle_t::CANCELLED;
-        }
-    } else {
+
+    if ((rc = tr.remove (jobid)) < 0) {
         if (is_existent_jobid (ctx, jobid)) {
+           // When this condition arises, we will be less likley
```
SteVwonder:
```suggestion
           // When this condition arises, we will be less likely
```

I'm not sure I'm parsing the first sentence of this comment correctly.  Where does the probability/likelihood in jobid reuse come in?  I thought that was a deterministic process.

resource/modules/resource_match.cpp
```diff
@@ -521,85 +528,93 @@ static int init_resource_graph (resource_ctx_t *ctx)
...
-static int track_schedule_info (resource_ctx_t *ctx, int64_t id, int64_t at,
-                                string &jspec, stringstream &R, double elapse)
+static int track_schedule_info (resource_ctx_t *ctx, int64_t id, int64_t now,
+                                int64_t at, string &jspec, stringstream &R,
+                                double elapse)
 {
     job_lifecycle_t state = job_lifecycle_t::INIT;
 
-    if (id < 0 || at < 0) {
+    if (id < 0 || now < 0 || at < 0) {
```
SteVwonder: might also be useful to check for and error out when `at < now`

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -27,11 +27,73 @@
...
+    if (m_params.find (k) != m_params.end ())
+        m_params.erase (k);
+    m_params.insert (std::pair<std::string, std::string>(k, v));
+    rc = 0;
+done:
+    return rc;
+}
+
+
+int queue_policy_base_t::set_params (const std::string &params)
```
SteVwonder: Does this handle nested argument lists?  For example, `sched-policy="hybrid-backfill",policy-params="queue-depth=100,reservation-depth=10"`  IIUC, this will result in `reservation-depth` being a parameter to the top level rather than being a part of the `policy-params` value.

If so, I think we either need to refactor this function or drop the usage of `policy-params` and just make the policy parameters part of the scheduler parameters (e.g., `sched-policy=hybrid-backfill,reservation-depth=10,queue-depth=100`).

qmanager/policies/queue_policy_hybrid_impl.hpp
```diff
@@ -0,0 +1,82 @@
...
+             != queue_policy_base_impl_t::m_params.end ()) {
+            unsigned int depth = std::stoi (i->second);
+            if (depth < MAX_QUEUE_DEPTH)
+                queue_policy_base_impl_t::m_queue_depth = depth;
+        }
+        if ((i = queue_policy_base_impl_t::m_params.find ("reservation-depth"))
+             != queue_policy_base_impl_t::m_params.end ()) {
+            unsigned int depth = std::stoi (i->second);
+            if (depth < MAX_RESERVATION_DEPTH)
+                queue_policy_bf_base_t<reapi_type>::m_reservation_depth = depth;
```
SteVwonder: If `depth > MAX_RESERVATION_DEPTH`, should `m_reservation_depth` be set to `MAX_RESERVATION_DEPTH` rather than remaining as the default?

Same for `MAX_QUEUE_DEPTH`.

qmanager/policies/queue_policy_hybrid_impl.hpp
```diff
@@ -0,0 +1,82 @@
...
+        HYBRID_RESERVATION_DEPTH;
+}
+
+template<class reapi_type>
+int queue_policy_hybrid_t<reapi_type>::apply_params ()
+{
+    int rc = -1;
+    try {
+        std::unordered_map<std::string, std::string>::const_iterator i;
+        if ((i = queue_policy_base_impl_t::m_params.find ("queue-depth"))
```
SteVwonder: As we discussed at coffee time, I wonder if there is some way to refactor these policy classes to reduce the amount of duplicate code, in particular in these `apply_params` methods.  One thought that came to mind which is less invasive than my original suggestions is putting the `queue-depth` handling in the `apply_params` of a parent class, since all four scheduling policies handle that parameter application in the exact same way currently.  Then each child could call it's parent's `apply_params` followed by it's own handling of additional parameters.

qmanager/policies/queue_policy_conservative_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+template<class reapi_type>
+int queue_policy_conservative_t<reapi_type>::apply_params ()
+{
+    int rc = -1;
+    try {
+        std::unordered_map<std::string, std::string>::const_iterator i;
+        if ((i = queue_policy_base_impl_t::m_params.find ("queue-depth"))
+             != queue_policy_base_impl_t::m_params.end ()) {
+            unsigned int depth = std::stoi (i->second);
+            if (depth < MAX_QUEUE_DEPTH)
```
SteVwonder: If `MAX_QUEUE_DEPTH` is greater than `MAX_RESERVATION_DEPTH` (which is currently the case), then this can result in the "conservative" policy producing non-conservative results.  Since this is already it's own class, it would be relatively easy to special-case the behavior of this policy to either limit the `queue-depth` to `MAX_RESERVATION_DEPTH` or allow `reservation-depth` to be set to `queue-depth` even if it exceeds `MAX_RESERVATION_DEPTH`.

This would make our "conservative" policy's behavior align with the textbook definition of conservative backfill.  For users that want a highly performant policy that is mostly conservative, they can always reach for the "hybrid" policy and manually specify a large reservation depth that is still smaller than the queue depth.

qmanager/policies/queue_policy_hybrid_impl.hpp
SteVwonder: This question/comment also applies to the `apply_params` of the other classes/policies in this PR.

dongahn:
@SteVwonder: just a heads-up. I will get to this today unless something else pops up.

qmanager/config/queue_system_defaults.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+// FIXME: These need to be coverted into a scheduler configuration file
+namespace Flux {
+namespace queue_manager {
+namespace detail {
+    const unsigned int MAX_QUEUE_DEPTH=16384;
+    const unsigned int DEFAULT_QUEUE_DEPTH=8192;
+    const unsigned int MAX_RESERVATION_DEPTH = 4096;
+    const unsigned int HYBRID_RESERVATION_DEPTH = 64;
+} // namespace detail
```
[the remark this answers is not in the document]
dongahn: Thanks. Interesting that SLURM's default reservation depth is 0. This is not even a backfill policy...

resource/modules/resource_match.cpp
dongahn: At the moment, I can't find the commit message. But a same jobid will have to be used as it often has to be reserved multiple times before it will finally be allocated.

resource/modules/resource_match.cpp
dongahn: Sure. Thanks.

qmanager/policies/base/queue_policy_base_impl.hpp
dongahn: Good feedback. 

From the implement point of view, this shouldn't be a problem. Both the queue parameters (specified with `queue-params`) and policy parameters (`queue-policy`) are parsed into a common map and both the generic queuing policy code and derived, specific policy code will share this common parameter map. 

As far as the queue parameters and policy parameters won't have the identically named key, (which should always be the case), this shouldn't be a problem.

dongahn: > As far as the queue parameters and policy parameters won't have the identically named key, (which should always be the case), this shouldn't be a problem.

To make this a bit more robust, I will use two distinct maps, one for `queue-param` and the other for `queue-policy`.

qmanager/policies/queue_policy_hybrid_impl.hpp
dongahn: If `depth` is invalid like `depth > MAX_RESERVATION_DEPTH`, the code currently uses a whole-sale approach of keeping the default value. 

https://github.com/flux-framework/flux-sched/blob/e8250ea29c666ef7f590f546ac6b603b28e01c83/qmanager/policies/queue_policy_hybrid_impl.hpp#L43

https://github.com/flux-framework/flux-sched/blob/e8250ea29c666ef7f590f546ac6b603b28e01c83/qmanager/policies/base/queue_policy_base.hpp#L137

Yes, I agree. If the value is larger than the max, we should set max. I will follow up.

qmanager/policies/queue_policy_hybrid_impl.hpp
dongahn: This is a good suggestion.

> One thought that came to mind which is less invasive than my original suggestions is putting the queue-depth handling in the apply_params of a parent class, since all four scheduling policies handle that parameter application in the exact same way currently. 

One possibility:

I can add the `queue-depth` selection in the `apply_params` method of the parent. Then, in an derived class like `queue_policy_hybrid_impl.hpp`, I can invoke this base class's `apply_params` from which its `apply_params`. I haven't  see this pattern much so let me do some study to see if there is any side-effect, though.

qmanager/policies/queue_policy_conservative_impl.hpp
dongahn: Ah. Excellent. Yes I will address this.

dongahn:
@SteVwonder: I rebased this PR to the current master and addressed all of your (excellent) review items. If CI turns green, this can use your second review. The only changes you will need to review are the ones marked with `[squash]`.

dongahn:
OK. Travis CI just turned green!

SteVwonder:
LGTM! Squash away and I'll push the button.

dongahn:
@SteVwonder: I squashed most of them but decide to keep two commits on their own. 

- https://github.com/flux-framework/flux-sched/pull/504/commits/30bb6920b02f7941a72f4941356eb8a7028e6f8d
- https://github.com/flux-framework/flux-sched/pull/504/commits/a93d02348cd6e6629d080e9f234b849543baf45a

I thought it would be reasonable to leave a trail about these changes. If you are okay with these change, this PR is ready to go in once Travis turns green.

dongahn:
OK. It is green now.

SteVwonder:
LGTM! Thanks @dongahn!

Issue #514, opened 2019-09-18
Variation-aware scheduling


tpatki:
First cut of the production version will focus on a small graph with a representative job queue with 15-20 jobs. It will leverage the recently developed `set-property` interface in `resource-query` for initial testing. 

Node-level scheduling will be assumed, and we will use ~3 performance classes for distributing nodes by their performance efficiency. We will also assume that users specify the execution time of their job based on the worst-case scenario (i.e., slowest node) and as a result we will not modify a job's runtime. In this version, we will place application ranks on similar-performing nodes and compare against the lowest-first or highest-first default policies. This will minimize rank-to-rank variation within the application.

One question is about potentially missing property values due to user entry error. Prior to scoring and matching, we will have to add a method to check that the associated property (in our case performance class) is available for each node and is in the expected range (the latter is harder to determine).  Not sure if we should do this for a first cut (comments, @dongahn)?

dongahn:
Thanks Tapasya:

> Node-level scheduling will be assumed

The graph-based scheduler doesn't make a distinct between node exclusive vs. non-node executive scheduling. It just schedule based on the resource specification in a jobspec. Does this break your assumption?

>  Prior to scoring and matching, we will have to add a method to check that the associated property (in our case performance class) is available for each node and is in the expected range (the latter is harder to determine).

I think it would be a reasonable assumption to treat those unannotated vertex as the lowest performing resource?

tpatki:
> Node-level scheduling will be assumed.
    The graph-based scheduler doesn't make a distinct between node exclusive vs. non-node executive scheduling. It just schedule based on the resource specification in a jobspec. Does this break your assumption?

I believe this will apply to testing, and the code will be general enough to handle any sort of variation. It is the input properties we provide externally that'll make a difference.
The properties (eg node perf_class) is important for physical mapping--  the real-world data we collect (be it based on manufacturing variation or for aging) will either be node-level or socket-level. For power, this assumption is important as we're looking at underlying power domains (i.e. individual subparts of chips where power caps can be placed). Assuming node-level granularity for the testing part because that's the data we have. 

> I think it would be a reasonable assumption to treat those unannotated vertex as the lowest performing resource?

[...]

PR #517, opened 2019-09-20
Support for variation-aware scheduler

tpatki:
Initial design for #514. 

Needs extensive error checks and additional tests, but posting for early review and feedback. A simple example has been tested and verified. 

Note: shall fix indentation, comments, etc in the final PR.


dongahn:
Thanks @tpatki! 

I took a brief look at this looking at where your changes are inserted and and structures etc. This seems like a great start. 

The only initial feedback is: at least for user visible names like the option argument for `--match-policy` of `resource-query`, I prefer `variation` instead of `var_aware`.

dongahn:
Also, now that I merged your set-/get-property support you can safely rebase.

tpatki:
Thanks @dongahn for the early feedback. I need to add more tests for (1) missing property values, (2) erroneous inputs where `stoi` fails, etc. I will change the name to `variation` instead of `var_aware`. 

I also have the example from quartz that we had tested earlier, it has variation data on 2417 nodes with 5 performance classes and ~200 jobs from quartz queue. I also have a helper script that plots graphs and compares the lowest ID first, highest ID first and variation-aware algorithms for figure of merit. 

Do you want me to add this quartz test as part of our production code or should I just commit it to a new local branch so we can work on analysis later?

dongahn:
> I also have the example from quartz that we had tested earlier, it has variation data on 2417 nodes with 5 performance classes and ~200 jobs from quartz queue. I also have a helper script that plots graphs and compares the lowest ID first, highest ID first and variation-aware algorithms for figure of merit.

Cool.

> Do you want me to add this quartz test as part of our production code or should I just commit it to a new local branch so we can work on analysis later?

I think these should stay in a research branch for our analysis later; our repo probably want to stay lean and mean as much as possible :-)

tpatki:
> I think these should stay in a research branch for our analysis later; our repo probably want to stay lean and mean as much as possible :-)

@dongahn: Agreed, just wanted to check. I'll create a new research branch in my fork based on the recent updates and maintain these there. Thanks!

dongahn:
@tpatki: Thanks Tapasya. I haven't looked at your logic yet but here are my initial set of change requests which will help me and @SteVwonder to review your PR further. This PR is relatively large but currently only have two commits. If they can be further splitter into topical commits, this would be very helpful. One suggestion is:

The first feature commit can be split as the following: 
1. commit that includes your policy selection bug fix (thank you again)!
2. commit that includes dfu_match_var_aware.hpp|cpp 
3. commit that integrate these files into the build system: `resoure/Makefile.am`
4. commit that integrates this into the factory class: `dfu_match_policy_factory.hpp|cpp`
5. commit that integrates this into the resource-query: `resource-query.cpp`
6. commit that integrates this into the resource match: `resource_match.cpp`

For testing commit, I would split it as the following:

1. Commit small.graphml on its own
2. Commit with cmds files
3. Commit with  output files
4. Commit with  jobspec files
5. Commit that introduces t3014-resource-var-aware.t 
6. Commit that introduces t4007-match-var-aware.t
7. Commit `t/Makefile.am` to tie those two sharness test scripts into the build system

I believe each commit can benefit from more detailed commit messages. Each commit has a specific topic, so hopefully it wouldn't be too painful to add a description. The most detail will be needed for your commit that contains changes in dfu_match_var_aware.hpp|cpp and perhaps the sharness tests.

dongahn:
@SteVwonder's schedutil change went in. This needs to be debased as well.

tpatki:
Something went wrong. Redoing this.

tpatki:
@dongahn, @SteVwonder: Split the commits are requested and this is ready for review.

dongahn:
@tpatki: Thanks. I will review this soon (probably tonight).

resource/policies/dfu_match_var_aware.cpp
```diff
@@ -0,0 +1,153 @@
...
+            unsigned int count = calc_count (c_resource, qc);
+            if (count == 0) {
+                score = MATCH_UNMET;
+                break;
+            }
+            dfu.choose_accum_best_k (subsystem, c_resource.type, count, comp);
+        }
+    }
+
+    for (p_it = g[u].properties.begin ();
```
dongahn: I think this is the core of your algorithm. Great that you were able to do this in such an isolated manner!

Having said that, here are two issues: 

1. I believe this code only needs to be executed when `score == MATCH_MET` so I would guard this code with the `if` statement. Otherwise this will be executed whether the visiting vertex is qualified or not. 

2. iterating through all of the keys in the property map seems a bit inefficient. `std::map` offers O(log N) access performance guarantee. Instead, I would use a lookup function. Now, because the graph object `g` is `const` and thus `operator[]` won't work. My guess is that's why you had used the iteration instead. If `key` doesn't exist, `std::map`'s `operator[]` creates a value object and inserts that with that key and this can often leads to difficult to debug problems. In this case, I would use `map::at()` method. It won't create an object when the key doesn't exist. But it does throw `std::out_of_range` exception. But you are already catch exceptions so this shouldn't change your code structure.

dongahn:
@tpatki: This PR looks pretty good! 

I left my review comments in-line to your code and testing inputs. Most of the comments are trivial, but you should look at the ones about the main algorithm and see if you agree. 

In addition, I generally find that your commit messages are a bit terse. I'd suggest you augment some of the commit messages (not on the title, but on the comment message body) with a bit more details. When someone other than us look at the log messages later for bug hunting of some sort, I am sure they will appreciate they could do that only by reading commit messages.

Once these issues are addressed, I believe this PR can be merged.

resource/policies/dfu_match_var_aware.cpp
dongahn: I forgot to mension. `std::stoi` can also throw `std::out_of_range`. Another reason to consider `std::at ()`.

dongahn:
@tpatki: PR #523 has been merged. So when you update your PR, please make sure you rebase to the upstream master.

tpatki:
@dongahn: sounds good! Will update this PR soon (in the next day).

resource/modules/resource_match.cpp
```diff
@@ -476,6 +476,9 @@ static int init_resource_graph (resource_ctx_t *ctx)
 {
     int rc = 0;
 
+    // Select the appropriate matcher based on CLI policy.
+    ctx->matcher = create_match_cb (ctx->args.match_policy);
```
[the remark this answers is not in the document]
tpatki: @dongahn: leaving this as-is for now as per your suggestion as we'll move to `shared_ptr` later.

resource/policies/dfu_match_var_aware.cpp
tpatki: @dongahn: Changed to use `std::map::at`. Note that we still need to catch the `invalid_argument` exception for cases where `perf_class` input is invalid, eg, `perf_class=randomstring`. `stoi` throws `out_of_range` only if the converted value is out of range for an integer.

tpatki:
@dongahn: This is ready for your review/merging, addressed all the changes you requested.

dongahn:
@tapasya: looks great to me. If you want to address the one last nit (a leading white space before stoi and add std namespace), let me know.

tpatki:
@dongahn: fixed the whitespace and `std::stoi` issue. Thanks for the thorough and detailed review of this! :)

dongahn:
LGTM! Please squash the last commit with the original amd let me know. I will merge.

tpatki:
@dongahn: Squashed the last commit in to original variation-aware policy files. Thanks!

dongahn:
Great work. Thank you! Merging.

(Just FYI next time when you squash, you probably don't need to include the description of the squashing commit because the file was newly introduced and from the perspective of the commit history that problem doesn't really exist :-)

PR #523, opened 2019-10-01
testsuite: update flux-sharness.sh to version from flux-core

grondo:
The flux-broker no longer supports FLUX_RC1_PATH and FLUX_RC3_PATH.
Update to flux-sharness.sh from flux-core so that test_under_flux
no longer uses these variables.

When testing this update I was getting some hangs, though I'm not sure how it is related to this PR. I'll keep poking at it, but wanted to propose this update sooner than later.

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -7,7 +7,7 @@
 #  Extra functions for Flux testsuite
 #
 run_timeout() {
-    perl -e 'use Time::HiRes qw( ualarm ) ; ualarm ((shift @ARGV) * 1000000) ; exec @ARGV' "$@"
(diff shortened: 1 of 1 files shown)


dongahn:
@grondo: This looks good to me. Is there a flux-core issue ticket or PR to which we can refer this? I wasn't aware of this change and I would be good to leave a trail in this PR.

grondo:
Good point. Referencing PR flux-framework/flux-core#2431 and issue flux-framework/flux-core#2304

grondo:
Since the validator module issue I was hitting is only on fluke without .notce, this PR is probably ok to go in.

dongahn:
Yap. Sounds good to me.

PR #525, opened 2019-10-01
rc3: Fix a bug in RC3 dir definition within configure.ac

### LINKED ISSUE #524: flux-sched rc3 script doesn't shut qmanager down cleanly
This is a minor issue with the scheme for overriding sched-simple in the `flux-sched` rc1,rc3 scripts. Actually, `rc1` seems to work fine, in that `qmanager` is cleanly loaded in place of `sched-simple`, however at shutdown the reverse is not true

```
$ exit
exit
2019-10-01T18:19:46.154151Z broker.err[0]: rc3: flux-module: cmb.rmmod[0] sched-simple: No such file or directory
flux-broker: module 'qmanager' was not cleanly shutdown
```

This seems to be reproducible, but someone let me know if it is only in my environment.

See also flux-framework/flux-core#2273.

dongahn:
Fix a copy-and-paste error in the rc3 directory definition for qmanager. Without a proper rc3, qmanager wasn't cleanly exited when the Flux instance is torn down:

```
module 'qmanager' was not cleanly shutdown
```

Resolve #524.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,7 +136,7 @@ AC_SUBST(fluxresourcerc3dir)
 AS_VAR_SET(fluxqmanagerrc1dir, $sysconfdir/flux/rc1.d)
 AC_SUBST(fluxqmanagerrc1dir)
 
-AS_VAR_SET(fluxresourcerc3dir, $sysconfdir/flux/rc3.d)
+AS_VAR_SET(fluxqmanagerrc3dir, $sysconfdir/flux/rc3.d)
 AC_SUBST(fluxqmanagerrc3dir)
 
 ##


grondo:
I restarted a builder that had a very strange error in the hwloc tests. (Couldn't even tell what went wrong, suddenly the broker was gone)

dongahn:
Yes, I didn't think it makes sense to reload sched-simple in qmanager's rc3.  Thank you for a quick merge.

rspavel:
Sorry to respond to an eleven month old PR but what was the solution for the ```sched-simple``` issue? Trying to follow the quick start and it sounds like this is a problem with a known solution that I can't find.

Thanks

dongahn:
@rspavel: 

> Trying to follow the quick start and it sounds like this is a problem with a known solution that I can't find.

I would be a bit better if you can open an Issue ticket and describe the problem you are having? We should be able to help you better with that. Thanks.

rspavel:
Yeah. on the to do and mostly just reproducing on a few different platforms. It just sounded like the ```sched-simple``` case was a single command that might have let me focus on testing rather than bringing up a known (and solved?) issue

Thanks

PR #536, opened 2019-10-12
Resource graph metadata update

milroy:
This PR fixes `resource-query` output messages and modifies `resource_graph_metadata_t`.

- Fixes whitespace and newline characters in `resource/utilities/resource-query.cpp` 

- Changes the `by_path` map value to `vtx_t` in `resource/store/resource_graph_store.hpp`.  A path will never be associated with more than one vertex, so `by_path` should be `map<string, vtx_t>`.


dongahn:
@milroy: This looks great to me. Merging.

PR #548, opened 2019-11-27
fix buffer overflow when handling slot type in a jobspec

milroy:
This PR integrates the patch @dongahn created to address the segfault @cmisale reported in issue #509.

- Add edge evaluator public method that returns the qualified granules size

- Add corresponding scoring API public method

- Add constraint check against qualified granules

This PR should be merged before merging #546, so the flux-sched 0.8.0 release contains the fix.

PR #546, opened 2019-11-20
Update README.md and NEWS.md

dongahn:
This PR updates README.md and NEWS.md in preparation for v0.8.0 tag and release

Address 2 issues:
- Issue #545 (@andre-merzky: So if you can give me a feedback on the new README.md, this should be very useful);
- Issue #544 (@cmisale: once this gets in, we should be able to tag/release 0.8.0 so that you can use a tagged name for your Kube+Flux work);

PR #549, opened 2019-12-02
Update jobspec command to be list instead of list or string

chu11:
Per RFC PR https://github.com/flux-framework/rfc/pull/215, and follow up in `flux-core` of https://github.com/flux-framework/flux-core/pull/2564, the command in a jobspec shall be a list and not a list OR string.

The work in this PR is "done", but I've labeled it as "WIP" until the RFC change is approved.  This was initially done mostly to see "how much do we have to change".  Ends up not too much.  Some parsing in `libjobspec` had to be tweaked, but that was it code wise.  Update jobspecs throughout the tests and added an extra test for a bad jobspec with command as a string.

### DIFF
--- a/resource/libjobspec/jobspec.cpp
+++ b/resource/libjobspec/jobspec.cpp
@@ -228,11 +228,9 @@ Task::Task (const YAML::Node &tasknode)
     }
     if (tasknode["command"].IsSequence()) {
         command = tasknode["command"].as<std::vector<std::string>>();
-    } else if (tasknode["command"].IsScalar()) {
-        command.push_back(tasknode["command"].as<std::string>());
     } else {
         throw parse_error (tasknode["command"],
-                           "\"command\" value must be a scalar or a sequence");
+                           "\"command\" value must be a sequence");
     }
 
     /* Import slot */
--- a/t/data/resource/jobspecs/advanced/test001.yaml
+++ b/t/data/resource/jobspecs/advanced/test001.yaml
@@ -22,7 +22,7 @@ attributes:
   system:
     duration: 3600
 tasks:
-  - command: default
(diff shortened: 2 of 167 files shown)


chu11:
Removed WIP as https://github.com/flux-framework/rfc/pull/215 has been merged.

Should wait for https://github.com/flux-framework/flux-core/pull/2564 to be merged first.

grondo:
We may want to wait to merge this PR until after v0.8 is tagged, just because we're not also tagging a version of flux-core with this change. (Though likelihood of a problem is low)

dongahn:
@chu11: Just FYI -- PR #548 added one additional jobspec.

dongahn:
PR #548 has been merged.

chu11:
re-based & re-pushed with updates to the new jobspects recently merged in (`update` and `granule` jobspecs)

dongahn:
On a second thought, it is okay to merge this in before 0.8 tag. The scheduler doesn't make use of the task section at all so this is harmless. Thiughts?

dongahn:
OK, I reviewed this PR and this looks great to me. I will merge this in and then rebase PR #546.

grondo:
> On a second thought, it is okay to merge this in before 0.8 tag. The scheduler doesn't make use of the task section at all so this is harmless. Thoughts?

I was more worried that the change to libjobspec would reject valid jobspec possibly generated by flux-core v0.13 (before the change to `command` key) when flux-sched-v0.8 was used with this version.

I do think the probability of that is quite low and it will be obvious what happened, so I'm not worried.

dongahn:
Sorry I jumped the gun. Yeah hopefully this isn't an issue in practice.

PR #561, opened 2020-01-11
t4000-match-params.t: suppress flux module unload failure

garlick:
Problem: flux-core changed flux module unload to return failure
if the module is not loaded.  This breaks flux-sched tests that
depended on that behavior.

Ignore that error where appropriate in the test.

### DIFF
--- a/t/t4000-match-params.t
+++ b/t/t4000-match-params.t
@@ -32,35 +32,48 @@ test_debug '
     echo ${malform}
 '
 
+# Unload the resource module
+# If module is not loaded, suppress error.
+unload_resource() {
+    flux module remove resource 2>rmmod.out
+    if test $? -ne 0; then
+        grep -q "No such file or directory" rmmod.out && return 0
+        cat rmmod.out >&2
+        return 1
+    fi
+}
+
 test_expect_success 'loading resource module with a tiny machine GRUG works' '
-    flux module remove resource &&
+    unload_resource &&
     flux module load resource load-file=${grug} \
 load-format=grug prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with an XML works' '
-    flux module remove resource &&
+    unload_resource &&
     flux module load resource load-file=${xml} \
 load-format=hwloc prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with no option works' '
-    flux module remove resource &&
+    unload_resource &&
     flux module load resource prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with a nonexistent GRUG fails' '
-    flux module remove resource &&
+    unload_resource &&
     test_expect_code 1 flux module load resource load-file=${ne_grug} \
 load-format=grug prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with a nonexistent XML fails' '
+    unload_resource &&
     test_expect_code 1 flux module load resource load-file=${ne_xml} \
 load-format=hwloc prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with known policies works' '
+    unload_resource &&
     flux module load resource policy=high &&
     flux module remove resource &&
     flux module load resource policy=low &&
@@ -69,7 +82,7 @@ test_expect_success 'loading resource module with known policies works' '
 '
 
 test_expect_success 'loading resource module with unknown policies is tolerated' '
-    flux module remove resource &&
+    unload_resource &&
     flux module load resource policy=foo &&
     flux module remove resource &&
     flux module load resource policy=bar


t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
+    unload_resource &&
     flux module load resource policy=foo &&
-    flux module remove resource &&
+    unload_resource &&
     flux module load resource policy=bar
 '
 
 test_expect_success 'removing resource works' '
-    flux module remove resource
+    unload_resource
```
dongahn: same

garlick: Sorry what's the suggestion?

dongahn: Sorry... somehow all of my suggestions except for this is gone. Let me repost.

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
 
+# Unload the resource module
+# If module is not loaded, suppress error.
+unload_resource() {
+    flux module remove resource || /bin/true
+}
+
 test_expect_success 'loading resource module with a tiny machine GRUG works' '
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: It feels like using the plain `flux module remove` without the error message suppression is a bit more future proof for normal cases? Hypothetically, if you add more error checking into `flux-module remove` in the future and it happens to find an error, we want this test to be failed and see that error message.

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource load-file=${grug} \
 load-format=grug prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with an XML works' '
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: same

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource load-file=${xml} \
 load-format=hwloc prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with no option works' '
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: same

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with a nonexistent GRUG fails' '
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: same

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     test_expect_code 1 flux module load resource load-file=${ne_grug} \
 load-format=grug prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with a nonexistent XML fails' '
+    unload_resource &&
```
dongahn: Good that you added this! This way, we can still run this test even if the previous test mistakenly loaded the module.

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     test_expect_code 1 flux module load resource load-file=${ne_xml} \
 load-format=hwloc prune-filters=ALL:core
 '
 
 test_expect_success 'loading resource module with known policies works' '
+    unload_resource &&
     flux module load resource policy=high &&
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: Maybe use the regular `flux module remove`?

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource policy=low &&
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: Same

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource policy=locality
 '
 
 test_expect_success 'loading resource module with unknown policies is tolerated' '
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: same.

t/t4000-match-params.t
```diff
@@ -32,51 +32,59 @@ test_debug '
...
     flux module load resource policy=foo &&
-    flux module remove resource &&
+    unload_resource &&
```
dongahn: same

t/t4000-match-params.t
garlick: Good point, I should at least only suppress the one error where the module being removed is not loaded.  Will propose a fix.

t/t4000-match-params.t
garlick: OK you are right, I got a little over-zealous with the wrapper script :-)

garlick:
Just forced a push
- function now suppresses only a "No such file or directory" error
- other errors are displayed and will be fatal in a test
- call the function only at the beginning of each test, to be resilient to the result of the previous test (this is still some of the places you called out @dongahn so please LMK if you still want those changed).

grondo:
I am going to merge this now so that other PRs can start passing travis. It seems like @garlick has addressed the review comments so hopefully @dongahn approves! (Sorry if not)

dongahn:
@grondo: yap. LGTM too. Thank you @garlick!

PR #565, opened 2020-01-12
Fix build/test issues and enable travis on CentOS 8

### LINKED ISSUE #563: flux-resource: command exited with 3, we wanted 16
I'm trying to get flux-sched to build on CentOS 8 with python 3.6.8 as the default `/usr/bin/python`.

After fixing various other small errors, some of the sched tests are failing similar to this:
```
I/O error(16): Device or resource busy
test_expect_code: command exited with 3, we wanted 16 flux resource match allocate /usr/src/t/data/resource/jobspecs/basics/test001.yaml
not ok 2 - After all resources are allocated, reserve still works
#	
#	    flux resource match allocate ${jobspec} &&
#	    flux resource match allocate ${jobspec} &&
#	    flux resource match allocate ${jobspec} &&
#	    flux resource match allocate ${jobspec} &&
#	    test_expect_code 16 flux resource match allocate ${jobspec} &&
#	    flux resource match allocate_orelse_reserve ${jobspec} &&
#	    flux resource match allocate_orelse_reserve ${jobspec} &&
#	    flux resource match allocate_orelse_reserve ${jobspec} &&
#	    flux resource match allocate_orelse_reserve ${jobspec}
#	

```

It appears that this block in `flux-resource` is not working the same under python 3.6 as it does under other versions:

```python
    #
    # Parse the args and call an action routine as part of that
    #
    try:
        args = parser.parse_args ()
        args.func (args)

    except IOError as e:
        print ("I/O error({0}): {1}".format (e.errno, e.strerror))
        exit (3)

    except EnvironmentError as e:
        print ("Environment error({0}): {1}".format (e.errno, e.strerror))
        if e.errno == errno.EBUSY:  # resource currently unavailable
            exit (16)
        if e.errno == errno.ENODEV: # unsatisfiable jobspec
            exit (19)
        else:
            exit (1)

    except yaml.YAMLError as e:
        print ("Parsing error: ", e)
        exit (2)
```

I'm not good enough with Python to understand why this is happening, but it would seem that `EBUSY` is being thrown as an `IOError` instead of an `EnvironmentError`, and thus the program is exiting with status `3` instead of the expected `16`.

grondo:
This PR enables a CentOS 8 build for flux-sched and fixes associated errors.

Many of the changes here were required due to the fact that Python 3.6 is the default /usr/bin/python in the centos8 docker image. I have no idea if I made the right fixes there, so feedback welcome!

Fixes #563

### DIFF
--- a/t/scripts/flux-jsonschemalint
+++ b/t/scripts/flux-jsonschemalint
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 import argparse
 import errno
 import json
@@ -9,16 +11,16 @@ import jsonschema
     Print json document and schema as a verbose option
 """
 def print_verbose (doc, schema):
-    print "=============================================="
-    print " Jsonschema "
-    print "=============================================="
-    print (json.dumps (schema, indent=4, sort_keys=True))
-    print ""
-    print "=============================================="
-    print " Json document "
-    print "=============================================="
-    print (json.dumps (doc, indent=4, sort_keys=True))
-    print ""
+    print("==============================================")
+    print(" Jsonschema ")
+    print("==============================================")
+    print(json.dumps (schema, indent=4, sort_keys=True))
+    print("")
+    print("==============================================")
+    print(" Json document ")
+    print("==============================================")
+    print(json.dumps (doc, indent=4, sort_keys=True))
+    print("")
 
 
 """
@@ -44,22 +46,22 @@ def main ():
 
         jsonschema.validate (doc, schema)
         valstr = "Document(\"{0}\") validates against schema (\"{1}\")"
-        print valstr.format (args.json, args.schema)
+        print (valstr.format (args.json, args.schema))
 
     except IOError as e:
-        print "I/O error({0}): {1}".format (e.errno, e.strerror)
+        print ("I/O error({0}): {1}".format (e.errno, e.strerror))
         exit (1)
     except EnvironmentError as e:
-        print "Environment error({0}): {1}".format (e.errno, e.strerror)
+        print ("Environment error({0}): {1}".format (e.errno, e.strerror))
         exit (1)
     except ValueError as estr:
-        print "Value error: {0}".format (str (estr))
+        print ("Value error: {0}".format (str (estr)))
         exit (1)
     except jsonschema.exceptions.ValidationError as e:
-        print "Jsonschema validation error: {0}".format (e.message)
+        print ("Jsonschema validation error: {0}".format (e.message))
         exit (1)
     except jsonschema.exceptions.SchemaError as e:
-        print "Jsonschema schema error: {0}".format (e.message)
+        print ("Jsonschema schema error: {0}".format (e.message))
         exit (1)
 
 if __name__ == "__main__":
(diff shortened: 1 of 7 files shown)


t/scripts/flux-jsonschemalint
```diff
@@ -9,16 +9,16 @@ import jsonschema
...
-    print " Jsonschema "
-    print "=============================================="
-    print (json.dumps (schema, indent=4, sort_keys=True))
-    print ""
-    print "=============================================="
-    print " Json document "
-    print "=============================================="
-    print (json.dumps (doc, indent=4, sort_keys=True))
-    print ""
+    print("==============================================")
```
SteVwonder: at the top of the file, you will want to add `from __future__ import print_function` to properly support this print syntax in py2.  It doesn't crash py2, but py2 interprets this as the printing of a newly constructed tuple containing a string.

t/t4003-cancel-info.t
```diff
@@ -71,7 +71,7 @@ test_expect_success 'resource-info on allocated jobs works' '
 '
 
 test_expect_success 'cancel on nonexistent jobid is handled gracefully' '
-    test_expect_code 1 flux resource cancel 100000
+    test_expect_code 3 flux resource cancel 100000
```
SteVwonder: Nice catch!

dongahn:
> Many of the changes here were required due to the fact that Python 3.6 is the default /usr/bin/python in the centos8 docker image. I have no idea if I made the right fixes there, so feedback welcome!

OK. I left my comment to #563

t/scripts/flux-jsonschemalint
grondo: Ah, thanks. Not sure why I didn't think of that when I added it to the other script. :man_facepalming:

dongahn:
@grondo: this can go in after adding "from __future__ import print_function" and rebase.

grondo:
Ok, rebased and force-pushed. Both python scripts should now have `from __future__ import print_function`

t/scripts/flux-jsonschemalint
```diff
@@ -9,16 +11,16 @@ import jsonschema
...
+    print(" Jsonschema ")
```
dongahn: LGTM. Just out of curiosity, we don't have that leading whitespace rule for python function call, right?

dongahn:
@grondo: looks good to me. I will merge it now. I just left one question about our python coding guide. If we put a code guide in place, we can have a global pass.

t/scripts/flux-jsonschemalint
grondo: No. In fact if we enabled the `black` formatter it would probably try to remove the leading spaces here. I tried to keep the existing format and only add parens as necessary, but may have inadvertently made some whitespace changes. Sorry!

grondo:
> @grondo: looks good to me. I will merge it now. I just left one question about our python coding guide. If we put a code guide in place, we can have a global pass.

Yeah, if you plan to add more python scripts in the future for flux-sched, you may want to adopt the style checks @trws added to flux-core `.travis.yml`. Alternately, a GitHub action to run `black` formatter could be used.

dongahn:
> Yeah, if you plan to add more python scripts in the future for flux-sched, you may want to adopt the style checks @trws added to flux-core .travis.yml. Alternately, a GitHub action to run black formatter could be used.

Good to know. So our python coding guy is "black formatter" then.

Issue #567, opened 2020-01-14
problems when building against flux-core configured with Python 3.6 when /usr/bin/python != 3.6


grondo:
I'm not able to build flux-sched RPMs against our new flux-core v0.14 RPMs. The flux-core RPMs are now successfully built against the system installation of Python 3.6, though obviously `/usr/bin/python` is still pointing to 2.7.

The first issue is that flux-sched `./configure` doesn't seem to obey `PYTHON_VERSION=3.6` and instead always detects the python version of `/usr/bin/python`. In the buildfarm, this means that even though the Python 3.6 `yaml` and `jsonschema` modules are pulled in as dependencies of `flux-core`, flux-sched does not detect them because it is looking for python 2.7 versions of the modules.

However, the Python dependencies in `flux-sched` are only required for `make check`, so perhaps it is ok to just install the python 2.7 versions of the modules to satisfy configure.

Even outside the buildfarm, however, I'm unable to get `flux-sched`'s `make check` to work when built against a `flux-core` installation that is configured with Python 3.6. For example, I'm getting this error from the `flux-resource` test script:

```
expecting success: 
    flux resource match allocate ${jobspec} &&
    flux resource match allocate ${jobspec} &&
    flux resource match allocate ${jobspec} &&
    flux resource match allocate ${jobspec}

Traceback (most recent call last):
  File "/g/g0/grondo/git/flux-sched.git/t/scripts/flux-resource", line 291, in <module>
    main ()
  File "/g/g0/grondo/git/flux-sched.git/t/scripts/flux-resource", line 272, in main
    args.func (args)
  File "/g/g0/grondo/git/flux-sched.git/t/scripts/flux-resource", line 74, in match_alloc_action
    r = ResourceModuleInterface ()
  File "/g/g0/grondo/git/flux-sched.git/t/scripts/flux-resource", line 27, in __init__
    self.f = flux.Flux ()
  File "/tmp/grondo/flux/lib/flux/python3.6/flux/__init__.py", line 17, in Flux
    import flux.core.handle
  File "/tmp/grondo/flux/lib/flux/python3.6/flux/core/handle.py", line 14, in <module>
    from flux.rpc import RPC
  File "/tmp/grondo/flux/lib/flux/python3.6/flux/rpc.py", line 14, in <module>
    from flux.future import Future
  File "/tmp/grondo/flux/lib/flux/python3.6/flux/future.py", line 13, in <module>
    from flux.util import check_future_error
[...]

PR #578, opened 2020-01-28
README: add copy-and-pastable dependency install line for ubuntu

SteVwonder:
also change the python-yaml dependency to python3-yaml, since flux-core has moved to python3 by default


SteVwonder:
I was a little too eager with the `python3` change.  Please don't merge this until #567 is fixed.

dongahn:
LGTM. Merging.

SteVwonder:
Thanks @dongahn!

PR #582, opened 2020-01-29
Add flux-tree support

dongahn:
This PR is experimental as I wanted to  get some feedback from end users (in particular @jamescorbett) who are waiting for this feature. 

This adds a new flux command called flux-tree, which creates Flux instance hierarchy according to the specified policies and schedule/run the specified number of Jobscripts at the last level of this hierarchy.

As a next step, in addition to gathering feedback from the users, I plan to heavily use the `--dry-run` option to add test cases.

### DIFF
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -1,0 +1,865 @@
+#! /bin/bash
+#
+###############################################################################
+#   Copyright (c) 2020 Lawrence Livermore National Security, LLC.  Produced at
+#   the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
+#   LLNL-CODE-658032 All rights reserved.
+# 
+#   This file is part of the Flux resource manager framework.
+#   For details, see https://github.com/flux-framework.
+# 
+#   This program is free software; you can redistribute it and/or modify it
+#   under the terms of the GNU General Public License as published by the Free
+#   Software Foundation; either version 2 of the license, or (at your option)
+#   any later version.
+# 
+#   Flux is distributed in the hope that it will be useful, but WITHOUT
+#   ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+#   FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+#   GNU General Public License for more details.
+# 
+#   You should have received a copy of the GNU General Public License along
+#   with this program; if not, write to the Free Software Foundation, Inc.,
+#   59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+#   See also:  http://www.gnu.org/licenses/
+###############################################################################
+
+set -o errexit
+set -o pipefail
+#set -o xtrace
+
+FT_TOPO='1'                    # store arg to --topology (e.g., 2x2)
+FT_QUEUE='default'             # store arg to --queue-policy (e.g., fcfs:easy)
+FT_PARAMS='default'            # store arg to --queue-params
+FT_MATCH='default'             # store arg to --match-policy (e.g., low:high)
+FT_PERF_OUT='%^+_no'           # store perf-out filename given by --perf-out
+FT_FXLOGS='%^+_no'             # dir in which flux logs are produced
+FT_LEAF='no'                   # is this a leaf instance (given by --leaf)?
+FT_G_PREFIX='tree'             # store hierarchical path given by --prefix
+FT_DRY_RUN='no'                # --dry-run for testing?
+FT_INTER='no'                  # is an option for internal levels given?      
+FT_MAX_EXIT_CODE=0             # maximum exit code detected
+FT_MAX_FLUX_JOBID=0            # maximum flux jobid that reports max exit code
+FT_MAX_TREE_ID=""              # FLUX_TREE_ID that reports max exit code
+FT_MAX_JOBSCRIPT_IX=""         # FLUX_TREE_JOBSCRIPT_INDEX reporting max code
+                               # --prefix is an internal-use-only option
+ORIG_FLUX_QMANAGER_OPTIONS=''  # 
+ORIG_FLUX_RESOURCE_OPTIONS=''  # to apply and unapply QMANAGER and RESOURCE
+ORIG_FLUX_QMANAGER_RC_NOOP=''  # module load options.
+ORIG_FLUX_RESOURCE_RC_NOOP=''  #
+
+declare -i FT_NJOBS=1          # store num of jobs to run, given by --njobs
+declare -i FT_NNODES=1         # store num of nodes assigned, given by --nnodes
+declare -i FT_NCORES=1         # store num of cores per node (--ncores-per-node)
+declare -i FT_NGPUS=0          # store num of gpus per node (--ngpus-per-node)
+declare -r top_prefix='tree'   # prefix name to identify the top Flux instance
+declare -r t_delim='x'         # topology delimiter
+declare -r p_delim=':'         # match policy delimiter
+declare -r perf_format='%-15s %20s %20s %20s %15s %15s %5s %4s %4s'
+declare -r perf_keys=' .treeid, .perf.elapse, .perf.begin, .perf.end,'\
+' .perf.match, .njobs, .my_nodes, .my_cores, .my_gpus '
+declare -r perf_heading=' TreeID Elapse(sec) Begin(Epoch) End(Epoch) Match(usec)'\
+' NJobs NNodes CPN GPN '
+declare -A mp_policies=(       # make sure to update this when match
+    [low]=1                    # policies are updated.
+    [high]=1
+    [locality]=1
+    [variation]=1
+    [default]=1
+)
+declare -A qp_policies=(       # make sure to update this when
+    [fcfs]=1                   # queuing policies are updated.
+    [easy]=1
+    [hybrid]=1
+    [conservative]=1
+    [default]=1
+)
+declare -A q_params=(          # make sure to update this when
+    [queue-depth]=1            # queuing parameters are updated.
+    [reservation-depth]=1
+    [default]=1
+)
+declare -a jobids             # array to store a set of submitted job IDs
+
+declare -r long_opts='help,leaf,flux-logs:,nnodes:,ncores-per-node:'\
+',ngpus-per-node:,topology:,queue-policy:,queue-params:,match-policy:'\
+',njobs:,perf-out:,prefix:,dry-run'
+
+declare -r short_opts='hlf:N:c:g:T:Q:P:M:J:o:X:d'
+declare -r prog=${0##*/}
+declare -r usage="
+Usage: ${prog} [OPTIONS] -- Jobscript\n\
+\n\
+Create a Flux instance hierarchy according to the specified\n\
+policies and schedule/run the specified number\n\
+of Jobscripts at the last level of this hierarchy.\n\
+\n\
+If --topology=2x4 and --njobs=32 are given, for instance,\n\
+2 Flux instances will be spawned from within the current instance,\n\
+each of which will in turn spawn 4 child Flux instances, totaling\n\
+8 instances at the last level of this hierarchy.\n\
+Once this is done, 4 jobs (of Jobscripts) will be scheduled\n\
+and executed at each of these 8 last-level Flux instances.\n\
+\n\
+The resources specified by --nnodes (total number of nodes) and\n\
+--ncores-per-node (total number of cores per node)\n\
+are recursively divided such that each sibling Flux instance\n\
+will be assigned to an equal split of the resources of their\n\
+parent instance. In addition, --ngpus-per-node can be given,\n\
+in which case the given GPU count will also be split.\n\
+If not given, it is assumed that there is no GPU on nodes.\n\
+\n\
+Jobscript is expected to submit one or more programs through\n\
+the flux-job submit command or its variants.\n\
+Jobscript is passed with five environment variables:\n\
+FLUX_TREE_ID, FLUX_TREE_JOBSCRIPT_INDEX, FLUX_TREE_NNODES,\n\
+FLUX_TREE_NCORES_PER_NODE and FLUX_TREE_NGPUS_PER_NODE.\n\
+FLUX_TREE_ID is an ID string uniquely identifing the hierarchical\n\
+path of the Flux instance on which Jobscript is being executed.\n\
+FLUX_TREE_JOBSCRIPT_INDEX is the integer ID of each jobscript\n\
+invocation local to the Flux instance. It starts from 1 and\n\
+sequentially increases.\n\
+FLUX_TREE_NNODES is the number nodes assigned to the instance.\n\
+FLUX_TREE_NCORES_PER_NODE is the number of cores per node\n\
+assigned to the instance.\n\
+FLUX_TREE_NGPUS_PER_NODE is the number of GPUs per node\n\
+assigned to the instance.\n\
+\n\
+If --queue-policy (additionally --queue-params) and/or\n\
+--match-policy are given, each level of this hierarchy will\n\
+be set to the specified queuing and matching policies and\n\
+parameters. Otherwise, all levels will be configured\n\
+to be used either the default policies or policies specified\n\
+through the FLUX_RESOURCE_OPTIONS and/or FLUX_QMANAGER_OPTIONS\n\
+environment variables.\n\
+\n\
+If any one of Jobscripts returns a non-zero exit code, flux-tree\n\
+detects the script invocation exited with the highest code and print\n\
+both that exit code and the outputs printed from executing the script.\n\
+In this case, FLUX_TREE_ID and FLUX_TREE_JOBSCRIPT_INDEX are also\n\
+reported in the from of \${FLUX_TREE_ID}@index[\${FLUX_TREE_JOBSCRIPT_INDEX}]\n\
+\n\
+Options:\n\
+ -h, --help                    Display this message\n\
+ -l, --leaf                    Leaf instance. Directly submit jobs\n\
+                               to enclosing Flux instance. Mutually-exclusive\n\
+                               with internal tree-node options like -T.\n\
+                                   (default=${FT_LEAF})\n\
+ -f, --flux-logs=DIR           Dump Flux logs for all instances into DIR\n\
+ -N, --nnodes=NNODES           Total num of nodes to use\n\
+                                   (default=${FT_NNODES})\n\
+ -c, --ncores-per-node=NCORES  Total num of cores per node to use\n\
+                                   (default=${FT_NCORES})\n\
+ -g, --ngpus-per-node=NGPUS    Total num of gpus per node to use\n\
+                                   (default=${FT_NGPUS})\n\
+ -T, --topology=HPOLICY        Topology of Flux instance hierarchy:\n\
+                                   e.g., 2x2 (default=${FT_TOPO})\n\
+ -Q, --queue-policy=QPOLICY    Queuing policy for each level of\n\
+                                   the hierarchy: e.g., easy:fcfs\n\
+ -P, --queue-params=QPARAMS    Queuing parameters for each level of\n\
+                                   the hierarchy: e.g.,\n\
+                                   queue-depth=5:reservation-depth=5\n\
+ -M, --match-policy=MPOLICY    Match policy for each level of\n\
+                                   the hierarchy: e.g., low:high\n\
+ -J, --njobs=NJOBS             Total num of Jobscripts to run\n\
+                                   (default=${FT_NJOBS})\n\
+ -o, --perf-out=FILENAME       Dump the performance data into\n\
+                                   the given file (default: don't print)\n\
+     --                        Stop parsing options after this\n\
+"
+
+die() { echo -e "${prog}:" "$@"; exit 1; }
+warn() { echo -e "${prog}: warning:" "$@"; }
+dr_print() { echo -e "${prog}: dry-run:" "$@"; }
+
+#
+# Write the performance data into the given file.  Should be called
+# by the upper layer when the performance data record for each and every
+# Flux instance has been aggregated. Then, this is recursively invoked.
+#
+dump_perf() {
+    local blurb="${1}"
+    local out="${2}"
+
+    [[ "x${blurb}" = "x" ]] && die "empty json"
+
+    # Parse perf json for the given level into a record
+    local tuple=''
+    local length=0
+    tuple=$(echo "${blurb}" | jq "${perf_keys}") || die "invalid json"
+    length=$(echo "${blurb}" | jq ' .child | length ') || die "invalid json"
+
+    # Print the record for this level Didn't quote $tuple to remove new lines
+    printf "${perf_format}\n" ${tuple} >> "${out}"
+
+    # Recurse down to parse the record for each child Flux instance
+    for i in $(seq 0 $(( length - 1 )))
+    do
+        local new_o=''
+        new_o=$(echo "${blurb}" | jq " .child[${i}] ")
+        dump_perf "${new_o}" "${out}"
+    done
+}
+
+
+#
+# Roll up the performance records for each Flux instance to the KVS
+# guest namesapce of the parent Flux instance or print them out if top level.
+#
+rollup() {
+    local prefix="${1}"
+    local blurb="${2}"
+    local out="${3}"
+
+    if [[ "${out}" != "%^+_no" ]]
+    then
+        # If output filename is given (e.g., top-level), dump the data
+        # to the file. Didn't quote $perf_heanding on purpose below.
+        rm -f "${out}"
+        printf "${perf_format}\n" ${perf_heading} > "${out}"
+        dump_perf "${blurb}" "${out}"
+    else
+        if [[ "${prefix}" != "${top_prefix}" ]]
+        then
+            if [[ "x${FT_DRY_RUN}" = "xno" ]]
+            then
+                # Put it to KVS parent namespace except when we are at top level
+                flux --parent kvs put -j "tree-perf=${blurb}" || warn "KVS error"
+            fi
+        fi
+    fi
+}
+
+
+#
+# Return a JSON string out of the performance data passed.
+#
+jsonify() {
+    local prefix="${1}"
+    local njobs="${2}"
+    local nnodes="${3}"
+    local ncores="${4}"
+    local ngpus="${5}"
+    local begin="${6}"
+    local end="${7}"
+    local avg=0
+    local avail="no"
+    local el_match=0
+
+    # Print resource match time only for internal study
+    # flux-resource isn't a public command
+    if [[ "x${FT_DRY_RUN}" = "xno" ]]
+    then
+        flux resource -h > /dev/null 2>&1 && avail="yes"
+    fi
+
+    if [[ "${avail}" = "yes" ]]
+    then
+        avg=$(flux resource stat | grep "Avg" | awk '{print $4}')
+        el_match=$(awk "BEGIN {print ${avg}*${njobs}*1000000.0}")
+    fi
+
+    local elapse=0
+    elapse=$(awk "BEGIN {print ${end} - ${begin}}")
+    echo "{\"treeid\":\"${prefix}\",\"njobs\":${njobs},\"my_nodes\":${nnodes},\
+\"my_cores\":${ncores},\"my_gpus\":${ngpus},\"perf\":{\"begin\":${begin},\
+\"end\":${end},\"elapse\":${elapse},\"match\":${el_match}}}"
+}
+
+
+#
+# Fetch the next topology parameter that will be passed to
+# the next-level Flux instances. E.g., If the current level topology
+# is 2x3x4, the topology handled at the next level will be 3x4. 
+#
+next_topo() {
+    local topo="${1}"
+    local nx=''
+    local nfields=0
+    nfields=$(echo "${topo}" | awk -F"${t_delim}" '{print NF}')
+    # Remove the first topo parameter 
+    [[ ${nfields} -gt 1 ]] && nx="${topo#*${t_delim}}"
+    echo "${nx}"
+}
+
+
+#
+# Fetch the next policy parameter that will be passed to
+# the next-level Flux instances. E.g., If the current policy parameter 
+# is high:low:locality, the policies handled at the next level
+# will be low:locality. 
+#
+next_policy_or_param() {
+    local policy_or_param="${1}"
+    local nx=""
+    local nfields=0
+    nfields=$(echo "${policy_or_param}" | awk -F"${p_delim}" '{print NF}')
+    [[ ${nfields} -gt 1 ]] && nx="${policy_or_param#*${p_delim}}"
+    echo "${nx}"
+}
+
+
+#
+# Check if the given queuing policy is valid 
+#
+qpolicy_check() {
+    local policy=${1%%${p_delim}*}
+    [[ "x${policy}" = "x" ]] && return 1
+    [[ "${qp_policies["${policy}"]:-missing}" = "missing" ]] && return 1
+    return 0
+}
+
+
+#
+# Check if the given match policy is valid 
+#
+mpolicy_check() {
+    local policy=${1%%${p_delim}*}
+    [[ "x${policy}" = "x" ]] && return 1
+    [[ "${mp_policies["${policy}"]:-missing}" = "missing" ]] && return 1
+    return 0
+}
+
+
+#
+# Check if the given queue param is valid 
+#
+qparams_check() {
+    local param=''
+    param=$(echo "${1}" | awk -F"${p_delim}" '{print $1}')
+    param=${1%%${p_delim}*}
+    local final_param=''
+    final_param=${param##*,}
+
+    for i in $(seq 1 10)
+    do
+        local token1=${param%%,*}
+        local token2=${token1%=*}
+        [[ "x${token2}" = "x" ]] && return 1
+        [[ "${q_params["${token2}"]:-missing}" = "missing" ]] && return 1
+        [[ "x${token1}" = "x${final_param}" ]] && break
+        param=${param#*,}
+    done
+    return 0
+}
+
+
+#
+# Calculate the number of jobs to execute based on the number of Flux instances
+# being used at a level and the rank of the instance amongst its siblings.
+#
+get_my_njobs(){
+    local njobs="${1}"
+    local size="${2}"          # rank starts from 1
+    local rank="${3}"
+    echo $(( njobs / size + (size + njobs % size)/(size + rank) ))
+}
+
+
+#
+# Calculate the total number of cores that will be assigned to a child
+# Flux instance based on the total number of nodes and cores per node
+# assigned to the current Flux instance as well as the size and rank parameter.
+#
+get_my_cores(){
+    local nnodes="${1}"
+    local ncores="${2}"
+    local size="${3}"
+    local rank="${4}"
+    local t_cores=$(( nnodes * ncores ))
+    echo $(( t_cores / size + (size + t_cores % size) / (size + rank) ))
+}
+
+
+#
+# Calculate the total number of GPUs that will be assigned to a child
+# Flux instance based on the total number of nodes and GPUs per node
+# assigned to the current Flux instance as well as the size and rank parameter.
+#
+get_my_gpus(){
+    local nnodes="${1}"
+    local ngpus="${2}"
+    local size="${3}"
+    local rank="${4}"
+    local t_gpus=$(( nnodes * ngpus ))
+    echo $(( t_gpus / size + (size + t_gpus % size) / (size + rank) ))
+}
+
+
+#
+# Adjust the number of Flux instances to spawn at the next level
+# if the amount of resources managed by the parent instance is small.
+#
+get_effective_size(){
+    local ncores="${1}"
+    local ngpus="${2}"
+    local size="${3}"
+    [[ ${ngpus} -ne 0 && ${ngpus} -lt ${size} ]] && size=${ngpus}
+    [[ ${ncores} -lt ${size} ]] && size=${ncores}
+    echo "${size}"
+}
+
+
+#
+# Calculate the total number of nodes that will be assigned to a child
+# Flux instance based on the total number of cores per node as well as
+# the total number of cores assigned to this child instance. Returns
+# minimum num of nodes required.
+#
+get_my_nodes(){
+    local ncores="${1}"
+    local m_cores="${2}"
+    echo $(( m_cores / ncores + (ncores + m_cores % ncores) / (ncores + 1 )))
+}
+
+
+#
+# Apply all of the policies for the target Flux instance
+# by setting environment variables.
+#
+apply_policies() {
+    local queue_policy="${1%%${p_delim}*}"
+    local queue_param="${2%%${p_delim}*}"
+    local match_policy="${3%%${p_delim}*}"
+
+    ORIG_FLUX_QMANAGER_OPTIONS=${FLUX_QMANAGER_OPTIONS:-none}
+    ORIG_FLUX_RESOURCE_OPTIONS=${FLUX_RESOURCE_OPTIONS:-none}
+    ORIG_FLUX_QMANAGER_RC_NOOP=${FLUX_QMANAGER_RC_NOOP:-none}
+    ORIG_FLUX_RESOURCE_RC_NOOP=${FLUX_RESOURCE_RC_NOOP:-none}
+    unset FLUX_QMANAGER_RC_NOOP
+    unset FLUX_RESOURCE_RC_NOOP
+
+    if [[ "${queue_policy}" != "default" ]]
+    then
+        export FLUX_QMANAGER_OPTIONS="queue-policy=${queue_policy}"
+    fi
+    if [[ "${queue_param}" != "default" ]]
+    then
+        local qo="${FLUX_QMANAGER_OPTIONS}"
+        export FLUX_QMANAGER_OPTIONS="${qo:+${qo},}queue-params=${queue_param}"
+    fi
+    if [[ "${match_policy}" != "default" ]]
+    then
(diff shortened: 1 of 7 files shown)


jamescorbett:
Wrong jamescorbett I'm afraid.

dongahn:
Tagging @jameshcorbett. (Sorry @jamescorbett)

jameshcorbett:
@dongahn any chance you could change line 725 of flux-tree to something like `[[ ! -x ${script} ]] && [[ ! -x $(which {script}) ]] && die "${script} does not exist or is not executable!"`? Or whatever would make it consistent with the other Flux utilities. Flux-tree complained when I set my jobscript to `python myscript.py`.

jameshcorbett:
Any chance that an environment variable like `FLUX_TREE_TOPOLOGY` could be exported to each jobscript with the same value as was passed in for `--topology`? With `FLUX_TREE_ID` it would allow jobscript to completely understand its place in the tree.

dongahn:
> @dongahn any chance you could change line 725 of flux-tree to something like [[ ! -x ${script} ]] && [[ ! -x $(which {script}) ]] && die "${script} does not exist or is not executable!"? Or whatever would make it consistent with the other Flux utilities. 

Sure. This can be done. 

> Flux-tree complained when I set my jobscript to python myscript.py.

> Also, can Flux-tree even recognize jobscript taking command line options? I'm looking at script="{$1}" on line 721... wouldn't that ignore any command line options that were supposed to be passed to jobscript? Shouldn't it be script="{$@}"?

Seems like you want the entire arguments (after parsing options) to be passed to the leaf level. Yes. This would be an good enhancement. I will add that mod to this PR.

dongahn:
> Any chance that an environment variable like FLUX_TREE_TOPOLOGY could be exported to each jobscript with the same value as was passed in for --topology? With FLUX_TREE_TOPOLOGY it would allow jobscript to completely understand its place in the tree.

The global topology isn't propagated to the lower levels. And the static global topology the user specific might not be the same as the instantiated topology. If there is lack of resources (and/or the number of jobs) to be divided up, `flux-tree` wouldn't create all the specify the topology.

What is the logic you want to program to `FLUX_TREE_TOPOLOGY` and `FLUX_TREE_TOPOLOGY`?

jameshcorbett:
> The static global topology the user specific might not be the same as the instantiated topology. If there is lack of resources (and/or the number of jobs) to be divided up, `flux-tree` wouldn't create all the specify the topology.
> 
> What is the logic you want to program to `FLUX_TREE_TOPOLOGY` and `FLUX_TREE_TOPOLOGY`?

Oh, that's right, I forgot about those complications. What I was trying to do isn't that important--I can easily do it without `FLUX_TREE_TOPOLOGY`. I just thought it would be easy and convenient.

dongahn:
> I just thought it would be easy and convenient.

I am just trying to see if I can get you to achieve your goal here because other users might have the same needs. So if you tell me how having `FLUX_TREE_TOPOLOGY` makes your task easy, this will be helpful for me to think through.

jameshcorbett:
That's fair. With `FLUX_TREE_TOPOLOGY` I would have used `FLUX_TREE_ID` to convert that tree ID into a zero-based increasing sequence of IDs (0, 1, 2, ...).

dongahn:
>  I would have used FLUX_TREE_ID to convert that tree ID into a zero-based increasing sequence of IDs (0, 1, 2, ...).

That's sort of what I thought! 

I think I can actually export zero-based monotonically increasing IDs at the leaf level where your job script is running. Let me give some thoughts to it.

jameshcorbett:
> That's sort of what I thought!
> 
> I think I can actually export zero-based monotonically increasing IDs at the leaf level where your job script is running. Let me give some thoughts to it.

Ok. I really don't need it, though, so I'd recommend you do it only if it's easy (or if someone else asks you to do it). I really only need the distinction between `ID==0` and `ID!=0`, which I already have with `FLUX_TREE_ID`; I use the specific numerical IDs only to make sure my code picks unique host/port pairs for the server it runs.

By the way, Flux-Tree seems great so far!

dongahn:
> By the way, Flux-Tree seems great so far!

Great to hear. I am glad there is an real use case and a user who kicks the tire for it before landing this new feature @jameshcorbett!

dongahn:
I just added some test cases and added minor changes to `flux-tree` that came out from running those tests. Forced a push. I will look into the following three remaining things after this:

- Test the given jobscript with search paths
- Add support such that the jobscript can be given its own options and arguments
- Add an environment variable so that each leaf-level instance can get a unique monotonically increasing integer ID. 

Once they are added, I will remove WIP tag and  this can be officially reviewed and merged. (as needed for near term work of @jameshcorbett and @SteVwonder.

jameshcorbett:
Thanks Dong! Quick question: does `flux-tree` take gpus into account when creating the hierarchy?

dongahn:
No it doesn"t. But it can be easily added with --ngpus-per-node option. If that option is not provided, flux tree can still use nnodes and ncores as the main resources to be divvied up. If given, all three resource types can be considered. Does it sound like what you want?

jameshcorbett:
Yeah, that sounds like it would be pretty useful.

dongahn:
>  But it can be easily added with --ngpus-per-node option. If that option is not provided, flux tree can still use nnodes and ncores as the main resources to be divvied up. If given, all three resource types can be considered.

@SteVwonder: I just added `--ngpus-per-node` but it doesn't look like the `-g` option isn't supported by the underlying `flux-jobspec srun`. I am not using `flux mini submit` interface.

Is it easy to add `-g` or whatever option (like `--gpus-per-task` to `flux-jobspec srun`?

dongahn:
@jameshcorbett and @SteVwonder:

I forced a push with support for 

> Test the given jobscript with search paths
> Add support such that the jobscript can be given its own options and arguments
> GPU support

But I didn't pursue the following yet because it is not needed currently. 

> Add an environment variable so that each leaf-level instance can get a unique monotonically increasing integer ID.

Also I didn't hook up the GPU support into `flux jobspec srun` so that the GPU binding is enforced by the underlying exec system (via `CUDA_VISIBLE_DEVICE`). 

This should be good to be reviewed and perhaps get merged with those two limitations addressed later...

src/cmd/flux-tree
```diff
@@ -0,0 +1,808 @@
...
+each of which will in turn spawn 4 child Flux instances, totaling\n\
+8 instances at the last level of this hierarchy.\n\
+Once this is done, 4 jobs (of Jobscripts) will be scheduled\n\
+and executed at each of these 8 last-level Flux instances.\n\
+\n\
+The resources specified by --nnodes and --ncores-per-node\n\
+are recursively divided such that each sibling Flux instance\n\
+will be assigned to an equal split of the resources of their\n\
+parent instance. In addition, --ngpus-per-node can be given,\n\
+in which case the given GPU count will also be splitted\n\
```
grondo: replace "splitted" with "split".

src/cmd/flux-tree
```diff
@@ -0,0 +1,808 @@
...
+\n\
+Options:\n\
+ -h, --help                    Display this message\n\
+ -l, --leaf                    Leaf (default=${leaf})\n\
+ -f, --flux-logs               Dump Flux logs for all instances\n\
+ -N, --nnodes=NNODES           Total num of nodes to use\n\
+                                   (default=${nnodes})\n\
+ -c, --ncores-per-node=NCORES  Total num of cores per node to use\n\
+                                   (default=${ncores})\n\
+ -c, --ngpus-per-node=NGPUS    Total num of gpus per node to use\n\
```
grondo: Did you mean `-g, --ngpus-per-node=NGPUS`?

src/cmd/flux-tree
```diff
@@ -0,0 +1,808 @@
...
+    local S="${cl}"
+    local rank=0
+
+    # Main Loop to Submit the Next-Level Flux Instances
+    size=$(get_effective_size "${ncores}" "${ngpus}" "${size}")
+    apply_policies "${3}" "${4}" "${5}"
+    for rank in $(seq 1 "${size}"); do
+        local my_cores=0
+        my_cores=$(get_my_cores "${nnodes}" "${ncores}" "${size}" "${rank}")
+	local my_gpus=0
```
grondo: mixed indent level here

src/cmd/flux-tree
```diff
@@ -0,0 +1,808 @@
...
+
+        # Flux instance rank-aware command-line options
+        local J="--njobs=${my_njobs}"
+        local o=''
+        [[ "${log}" = "yes" ]] && o="-o,-S,log-filename=${prefix}.${rank}.log"
+        local N=0
+        N=$(get_my_nodes "${ncores}" "${my_cores}")
+        local c=0
+        c=$((my_cores/N + (my_cores + my_cores % N)/(my_cores + 1)))
+	local g=0
```
grondo: More lines with indent mismatch

src/cmd/flux-tree
```diff
@@ -0,0 +1,808 @@
...
+    blurb=$(jsonify "${X}" "${S}" "${N}" "${c}" "${g}" "${B}" "${E}")
+    local child=''
+    local kvspath=''
+    local rank=0
+
+    for rank in $(seq 1 "${S}");
+    do
+        [[ "x${jobids[${rank}]}" = "x" || "x${dry_run}" = "xyes" ]] && break
+        kvspath=$(flux job id --from=dec --to=kvs "${jobids[${rank}]}")
+        child=$(flux kvs get -j "${kvspath}.guest.tree-perf")
```
grondo: You can also fetch kvs data under `guest` with `flux job info [JOBID] [KEY]`. This not only allows you to skip the `flux job id`, but also has the benefit of working for users that aren't the instance owner at this level in the tree (didn't follow the script enough to know if that is always the case here).

in this case `flux job info ${jobids[${rank}]} tree-perf` will work (I think)

Further down the road, we may want to figure out how to parallelize the fetch of child data. (Possibly a challenge from a shell script)

dongahn:
Great feedback across the board. I will address all of then soon.

dongahn:
> Add an environment variable so that each leaf-level instance can get a unique monotonically increasing integer ID.

It turns out this is a bit difficult to implement as the tree shape is not entire static. If there is lack of resources or jobs, flux-tree won't spawn the children instances further...

dongahn:
I also realized that I need a few test casse where `flux-tree` is run in real mode (not in dry-run mode).

src/cmd/flux-tree
dongahn: > flux job info ${jobids[${rank}]} tree-perf

@grondo: I am getting an error with this change:

```
flux-job: job 2650313588736 id or key not found
```

Maybe the flux-core I'm using on my dev system is too old?

grondo: `flux job info` has been in flux-core for quite awhile, maybe I messed up the syntax. Let me try it out and see if I got it wrong.

grondo: @dongahn, oops I forgot you need to prepend `guest.` to keys stored in the job's KVS namespace.
```
grondo@peep:~/git/f.git$ flux mini run -v flux kvs put tree-perf="data"
jobid: 520144027648
grondo@peep:~/git/f.git$ flux job info 520144027648 guest.tree-perf
data

```

dongahn: Works great. Thanks!

dongahn:
OK. I believe I addressed all of the review comments that came out so far. In particular I used `flux mini run --dry-run` interface to hook up the GPU support. So this should have nearly everything that @jameshcorbett needs. 

I also added a few more real-mode test cases.

W/ one more quick review by someone like @SteVwonder, this can go in from my perspective.

SteVwonder:
> W/ one more quick review by someone like @SteVwonder, this can go in from my perspective.

:+1: Sure thing.  I'll being poking around with `flux-tree` today and Sunday.  I'll use it and give any relevant feedback.

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+of Jobscripts at the last level of this hierarchy.\n\
+\n\
+If --topology=2x4 and --njobs=32 are given, for instance,\n\
+2 Flux instances will be spawned from within the current instance,\n\
+each of which will in turn spawn 4 child Flux instances, totaling\n\
+8 instances at the last level of this hierarchy.\n\
+Once this is done, 4 jobs (of Jobscripts) will be scheduled\n\
+and executed at each of these 8 last-level Flux instances.\n\
+\n\
+The resources specified by --nnodes and --ncores-per-node\n\
```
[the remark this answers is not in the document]
dongahn: Yes, that the total number of cores on each node. This same options are used recursively all the way down to the leaf level Flux instance so the latter is also right when flux-tree is invoked at that level.

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+be set to the specified queuing and matching policies and\n\
+parameters. Otherwise, all levels will be configured\n\
+to be used either the default policies or policies specified\n\
+through the FLUX_RESOURCE_OPTIONS and/or FLUX_QMANAGER_OPTIONS\n\
+environment variables.\n\
+\n\
+Options:\n\
+ -h, --help                    Display this message\n\
+ -l, --leaf                    Leaf (default=${leaf})\n\
+ -f, --flux-logs               Dump Flux logs for all instances\n\
```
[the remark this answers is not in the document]
dongahn: Yeah this is a good idea. Will follow up.

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+--match-policy are given, each level of this hierarchy will\n\
+be set to the specified queuing and matching policies and\n\
+parameters. Otherwise, all levels will be configured\n\
+to be used either the default policies or policies specified\n\
+through the FLUX_RESOURCE_OPTIONS and/or FLUX_QMANAGER_OPTIONS\n\
+environment variables.\n\
+\n\
+Options:\n\
+ -h, --help                    Display this message\n\
+ -l, --leaf                    Leaf (default=${leaf})\n\
```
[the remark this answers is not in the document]
dongahn: I am ambivalent. I can imagine users user this option to repeat a jobscript (similar to jobarray).

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+                                   e.g., 2x2 (default=${topo})\n\
+ -Q, --queue-policy=QPOLICY    Queuing policy for each level of\n\
+                                   the hierarchy: e.g., easy:fcfs\n\
+ -P, --queue-params=QPARAMS    Queuing parameters for each level of\n\
+                                   the hierarchy: e.g.,\n\
+                                   queue-depth=5:reservation-depth=5\n\
+ -M, --match-policy=MPOLICY    Match policy for each level of\n\
+                                   the hierarchy: e.g., low:high\n\
+ -J, --njobs=NJOBS             Total num of Jobscripts to run\n\
+                                   (default=${njobs})\n\
```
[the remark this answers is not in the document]
dongahn: @SteVwonder: Good point. I will change the global variable names to minimize a collision with environment variables. My first thought is to add `ft_` prefix. I don't see why they need to be capitalized though? Is it a common practice to use capital letters for shell globals?

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+        local o=''
+        [[ "${log}" = "yes" ]] && o="-o,-S,log-filename=${prefix}.${rank}.log"
+        local N=0
+        N=$(get_my_nodes "${ncores}" "${my_cores}")
+        local c=0
+        c=$((my_cores/N + (my_cores + my_cores % N)/(my_cores + 1)))
+        local g=0
+	g=$((my_gpus/N + (my_gpus + my_gpus % N)/(my_gpus + 1)))
+	local G=''
+	[[ ${g} -gt 0 ]] && G="-g ${g}"
```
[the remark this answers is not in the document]
dongahn: Ugg. The default vi config on my flux-sched docker add tabs when I add a line. Because I don't mount my home directory for this container, I often forget to customize the config everytime. I should develop a habit of not using vi on that container. Will fix.

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+        local X="--prefix=${prefix}.${rank}"
+
+        if [[ "x${dry_run}" = "xyes" ]]
+        then
+            dr_print "Rank=${rank}: N=${N} c=${c} ${G:+g=${G}} ${o:+o=${o}}"
+	    dr_print "Rank=${rank}: ${T:+T=${T}}"
```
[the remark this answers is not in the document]
dongahn: Will fix.

dongahn:
> FAIL: t2000-tree-basic.t 26 - flux-tree: -T4 on 4 nodes/4 cores/4 GPUs work
2109FAIL: t2000-tree-basic.t 27 - flux-tree: -T4 on 4 nodes/4 cores/2 GPUs work
2110FAIL: t2000-tree-basic.t 28 - flux-tree: -T4 on 4 nodes/4 cores/2 GPUs 1 job works
2111FAIL: t2000-tree-basic.t 31 - flux-tree: --leaf in real mode
2112FAIL: t2000-tree-basic.t 32 - flux-tree: -T1 in real mode

CI is failing this test case and I need to take a look at it as well.

src/cmd/flux-tree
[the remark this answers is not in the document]
SteVwonder: > I can imagine users user this option to repeat a jobscript (similar to jobarray).

Ah. That makes sense. Maybe the description in the usage message be can updated to something like: `Leaf instance.  Will directly submit jobs to enclosing instance scheduler.  Mutually-exclusive with internal tree-node options like --topology. (default=${leaf})`

src/cmd/flux-tree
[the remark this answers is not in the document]
SteVwonder: > Is it a common practice to use capital letters for shell globals?

I'm not 100% sure, so I originally formed the comment as a question.  All of my experience with shell scripting and env vars has lead me to believe that any input or output variables to a shell script should be capitalized.   More specifically, I believe this `FT_NJOBS=1024 flux tree -- sleep 1` is more the norm than `ft_njobs=1024 flux tree -- sleep 1`.  But I don't think there are any formal rules or conventions, other than [SO posts](https://stackoverflow.com/a/42290320) about it.

Either way though, we should probably add any decisions we make on these conventions to our RFC coding style guide.

SteVwonder:
One other random thought I had today was that we could always run this through a shell linter like [ShellCheck](https://www.shellcheck.net/) if you want some static analysis done on it.  There is already a [Github Action for it](https://github.com/marketplace/actions/shell-linter), so it would be trivial to add to the CI if we wanted.

grondo:
One warning: shellcheck is *very* pedantic :wink: 

I hate to suggest it, but was there any thought to writing a version of `flux-tree` in Python (possibly a future enhancement)? Some things would be much easier, like interacting with flux asynchronously, and overall the script might be more maintainable and extensible down the road. (I've found in past that you can only push shell scripts so far and still keep your sanity)

SteVwonder:
> One warning: shellcheck is very pedantic wink

Good to know.  Thanks!  Maybe we avoid adding it to the CI then.

> I hate to suggest it, but was there any thought to writing a version of flux-tree in Python (possibly a future enhancement)? Some things would be much easier, like interacting with flux asynchronously, and overall the script might be more maintainable and extensible down the road. (I've found in past that you can only push shell scripts so far and still keep your sanity)

I think that makes a lot of sense.  I agree that extending this shell script to do more advanced things like scheduling heterogeneous jobs or automatic load balancing of jobs (either through lazy pulling of jobs from the root or work stealing) would get hairy quickly.

Maybe we merge this PR with the shell scripting so that we have a "stake-in-the-ground" (so to speak) on the interface, and then open an issue to track replacing it with a Python script? @dongahn what are your thoughts?

dongahn:
> One warning: shellcheck is very pedantic 

I already pushed a version of flux-tree to that and changed a bunch. But there are a few that didn't make sense to change so I didn't take it. We probably don't want to require it to make a PR passing.

dongahn:
> I hate to suggest it, but was there any thought to writing a version of flux-tree in Python (possibly a future enhancement)?

I initially thought about Python, but there were certain things (like calling flux CLI) for which a shell script made better sense at that time. 

> Maybe we merge this PR with the shell scripting so that we have a "stake-in-the-ground" (so to speak) on the interface, and then open an issue to track replacing it with a Python script? @dongahn what are your thoughts?

Yes. I definitely don't want to spend more time on this at this point as I have bigger fish to fry in the next few months. This should address most of the initial use cases. We can decide whether we want to port the logic to Python as this capability is used more and we get more feedback.

dongahn:
> then open an issue to track replacing it with a Python script? @dongahn what are your thoughts?

Don't know if we want to open up an issue for this yet. I would say, let's get some more feedback from early users first.

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+            continue
+        fi
+
+        if ! eval "${jobscript_cl}"
+        then
+            warn "Error executing" "${jobscript_cl}"
+        fi
+    done
+
+    [[ "x${dry_run}" = "xno" ]] && flux job drain
```
SteVwonder:
```suggestion
    [[ "x${dry_run}" = "xno" ]] && flux queue drain
```

`flux job drain` was recently changed to `flux queue drain` in this PR: https://github.com/flux-framework/flux-core/pull/2687#issuecomment-580844346

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+            dr_print "Rank=${rank}: ${X:+X=${X}} ${J:+J=${J}} ${S:+S=${S}}"
+            dr_print ""
+            continue
+        fi
+        jobid=$(flux mini run --dry-run -N${N} -n${N} -c${c} ${G} flux start \
+${o} flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${X} ${J} ${S} \
+| flux job submit -)
+        jobids["${rank}"]="${jobid}"
+    done
+    [[ "x${dry_run}" = "xno" ]] && flux job drain
```
SteVwonder:
```suggestion
    [[ "x${dry_run}" = "xno" ]] && flux queue drain
```

`flux job drain` was recently changed to `flux queue drain` in this PR: https://github.com/flux-framework/flux-core/pull/2687#issuecomment-580844346

src/cmd/flux-tree
SteVwonder: @dongahn: I'm using the latest version of flux-tree (i.e., you already added  the `guest.tree-perf`), but I'm still getting the error you got above:

```
++ flux job info 16726884352 guest.tree-perf
flux-job: job 16726884352 id or key not found
```

Note: the `++` are from adding bash's xtrace to flux-tree for debugging.

Here's my full run line:
```
srun --mpi=none -N 2 flux start \
            flux tree \
            --nnodes=2 \
            --ncores-per-node=36 \
            --njobs=1 \
            --perf-out=${RESULTS_DIR}/perf.out \
            -- sleep 0
```

SteVwonder: Ah. I figured it out.  I was not turning off mpibind, and so I flux was only seeing 18 cores per node.  When `flux-tree` went to submit the sub-instance, it requested 36 cores-per-node (as I specified), and that request was ultimately unsatisfiable and never ran.  In the future, it would be useful to check for sub-instance failures and print a warning, rather than erroring out when collecting the performance results.

After turning off mpibind, it works as expected :)

src/cmd/flux-tree
SteVwonder:
```suggestion
    [[ "x${dry_run}" = "xno" ]] && flux queue drain
```

`flux job drain` was recently changed to `flux queue drain` in this PR: https://github.com/flux-framework/flux-core/pull/2687#issuecomment-580844346

There is another use further down too.

dongahn:
> Ah. I figured it out. I was not turning off mpibind, and so I flux was only seeing 18 cores per node. When flux-tree went to submit the sub-instance, it requested 36 cores-per-node (as I specified), and that request was ultimately unsatisfiable and never ran. In the future, it would be useful to check for sub-instance failures and print a warning, rather than erroring out when collecting the performance results.

I have https://github.com/flux-framework/flux-sched/issues/589 as a future item. But we need to think through. 

1. We may need to capture the outputs from the nested instances and bubble up to the top. But if we are not careful with it, flux-tree is not going to be scalable. 

2. While we can capture the error code from the script itself as well as flux-tree itself, this won't bubble up to their parent instance (flux start process). So I don't know what will be the end results.

Perhaps we can discuss this a bit at our 11AM meeting.

dongahn:
> After turning off mpibind, it works as expected :)

Glad to hear!

grondo:
> While we can capture the error code from the script itself as well as flux-tree itself, this won't bubble up to their parent instance (flux start process).

`flux start` (or the flux-broker spawned as a result thereof) will pass along the final exit status of its "initial program", so if the intial program of each instance (presumably this is a flux-tree at each level?) then the exit code *will* bubble all the way up to the parent `flux start`, e.g.

```console
$ srun -N2 flux start sh -c 'exit 2'
2020-02-20T17:23:36.965452Z broker.err[0]: Run level 2 Exited with non-zero status (rc=2) 0.0s
srun: error: fluke88: task 0: Exited with exit code 2
```

src/cmd/flux-tree
```diff
@@ -0,0 +1,807 @@
...
+            dr_print "Rank=${rank}: N=${N} c=${c} ${G:+g=${G}} ${o:+o=${o}}"
+	    dr_print "Rank=${rank}: ${T:+T=${T}}"
+            dr_print "Rank=${rank}: ${Q:+Q=${Q}} ${P:+P=${P}} ${M:+M=${M}}" 
+            dr_print "Rank=${rank}: ${X:+X=${X}} ${J:+J=${J}} ${S:+S=${S}}"
+            dr_print ""
+            continue
+        fi
+        jobid=$(flux mini run --dry-run -N${N} -n${N} -c${c} ${G} flux start \
+${o} flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${X} ${J} ${S} \
+| flux job submit -)
```
grondo: Just noticed you pipe `flux mini run --dry-run` output directly to `flux job submit`. Any reason to not just use `flux mini submit`?

Also, if you want to collect exit status from sub-jobs, consider adding `--flags waitable` to `flux mini submit` or `flux job submit`. Then below,  after `flux queue drain` returns, fetch the highest exit code from all jobs via something like

```shell
EXITCODE=0
for id in $(flux jobs -ano {id}); do
    flux job wait $id >/dev/null 2>&1
    rc=$?
    [[ $rc -gt $EXITCODE ]] && EXITCODE=$rc
done
```

If you don't want to submit jobs as waitable, then presently you can get the exit status by either `flux job attach JOBID >/dev/null 2>&1`, or by parsing the `finish` event  `flux job wait-event JOBID finish`, e.g.

```
$ flux job wait-event 291577394102272 finish | sed 's/.*status=//'
256
```

Eventually, the status will be available from the `job-info.list` RPC, so you may be able to fetch all job status like

```
$ flux jobs -ano {id},{status}
```
Seems like we may want to add a `flux job status` helper in the meantime. Should be trivial.

dongahn: > Just noticed you pipe flux mini run --dry-run output directly to flux job submit. Any reason to not just use flux mini submit?

Yap this was to replace the `flux jobspec srun` with `flux mini run --dry-run`. I will change this to flux mini submit.

dongahn: > Also, if you want to collect exit status from sub-jobs, consider adding --flags waitable to flux mini submit or flux job submit. Then below, after flux queue drain returns, fetch the highest exit code from all jobs via something like

Yes this is a good idea. 

I am thinking, `flux-tree` will return the highest exit code from its children. This means, at the leaf `flux-tree` node, it will detect the highest exit codes resulted from executing job scripts. At an internal (and root) `flux-tree` node, it will detect the highest exit code resulted from its subtree.

dongahn: As to printing out "reduced error output" at the top-level, I can bubble up the stdout and stderr  from the job with the highest exit code. I think this would capture a "representative" output without having to hamper scalability.

dongahn:
@grondo:

> Also, if you want to collect exit status from sub-jobs, consider adding --flags waitable to flux mini submit or flux job submit. 

I like this idea a lot. But at the leaf flux-tree instance, I have no control over what flags the user-provided script would use. My guess is 99% of them would use default flux mini submit without --flags options. So I can't reliably harvest the exit code from user-submitted apps.

It seems an environment variable or similar which allow flux-tree to add --flags writable to subsequence submit would be useful. 

Also, is there a reason not to make writable jobs be default? I can imagine some performance loss but would that make significant enough difference with the scalable aggregation of exit codes?

grondo:
> I like this idea a lot. But at the leaf flux-tree instance, I have no control over what flags the user-provided script would use. 

At the time I made that suggestion, I was under the (wrong) impression that flux-tree was responsible for submission of work to the leaf instance. Since flux-tree doesn't control what and how jobs are run at the leaf, performing any operation "over all jobs" is possible not practical.

Perhaps you can just document that flux-tree will report the highest exit status of all user-provided scripts. The user script would then be responsible for exiting with nonzero status on "failure", putting the burden of collecting individual job exit status on the user script.

Then, flux-tree can submit non-leaf instances with the `waitable` flag and collect their exit status as described.

dongahn:
That's the approach I'm currently taking. 

Initially I wanted to detect the maximum exit code between the user provided script and its submitted jobs. If the max code is from a submitted job, I wanted to do `flux job attach` on that job to bubble up its output.

grondo:
> Initially I wanted to detect the maximum exit code between the user provided script and its submitted jobs. If the max code is from a submitted job, I wanted to do flux job attach on that job to bubble up its output.

You could do this with the new `flux job status` which can fetch exit status from any job, waitable or not. It is not fast though since it is fetching job eventlogs, but something like this would "work":
(untested)

```sh
for id in $(flux jobs -ano {id} -c 0); do
    flux job status -v ${id}
    if [[ $? -lt 0 ]]; then
        flux job attach $id >/dev/null
    fi
done
```

`waitable` jobs won't be reliable here -- the user script may use `flux job wait` to synchronize on jobs and, like the `wait(2)` system call, a job can only have `flux job wait` called on it once.

dongahn:
@grondo: 

```
ahn1@51a18fd18818:/usr/src$ flux mini submit -N 1 -n1 --flags waitable sh -c "exit 4"
417081589760
ahn1@51a18fd18818:/usr/src$ flux job wait 417081589760
flux-job: task(s) exited with exit code 4
ahn1@51a18fd18818:/usr/src$ echo $?
1
```

It looks like `flux job wait <jobid>` doesn't propagate the return code to its own return code. Is this intended?

dongahn:
> flux job status -v ${id}

Is this fairly new? My flux-core version doesn't seem to have it.

grondo:
> It looks like flux job wait <jobid> doesn't propagate the return code to its own return code. Is this intended?

Yeah, I guess it is only success/failure. The response payload in the RPC is `{"success":boolean, "errstr":string}`

> > flux job status -v ${id}
>
> Is this fairly new? My flux-core version doesn't seem to have it.

Yes, I think the PR was merged this week. It made it into flux-core v0.16.

dongahn:
Thanks. I will just use 

`flux job wait ${job} 2>&1 | awk '{print $7}'`

for now.

src/cmd/flux-tree
```diff
@@ -132,7 +132,7 @@ environment variables.\n\
 Options:\n\
  -h, --help                    Display this message\n\
  -l, --leaf                    Leaf (default=${leaf})\n\
- -f, --flux-logs               Dump Flux logs for all instances\n\
+ -f, --flux-logs=DIR           Dump Flux logs for all instances int DIR\n\
```
SteVwonder:
```suggestion
 -f, --flux-logs=DIR           Dump Flux logs for all instances into DIR\n\
```

grondo:
> Thanks. I will just use
> `flux job wait ${job} 2>&1 | awk '{print $7}'`
> for now.

Careful, this will break if the job got an exception, in which case `flux job wait` prints the exception message.

Here's some example output:
```
grondo@asp:~/git/f.git$ flux job wait -v --all
2498177794048: Fatal exception type=exec task 0: start failed: nosuchthing: No such file or directory
1917870669824: task(s) exited with exit code 1
2078478958592: task(s) exited with exit code 42
1824924893184: job completed successfully
2243466100736: task(s) Killed
```

dongahn:
> Careful, this will break if the job got an exception, in which case flux job wait prints the exception message.

Ugg. How about the new command:

> flux job status -v ${id}

What would this report if the job got an exception?

grondo:
> `flux job status -v ${id}`
> What would this report if the job got an exception?

This command ignores job exceptions and exits with the same exit code as would be reported by `flux job attach ${id}`. This exit code is reported in the job `finish` event `status=` context.

Without the `-v` it will just exit with same exit code as the job. With `-v` it will report an informational message to stderr.

(BTW, this command can also take multiple JOBIDs on the cmdline. It will exit with the highest exit code of all jobs)

dongahn:
@grondo: Thanks. I updated my flux-core docker images and decided to use `flux job status`. The exit status rollup support is now at 89dcc339af6f2ed0a0302667e466f3b65399c797. I added one test for this one as well.

@SteVwonder: I added a fix for the typo you found and also added some changes to pass Travis CI. Could you take a look at the last few commits (the ones that came after your review.)

If all is good, I will squash those intermediate commits to land this!

SteVwonder:
Thanks @dongahn for all the changes.   It looks like the arg parsing is chewing into the arguments passed to `jobscript`.

```
(1) herbein1@opal186 ~/.../sc2020/libexec/flux/cmd
❯             flux tree \                                                                                                                                                                                         22:52:11 (sc2020-spack-env)
             --leaf \
            --nnodes=1 \
            --ncores-per-node=36 \
            --njobs=1 \
            --perf-out=/tmp/perf.out \
            --flux-logs \
--     flux mini submit  sleep 0
1849184747520

herbein1@opal186 ~/.../sc2020/libexec/flux/cmd
❯             flux tree \                                                                                                                                                                                         22:52:44 (sc2020-spack-env)
             --leaf \
            --nnodes=1 \
            --ncores-per-node=36 \
            --njobs=1 \
            --perf-out=/tmp/perf.out \
            --flux-logs \
--     flux mini submit  -n1 sleep 0
flux-tree: invalid option -- 'n'
flux-tree: invalid option -- '1'

(1) herbein1@opal186 ~/.../sc2020/libexec/flux/cmd
❯             flux tree \                                                                                                                                                                                         22:52:47 (sc2020-spack-env)
             --leaf \
            --nnodes=1 \
            --ncores-per-node=36 \
            --njobs=1 \
            --perf-out=/tmp/perf.out \
            --flux-logs \
--  -n1 sleep 0
flux-tree: invalid option -- 'n'
flux-tree: invalid option -- '1'
```

Here is a more minimal reproducer, since it looks like the issue is with `getopt`: 

```
++ /usr/bin/getopt -u -o hlf:N:c:g:T:Q:P:M:J:o:X:d -l help,leaf,flux-logs:,nnodes:,ncores-per-node:,ngpus-per-node:,topology:,queue-policy:,queue-params:,match-policy:,njobs:,perf-out:,prefix:,dry-run -n flux-tree -- --leaf --nnodes=1 --ncores-per-node=36 --njobs=1 --perf-out=/tmp/perf.out --flux-logs -- -n1 sleep 0
flux-tree: invalid option -- 'n'
flux-tree: invalid option -- '1'
```

I'll take another crack at debugging this tomorrow night unless someone else beats me to it.

dongahn:
>  It looks like the arg parsing is chewing into the arguments passed to jobscript.

Great finding! I will take a look at it tomorrow.

dongahn:
Actually, I think this is because of the latest update to the `--flux-logs` option. It now requires an argument: `--flux-log=DIR`. 

```
ahn1@7eb007aae729:/usr/src$ /usr/bin/getopt -u -o hlf:N:c:g:T:Q:P:M:J:o:X:d -l help,leaf,flux-logs:,nnodes:,ncores-per-node:,ngpus-per-node:,topology:,queue-policy:,queue-params:,match-policy:,njobs:,perf-out:,prefix:,dry-run -n flux-tree -- --leaf --nnodes=1 --ncores-per-node=36 --njobs=1 --perf-out=/tmp/perf.out --flux-logs=./logs -- -n1 sleep 0
 --leaf --nnodes 1 --ncores-per-node 36 --njobs 1 --perf-out /tmp/perf.out --flux-logs ./logs -- -n1 sleep 0
```

So either drop `--flux-logs` or use `--flux-logs=./` if you want the log files dumped into the `cwd`.

dongahn:
Rebased and forced a push.

dongahn:
Per #602, I removed `2> /dev/null` from `flux job attach` so that the stderr message is printed as well. Also, I removed `--flags=waitable` from `flux mini submit`. This isn't needed any longer because I am now using `flux job status` instead of `flux job wait`.

dongahn:
Meant to say, I forced a push.

dongahn:
Because https://github.com/flux-framework/flux-core/pull/2784 went it, I will make one line change to make use of the exit code setting option, rebase and force a push.

SteVwonder:
@dongahn: Good news; it looks like having `flux kvs put` read from stdin is working!  I'll post it as an inline suggestion.  Bad news; it now breaks at the instance that tries to combine the various child JSONs into one big one: `/usr/WS2/herbein1/packages/toss3/toss3/spack/opt/spack/linux-rhel7-broadwell/gcc-4.9.3/flux-core-master-bllpjq7777lz4jaqj7mdnlmv3co753rd/libexec/flux/cmd/flux-tree: line 717: /usr/bin/jq: Argument list too long`

The offending line (with context):
```
        child=$(flux job info ${jobids[${rank}]} guest.tree-perf)
        blurb=$(echo "${blurb}" | jq " .child += [${child}]")
```

I'm not sure how to easily workaround this one.

---

When I was watching it run interactively, I noticed that was taking quite a while.  I `tail`'d all of the `tree.X.log` files, and they had `2020-03-03T05:20:39.892250Z broker.info[0]: Run level 3 Exited (rc=0) 1.0s` at the bottom; so I assume they all finished and it was stuck in the top instance processing something (unless there was a straggler I missed).

Output from the time wrapping the whole `flux start ... flux tree` call:
```
212.08user 74.79system 3:39.69elapsed 130%CPU (0avgtext+0avgdata 41136maxresident)k
162104inputs+6424outputs (117major+9801980minor)pagefaults 0swaps
```

Head of perf.out:

```
TreeID                   Elapse(sec)           Begin(GTD)             End(GTD)     Match(usec)           NJobs NNodes  CPN  GPN
"tree"                        49.527   1583212804.2368872   1583212853.7638848               0            1152    32   36    0
"tree.1"                     27.6259   1583212807.1609085    1583212834.786855               0              36     1   36    0
"tree.1.1"                   1.04054   1583212810.9444387    1583212811.984977               0               1     1    1    0
```

Given the large discrepency between the total runtime and the runtime reported in the perf.out, I wonder if the major cost was in the performance data collection at the end.  I don't have enough data to fully support that hypothesis though.  It could also be data slipping into stdout causing lots of KVS activity or a whole host of other things.

---

Also, I accidentally ran our 48 core per node test from earlier today on Opal and the unsatisfiability error works well :)
```
Checking status of 68753031168
flux-job: 68753031168: exception type=alloc
0.022s: job.exception type=alloc severity=0 alloc denied due to type="unsatisfiable"
```

src/cmd/flux-tree
```diff
@@ -0,0 +1,865 @@
...
+        rm -f "${out}"
+        printf "${perf_format}\n" ${perf_heading} > "${out}"
+        dump_perf "${blurb}" "${out}"
+    else
+        if [[ "${prefix}" != "${top_prefix}" ]]
+        then
+            if [[ "x${FT_DRY_RUN}" = "xno" ]]
+            then
+                # Put it to KVS parent namespace except when we are at top level
+                flux --parent kvs put -j "tree-perf=${blurb}" || warn "KVS error"
```
SteVwonder:
```suggestion
                flux --parent kvs put --raw "tree-perf=-" <<< "${blurb}" || warn "KVS error"
```

dongahn:
Maybe this is the area where our performance modeling can play a role to make this scalable. Instead of aggregating all individual performance data points, perhaps we can collect the performance data *only on the critical path*. This way, the amount of data we collect will always be scalable?

SteVwonder:
> Maybe this is the area where our performance modeling can play a role to make this scalable. Instead of aggregating all individual performance data points, perhaps we can collect the performance data only on the critical path. This way, the amount of data we collect will always be scalable?

Yeah, that's a good idea.  Maybe collect the ID and times of just the slowest child and propagate only that up to the parent.  We could even use that data to verify our hypothesis that the critical path should be the last child launched at each level.  Could be the right stopgap solution so that we can get this PR merged and in.

Ultimately though, I don't think the size of the data is the main issue.  After running the `32x36` topology example, the total `perf.out` size is only 148K on disk.  That seems like a small amount of data to break on.  I think the limiting factor here is the use of CLI arguments to pass data into programs.

(Below is just brainstorming, I don't think it should necessarily block this PR; I'm just trying to think through how to generally solve this problem going forward).

Definitely re-writing the whole thing in Python would avoid this issue (but that's probably at least a week's worth of work, maybe two).  One idea for an incremental move towards Python would be to still use bash and the current flux-tree to spin up the hierarchy and launch the jobs, but before exiting, have flux-tree launch a python script (flux-tree-helper.py?) to do the data collection, reduction, propagation, formatting, and saving out to disk.  That python script could then gradually evolve to do more and more of the heavy lifting, and ultimately replace flux-tree entirely.

dongahn:
>  I wonder if the major cost was in the performance data collection at the end.

It could be `jq` performance on the critical path. 

> have flux-tree launch a python script (flux-tree-helper.py?) to do the data collection, reduction, propagation, formatting, and saving out to disk. That python script could then gradually evolve to do more and more of the heavy lifting, and ultimately replace flux-tree entirely.

Good idea. If this is on the performance critical path, we can write this in C or C++ as well. 

Maybe there is a chance to make this a general command, a command that pushes key-values to the parent's guest namespace and gathers the key-values of all children into a `.child` key as an array. All has to be done via `STDIN` as opposed to command line of course. This can come in handy to support developing a hierarchical algorithm like this.

dongahn:
Thank you!

PR #584, opened 2020-01-30
New metadata for elastic jobs

milroy:
This PR adds necessary metadata to enable elastic jobs.  It defines an attribute `jobtype` that is attached to each resource vertex and mapped by `jobid`.  Furthermore, since a vertex can have multiple jobs allocated, job shrink will need a count of the number of elastic jobs scheduled on the vertex.

The PR also includes logic to read the jobtype attribute from a jobspec and attach it to `jobmeta`.

### DIFF
--- a/resource/traversers/dfu_impl_update.cpp
+++ b/resource/traversers/dfu_impl_update.cpp
@@ -148,17 +148,11 @@ int dfu_impl_t::upd_plan (vtx_t u, const subsystem_t &s, unsigned int needs,
             }
             return -1;
         }
-        if (jobmeta.allocate) {
-            (*m_graph)[u].schedule.allocations.spantype[jobmeta.jobid].span = span;
-            (*m_graph)[u].schedule.allocations.spantype[jobmeta.jobid].jobtype = jobmeta.jobtype;
+        if (jobmeta.allocate) 
+            (*m_graph)[u].schedule.allocations.insert (jobmeta.jobid, span, jobmeta.jobtype);
 
-            if (jobmeta.jobtype == "elastic")
-                ++(*m_graph)[u].schedule.allocations.elasticcount;
-       }
-        else {
-            (*m_graph)[u].schedule.reservations.spantype[jobmeta.jobid].span = span;
-            (*m_graph)[u].schedule.reservations.spantype[jobmeta.jobid].jobtype = jobmeta.jobtype;
-        }
+        else
+            (*m_graph)[u].schedule.reservations.insert (jobmeta.jobid, span, jobmeta.jobtype);
     }
     return 0;
 }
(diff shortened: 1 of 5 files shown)


resource/traversers/dfu_impl_update.cpp
```diff
@@ -148,10 +148,17 @@ int dfu_impl_t::upd_plan (vtx_t u, const subsystem_t &s, unsigned int needs,
...
-        if (jobmeta.allocate)
-            (*m_graph)[u].schedule.allocations[jobmeta.jobid] = span;
-        else
-            (*m_graph)[u].schedule.reservations[jobmeta.jobid] = span;
+        if (jobmeta.allocate) {
+            (*m_graph)[u].schedule.allocations.spantype[jobmeta.jobid].span = span;
+            (*m_graph)[u].schedule.allocations.spantype[jobmeta.jobid].jobtype = jobmeta.jobtype;
+
+            if (jobmeta.jobtype == "elastic")
+                ++(*m_graph)[u].schedule.allocations.elasticcount;
```
SteVwonder: It would be nice if we didn't have to have special increment/decrement code for every hardcoded jobtype.   If this could be a generic map that tracks all jobtypes, I think that will be more general and helpful for future use-cases.

milroy: I was considering that but was worried about the memory overhead added to each vertex.  After more consideration I'm thinking the potential cost is outweighed by the benefits of maintainability and generality.

milroy: Ok, I updated `allotment_t` in my test branch.  I think it's easier to extend, and cleans up the code:

```
struct allotment_t {
    allotment_t ();
    void insert (const int64_t jobid, const int64_t span,
                        const std::string &jobtype);
    void erase (const int64_t jobid);
    ~allotment_t ();

    std::map<int64_t, spantype_t> id2spantype;
    std::map<std::string, std::set<int64_t>> type2idspan;
};
```

So inserting a new allocation becomes one line:
```
g[v].schedule.allocations.insert (jobid, span, jobtype);
```
@dongahn do you think that's excessive metadata per vertex?

milroy:
After making progress on an upcoming, related PR I think that the metadata proposed here may be better stored in the planner. @dongahn and @SteVwonder let's put this PR on pause until we have time to discuss the architectural changes I'm considering.

milroy:
I've updated the PR with the new metadata structure in anticipation of upcoming, more substantial PRs.  I'll need to squash and amend commits before this gets merged.

milroy:
After discussions with @dongahn and @SteVwonder I'm going to close this PR.  The latest thinking is we can get by with less new metadata (too little to warrant a PR) since we won't be adding another planner.

The new metadata will be included in an upcoming PR.

dongahn:
Sounds good to me. Thank you for going for the most elegant solution here!

PR #595, opened 2020-02-14
README: clarify install instruction to better align with flux-core

### LINKED ISSUE #594: Make README.md install instructions consistent with flux-core
As noted by @fmuelle4711

> 
> One more comment about flux-core README.md:
> 
> PYTHON_VERSION=3.6 ./configure
> 
> is inconsistent w/ flux-sched README.md:
> 
> PYTHON_VERSION=3.6 ./configure --prefix=$HOME/local
> 
> I'd suggest to use one or the other throughout the READMEs, mixing them causes problems during building/testing.

We should probably just drop the `--prefix` from flux-sched.

SteVwonder:
Reference the flux-core install instructions, rather than copying them.

Provide flux-sched install instructions for both the cases where flux-core is "system installed" and "side installed".

Add a note about the convenience of having flux-core and flux-sched installed to the same prefix.

Closes #594

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,6 @@ python3-yaml >= 3.10
 apt install libhwloc-dev libboost-dev libboost-system-dev libboost-filesystem-dev libboost-thread-dev libboost-graph-dev libboost-regex-dev libxml2-dev libyaml-cpp-dev python3-yaml
 ```
 
-
-The sched module contains
-a bifurcated structure of a core framework that has all the basic
-functionality plus loadable modules that implement specific
-scheduling behaviors.
-
 If you did not provide an install prefix when building flux-core or are building
 against a system-installed flux-core, then you can build flux-sched with:
 
--- a/README.md
+++ b/README.md
@@ -39,24 +39,10 @@ challenging scheduling objectives.
 
 #### Building flux-sched
 
-flux-sched can only be built against an installed flux-core build.  If
-the flux-core is side installed, the PKG_CONFIG_PATH environment
-variable must include the location to flux-core's pkgconfig directory.
+flux-sched can only be built against an installed flux-core build.  Instructions
+for installing flux-core can be found in [the flux-core
+README](https://github.com/flux-framework/flux-core/blob/master/README.md).
 
-The example below is of a side install of flux-core.  In this example,
-the flux-core and flux-sched repos are under the $HOME directory.  The
-flux-core repo is cloned, configured with a prefix that identifies the
-install directory ($HOME/local), built and installed into that
-directory.
-
-```
-git clone <flux-core repo of your choice>
-cd ~/flux-core
-./autogen.sh
-./configure --prefix=$HOME/local
-make
-make install
-```
 
 The next step is to build the sched module.
 flux-sched requires the following packages to build:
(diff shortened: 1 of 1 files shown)


dongahn:
LGTM. Will merge when Travis turns green.

grondo:
I wonder if we want to add a note that `flux-sched` should be installed into the same `--prefix` as `flux-core` though...

garlick:
That breaks the instructions since the point is to side install flux-core so flux-sched can find it right?

dongahn:
I think this should work as both flux-core and flux-sched will be installed into the default location. But I like the idea of @grondo to leave a note about this requirement.

garlick:
This part should be changed then:
```
export PKG_CONFIG_PATH=$HOME/local/lib/pkgconfig:$PKG_CONFIG_PATH
git clone <flux-sched repo of your choice>
cd ~/flux-sched
./autogen.sh
./configure --prefix=$HOME/local
make
make check
make install
```

SteVwonder:
All good points.  Sorry for the hasty PR.  I'll push a copy up where both prefixes match and add a note that they should match if the user wants `qmanager` to be auto-loaded when flux is started.  Maybe mentioning the `FLUX_MODULE_PATH` if the prefixes are different.

grondo:
Ah, I hadn't actually looked at the full README, my suggestion would be to reference the `flux-core` installation instructions, not copy them here. 

Then say something to the effect of "If flux-core is installed in a non-standard PREFIX, then run

```
$ PKG_CONFIG_PATH=${PREFIX}/lib/pkgconfig ./configure --prefix=$PREFIX
```

This ensures that flux-sched is installed to the same prefix as flux-core so that flux-sched components may be found automatically by flux-core components.

grondo:
I've often wondered if we should set `PKG_CONFIG_PATH` in the `flux(1)` command driver such that this would work

```
$ /my/flux/path/bin/flux ./configure --prefix=/my/flux/path
```

SteVwonder:
Ok. Take two.

Referenced the flux-core install instructions as @grondo suggested.  Also provided flux-sched install instructions for both the cases where flux-core was "system installed" and "side installed".

Added a note about the convenience of having flux-core and flux-sched installed to the same prefix. I started going down the path of extending the example for the case where `flux-core` and `flux-sched` are installed into different directories, but that ended up being way more than seems reasonable for a "quick start" guide in the README.  Probably something we can cover in more depth in the docs repo.

README.md
```diff
@@ -39,24 +39,10 @@ challenging scheduling objectives.
...
-```
-git clone <flux-core repo of your choice>
-cd ~/flux-core
-./autogen.sh
-./configure --prefix=$HOME/local
-make
-make install
-```
 
 The next step is to build the sched module.
```
grondo: Suggestion: instead of "the next step", say "Once the target version of flux-core has been built and installed, the flux-sched package may be configured with the following steps" (or something like that)

README.md
```diff
@@ -87,14 +73,32 @@ a bifurcated structure of a core framework that has all the basic
 functionality plus loadable modules that implement specific
```
grondo: Sorry, couldn't comment on the exact line since it wasn't modified, but since you're in here, you might want to remove all references to "The sched module" and replace with `flux-sched` throughout.

SteVwonder:
Thanks @grondo!  Let me know if the latest commits address your concerns.  If so, I'll squash the one.

grondo:
Yeah, all my concerns have been addressed. Thanks!

dongahn:
> I'd suggest we go through this one more time in the near future. The only thing I might fix now is to replace the suggested use of flux job list with flux jobs. But I think that can wait.

I will need to get to README when I change the name to fluxion and this can be part of this. Merging.

garlick:
Nice improvement!

SteVwonder:
Thanks all!

Minor thing: looks like this was merged before I had a chance to squash the `fixup!` commit.  Do we want to try and handle that in `master` now?  Or just let it go?

dongahn:
Ugg. Sorry about that. I would say just let it go. (I thought we set things up so that we can't merge when we have certain tags though...)

grondo:
A project owner can merge, even if no tests are passing. We have it set up so mergify can't merge if certain tests aren't passing. Maybe a good habit is to approve and set merge-when-passing and avoid merging manually, except when necessary.

dongahn:
> Maybe a good habit is to approve and set merge-when-passing and avoid merging manually, except when necessary.

Yap. I will do this next time on as much as possible.

Issue #591, opened 2020-02-12
planner_t to planner_multi_t refactor


milroy:
Based on feedback on issue #590, I'm refactoring flux-sched to use `planner_multi_t` instead of `planner_t`.  The purpose of this change is to enable clean handling of a planner per job type per resource vertex.  This introduces a difficulty related to an extension/abuse of `planner_multi_t`: 

```
planner_multi_t *planner_multi_new (int64_t base_time, uint64_t duration,
                                    const uint64_t *resource_totals,
                                    const char **resource_types, size_t len);
```

In particular, `resource_types` is a "string array of size `len` where each element contains the resource type corresponding to each corresponding element in the resource_totals array."  Considering `resource_types` to be job types, e.g., `{"rigid", "elastic"}` will probably work fine, but the planners will lose information about the actual resource types.

@dongahn or @SteVwonder do you recommend that I create another entry in the `planner_multi` struct for job types (and related calls), or is there a better approach?

dongahn:
@milroy: does `planner_multi_t` need to know whether a planner at index I is rigid or elastic? Can `resource_types` still be like `{"node", "node"}` and elasticity semantics be implemented by the user of `planner_multi_t`? IOW, is there any logic you need to implement within `planner_multi_t` that depends on whether a planner in it has the resource state from the rigid jobs vs. that from the elastic jobs?

milroy:
@dongahn: the answer to both of your questions is yes, but upon further consideration/experimentation I think the better approach is to create a new type.  While `planner_multi_t` can be extended to perform the needed functions, it carries a lot of unneeded overhead in the form of data structures and more dimensions (e.g. `request iter`, `char **resource_types`, `uint64_t *resource_totals`).

dongahn:
This probably is old at this point. We can reopen if needed of course.

Issue #590, opened 2020-02-11
planner: store jobid and jobtype in scheduled_point_t


milroy:
This question and potential enhancement is related to (and may supersede) PR #584, as well as issues #576, #558, and #552.

Scheduling of adaptive and elastic jobs extends the intention of: https://github.com/flux-framework/flux-sched/blob/05243593bd908f96599700eab5000e44ac4655d0/resource/planner/planner.c#L823  Since different job types have different relative priorities, the number of available resources at each scheduled point depends on the job type being matched.  Furthermore, the score of each vertex should depend on the number of elastic, adaptive, and rigid jobs allocated or reserved at the point in time in question.

So far I've been refactoring flux-sched to accommodate a `planner_t` for each job type.  This approach has undesirable effects: the additional overhead of multiple planners, the code bloat of added logic related to checking which job types are running at various times (and their relative priorities), and the inelegance of querying the `g[v].schedule.allocations` or `g[v].schedule.reservations` to get `jobid`s.  Note that my testing indicates that the exclusive planner will also need a separate instance per job type.

Beyond that, if we allow adaptive jobs to hold reservations: https://github.com/flux-framework/flux-sched/blob/af4447ffd27400cc33a908c62201380a98075571/resource/traversers/dfu_impl_update.cpp#L122 will need to check which and how many job types are allocated/reserved during `jobmeta.at`.  Such functionality is ~~not possible~~ clumsy with the current planner design.

Should we extend planner and `scheduled_point_t` to store `jobid`s and `jobtype`s?  An inspection of `planner.c` suggests that this could be an involved undertaking.

milroy:
@dongahn and I discussed this issue today.  We decided that modifying planner to include the specified metadata will add significant, undesirable complexity.  @dongahn suggested that I subsume this information and the logic checks for availability into `planner_multi_t`.

dongahn:
Also probably old.

Issue #576, opened 2020-01-25
Skip planner update for elastic job type


milroy:
Resources allocated to elastic jobs should be considered available here: https://github.com/flux-framework/flux-sched/blob/c03c083a7b4afb34fbfe2ed76e69309ed42732c8/resource/traversers/dfu_impl.cpp#L554
which means `planner_avail_resources_during` should return a positive integer.  I think the simple solution is to add a check for vertex job type before this line: https://github.com/flux-framework/flux-sched/blob/c03c083a7b4afb34fbfe2ed76e69309ed42732c8/resource/traversers/dfu_impl_update.cpp#L141
If the job type is elastic, the `planner_add_span` block should be skipped.

This is a sub-issue of #554.

milroy:
@dongahn could skipping over `planner_add_span` cause unexpected behavior?

dongahn:
Seems possible. But it is difficult to say whether this will work with no side effect without seeing some sort of a prototype. 

> @dongahn could skipping over planner_add_span cause unexpected behavior?

This is like not adding a schedule to the visiting resource. So when we visit this resource to schedule it for the next job, the resource state will be the same as there is no job scheduled. 

What happens to the other states? I have no idea without actually making some changes and running things. I certainly didn't design this subsystem with this use case in mind. But at the same time, this may work (potentially with some other changes).

In general, the main question would be to understand what fields of visiting resource (as well as its ancestor vertices) still needs to be updated and what fields must not be updated to add an elastic job. We are adding a whole new 3rd class job (no job scheduled, rigid job scheduled and elastic job scheduled). So, we need to be really clear about what a resource vertex and other metadata allow us to add this new class of job. Or even possible with the current infrastructure? Because this is pretty complex, we need some design exploration here.

I would prototype what you suggested and actually see the behaviors (to start with, what happens to all the test cases we have etc. then, new test cases...) and iterate.

milroy:
>This is like not adding a schedule to the visiting resource. So when we visit this resource to schedule it for the next job, the resource state will be the same as there is no job scheduled.

That is my intention, and I'm hoping it won't break anything else and won't make future design more difficult.

> What happens to the other states? I have no idea without actually making some changes and running things. I certainly didn't design this subsystem with this use case in mind. But at the same time, this may work (potentially with some other changes).

I wanted to get your input to make sure that this idea wouldn't create obvious issues before jumping in.

[...]

Issue #551, opened 2019-12-22
Need job type in resource_graph_metadata_t


milroy:
To enable elastic job shrink and expand, `resource_graph_metadata_t` needs a data structure that permits quick lookups of job type keyed by `jobid`.  This will allow resources that are currently running a job of type "elastic" to be allocated by another job.

dongahn:
Sounds good to me.

dongahn:
This probably didn't work out. Please reopen if I'm mistaken.

PR #600, opened 2020-02-22
Add support for elastic jobs

milroy:
This PR lays the foundation for elastic job support.  It consists of three contributions:

1. A small amount of additional metadata to identify a job's type ("rigid" by default, optionally "elastic").
2. Changes to traverser to allow for the bypassing of elastic job insertion into planner.  It also handles rigid and elastic job removal.
3. Nine tests for correct behavior of elastic jobs, including to make sure rigid jobs take precedence, and errors on unrecognized job types.

The changes here lead to deprioritization of resource vertices with allocated elastic jobs in the upcoming elastic matcher PR.  The complexities of this behavior will be discussed in the corresponding PR.  

This PR addresses issues #551, #576, #590, and #591.


milroy:
The Ubuntu: clang-6.0 chain-lint Travis test is failing with a weird error:
```
t4011-elastic-jobs.t: PASS: N=3   PASS=3   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
ERROR: t4011-elastic-jobs.t - missing test plan
ERROR: t4011-elastic-jobs.t - exited with status 1
```
I'll investigate, but this appears to be a testsuite configuration error.

grondo:
Try running `./t4011-elastic-jobs -d -v --chain-lint`

From the test-suite.log:
```
error: bug in the test script: broken &&-chain: 
    cat >cmds004 <<-EOF &&
	match allocate ${unit_job}
	match allocate ${unit_job}
	match allocate ${unit_job}
	match allocate ${unit_job}
	quit
EOF
```

milroy:
> Try running `./t4011-elastic-jobs -d -v --chain-lint` [...]

Thanks @grondo!  I missed a `&&` as the message indicates.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -156,9 +156,16 @@ int dfu_impl_t::upd_plan (vtx_t u, const subsystem_t &s, unsigned int needs,
...
         if (jobmeta.allocate) {
             (*m_graph)[u].schedule.allocations[jobmeta.jobid] = span;
-            // update the allocated elastic job bool
-            (*m_graph)[u].schedule.elastic_job = (jobmeta.job_type == 
-                                                 "elastic") ? true : false;
+            // update the allocated elastic job metadata
+            if (jobmeta.job_type == "elastic") {
+                (*m_graph)[u].schedule.elastic_job = true;
+                (*m_graph)[u].schedule.elastic_at = jobmeta.at;
+                (*m_graph)[u].schedule.elastic_at = jobmeta.elastic_duration;
```
SteVwonder:
```suggestion
                (*m_graph)[u].schedule.elastic_duration = jobmeta.elastic_duration;
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -156,9 +156,16 @@ int dfu_impl_t::upd_plan (vtx_t u, const subsystem_t &s, unsigned int needs,
...
+            } else {
+                (*m_graph)[u].schedule.elastic_job = false;
+                (*m_graph)[u].schedule.elastic_at = -1;
+                (*m_graph)[u].schedule.elastic_at = 0;
```
SteVwonder:
```suggestion
                (*m_graph)[u].schedule.elastic_duration = 0;                
```

milroy:
Thanks @dongahn and @SteVwonder for your initial offline feedback.  To note here: we've determined that the Codecov relative scope coverage result is due to error checks changing line numbers, which is not indicative of a problem.

Once Travis is green this PR is ready for your review.

SteVwonder:
Thanks @milroy!  It looks like the github action to validate commits is failing because there is not a newline between the commit subject and the commit body.  Other than that, those commit messages look valid.

resource/traversers/dfu_impl.cpp
```diff
@@ -558,6 +558,16 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta, vtx_t u,
         m_err_msg += strerror (errno);
         m_err_msg += ".\n";
         goto done;
+    } else if ((meta.job_type == "elastic") && (*m_graph)[u].schedule.elastic_job) {
```
SteVwonder: Super minor nit: not sure about the proper style here but elsewhere it seems like there is a space between the paren.s

```suggestion
    } else if ( (meta.job_type == "elastic") && (*m_graph)[u].schedule.elastic_job) {
```

milroy:
> Thanks @milroy! It looks like the github action to validate commits is failing because there is not a newline between the commit subject and the commit body. Other than that, those commit messages look valid.

Yes, I'd forgotten to leave a blank line between the commit subject and body.  The commit validator is happy now.

resource/schema/sched_data.hpp
```diff
@@ -40,6 +40,10 @@ struct schedule_t {
     std::map<int64_t, int64_t> allocations;
     std::map<int64_t, int64_t> reservations;
     planner_t *plans = NULL;
+    bool elastic_job = false;
+    bool elastic_job_running_at = false;
```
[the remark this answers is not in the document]
milroy: Yes, I'm checking the `elastic_job_running_at` in the upcoming elastic matcher PR.  The other way to accomplish the same thing is to add a parameter to `dom_finish_vtx`, but that requires changing the interface.

resource/traversers/dfu_impl.cpp
milroy: Good catch, I'll make that change.

milroy:
> Generally LGTM! Two comments below. I'm also going to be honest and say I got a little lost on the tests that used the `update allocate` `resource-query` command, mainly due to my own unfamiliarity with that functionality. So maybe we can chat briefly offline about that over coffee tomorrow?

I'm happy to walk you though the `update_allocate` changes tomorrow afternoon.

resource/schema/sched_data.hpp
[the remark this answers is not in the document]
SteVwonder: Ok.  Makes sense.  Do you mind adding a comment above these bools to highlight the difference between them?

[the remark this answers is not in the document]
milroy: That's a good idea- I'll be sure to add a comment before this gets merged.

dongahn:
@milroy: good work! A couple of suggestions to make our reviews a bit easier. 

- The third commit is pretty big and probably can be split. Please consider splitting it to
    1. Test outputs (i.e., expected R files)
    2. Test input job spec files
    3. The test itself

- Try to make your commit message imperative. For example, instead of
"This commit enables elastic jobs by adding metadata to keep track of exclusive job types", 
"Enable enable elastic jobs by adding metadata to keep track of exclusive job types. This can make our logs appear more consistent.

resource/schema/sched_data.cpp
```diff
@@ -50,7 +50,6 @@ schedule_t &schedule_t::operator= (const schedule_t &o)
 {
     int64_t base_time = 0;
     uint64_t duration = 0;
-    size_t len = 0;
```
dongahn: Thank you for detecting this deadwood. This can go to its own commit.

resource/schema/sched_data.hpp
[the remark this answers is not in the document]
dongahn: Can you use `elastic_job` instead of `elastic_job_running_at` within your upcoming match policy PR, though. My guess is you are using  `elastic_job_running_at` to determine whether `elastic_at` and `elastic_duration` are valid. But I can't see a reason why `elastic_job` cannot be used for this check...

resource/traversers/dfu_impl.hpp
```diff
@@ -57,6 +57,7 @@ struct jobmeta_t {
     int64_t jobid = -1;
     int64_t at = -1;
     uint64_t duration = SYSTEM_DEFAULT_DURATION; // will need config ultimately
+    std::string job_type = "rigid";
```
dongahn: This will soon be updated with configuration file support for resource. I have a slight preference to add this "default" resource type into "resource/config/system_defaults.hpp".

resource/traversers/dfu_impl.hpp
```diff
@@ -68,6 +69,11 @@ struct jobmeta_t {
             duration = SYSTEM_MAX_DURATION; // need config support ultimately
         else
             duration = (int64_t)jobspec.attributes.system.duration;
+
+        std::unordered_map<std::string, YAML::Node>::const_iterator jtype =
```
dongahn: I forgot. Has our canonical job spec RFC updated for this?

dongahn:
It seems the dfu code changes in c0e600136741a2a0b36a3ef405ae141b468e89a1 would better fit the second commit (2d129a74e069e6098a8197c442370243998f554e)?

resource/traversers/dfu.cpp
```diff
@@ -78,22 +78,27 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
-        t = meta.at + 1;
-        p = (*get_graph ())[root].idata.subplans.at (dom);
-        len = planner_multi_resources_len (p);
-        duration = meta.duration;
-        detail::dfu_impl_t::count_relevant_types (p, dfv, agg);
-        for (t = planner_multi_avail_time_first (p, t, duration, &agg[0], len);
-             (t != -1 && rc && !errno); t = planner_multi_avail_time_next (p)) {
-            meta.at = t;
-            rc = detail::dfu_impl_t::select (jobspec, root, meta, x);
+        if (meta.job_type == "rigid") {
```
dongahn: Now that I think about this, this will have impact when we choose backfill scheduling policies like EASY or CONSERVATIVE. What's going to happen if a user submits a bunch of elastic jobs, and our backfill scheduler tried to reserve them to calculate backfill?

dongahn:
It is too late and I am not thinking clearly. I will look at the rest of the PR tomorrow.

resource/traversers/dfu.cpp
milroy: My intention with `meta.job_type == "rigid"` is to protect against reserving elastic jobs by triggering an `EINVAL` when the check fails.  Will this not work as intended?

resource/traversers/dfu_impl.hpp
milroy: I can certainly do that instead.

resource/traversers/dfu_impl.hpp
milroy: No, but since I've placed `job_type` in `system` attributes the RFC doesn't need to change.  We could make a new attribute or modify another existing one if there's a consensus preference.

milroy:
> The third commit is pretty big and probably can be split. Please consider splitting it to [...]

I'll split up the third commit as you indicate.

>Try to make your commit message imperative. For example, instead of
"This commit enables elastic jobs by adding metadata to keep track of exclusive job types",
"Enable enable elastic jobs by adding metadata to keep track of exclusive job types. This can make our logs appear more consistent.

That's valuable feedback, I'll change my messages and make sure to do that in the future.

resource/schema/sched_data.hpp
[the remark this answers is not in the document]
milroy: I created `elastic_job_running_at` with the intention of simplifying the code additions to `dfu_impl_t::dom_dfv`.  Upon further thought, I can add a temp variable in `dfu_impl_t::dom_dfv` with a couple more lines of code to update and reset `elastic_job`.  That should work fine and I won't need `elastic_job_running_at`.

milroy:
> It seems the dfu code changes in c0e6001 would better fit the second commit (2d129a7)?

@dongahn: yes, that makes sense.  I can put it into 2d129a7.

resource/traversers/dfu.cpp
milroy: I do have a test to make sure elastic jobs don't get reserved, but does that not take into account different backfill policies?

dongahn: Yes I saw that. But I think the problem will occur when you hook this into `resource`. Consider the following scenario.

1. EASY is selected. 
2. an elastic job is the highest priority job which cannot be "allocated" so EASY will try to reserve it to the earliest time point (through `allocate_orelse_reserve`)
3. but this will get `EINVAL`, which means an error was encountered.

At this point, I think EASY won't know what to do.

milroy: Ok, I see.  I hadn't considered how the backfill policies handle errors.

dongahn: I think we need to go back to the drawing board and think about the relationship between elastic job support and queuing policies. 

One can imagine we limit our elastic job support to FCFS thereby only `allocate` is used and thus elastic job support is clearly defined. However, the concepts of queuing policies and jobspecs are completely orthogonal at this point. IOW, all policies are clearly defined for *any* jobspec and thus good separation of concerns. So I hate to make a special case like this. 

It would be best if the concept of elastic jobs is such that it is clearly defined against all of the queueing policies. Like it is with regular jobs. -- Seems like a worthwhile topic to discuss.

SteVwonder: One thought, maybe we could combine this functionality with something like #588 so that `allocate` is called on a preemptible job and never `allocate_orelse_reserve`.  But you could still use EASY backfilling etc.

Another thought, we could leverage the multiple queue idea in https://github.com/flux-framework/flux-sched/issues/529 to allow multiple scheduling policies depending on the jobs label (e.g. `rigid`, `elastic`, `batch`, `debug`).

milroy:
I'm closing this PR.  Some of the features might make it into future PRs so I'm keeping the branch on my fork.

PR #603, opened 2020-02-28
build: make it easier to build flux-sched against side-installed flux-core

### LINKED ISSUE #609: Add a `--with-flux-core-prefix` configure option
Adding @grondo's suggestion from https://github.com/flux-framework/flux-docs/pull/23, it would nice to have a option like `--with-flux-core-prefix=PREFIX` as opposed to setting `PKG_CONFIG_PATH=$PREFIX/lib/pkgconfig`.

grondo:
Currently, building flux-sched against a side installed flux-core requires `PKG_CONFIG_PATH` be set appropriately, along with `--prefix` if, after `make install`, you want to be able to use flux-sched without jumping through further hoops.

This PR eases this process by making the assumption that almost all the time, you want to install flux-sched to the same `--prefix` as the flux-core you are building against. The changes include:

 * Default prefix, if not supplied by `--prefix` is the same prefix as the first `flux` binary found in PATH
 * `${prefix}/lib/pkgconfig` is always prepended to `PKG_CONFIG_PATH` before search for libflux-core pkg-config files. This obviates the need to set PKG_CONFIG_PATH by hand in most cases.
 * Fixes a bug where the `FLUX` variable (path to `flux` binary) could be from a different flux-core install than the flux libraries found by `pkg-config`. (`FLUX_PREFIX/bin` is prepended to `PATH` when searching for `flux`)

The end result is that building against a side installed flux looks like this:

```console
$ ./configure --prefix=$HOME/flux
```
or, this will work:
```console
$ ${HOME}/flux/bin/flux env ./configure
```

since the default `--prefix` is the same as that of the first `flux` found in PATH, `configure` will always build against the Flux install that you are actively using.
This should be less surprising in most cases.

Closes #609


SteVwonder:
Now that #609 is in the right repo, can you add it to the OP or a commit?  Sorry about that @grondo.

dongahn:
For my own edification, what is the OP?

SteVwonder:
> For my own edification, what is the OP?

Sorry, I guess it's more a Reddit initialism than a Github one.  It stands for Original Post.

grondo:
Feel free to edit the initial comment @SteVwonder. GitHub should allow you
to do that. I won't be available until this evening.

SteVwonder:
Thanks @grondo for putting this together! This is a massive ergonomics improvement, and will make our install instructions very streamlined (feel free to remove the PKG_CONFIG_PATH line from the README.md if you want, otherwise we can do that in a different PR).

This change seems to work well for detecting the flux-core path, but it seems to break on `flux-idset`.  Assuming I'm not doing something wrong on my end, I think that [restoring the saved PKG_CONFIG_PATH](https://github.com/flux-framework/flux-sched/pull/603/files#diff-95b54d2de3401d64f74c664fe19f8ca1R63) is what is causing this issue since the [flux-idset check using PKG_CHECK_MODULES and is outside the `ax_flux_core.m4`](https://github.com/flux-framework/flux-sched/blob/master/configure.ac#L62).  I am not sure what the right solution is:

- don't restore the saved `PKG_CONFIG_PATH`, just keep the `${prefix}/lib/pkgconfig` path prepended for the rest of the configure script?
- move the `flux-idset` and `flux-schedutil` detection into `ax_flux_core.m4`?
- something else?

```
herbein1@asgardia ~/Repositories/flux-framework/flux-sched (configure-improvements ?S)
❯ ./configure --prefix=$HOME/opt/flux-core/master-2020-02-29
<snip>
checking for FLUX_CORE... yes
checking for flux... /home/herbein1/opt/flux-core/master-2020-02-29/bin/flux
checking for YAMLCPP... yes
checking whether yaml-cpp/yaml.h is usable... yes
checking whether yaml-cpp has YAML::Node::Mark... no
checking for boostlib >= 1.53.0 (105300)... yes
checking for buggy boostlib versions (105400-105800)... no
checking whether the Boost::System library is available... yes
checking for exit in -lboost_system... yes
checking whether the Boost::Filesystem library is available... yes
checking for exit in -lboost_filesystem... yes
checking whether the Boost::Graph library is available... yes
checking for exit in -lboost_graph... yes
checking whether the Boost::Regex library is available... yes
checking for exit in -lboost_regex... yes
checking for IDSET... configure: error: Package requirements (flux-idset) were not met:

No package 'flux-idset' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables IDSET_CFLAGS
and IDSET_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

```
herbein1@asgardia ~
❯ $HOME/opt/flux-core/master-2020-02-29/bin/flux env ./configure
<snip>
checking for FLUX_CORE... yes
checking for flux... /home/herbein1/opt/flux-core/master-2020-02-29/bin/flux
checking for YAMLCPP... yes
checking whether yaml-cpp/yaml.h is usable... yes
checking whether yaml-cpp has YAML::Node::Mark... no
checking for boostlib >= 1.53.0 (105300)... yes
checking for buggy boostlib versions (105400-105800)... no
checking whether the Boost::System library is available... yes
checking for exit in -lboost_system... yes
checking whether the Boost::Filesystem library is available... yes
checking for exit in -lboost_filesystem... yes
checking whether the Boost::Graph library is available... yes
checking for exit in -lboost_graph... yes
checking whether the Boost::Regex library is available... yes
checking for exit in -lboost_regex... yes
checking for IDSET... configure: error: Package requirements (flux-idset) were not met:

No package 'flux-idset' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables IDSET_CFLAGS
and IDSET_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

```
(130) herbein1@asgardia ~
❯ ls $HOME/opt/flux-core/master-2020-02-29/lib/pkgconfig
flux-core.pc  flux-idset.pc  flux-optparse.pc  flux-pmi.pc  flux-schedutil.pc

herbein1@asgardia ~
❯ cat $HOME/opt/flux-core/master-2020-02-29/lib/pkgconfig/flux-idset.pc
prefix=/home/herbein1/opt/flux-core/master-2020-02-29
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: flux-idset
Description: Flux Resource Manager ID Set Library
Version: 0.16.0
Libs: -L${libdir} -lflux-idset
Cflags: -I${includedir}
```

grondo:
Ah, yeah, sorry. I was being paranoid about restoring PKG_CONFIG_PATH, but it is probably ok to keep it for the duration of `configure`. Though I think your idea of moving the check for `libflux-idset` and `schedutil` into `AX_FLUX_CORE` has merit too.

Also, I realized we can get the libflux-core version from `pkg-config` as well. Perhaps we could use this to ensure the version of `libflux-core` and `flux(1)` match?

grondo:
@SteVwonder, I've pushed some changes here going with your suggestion of moving the detection of `libflux-idset` and `libflux-schedutil` under `AX_FLUX_CORE`. One benefit of this approach is another project could pick up `ax_flux_core.m4` from flux-sched and use it (if there's ever another C autoconf project that needs to build against flux-core).

The macro now searches for all of `libflux-idset`, `libflux-schedutil`, and `libflux-optparse` to be more generic, though flux-sched only appears to use `libflux-schedutil` at this time.

Additionally, `LIBFLUX_VERSION` is now defined, eventually this can be used to add a version test to the macro. For now, `LIBFLUX_VERSION` is added to the info blurb at the end to hopefully catch misconfigurations.

I've also updated the `README.md` with something I hope makes sense.

grondo:
Another idea: It doesn't make much sense to configure flux-sched against a different version of Python than flux-core. We could set `PYTHON_VERSION` after `AX_FLUX_CORE` like this:
```
# Export PYTHON_VERSION
PYTHON_VERSION=$($FLUX python -c 'import sys; print(".".join(map(str, sys.version_info[[:2]])))')
```

grondo:
Ok, with the latest push, `AX_FLUX_CORE` now sets `FLUX_PYTHON_VERSION` to the version of `$FLUX python`. `PYTHON_VERSION` is initialized to this value if `PYTHON_VERSION` not already set by the user.

Also, generate a `configure` error if `flux` executable isn't found by `configure`, instead of just setting `FLUX="flux"`

grondo:
If this all looks good I'll want to squash everything down before merging.

SteVwonder:
> It doesn't make much sense to configure flux-sched against a different version of Python than flux-core. We could set PYTHON_VERSION after AX_FLUX_CORE like this:

Sounds good to me 😁

grondo:
Ok, squashed things a little bit and force-pushed.

grondo:
Found one more thing in flux-sched's configure and pushed a fix here. There was some code to remove `$cwd/src/cmd` from the current `PATH` which doesn't apply in flux-sched. The commit removes this unnecessary code.

grondo:
Thanks!

Issue #604, opened 2020-02-29
rc scripts do not use `flux module remove -f`


SteVwonder:
The `qmanager-start` rc1 script and the `*-stop` rc3 scripts do not use `flux module remove -f`.  So if `sched-simple` is not already loaded when `qmanager-start` runs or if `qmanager` or `resource` are unloaded before the rc3 scripts run, you get the following errors:

```
2020-02-29T03:40:04.829955Z broker.err[0]: rc1: flux-module: cmb.rmmod sched-simple: No such file or directory
2020-02-29T03:40:04.831564Z broker.err[0]: Run level 1 Exited with non-zero status (rc=1) 0.6s
2020-02-29T03:40:04.943633Z broker.err[0]: rc3: flux-module: cmb.rmmod qmanager: No such file or directory
2020-02-29T03:40:04.960031Z broker.err[0]: rc3: flux-module: cmb.rmmod resource: No such file or directory
```

I believe this also affects the error code of the exiting broker (causing test cases to fail unnecessarily).

dongahn:
Yap it is a bug...Thanks.

SteVwonder:
I have the changes locally, I can post a PR shortly.

PR #605, opened 2020-02-29
rc: add numbered prefixes

### LINKED ISSUE #587: rc scripts don't have number prefix
After building and installing `flux-core` and `flux-sched`, the `etc/rc1.d` directory looks like:

```
→ ll rc1.d
total 16K
-rwxr-xr-x 1 root   root   571 Jan 30 21:33 01-enclosing-instance
-rwxr-xr-x 1 root   root   632 Feb  1 01:17 qmanager-start
-rwxr-xr-x 1 root   root   683 Feb  1 01:17 resource-start
```

I think they should be numbered to ensure an ordering, especially when user/site rc scripts start getting added into the mix.

SteVwonder:
- Add numbered prefixes to rc scripts
- Removes an unused sharness environment variable: FLUXION_RESOURCE_RC_PATH

Closes #587


t/t0003-basic-install.t
```diff
@@ -20,7 +20,7 @@ fi
 # test_under_flux is under sharness.d/
 #
 SIZE=2
-if [ test ${FLUX_FLUX_CO_INST} = co ]; then
+if [ test ${FLUX_SCHED_CO_INST} != co ]; then
```
dongahn: I think this will still fail, if a version of `flux-sched` was previously installed into `flux-core` you are testing this against...

If `FLUX_SCHED_CO_INST` == "co", this will use the installed RC scripts, which you probably want to avoid. 

I think this is the same issue as the problem you hit when you test PR #599...

.travis.yml
```diff
@@ -27,6 +27,7 @@ matrix:
       env:
        - ARGS="--prefix=/usr"
        - TEST_INSTALL=t
+       - FLUX_SCHED_TEST_INSTALLED=t
```
dongahn: In my PR #599, I actually added this to `travis_run.sh` as it was exporting another variable. Perhaps doing it at one level down would be a bit cleaner?

t/t0003-basic-install.t
SteVwonder: Correct, this PR does not generically fix #607.  It enforces that only one set of flux-sched rc scripts runs, but unfortunately the ones it allows to run are the installed ones (not the in-tree ones).  I wasn't necessarily shooting to fix #607 with this PR.  Just patch up this test enough so that we have something to prevent a regression of the fix for #604.  We should definitely revisit this bit of logic once #607 is solved generically.

.travis.yml
SteVwonder: Ah! Perfect!  I'll pull this line then, since your changes covers it.

SteVwonder:
Hmmm.  Got a failure just on Centos 7.  I'm tempted to say it's unrelated to this PR and might be a racy test.  The `stop due to alloc response error` below look ominous.  Going to try and restart the builder to see if it comes back.  Here are the relevant logs in case it doesn't reproduce.

test-script output:
```
expecting success: 

    flux module remove qmanager &&
    flux module load qmanager &&
    flux job list > jobs.list2 &&
    diff jobs.list2 jobs.list

2d1
< {"id": 11442061312, "userid": 2000, "priority": 16, "t_submit": 1582953004.6481304, "state": 4, "name": "hostname", "ntasks": 1, "t_depend": 1582953004.6481304, "t_sched": 1582953004.6600838}
not ok 5 - qmanager: restart keeps the main job queue intact
```

dmesg:
```
2020-02-29T05:10:04.619644Z qmanager.debug[0]: alloc success (id=10720641024)
2020-02-29T05:10:04.639460Z job-manager.debug[0]: submit_cb: added 1 jobs
2020-02-29T05:10:04.660067Z job-manager.debug[0]: submit_cb: added 1 jobs
2020-02-29T05:10:04.679174Z broker.debug[0]: rmmod qmanager
2020-02-29T05:10:04.679250Z job-manager.debug[0]: alloc: stop due to alloc response error: Function not implemented
2020-02-29T05:10:04.679278Z qmanager.debug[0]: service_unregister
2020-02-29T05:10:04.679342Z broker.debug[0]: module qmanager exited
2020-02-29T05:10:04.690664Z broker.debug[0]: insmod qmanager
2020-02-29T05:10:04.691011Z qmanager.debug[0]: service_register
2020-02-29T05:10:04.691084Z job-manager.debug[0]: scheduler: hello
2020-02-29T05:10:04.693787Z job-manager.debug[0]: scheduler: ready single
```

t/t0003-basic-install.t
SteVwonder: >  this will use the installed RC scripts, which you probably want to avoid.

Scratch my last response, as I think more about it, the whole intent of this script is to test the installed RC scripts and thus, by definition, ignore the rc scripts in-tree.  So we actually do want this behaviour for this test, but we don't always want this test to run (i.e., when your in-tree and install rc scripts differ).  We only want it to run when we intend to test the install scripts (i.e., in Travis CI), hence why this script is gated by the `FLUX_SCHED_TEST_INSTALLED` variable.  Sorry I didn't catch that earlier, it's getting late and I should probably step away from coding now :laughing:.

SteVwonder:
It didn't reproduce, so it looks like we have a race-condition bug on our hands.  I'll open an issue.

dongahn:
Probably the module got removed before the submitted job finished and we need to utilize some of the new sync primitives like `flux job wait`...

t/t0003-basic-install.t
dongahn: Assuming there will be a fix for ignoring installed third party RCs, this code appears correct to me.

dongahn:
Once you remove your FLUX_SCHED_TEST_INSTALLED export commit and rebase, and travis CI passes, this looks good to me.

SteVwonder:
Just rebased and force-pushed, but it looks like the t0003-basic-install.t is not marked as failed when the broker encounters an error in rc3 scripts (i.e., when I revert 5191f03).  Lemme poke at this some to see if I can get the test to fail when that happens before we merge, otherwise it isn't really all that useful.

SteVwonder:
With the addition of b9ca320, `t0003-basic-install.t` will fail if 5191f03 is reverted, but it has the nasty side-effect of running every script in the installed rc3.d directory.  If flux-core puts any scripts there in the future, it will run them as well.  Not sure if it is the right solution.  Maybe the fix in #607 will help?

dongahn:
@SteVwonder: this is getting long in the tooth. Do we want to pick up something from this PR at this point?

SteVwonder:
Just rebased onto the latest master, dropped the added test which I believe we determined in no longer needed, and a commit that was superseded by other PRs.  Leaving the rc numbering change and the deletion of an obsolete variable.  Relatively small PR.

SteVwonder:
@dongahn, the numbered rc scripts has been added to #694, correct? If so, I think we can close this PR.

dongahn:
Ah yes. Closing this makes sense

Issue #383, opened 2018-08-30
JSON in C++


SteVwonder:
Now that more of flux-sched is becoming C++, do we want to keep using the jansson C interface to interact with JSON objects from within C++, or do we want to pull in a C++ JSON library?  @trws recommended [JSON for modern C++](https://github.com/nlohmann/json) (which would require a patch to compile under gcc 4.8.5) or [RapidJSON](https://github.com/Tencent/rapidjson).

dongahn:
I guess when we have real use cases. I haven't had one yet. Flux's pack and unpack has been sufficient for me so far.

dongahn:
#455 clearly indicates that we will absolute need this for better code maintainability.

trws:
One thought I was reminded of earlier this week.  I think either of those fit the bill, but it would be really good to pick one that interoperates well with one of the jsonschema implementations for C++ so we don't need to end up with multiple.  Might be good to replace the python dep on jsonschema in the long run too if that becomes a performance issue.

dongahn:
Makes sense. 

FWIW, for sched for now, python jsonschema isn't on the critical path as I just used it for testing. For actual execution, @grondo and I decided we won't pass an R through jsonschema validation as this is an internal operation and a malformed R will result in an exception raised from the execution system. 

But in general, it does make sense to have "embedded jsonschema" support as this seems to become trendy with REST API and etc.

SteVwonder:
> it would be really good to pick one that interoperates well with one of the jsonschema implementations for C++

That's a good point.  Here are the C++ json-schema validators that I could find (via the [json-schema.org implementations list](http://json-schema.org/implementations.html) and from googling) and the C++ json parser that they are compatible with:

| jsonschema lib | compatible json parsers | notes |
| -------------- | ----------------------- | ----- |
| [Valijson](https://github.com/tristanpenman/valijson) | [many](https://github.com/tristanpenman/valijson#supported-parsers) | requires C++11, supports json-schema draft-4 |
| [RapidJSON](http://rapidjson.org/md_doc_schema.html) | RapidJSON | supports json-schema draft-4 |
| [pboettch's](https://github.com/pboettch/json-schema-validator) | [nlohmann's JSON for Modern C++](https://github.com/nlohmann/json) | compatible JSON library won't compile with gcc 4.8.5, supports draft-7 |
| [KayEss's](https://github.com/KayEss/json-schema) | ? | requires C++17 |

dongahn:
[...]

PR #613, opened 2020-03-04
Use Jansson to emit JSON-based writers + time keys in RV1

### LINKED ISSUE #610: Emit starttime and expiration from our RV1 emitter
To be compliant with the latest [RFC 20](https://github.com/flux-framework/rfc/blob/master/spec_20.rst), our RV1 emitter needs to generate `starttime` and `expiration` keys.  This is a prerequisite for the Issue #470 (an ability to reconstruct the resource state upon qmanager and resource reload).

dongahn:
This PR reworks the match writers to use Jansson for emitting JSON-based writers. 

Currently, the writers commonly use C++ `std::stringstream` to emit resource set information regardless of emitting formats: i.e., plain-text formats or JSON-based formats alike. For a JSON-based format, however, this approach is becoming increasingly more complex to modify when we change the format.

Use Jansson to emit all of the JSON-based resource set information including `rlite`, `jgf`, `rv1` and `rv1_nosched` formats.

Fix #610 and #383.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1,5 +1,6 @@
 /*****************************************************************************\
- *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  Copyright (c) 2014 - 2020 Lawrence Livermore National Security, LLC.
+ *  Produced at
  *  the Lawrence Livermore National Laboratory (cf, AUTHORS, DISCLAIMER.LLNS).
  *  LLNL-CODE-658032 All rights reserved.
  *
@@ -672,7 +673,6 @@ static int run_match (std::shared_ptr<resource_ctx_t> &ctx, int64_t jobid,
     struct timeval end;
 
     gettimeofday (&start, NULL);
-    ctx->writers->reset ();
 
     if (strcmp ("allocate", cmd) != 0
         && strcmp ("allocate_orelse_reserve", cmd) != 0
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -142,7 +142,7 @@ int resource_reader_jgf_t::unpack_vtx (json_t *element, fetch_helper_t &f)
     if ( (json_unpack (element, "{ s:s }", "id", &f.vertex_id) < 0)) {
         errno = EPROTO;
         m_err_msg += __FUNCTION__;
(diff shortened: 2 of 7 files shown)


dongahn:
> One quick comment: some jansson calls that can return errors are not being checked, such as json_array(), json_dumps(), json_object_set_new() , json_object(), json_string().

Ah. I tried to check errors from these calls as much as possible but apparently there were some blind spots. Thank you for catching that!  E.g., `json_dumps()` and that embedded `json_string()`. I will look at it more closely and fix them. 

> I guess advantage of a native C++ json library here would be presumably it would throw exceptions instead of returning some value you have to check for...

That would be one advantage or disadvantage depending on how you see it. I have seen a large C++ project that disallows use of exception once and for all because of its various side effects:-)

But one advantage of a native C++ json would be that it can allow me to follow the [RAII](https://en.cppreference.com/w/cpp/language/raii) pattern a bit better. 

```C++
class writer {

private:
    Json m_out1; // Not a raw pointer or
    std::shared_ptr<Json> m_out2; // use a C++ shared pointer or unique pointer  
};
```

I introduced the `initialize` method to do this resource acquisition,  but the ideal pattern would be to do this within constructors. I could actually turn `ENOMEM` condition into my a custom exceptions within the constructors, but I have generally avoided prevalent use of exceptions.

BTW, this also reminds me that I need to check the error code from the callsite of `initialize`, which I will do as well.

dongahn:
> Ah. I tried to check errors from these calls as much as possible but apparently there were some blind spots. Thank you for catching that! E.g., json_dumps() and that embedded json_string(). I will look at it more closely and fix them.

Well, I also found the PR actually includes one interim prototype commit which I shouldn't have included where some of those calls are not error checked at all. Let me remove that commit. Sorry about that.

dongahn:
>> I introduced the initialize method to do this resource acquisition, but the ideal pattern would be to do this within constructors. I could actually turn ENOMEM condition into my a custom exceptions within the constructors, but I have generally avoided prevalent use of exceptions.

> BTW, this also reminds me that I need to check the error code from the callsite of initialize, which I will do as well.

Now that I think about this, there should be only one call-site of `initialize`, the factory method for a writer where we do catch the `std::bad_alloc` exception. So I can have the writer constructors raise this exception when `ENOMEM` is encountered without having to worry about that exception spilled up the stack. 

Let me close this PR and repost with that change. Sorry for the noise.

PR #614, opened 2020-03-04
Use Jansson for JSON-based writers + time keys in RV1

### LINKED ISSUE #610: Emit starttime and expiration from our RV1 emitter

dongahn:
This PR reworks the match writers to use Jansson for emitting JSON-based writers.

Currently, the match writers commonly use C++ `std::stringstream` to emit resource set information regardless of emitting formats: i.e., plain-text formats or JSON-based formats alike. For a JSON-based format, however, this approach is becoming increasingly more complex to modify when we change the format.

Use Jansson to emit all of the JSON-based resource set information including `rlite`, `jgf`, `rv1` and `rv1_nosched` formats.

This should address the issues identified from my previous attempt: #613.

Fix #610 and #383.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -686,7 +686,11 @@ static int run_match (std::shared_ptr<resource_ctx_t> &ctx, int64_t jobid,
     if ((rc = run (ctx, jobid, cmd, jstr, at)) < 0)
         goto done;
 
-    ctx->writers->emit (o);
+    if ((rc = ctx->writers->emit (o)) < 0) {
+        flux_log_error (ctx->h, "%s: writer can't emit", __FUNCTION__);
+        goto done;
+    }
+
     gettimeofday (&end, NULL);
     *ov = get_elapse_time (start, end);
     update_match_perf (ctx, *ov);
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -672,7 +672,6 @@ static int run_match (std::shared_ptr<resource_ctx_t> &ctx, int64_t jobid,
     struct timeval end;
 
     gettimeofday (&start, NULL);
-    ctx->writers->reset ();
 
     if (strcmp ("allocate", cmd) != 0
         && strcmp ("allocate_orelse_reserve", cmd) != 0
--- a/resource/traversers/dfu_impl_update.cpp
+++ b/resource/traversers/dfu_impl_update.cpp
@@ -477,7 +477,15 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers,
     unsigned int needs = m_graph_db->metadata.v_rt_edges[dom].get_needs ();
     m_color.reset ();
 
-    rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu);
+    if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu)) > 0) {
+         uint64_t starttime = jobmeta.at;
+         uint64_t endtime = jobmeta.at + jobmeta.duration;
+         if (writers->emit_tm (starttime, endtime) == -1) {
+             m_err_msg += __FUNCTION__;
+             m_err_msg += ": emit_tm returned -1.\n";
+         }
+     }
+
     return (rc > 0)? 0 : -1;
 }
 
--- a/resource/writers/match_writers.cpp
+++ b/resource/writers/match_writers.cpp
@@ -690,6 +690,13 @@ int rv1_match_writers_t::emit_edg (const std::string &prefix,
     return rc;
 }
 
+int rv1_match_writers_t::emit_tm (uint64_t start_tm, uint64_t end_tm)
+{
+    m_starttime = start_tm;
+    m_expiration = end_tm;
+    return 0;
+}
+
 
 /****************************************************************************
  *                                                                          *
@@ -744,6 +751,13 @@ int rv1_nosched_match_writers_t::emit_vtx (const std::string &prefix,
                                            bool exclusive)
 {
     return rlite.emit_vtx (prefix, g, u, needs, exclusive);
+}
+
+int rv1_nosched_match_writers_t::emit_tm (uint64_t start_tm, uint64_t end_tm)
+{
+    m_starttime = start_tm;
+    m_expiration = end_tm;
+    return 0;
 }
 
 
--- a/resource/writers/match_writers.cpp
+++ b/resource/writers/match_writers.cpp
@@ -642,10 +642,12 @@ int rv1_match_writers_t::emit (std::stringstream &out)
         json_decref (rlite_o);
         goto ret;
     }
-    if (!(o = json_pack ("{s:i s:{s:o s:o}}",
+    if (!(o = json_pack ("{s:i s:{s:o s:I s:I} s:o}",
                              "version", 1,
                              "execution",
                              "R_lite", rlite_o,
+                             "starttime", m_starttime,
+                             "expiration", m_expiration,
                              "scheduling", jgf_o))) {
         json_decref (rlite_o);
         json_decref (jgf_o);
@@ -720,10 +722,12 @@ int rv1_nosched_match_writers_t::emit (std::stringstream &out)
         goto ret;
     if ((rc = rlite.emit_json (&rlite_o)) < 0)
         goto ret;
-    if (!(o = json_pack ("{s:i s:{s:o}}",
+    if (!(o = json_pack ("{s:i s:{s:o s:I s:I}}",
                              "version", 1,
                              "execution",
-                             "R_lite",  rlite_o))) {
+                             "R_lite",  rlite_o,
+                             "starttime", m_starttime,
+                             "expiration", m_expiration))) {
         json_decref (rlite_o);
         rc = -1;
         errno = ENOMEM;
--- a/resource/writers/match_writers.cpp
+++ b/resource/writers/match_writers.cpp
@@ -621,31 +621,51 @@ ret:
  *                                                                          *
  ****************************************************************************/
 
-void rv1_match_writers_t::reset ()
+bool rv1_match_writers_t::empty ()
 {
-    rlite.reset ();
-    jgf.reset ();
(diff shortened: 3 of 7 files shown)


dongahn:
codecov raised a flag, but it looks like this is just because this PR has more failure handling code and line number changes.

resource/writers/match_writers.cpp
```diff
@@ -97,140 +99,433 @@ int sim_match_writers_t::emit_vtx (const std::string &prefix,
...
+                             "graph",
+                                 "nodes", m_vout,
+                                 "edges", m_eout))) {
+        errno = ENOMEM;
+        goto ret;
+    }
+    if (!(m_vout = json_array ())) {
+        json_decref (o);
+        o = NULL;
+        errno = ENOMEM;
```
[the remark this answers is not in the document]
dongahn: Good catch. Yes this should return. Thanks.

resource/writers/match_writers.cpp
```diff
@@ -97,140 +99,433 @@ int sim_match_writers_t::emit_vtx (const std::string &prefix,
...
+            json_decref (o);
+            o = NULL;
+            errno = ENOMEM;
+            goto out;
+        }
+        out << json_str << std::endl;
+        free (json_str);
+        json_decref (o);
+    }
+out:
```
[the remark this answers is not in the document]
dongahn: Ah... yes good catch. A different label should be far better for readability.

dongahn:
> The jansson conversion commit is really huge, and appears to do a lot more than just change the way the output is generated. Can it be broken down into commits that do one thing each?

Good comment. Yeah I was a bit concerned about the side but didn't really have a good idea to split them. I will think about this some more.

dongahn:
> In a couple places, assertions are raised when json_ functions fail, and most others, the errors are passed back to the caller. Was it intentional to handle it both ways?

Are you talking about the throw bad_alloc exception statements in the constructors? If so, yes this is intentional. C++ constructors cannot return a code, this is about the best way to handle this case. 

If you are talking about other places then this would not be intentional and I should fix it. I see one such place from your online comment. Is it what you are referring to?

resource/writers/match_writers.cpp
```diff
@@ -97,140 +99,433 @@ int sim_match_writers_t::emit_vtx (const std::string &prefix,
  *                                                                          *
  ****************************************************************************/
 
-int jgf_match_writers_t::emit (std::stringstream &out, bool newline)
+jgf_match_writers_t::jgf_match_writers_t ()
 {
-    size_t vout_size = m_vout.str ().size ();
-    size_t eout_size = m_eout.str ().size ();
+    if (!(m_vout = json_array ()))
+        throw std::bad_alloc ();
```
garlick: This is the `throw` (and the one after it) that I was referring to in my general comment.  Sorry for not pointing it out.

resource/modules/resource_match.cpp
```diff
@@ -1,5 +1,6 @@
 /*****************************************************************************\
- *  Copyright (c) 2014 Lawrence Livermore National Security, LLC.  Produced at
+ *  Copyright (c) 2014 - 2020 Lawrence Livermore National Security, LLC.
```
garlick: Updating copyright dates adds churn to the history and, as far as I understand, has no useful legal effect, since the  initial claim doesn't expire until well past the usable life of the software (120 years in the USA).

https://en.wikipedia.org/wiki/Copyright_law_of_the_United_States

dongahn: Oh. I had no idea. Great to know.

resource/writers/match_writers.cpp
dongahn: Yeah. This is intended. 

This code is following a good C++ practice called RAII whereby all of the resources needed by a class is acquired by its constructors. Because there isn't a way to return a value from a C++ constructor, throwing a `std::bad_alloc` exception on `ENOMEM` is just about the best way I can think of to use the C Jansson library while following this principle.

Like mentioned before, I had tried to minimize the use of exceptions, but in this case I figured it makes sense. All of the writer objects are constructed through [a factory method](https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.cpp#L414) which actually handles this exception type already. `make_shared` has no side effect when the creating object throws an exception.

dongahn:
> Good comment. Yeah I was a bit concerned about the side but didn't really have a good idea to split them. I will think about this some more.

For the record, I'm currently acting on this (great) comment. So hold off your review if you have't started. I am trying to convert one writer at a time in a separate commit. This requires a bit more effort (adjusting its users as well), but this is definitely worth it.

dongahn:
I believe this is ready for another review. I am going to need this PR to make progress with #470.

resource/modules/resource_match.cpp
milroy: TIL. Definitely good to know!

dongahn:
Thanks @milroy. Let me look at the reduced code coverage stats.

dongahn:
> That said, let's figure out why the differential coverage is low before the PR gets merged.

There are just too many error handling branches with the introduction of Jansson, compared to the previous code that used C++ `std::stringstream`. So,  the source lines that were added as part of this PR received a low testing coverage -- a lot of those lines look like the following code segment.

Screen Shot 2020-03-18 at 6 45 47 PM

I tried my best to manually test each and every branch (including memory leaks w/ valgrind).

Still, I generally felt that `flux-sched` need some more quality time to look at its behaviors on various types of errors so I had created a new issue #618. This should be a part of stabilization/hardening effort. Hopefully this addresses your concerns on this PR about the coverage @milroy.

dongahn:
@Mergifyio rebase

milroy:
> Hopefully this addresses your concerns on this PR about the coverage @milroy.

Thanks!  It certainly does, and I suspected the error handling branches given my previous investigation into codecov/patch.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -477,7 +477,15 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers,
     unsigned int needs = m_graph_db->metadata.v_rt_edges[dom].get_needs ();
     m_color.reset ();
 
-    rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu);
+    if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu)) > 0) {
+         uint64_t starttime = jobmeta.at;
+         uint64_t endtime = jobmeta.at + jobmeta.duration;
+         if (writers->emit_tm (starttime, endtime) == -1) {
+             m_err_msg += __FUNCTION__;
+             m_err_msg += ": emit_tm returned -1.\n";
```
SteVwonder: Nit: as far as I can tell, this function only ever returns 0 as of now, but in the future, it could return any `int`.  Might be a little more resilient to check for < 0:
```suggestion
         if (writers->emit_tm (starttime, endtime) < 0) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned <0.\n";
```

dongahn: Good point.

dongahn:
> I did have an over-arching question about handling allocation failures in C++. It looks like throughout this PR (expect in the constructor as you noted), the C-style of error handling is being used (i.e., use the return value to signal an error occured and errno to provide more information). What's the reasoning behind doing that over say throwing a std::bad_alloc exception, catching that exception at a higher-level (like in the main cb function) and doing the errno setting (or msg response) there? 

Mainly because I didn't have @trws's C++ wrapper yet :-). His change will be a part of Issue #618 needed for stabilization. 

> Something along the lines of what @trws did in his C++ interface PR. It seems like it would save quite a few lines of code (at least 2 or 3 per error condition), but I'm not sure what the cons list is. I ask, because it's what we are doing in the python bindings and was curious as to why it isn't done in C++.

Also in general, I have been trying to minimize the use of C++ exceptions in the flux-sched code unless I have to and also I know it is absolutely safe (e.g., the case of constructors in this PR). Without a safe net like @trws's wrapper, you can miss an exception and that can crash the entire flux instance. 

But w/ the introduction @trws's wrapper, I do hope the use of exceptions would become easier although with the existing code and significantly use of c-based libraries, C++ exception handling will be less prevalent than C style still in the end.

resource/traversers/dfu_impl_update.cpp
dongahn: Actually, this is a part of `emit` methods (we have `emit_vtx` and `emit_edg`) and the error checking is all done against -1.  To be consistent, it seems to make sense to keep this code as is for now and then later when we address #618, we can reevaluate and change all of the callosities if necessary, instead of making one off change?

trws:
As a follow-on to @SteVwonder's question, I don't see it as a problem necessarily, but why jansson here rather than working in the C++ wrapper PR, or at least part of it, and using the nlohmann/json library I pulled in there? It would be less code, less error prone, and faster, though admittedly different from our C json handling.

dongahn:
> As a follow-on to @SteVwonder's question, I don't see it as a problem necessarily, but why jansson here rather than working in the C++ wrapper PR, or at least part of it, and using the nlohmann/json library I pulled in there? It would be less code, less error prone, and faster, though admittedly different from our C json handling.

Thank you for bringing this up. I actually seriously considered your library or other C++ JSON libraries here as well.

But as I mentioned in this PR and a commit, I decided to go with Jansson with a thought that it is well understood compared to a brand new library. It is used in everywhere else. I was risk averse given the time limit we are facing. 

Now, even with this well-understood Jansson, I found things like https://github.com/flux-framework/flux-core/issues/2803#issuecomment-601236699. W/ the use of a common infrastructure, not only me but also other Flux developers can commonly pursue issues.

We can certainly explore C++-based one perhaps after we achieve our major milestone.

dongahn:
Thanks.

dongahn:
File https://github.com/flux-framework/flux-sched/issues/628

PR #617, opened 2020-03-07
testsuite: break a race in a qmgr reload test

dongahn:
Problem: A qmanager reload test is randomly
failing in Travis CI, which tests jobs in
the job-manager queue aren't lost during
the reload. This test enqueues a few jobs
and removes/reloads qmanager before checking
if those jobs are still in the queue.
But because this test only gives
1 second walltime for those jobs, there is
a chance that a job can complete before
the reload is done.

Use 100 seconds as the walltime for
those jobs to break those jobs "effectively".


SteVwonder:
Thanks @dongahn!

PR #619, opened 2020-03-07
rc: use reload -f to load qmanager/resource

### LINKED ISSUE #607: `make check` fails when run after `make install`
If flux-sched is configured with the same prefix as flux-core and you attempt to run `make check` after a `make install`, `t1005-qmanager.t` will fail because the flux-sched rc.d scripts are run twice. The first time is the installed rc.d scripts being run by the installed rc script (from flux-core) and the second time is the in-tree rc.d scripts being run by the test (via FLUX_RC_EXTRA).  We need some way to disable installed flux-sched rc scripts when running certain tests.

One suggestion (which would have to be implemented in flux-core I believe), would be to add a variable like `FLUX_IGNORE_INSTALLED_SCHED_RC` that disables the installed sched rc.d scripts (but still runs the installed flux-core rc scripts) without affecting the in-tree sched rc scripts.

dongahn:
This PR changes the behavior of the rc1 scripts of flux-sched modules. 

Currently, some of the in-tree tests are failing when flux-sched is  already installed within flux-core. Their rc1 scripts are executed and load those "installed" modules instead of "in-tree" modules.

Use "flux reload -f <qmanager|resource>" so that when these failing tests load them via the FLUX_RC_EXTRA environment variable, the in-tree versions are loaded.

Discussed at https://github.com/flux-framework/flux-core/issues/2788#issuecomment-593131902. 

Fix #607 and #604 .

### DIFF
--- a/etc/qmanager-start
+++ b/etc/qmanager-start
@@ -10,9 +10,16 @@
 # Finally, if FLUX_QMANAGER_RC_NOOP=1, flux-core
 # won't load in or remove qmanager as part of runlevel 1 and 3.
 #
+# Note that we use "flux reload -f qmanager" instead of "flux load qmanager"
+# to ease our testing.
+# Without this, some of the in-tree tests can fail when flux-sched is already
+# installed with flux-core. These installed rc1 scripts are
+# executed first and disallow the "in-tree" modules from loading.
+# We now use "flux reload -f qmanager" so that when these failing tests
+# load it via FLUX_RC_EXTRA, the in-tree version can be loaded.
 
 if [ -z ${FLUX_QMANAGER_RC_NOOP} ]; then
-    flux module remove sched-simple
-    flux module load qmanager ${FLUX_QMANAGER_OPTIONS}
+    flux module remove -f sched-simple
+    flux module reload -f qmanager ${FLUX_QMANAGER_OPTIONS}
 fi
 
--- a/etc/resource-start
+++ b/etc/resource-start
@@ -10,9 +10,16 @@
 # Finally, if FLUX_RESOURCE_RC_NOOP=1, flux-core
 # won't load in or remove resource as part of runlevel 1 and 3.
 #
+# Note that we use "flux reload -f resource" instead of "flux load resource"
+# to ease our testing.
+# Without this, some of the in-tree tests can fail when flux-sched is already
+# installed with flux-core. These installed rc1 scripts are
+# executed first and disallow the "in-tree" modules from loading.
+# We now use "flux reload -f resource" so that when these failing tests
+# load it via FLUX_RC_EXTRA, the in-tree version can be loaded.
 
 if [ -z ${FLUX_RESOURCE_RC_NOOP} ]; then
     FLUX_RESOURCE_OPTIONS=${FLUX_RESOURCE_OPTIONS:-"load-whitelist=node,core,gpu"}
-    flux module load resource ${FLUX_RESOURCE_OPTIONS}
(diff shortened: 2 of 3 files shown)


grondo:
Instead of using `awk "END{print}"`, a suggestion would to clear the dmesg ring buffer (`flux dmesg -c`) before your test to ensure you aren't matching dmesg output from a previous load of the qmanager. A potential problem with accepting the "last" match is that it could generate a false positive test result in the case where the module failed in the current test.

dongahn:
> Instead of using awk "END{print}", a suggestion would to clear the dmesg ring buffer (flux dmesg -c) before your test to ensure you aren't matching dmesg output from a previous load of the qmanager. 

I thought about that. But because both of the log records are generated during the instance startup time, this isn't a way to clear up the first log out... I don't have the right control at the test body. 

> A potential problem with accepting the "last" match is that it could generate a false positive test result in the case where the module failed in the current test.

This is a good point and I have no good solution for this. I will think about about this some more.

dongahn:
> This is a good point and I have no good solution for this. I will think about about this some more.

Well I can't think of any other clean way unless we have an RPC to fetch the current parameter settings. 

We wrote the `dmesg` based hack as a stopgap before this RPC and a chance that a bug can creep into our code based because of the false positive you identified (this problem will be caught by other test configuration -- in-tree without previous install and Travis CI), perhaps practically this is still an effective solution. For midterm this will move to the RPC version anyway. 

I am open to other ideas though.

grondo:
Can you effectively test the conf load by something like:
```
flux dmesg -c
flux module reload qmanager
flux dmesg | grep "conf_parameter_name" | ...
```

dongahn:
Good thought. However, by the time you get to `flux dmesg` you won't have your flux broker, though. We also couldn't write this test  w/ `flux module reload` because you can't change `CONF` directory between `flux module load`s.

dongahn:
Maybe this can go to the initial program...

```
flux dmesg -c && flux module reload -f qmanager && flux dmseg
```

We will see if this doesn't have other side effects..

grondo:
Ah, I see. Sorry I didn't notice these tests weren't running under `test_under_flux`! (sorry seems obvious in retrospect)

dongahn:
@grondo : OK, doing this as the initial program seems to work. The only needed change was to use `dmesg -C` instead of `dmesg -c` as the latter still prints the ring buffer before clearing it up. 

I forced a push. As far as I'm concerned, this can go in.

grondo:
One chain-lint error:

```
error: bug in the test script: broken &&-chain: 
    conf_name="15-no-qmanager" &&
    export FLUX_CONF_DIR=${conf_base}/${conf_name} &&
    flux broker bash -c \
"flux dmesg -C &&  flux module reload -f qmanager && flux dmesg" \
> ${conf_name}.out
    qp=$(cat ${conf_name}.out | grep "queue params" | awk "{ print \$6}") &&
    pp=$(cat ${conf_name}.out | grep "policy params" | awk "{ print \$6}") &&
    test "${qp}" = "queue-depth=8192,max-queue-depth=1000000" &&
    test "${pp}" = "reservation-depth=64,max-reservation-depth=100000" &&
    unset FLUX_CONF_DIR

```

dongahn:
Oops. Thanks.

dongahn:
> Only comment I would make is to perhaps add a comment in rc scripts that explains why reload -f is used, for the casual observer.

Great idea. I will make a quick comment and forced a push.

dongahn:
Done.

PR #621, opened 2020-03-09
Flux tree helper

SteVwonder:
Problem: constructing and manipulating JSON blobs via the command line
has a high overhead and quickly exceeds the limit for command line
argument length set in the linux kernel.

Solution: Add a helper written in python that performs the same
action but using python's standard library utilities and the flux python
bindings.  This reduces the number of processes launched, eliminates the
need for using command line arguments to pass JSON blobs between
programs, and reduces the total number of times the JSON needs to
be (de-)serialized.

Results:
Before the helper, for a tree topology of 32x36, the perf
results record a runtime of 45 seconds, but the `time` utility used on
the flux-tree command directly reports a runtime of 3 minutes 45
seconds.  After the introduction of the helper, for the same tree
topology, the perf results record a runtime of 34 seconds, and the
`time` utility reports a runtime of 47 seconds.  So the helper not only
speeds up the post-processing time, but it also has a measurable impact
on the reported results as well.

Before the helper, for a tree topology of 1x32x36, the root instance in
the tree topology (i.e., not the instance that flux-tree is directly run
inside of), would hit the linux kernel limit for command line argument
length whilst processing the JSON performance data. This would cause the
script to error out when collecting performance results and fails to
write any performance data out to the filesystem. After the helper, the
same tree topology runs to completion, successfully saves out the
performance data, and does so with a total runtime (as reported by
`time`) of 53 seconds.

Depends on https://github.com/flux-framework/flux-core/pull/2804

### DIFF
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -165,11 +165,11 @@ Options:\n\
                                    (default=${FT_NJOBS})\n\
  -o, --perf-out=FILENAME       Dump the performance data into\n\
                                    the given file (default: don't print)\n\
- --perf-format=FORMAT          Dump the performance data with the given\n\
+     --perf-format=FORMAT      Dump the performance data with the given\n\
                                    format. Uses the python format\n\
                                    specification mini-language.\n\
                                    Example: \"{treeid:<15s},{elapse:>20f}\"\n\
- --job-name=NAME               Name to use when submitting child jobs\n\
+     --job-name=NAME           Name to use when submitting child jobs\n\
      --                        Stop parsing options after this\n\
 "
 
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -33,6 +33,9 @@ FT_QUEUE='default'             # store arg to --queue-policy (e.g., fcfs:easy)
 FT_PARAMS='default'            # store arg to --queue-params
 FT_MATCH='default'             # store arg to --match-policy (e.g., low:high)
 FT_PERF_OUT='%^+_no'           # store perf-out filename given by --perf-out
+FT_PERF_FORMAT='{treeid:<15s} {elapse:>20f} {begin:>20f} {end:>20f} {match:>15f} '\
+'{njobs:>15d} {my_nodes:>5d} {my_cores:>4d} {my_gpus:>4d}'
+                               # store perf-out with format given by --perf-format
 FT_FXLOGS='%^+_no'             # dir in which flux logs are produced
 FT_LEAF='no'                   # is this a leaf instance (given by --leaf)?
 FT_G_PREFIX='tree'             # store hierarchical path given by --prefix
@@ -56,10 +59,6 @@ declare -r top_prefix='tree'   # prefix name to identify the top Flux instance
 declare -r t_delim='x'         # topology delimiter
 declare -r p_delim=':'         # match policy delimiter
 declare -r perf_format='%-15s %20s %20s %20s %15s %15s %5s %4s %4s'
-declare -r perf_keys=' .treeid, .perf.elapse, .perf.begin, .perf.end,'\
-' .perf.match, .njobs, .my_nodes, .my_cores, .my_gpus '
-declare -r perf_heading=' TreeID Elapse(sec) Begin(Epoch) End(Epoch) Match(usec)'\
-' NJobs NNodes CPN GPN '
 declare -A mp_policies=(       # make sure to update this when match
     [low]=1                    # policies are updated.
     [high]=1
@@ -83,7 +82,7 @@ declare -a jobids             # array to store a set of submitted job IDs
 
 declare -r long_opts='help,leaf,flux-logs:,nnodes:,ncores-per-node:'\
 ',ngpus-per-node:,topology:,queue-policy:,queue-params:,match-policy:'\
-',njobs:,perf-out:,prefix:,dry-run'
+',njobs:,perf-out:,perf-format:,prefix:,dry-run'
 
 declare -r short_opts='hlf:N:c:g:T:Q:P:M:J:o:X:d'
 declare -r prog=${0##*/}
@@ -165,42 +164,16 @@ Options:\n\
                                    (default=${FT_NJOBS})\n\
  -o, --perf-out=FILENAME       Dump the performance data into\n\
                                    the given file (default: don't print)\n\
+ --perf-format=FORMAT          Dump the performance data with the given\n\
+                                   format. Uses the python format\n\
+                                   specification mini-language.\n\
+                                   Example: \"{treeid:<15s},{elapse:>20f}\"\n\
      --                        Stop parsing options after this\n\
 "
 
 die() { echo -e "${prog}:" "$@"; exit 1; }
 warn() { echo -e "${prog}: warning:" "$@"; }
 dr_print() { echo -e "${prog}: dry-run:" "$@"; }
-
-#
-# Write the performance data into the given file.  Should be called
-# by the upper layer when the performance data record for each and every
-# Flux instance has been aggregated. Then, this is recursively invoked.
-#
-dump_perf() {
-    local blurb="${1}"
-    local out="${2}"
-
-    [[ "x${blurb}" = "x" ]] && die "empty json"
-
-    # Parse perf json for the given level into a record
-    local tuple=''
-    local length=0
-    tuple=$(echo "${blurb}" | jq "${perf_keys}") || die "invalid json"
-    length=$(echo "${blurb}" | jq ' .child | length ') || die "invalid json"
-
-    # Print the record for this level Didn't quote $tuple to remove new lines
-    printf "${perf_format}\n" ${tuple} >> "${out}"
-
-    # Recurse down to parse the record for each child Flux instance
-    for i in $(seq 0 $(( length - 1 )))
-    do
-        local new_o=''
-        new_o=$(echo "${blurb}" | jq " .child[${i}] ")
-        dump_perf "${new_o}" "${out}"
-    done
-}
-
 
 #
 # Roll up the performance records for each Flux instance to the KVS
@@ -210,23 +183,14 @@ rollup() {
     local prefix="${1}"
     local blurb="${2}"
     local out="${3}"
+    local num_children="${4}"
+    local format="${5}"
 
-    if [[ "${out}" != "%^+_no" ]]
-    then
-        # If output filename is given (e.g., top-level), dump the data
-        # to the file. Didn't quote $perf_heanding on purpose below.
-        rm -f "${out}"
-        printf "${perf_format}\n" ${perf_heading} > "${out}"
-        dump_perf "${blurb}" "${out}"
+    if [[ "${prefix}" == "${top_prefix}" &&  "${out}" != "%^+_no" ]]; then
+        flux tree-helper --perf-out="${out}" --perf-format="${format}" \
+             ${num_children} "tree-perf" <<< "${blurb}"
     else
-        if [[ "${prefix}" != "${top_prefix}" ]]
-        then
-            if [[ "x${FT_DRY_RUN}" = "xno" ]]
-            then
-                # Put it to KVS parent namespace except when we are at top level
-                flux --parent kvs put --raw "tree-perf=-" <<< "${blurb}" || warn "KVS error"
-            fi
-        fi
+        flux tree-helper ${num_children} "tree-perf" <<< "${blurb}"
     fi
 }
 
@@ -563,6 +527,7 @@ leaf() {
     local njobs="${5}"
     local cl="${6}"
     local perfout="${7}"
+    local format="${8}"
 
     # Begin Time Stamp
     local B=''
@@ -578,7 +543,7 @@ leaf() {
 
     o=$(jsonify "${prefix}" "${njobs}" "${nnodes}" "${ncores}" \
 "${ngpus}" "${B}" "${E}")
-    rollup "${prefix}" "${o}" "${perfout}"
+    rollup "${prefix}" "${o}" "${perfout}" "0" "${format}"
 }
 
 
@@ -605,8 +570,9 @@ rollup_exit_code() {
 }
 
 #
-# Submit the specified number of Flux instances at the next level
-# of the calling instance. Use flux-tree recursively.
+# Submit the specified number of Flux instances at the next level of the calling
+# instance. Use flux-tree recursively. Instances that have 0 jobs assigned are
+# not launched.
 #
 submit() {
     local prefix="${1}"
@@ -700,23 +666,15 @@ coll_perf() {
     local begin="${6}"
     local end="${7}"
     local perfout="${8}"
+    local nchildren="${9}"
+    local format="${10}"
 
     #
     # Make a JSON string from the performance data
     #
     local blurb=''
     blurb=$(jsonify "${prefix}" "${njobs}" "${nnodes}" "${ncores}" "${ngpus}" "${begin}" "${end}")
-    local child=''
-    local rank=0
-
-    for rank in $(seq 1 "${njobs}");
-    do
-        [[ "x${jobids[${rank}]}" = "x" || "x${FT_DRY_RUN}" = "xyes" ]] && break
-        child=$(flux job info ${jobids[${rank}]} guest.tree-perf)
-        blurb=$(echo "${blurb}" | jq " .child += [${child}]")
-    done
-
-    rollup "${prefix}" "${blurb}" "${perfout}"
+    rollup "${prefix}" "${blurb}" "${perfout}" "${nchildren}" "${format}"
 }
 
 
@@ -732,6 +690,7 @@ internal() {
     local ngpus="${8}"
     local njobs="${10}"
     local perfout="${13}"
+    local format="${14}"
 
     # Begin Time Stamp
     local B=''
@@ -743,8 +702,13 @@ internal() {
     local E=''
     E=$(date +%s.%N)
 
+    if [[ "x${FT_DRY_RUN}" = "xyes" ]]; then
+        nchildren=0
+    else
+        nchildren=${#jobids[@]}
+    fi
     coll_perf "${prefix}" "${nnodes}" "${ncores}" "${ngpus}" \
-"${njobs}" "${B}" "${E}" "${perfout}"
+"${njobs}" "${B}" "${E}" "${perfout}" "${nchildren}" "${format}"
 }
 
 
@@ -768,6 +732,7 @@ main() {
     local cl="${11}"           # jobscript commandline
     local flogs="${12}"        # flux log output option
     local out="${13}"          # perf output filename
+    local format="${14}"      # perf output format
     local size=0
 
     if [[ ${leaf} = "yes" ]]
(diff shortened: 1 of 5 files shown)


dongahn:
Great work @SteVwonder !

> After the introduction of the helper, for the same tree
topology, the perf results record a runtime of 34 seconds, and the
time utility reports a runtime of 47 seconds. So the helper not only
speeds up the post-processing time, but it also has a measurable impact
on the reported results as well.

Yeah. I believe this is because the overhead of an upper-level Flux instance in flux-tree includes the end-to-end overhead of its child instance, which includes `jq`-based JSON handling. 

Great that this python-based handler solves both performance and a scalability bug.

dongahn:
CI tests are either hung or failed at the flux-tree though.

src/cmd/flux-tree
```diff
@@ -165,43 +167,15 @@ Options:\n\
                                    (default=${FT_NJOBS})\n\
  -o, --perf-out=FILENAME       Dump the performance data into\n\
                                    the given file (default: don't print)\n\
+ --perf-format=FORMAT        Dump the performance data with the given\n\
+                                   format (uses python mini format syntax)\n\
```
dongahn: Like it!

But not every user of flux-tree would know what python mini format syntax. Maybe example or better yet a URL to python mini format?

src/cmd/flux-tree
```diff
@@ -210,23 +184,14 @@ rollup() {
...
     else
-        if [[ "${prefix}" != "${top_prefix}" ]]
-        then
-            if [[ "x${FT_DRY_RUN}" = "xno" ]]
-            then
-                # Put it to KVS parent namespace except when we are at top level
-                flux --parent kvs put --raw "tree-perf=-" <<< "${blurb}" || warn "KVS error"
-            fi
-        fi
+        flux tree-helper.py ${num_children} "tree-perf" <<< "${blurb}"
```
dongahn: Do we need `.py`? I vaguely remember the top flux command handles python-based sub-commands without `.py`...

src/cmd/flux-tree
```diff
@@ -633,6 +601,7 @@ submit() {
     [[ "x${log}" != "x%^+_no" ]] && F="--flux-logs=${log}"
     local S="${cl}"
     local rank=0
+    NUM_CHILDREN_SUBMITTED=0
```
dongahn: I see why you need this variable. But can this be avoided if you fetch the number of elements in the jobid array?

```bash
${#jobids[@]}
```

I add each jobid created from `flux-mini-submit` to an array.

src/cmd/flux-tree
```diff
@@ -819,6 +788,7 @@ while true; do
       -M|--match-policy)           FT_MATCH="${2}"; FT_INTER="yes";    shift 2 ;;
       -J|--njobs)                  FT_NJOBS=${2};                shift 2 ;;
       -o|--perf-out)               FT_PERF_OUT="${2}";           shift 2 ;;
+      --perf-format)               FT_PERF_FORMAT="${2}";       shift 2 ;;
```
dongahn: super minor knit. shift can be aligned with the previous line.

src/cmd/flux-tree-helper.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    Store a parsed version of the program's output format,
+    allowing the fields to iterated without modifiers, building
+    a new format suitable for headers display, etc...
+    """
+
+    #  List of legal format fields and their header names
+    headings = dict(
+        treeid="TreeID",
+        elapse="Elapsed",
```
dongahn: Previous I had units for some metrics (`secs`) in this case, was it intentional to remove the units?

src/cmd/flux-tree-helper.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    Get the jobids of num_children instances.  Will repeatedly query the
+    job-info module until num_children jobids are collected, with sleeps
+    inbetween queries.
+    """
+    jobids = set()
+    since = 0.0
+    logger.debug("Getting inactive children's IDs")
+    while True:
+        for job in flux.job.job_list_inactive(
```
dongahn: Does `since=0.0` give you `num_children` most recent job IDs? Or does it give you all of the jobs?

At the top level, this can be a problem when the submitting Flux instance runs jobs that don't belong to `flux-tree`

If former, there is a very slight race as another job can be submitted and run while `flux-tree` is running its jobs.

If latter, this will give you the list of jobs that ran before `flux-tree` was invoked.

dongahn:
Thanks @SteVwonder. Looks generally good. I added some inlined comments above. A big question is the correctness of using `job_list_inactive` at the top-level Flux when it isn't solely dedicated to `flux-tree`. It would also be good not to introduce `NUM_CHILDREN_SUBMITTED`. 

Other comments are super minor and/or questions.

SteVwonder:
> CI tests are either hung or failed at the flux-tree though.

Yeah, this PR depends on flux-framework/flux-core#2804, which hasn't been merged yet.  Due to the missing functions/classes in the python bindings, the CI will continue to fail until that gets in.  There is also an implicit dependence on https://github.com/flux-framework/flux-core/pull/2805 right now, so the helper was written with python3 in mind, which is probably the secondary source of failures (the flux-sched CI currently only uses python2).

src/cmd/flux-tree-helper.py
SteVwonder: > Does since=0.0 give you num_children most recent job IDs?

Yeah, it gives you up-to the last `num_children` recent job IDs.

> If former, there is a very slight race as another job can be submitted and run while flux-tree is running its jobs.

Yeah, that's a good point.  I guess we will have to find a way to pass the `jobids` bash array down into the helper script.  We obviously want to avoid using CLI arguments so that we don't run into the same issue that we had before.  We could include the child ids in the JSON that is sent to stdin, but we'd have to construct that JSON blob in a clever way to avoid the argument length issue.

It is a bit kludgy, but my current idea that doesn't involve CLI arguments is to send the jobids in directly (i.e., not formatted as JSON) via stdin.  The first line to stdin could be all the jobids (space delimited) and the second line (and beyond) could be the JSON perf blob. Since we control both ends of this simple "protocol", I don't think it would be fragile. Thoughts?

src/cmd/flux-tree-helper.py
SteVwonder: It was not intentional.  I'll add those back in. Thanks!

src/cmd/flux-tree
SteVwonder: Great suggestion! 

The URL is unfortunately quite long and doesn't fit within 80 characters with the indent level of the help text: https://docs.python.org/3/library/string.html#format-specification-mini-language

So I opted for an example as well as bringing the terminology exactly in-line with Python's for better googlability: `python mini format syntax` -> `python format specification mini-language`

SteVwonder:
> Thanks @SteVwonder. Looks generally good. I added some inlined comments above. A big question is the correctness of using job_list_inactive at the top-level Flux when it isn't solely dedicated to flux-tree. It would also be good not to introduce NUM_CHILDREN_SUBMITTED.
> Other comments are super minor and/or questions.

Thanks @dongahn for the review!  I left a response to your `job_list_inactive` comment inline above.  I'll wait to implement a fix to that until I hear back from you on the suggested solution.  We may need to iterate on that a bit.  I addressed your other comments/questions in a `fixup!` commit that I just pushed.  Note: that commit also removes the `perf_keys` and `perf_heading` variables declared at the top of `flux-tree` which are no longer used.

src/cmd/flux-tree
SteVwonder: Excellent!  Yeah, that is a much cleaner solution.  Thanks!

src/cmd/flux-tree-helper.py
dongahn: > It is a bit kludgy, but my current idea that doesn't involve CLI arguments is to send the jobids in directly (i.e., not formatted as JSON) via stdin. The first line to stdin could be all the jobids (space delimited) and the second line (and beyond) could be the JSON perf blob. Since we control both ends of this simple "protocol", I don't think it would be fragile. Thoughts?

Yes, this sounds good to me. `flux-tree-helper.py` is not public, so we can have a custom protocol. If you are worried about a long line that can be formed with lots of JOBIDs, you can even introduce custom delimiter like:  

```
[Perf]
{"elapse": 0.111, ...}
[JOBID]
19833399
19922900
```

grondo: Just a silly idea, but could you use the job name field to filter out jobs submitted by the current instance of `flux-tree`?

dongahn: Good idea, if you can create a "uniquely" identifiable name for each `flux-tree` invocation, you can solve this problem that way too.

dongahn: On a second thought: users can run `flux-tree --leaf` at the existing Flux instance? Then there isn't a good way to generate the name as we have no control over `flux mini submit` or similar issued by their scripts?

grondo: But at a leaf don't you know *all* jobs are flux-tree jobs?
Sorry, I perhaps shouldn't have spoken up since I don't know how flux-tree works exactly.

src/cmd/flux-tree
```diff
@@ -167,8 +163,10 @@ Options:\n\
                                    (default=${FT_NJOBS})\n\
  -o, --perf-out=FILENAME       Dump the performance data into\n\
                                    the given file (default: don't print)\n\
- --perf-format=FORMAT        Dump the performance data with the given\n\
-                                   format (uses python mini format syntax)\n\
+ --perf-format=FORMAT          Dump the performance data with the given\n\
+                                   format. Uses the python format\n\
+                                   specification mini-language.\n\
+                                   Example: \"{treeid:<15s},{elapse:>20f}\"\n\
```
dongahn: I am not so familiar with the python mini format. Is there a difference between `"{treeid:<15s},{elapse:>20f}" and "{treeid:<15s} {elapse:>20f}"? The latter is used in the code and the former is used in the example. If this should be so obvious to python programers, sorry for the comment.

dongahn:
@SteVwonder: `fixup!` commit looks good though I left one naive question.

src/cmd/flux-tree-helper.py
dongahn: Sorry -- You can type in `flux tree --leaf` on the existing Flux instance in which case the user script will be run on the exiting instance. This isn't a normal mode of operations though.

src/cmd/flux-tree
SteVwonder: > Is there a difference between `"{treeid:<15s},{elapse:>20f}" and "{treeid:<15s} {elapse:>20f}"?

Yeah. Everything outside the curly-braces is treated as a string literal.   Everything inside the curly-braces is treated as directive to the string formatter.  So the former example gets you comma-delimited output and the latter gets you space-delimited output.  

> The latter is used in the code and the former is used in the example.

Yeah. I did that to show some diversity in the usages and highlight that you can have more than just spaces in between.   Let me know if you would prefer the example to have a space rather than a comma.  I realize it may be a bit confusing or unclear.

src/cmd/flux-tree-helper.py
dongahn: > Sorry, I perhaps shouldn't have spoken up since I don't know how flux-tree works exactly.

The comment was actually great. This got me to think more. 

I may be mistaken, but the current code seems to create a performance JSON per each invocation of a job script at the leaf level. The original behavior was to create one per JSON for all the job script invocations at each leaf. @SteVwonder: Is this change intentional? I can see both side of the arguments but we should at least discuss...

src/cmd/flux-tree
dongahn: Awesome! It's good as is then.

src/cmd/flux-tree-helper.py
dongahn: On a second though (sorry -- again), we probably don't want to create a perf JSON per script invocation -- people like @jameshcorbett is doing 1 million jobs and we probably don't want 1 million JSONs...

SteVwonder: > Just a silly idea, but could you use the job name field to filter out jobs submitted by the current instance of flux-tree?

I like this idea, especially since the edit distance seems to be small from the code already in this PR, but the more I think about it, the more code I think it requires.

Is there still a race between when a script gets a `wait` response from the `job-manager` and when `job-info` serves up the job in the inactive listing?  If so, when we query `job-info` and get fewer jobs with the unique name than we were looking for, I think we won't be sure whether we need to just wait for the job to get added to the top of the `job-info` list, or if we need to search further back in the `job-info` inactive history.

It also isn't clear to me how to easily and efficiently search backwards through the job history using the inactive job listing.  You can always just increase the `max_entries` value, but that will get slow quickly. For this specific use-case, it would be handy to have the converse of `since` in the inactive job listing (i.e., get me N jobs that became inactive before this time).  In that case, you can just grab that earliest inactive time from the previous query and issue a new query with that time as the **before** value.  Not sure how useful that functionality would be generally.

Maybe I am just overthinking this and missing the obvious answer.  If not, maybe we go with the stdin solution for now and once the inactive job listing race is closed up and searching backwards is faster, we can switch to the job listing method?

grondo: > Is there still a race between when a script gets a wait response from the job-manager and when job-info serves up the job in the inactive listing? 

Yes. And there is also a race between the wait response and the `job.list-id` RPC, so you may need an alternate synchronization scheme or busywaiting in either case.

> It also isn't clear to me how to easily and efficiently search backwards through the job history using the inactive job listing. You can always just increase the max_entries value, but that will get slow quickly. 

How many jobs are you talking about? Do you really expect a lot of non-flux-tree based jobs you have to filter out?

There currently isn't a `start` timestamp at which to start searching backwards, but that is planned to be added afaik.

We had discussed improving the performance in another issue, perhaps @chu11 remembers which one. Maybe the real long term fix would be to allow filtering on the job name as well as states.

grondo: > In that case, you can just grab that earliest inactive time from the previous query and issue a new query with that time as the before value. Not sure how useful that functionality would be generally.

Yeah, as alluded to above, we already have plans for this.. I think it will be required in order to support a hard maximum for `max_entries` to avoid DoS of the job-info module.

SteVwonder: > so you may need an alternate synchronization scheme or busywaiting in either case.

Yeah, this PR currently has a "busywait" that sleeps and issues another query if the number of jobs returned is less than the number of children, to handle this race.  But if we have to handle other non-flux-tree jobs coming in, it isn't clear if the number mismatch is due to the race or other interleaved jobs.

> How many jobs are you talking about? Do you really expect a lot of non-flux-tree based jobs you have to filter out?

In a user-instance, I wouldn't expect it but I wouldn't rule it out either. In a multi-user instance, I would definitely expect the need for filtering out other jobs.  In the general case though, I think you are spot on:  most users will be running flux-tree in their own instance, so we can request exactly the number of inactive jobs as their are children and expect it to succeed.  We can add logic that increases the window size with each query to ensure we handle the case of interleaved jobs.

chu11: > There currently isn't a start timestamp at which to start searching backwards, but that is planned to be added afaik.

So I think the idea is, specify a timestamp and return "count" entries before that timestamp?  Wasn't on my current todo list.  But can definitely be added.  Unsure if it should be in the same RPC or another one, gotta think about that.  Would have to think of a way to make the performance not as horrible.  Maybe internally should be a skip list or something.

> We had discussed improving the performance in another issue, perhaps @chu11 remembers which one. 

I think you're thinking of the https://github.com/flux-framework/flux-core/issues/2776 discussion?  Which lead to https://github.com/flux-framework/flux-core/issues/2778?

> Maybe the real long term fix would be to allow filtering on the job name as well as states.

Shouldn't be too hard to add either.

SteVwonder: > Maybe the real long term fix would be to allow filtering on the job name as well as states.

Yeah, that would be awesome! (and if I'm being greedy, filtering based on arbitrary user attributes within the jobspec).

chu11: An arbitrary filter is probably a bit much, requiring us to develop a query language of sorts?  That's probably a whole other can of worms

grondo: Maybe when we have a sqlite backend for job-info. We'll need this eventually I'm sure, but not in the near future.

> So I think the idea is, specify a timestamp and return "count" entries before that timestamp? Wasn't on my current todo list. But can definitely be added. Unsure if it should be in the same RPC or another one, gotta think about that. Would have to think of a way to make the performance not as horrible. Maybe internally should be a skip list or something.

Yeah, I thought this was how we had proposed implementing a restartable RPC.
We'll probably need this to avoid DoS of the job-info module.

chu11: > Yeah, I thought this was how we had proposed implementing a restartable RPC.

Ahh, I did not mentally equate this that much to the restartable RPC, but you are right, it's far more similar than I had originally thought.

dongahn:
@SteVwonder: what still remains to be done? From my perspective, adding job IDs to `stdin` and not generating performance data for each jobscript invocation should do. Do you plan to tackle this some other ways? Note that @jameshcorbett will need this soon.

SteVwonder:
> Do you plan to tackle this some other ways?

@dongahn: just pushed some new commits that use the new job listing name filter to solve this issue.  We just need to land https://github.com/flux-framework/flux-core/pull/2818 and https://github.com/flux-framework/flux-core/pull/2804 (and transitively https://github.com/flux-framework/flux-core/pull/2805) before this can be merged.  The Travis CI will fail until these are merged.

I also squashed in that fixup! commit.

src/cmd/flux-tree
```diff
@@ -167,6 +168,7 @@ Options:\n\
                                    format. Uses the python format\n\
                                    specification mini-language.\n\
                                    Example: \"{treeid:<15s},{elapse:>20f}\"\n\
+ --job-name=NAME               Name to use when submitting child jobs\n\
```
dongahn: Minor nit: would be good to align the long option to other long options. I see the same issue with `--perf-format` as well.

dongahn:
Thanks @SteVwonder. Pretty neat. Because you are filtering jobs by their names, you won't collect the performance data from those jobs submitted by the user-provided job script right? 

I left one minor inline comment. And I also noticed you have long lines that exceed 80 characters, I would break them unless that significantly decrease readability or disallow you from passing checks like "blacken".

SteVwonder:
> Because you are filtering jobs by their names, you won't collect the performance data from those jobs submitted by the user-provided job script right?

I may be misunderstanding or misremembering the control flow (pre-apologies if so), but I believe the flux-tree-helper's spawned by flux-tree leaves are given a `num_children` of 0, so they shouldn't even poll for inactive jobs.  They should just "consume" the json performance blob from stdin and dump it into their parent's KVS.

SteVwonder:
> And I also noticed you have long lines that exceed 80 characters, I would break them unless that significantly decrease readability or disallow you from passing checks like "blacken".

Yep, sorry about that.  I at one point thought the line breaks were causing an issue and late at night forgot to add them back in.  Thanks for the reminder!  I'll go through and rectify that now.

SteVwonder:
Lines >80 characters should be taken care of now.  For the flux-tree-helper.py, I just used whatever py-black produced.  (I think py-black allows a little extra leeway with line length, 88 characters?).

SteVwonder:
@mergifyio rebase

SteVwonder:
@mergifyio rebase

SteVwonder:
Looks like the previous travis failure was just an internal Travis hiccup.  Forcing a rebase cleared everything up.

@dongahn, I believe this PR is ready for merging (unless you or anyone else want to give it a second pass).

dongahn:
This looks great. Significantly enhancement to flux-tree. Thanks @SteVwonder!

PR #622, opened 2020-03-11
flux-tree: add rundir support

dongahn:
This PR adds `--flux-rundir` support to `flux-tree`.

Problem: For a large flux-tree test where lots of flux instances are used, the content store back-ends can use lots of memory leading to OOM.

Provide `--flux-rundir=DIR` option which sets the rundir to DIR so that the content stores can be redirected to a subdirectory in DIR for this case. Use the rundir attribute because upon the exit of each Flux instance, its content store sqlite file is removed.

Add two test cases that uses `--dry-run`

### DIFF
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -37,6 +37,7 @@ FT_PERF_FORMAT='{treeid:<15s} {elapse:>20f} {begin:>20f} {end:>20f} {match:>15f}
 '{njobs:>15d} {my_nodes:>5d} {my_cores:>4d} {my_gpus:>4d}'
                                # store perf-out with format given by --perf-format
 FT_FXLOGS='%^+_no'             # dir in which flux logs are produced
+FT_FXRDIR='%^+_no'             # flux rundir attribute to pass
 FT_LEAF='no'                   # is this a leaf instance (given by --leaf)?
 FT_G_PREFIX='tree'             # store hierarchical path given by --prefix
 FT_DRY_RUN='no'                # --dry-run for testing?
@@ -81,11 +82,10 @@ declare -A q_params=(          # make sure to update this when
 )
 declare -a jobids             # array to store a set of submitted job IDs
 
-declare -r long_opts='help,leaf,flux-logs:,nnodes:,ncores-per-node:'\
-',ngpus-per-node:,topology:,queue-policy:,queue-params:,match-policy:'\
-',njobs:,perf-out:,perf-format:,prefix:,job-name:,dry-run'
-
-declare -r short_opts='hlf:N:c:g:T:Q:P:M:J:o:X:d'
+declare -r long_opts='help,leaf,flux-logs:,flux-rundir:,nnodes:'\
+',ncores-per-node:,ngpus-per-node:,topology:,queue-policy:,queue-params:'\
+',match-policy:,njobs:,perf-out:,perf-format:,prefix:,job-name:,dry-run'
+declare -r short_opts='hlf:r:N:c:g:T:Q:P:M:J:o:X:d'
 declare -r prog=${0##*/}
 declare -r usage="
 Usage: ${prog} [OPTIONS] -- Jobscript\n\
@@ -146,6 +146,9 @@ Options:\n\
                                with internal tree-node options like -T.\n\
                                    (default=${FT_LEAF})\n\
  -f, --flux-logs=DIR           Dump Flux logs for all instances into DIR\n\
+ -r, --flux-rundir=DIR         Set the rundir attribute of each Flux tree instance\n\
+                                   into a subdirectory within DIR. The content\n\
+                                   stores will be redirected to them as well\n\
  -N, --nnodes=NNODES           Total num of nodes to use\n\
                                    (default=${FT_NNODES})\n\
  -c, --ncores-per-node=NCORES  Total num of cores per node to use\n\
@@ -590,7 +593,7 @@ submit() {
     local njobs="${10}"
     local cl="${11}"
     local log="${12}"
-    local out="${13}"
+    local rdir="${13}"
 
     # Flux instance rank-agnostic command-line options for the next level
     local T="${nx_topo:+--topology=${nx_topo}}"
@@ -601,6 +604,8 @@ submit() {
     local F=''
     [[ "x${log}" != "x%^+_no" ]] && F="--flux-logs=${log}"
     local S="${cl}"
+    local R=''
+    [[ "x${rdir}" != "x%^+_no" ]] && R="--flux-rundir=${rdir}"
     local rank=0
 
     # Main Loop to Submit the Next-Level Flux Instances
@@ -621,8 +626,20 @@ submit() {
         local o=''
         if [[ x"${log}" != "x%^+_no" ]]
         then
-            mkdir -p "${log}"
-            o="-o,-S,log-filename=${log}/${prefix}.${rank}.log"
+            if [[ "x${FT_DRY_RUN}" != "xyes" ]]
+            then
+                mkdir -p "${log}"
+            fi
+            o="-o,-Slog-filename=${log}/${prefix}.${rank}.log"
+        fi
+        if [[ x"${rdir}" != "x%^+_no" ]]
+        then
+            if [[ "x${FT_DRY_RUN}" != "xyes" ]]
+            then
+                rm -rf "${rdir}/${prefix}.${rank}.pfs"
+                mkdir -p "${rdir}/${prefix}.${rank}.pfs"
+            fi
+            o="${o:+${o} }-o,-Srundir=${rdir}/${prefix}.${rank}.pfs"
         fi
         local N=0
         N=$(get_my_nodes "${ncores}" "${my_cores}")
@@ -646,7 +663,7 @@ submit() {
         jobid=$(\
 flux mini submit --job-name=${FT_JOB_NAME} -N${N} -n${N} -c${c} ${G} \
      flux start ${o} \
-     flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${X} ${J} -- ${S})
+     flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${R} ${X} ${J} -- ${S})
         jobids["${rank}"]="${jobid}"
     done
 
@@ -693,8 +710,8 @@ internal() {
     local ncores="${7}"
     local ngpus="${8}"
     local njobs="${10}"
-    local perfout="${13}"
-    local format="${14}"
+    local perfout="${14}"
+    local format="${15}"
 
     # Begin Time Stamp
     local B=''
@@ -735,8 +752,9 @@ main() {
     local njobs="${10}"        # num of jobs assigned to this Flux instance
     local cl="${11}"           # jobscript commandline
     local flogs="${12}"        # flux log output option
-    local out="${13}"          # perf output filename
-    local format="${14}"      # perf output format
+    local frdir="${13}"        # flux rundir attribute
+    local out="${14}"          # perf output filename
+    local format="${15}"      # perf output format
     local size=0
 
     if [[ ${leaf} = "yes" ]]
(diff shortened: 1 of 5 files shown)


dongahn:
OK. This can be reviewed now. I know this will conflict with #621 work, but this is also required by @jameshcorbett's work soon.

src/cmd/flux-tree
```diff
@@ -146,6 +147,9 @@ Options:\n\
                                with internal tree-node options like -T.\n\
                                    (default=${FT_LEAF})\n\
  -f, --flux-logs=DIR           Dump Flux logs for all instances into DIR\n\
+ -r, --flux-rundir=DIR         Dump the rundir attribute of each Flux tree instance\n\
```
SteVwonder: Does this also dump the value of the attribute into the DIR?  Or just set the attribute?

src/cmd/flux-tree
```diff
@@ -621,7 +625,8 @@ submit() {
     local njobs="${10}"
     local cl="${11}"
     local log="${12}"
-    local out="${13}"
+    local rdir="${13}"
+    local out="${14}"
```
SteVwonder: I don't think `out` is used in this function at all since the children don't use the perf-out value at all.  If it is not being used, I believe it can just be dropped from this function header part (removing it as an argument entirely is harder given this is passed "$@").

src/cmd/flux-tree
dongahn: This just sets the attribute.

src/cmd/flux-tree
dongahn: Good finding. I will remove `out` but I would like to keep "$@" at the callee per its brevity.

dongahn:
> Does flux automatically clean up the run directories after it closes? If it is a clean shutdown, I am pretty sure it does. If not (or in the case of an instance crash leaving them behind), we may need to cleanup the rundirs so that the PFS isn't littered with content stores from all the instances.

This would be a question for @grondo or @garlick. I am sure the content store is removed (and that's because I used `rundir` attribute), but I am unsure about other data in there even for a clean exit.

Regardless, it seems flux-tree should remove the `rundir` by itself.

src/cmd/flux-tree
SteVwonder: Ok, minor nit then:
```suggestion
 -r, --flux-rundir=DIR         Set the rundir attribute of each Flux tree instance\n\
```

SteVwonder:
Looks like the broker does clean up after itself even if the initial program errors out, but if the broker is force-ably killed or crashes, the run directory gets left behind with the content store etc.   If the root instance dies, there isn't much we can do, but if a child in the tree crashes, should we have the root try and clean up after it?

dongahn:
> On the one hand, those content stores could be quite larger, on the other hand, they might have useful debugging info.

Now that I think about this more, `flux-tree` should not try to change the cleanup semantics of flux instances being used. If leaving large contents around on flux error becomes a problem (compared to debuggability benefits), this should be handled by flux broker itself (for example, by installing a signal handler on anomalous signals and removubg rundirs as part of it). It seem best if `flux-tree` remains to be a convenience utility layer that respects the semantics of flux instances and schedulers of all the Flux instances it employs as much as possible.

SteVwonder:
Ok.  I think this PR is ready to go in once the two inline comment threads above are resolved.

SteVwonder:
> Now that I think about this more, flux-tree should not try to change the cleanup semantics of flux instances being used. If leaving large contents around on flux error becomes a problem

It looks like it does pose a problem for back-to-back runs of `flux-tree`.  After running once with the `--rundir` option, all subsequent runs fail until you delete the rundir subdirectories:

```
herbein1@opal63 ~/workspace/flux-tree-rundir-test
❯ ll                                                   
total 0
                                                                                                                                                   
herbein1@opal63 ~/workspace/flux-tree-rundir-test
❯ flux tree --topology=2x2 -r $PWD -J 4 -f $PWD/logs -o $PWD/perf.out -- flux mini run -n1 -c1 hostname
                                                                                                                                                   
6s herbein1@opal63 ~/workspace/flux-tree-rundir-test
❯ ll
total 12K
drwxr-sr-x 2 herbein1 herbein1 4.0K Mar 27 19:55 logs
-rw-r--r-- 1 herbein1 herbein1  509 Mar 27 19:55 perf.out
drwxr-sr-x 3 herbein1 herbein1 4.0K Mar 27 19:55 tree.1.pfs
                                                                                                                                                   
herbein1@opal63 ~/workspace/flux-tree-rundir-test
❯ flux tree --topology=2x2 -r $PWD -J 4 -f $PWD/logs -o $PWD/perf.out -- flux mini run -n1 -c1 hostname
flux-job: task(s) exited with exit code 1
flux-broker: create_broker_rundir: mkdir (/g/g0/herbein1/workspace/flux-tree-rundir-test/tree.1.pfs/0): File exists
flux-broker: bootstrap failed
flux-job: job 10590835703808 id or key not found
```

I am not sure what we should do here (if anything at all).  One thought is to have `flux-tree` check for existing subdirectories in the rundir matching the `$ID.$RANK.pfs` format and delete them (maybe guard this functionality behind a `-f, --force` flag).  Another thought is to have flux-tree check for the existing directories and then print an error before launching any children.  My final thought would be to leave the behavior as is.

dongahn:
> I am not sure what we should do here (if anything at all). One thought is to have flux-tree check for existing subdirectories in the rundir matching the $ID.$RANK.pfs format and delete them (maybe guard this functionality behind a -f, --force flag). Another thought is to have flux-tree check for the existing directories and then print an error before launching any children. My final thought would be to leave the behavior as is.

Because this can only happen when `--flux-rundir=DIR` is given, maybe the top-level `flux-tree` should delete the entire DIR if it exists. Seems introducing another `-f` option is an overkill when `--flux-rundir=DIR` is unlikely to be used by default?

SteVwonder:
> maybe the top-level flux-tree should delete the entire DIR if it exists

I am hesitant to have the default behavior be to delete the entire directory pointed to by `--flux-rundir`.  For example, if I run flux tree with `--flux-rundir=/tmp/herbein1`, I wouldn't want the entirety of `/tmp/herbein1` to be wiped out by default.

> Seems introducing another -f option is an overkill when --flux-rundir=DIR is unlikely to be used by default?

Sure, if we can eliminate any collateral from the default behavior, then avoiding the extra argument sounds good to me.

dongahn:
> I am hesitant to have the default behavior be to delete the entire directory pointed to by --flux-rundir. For example, if I run flux tree with --flux-rundir=/tmp/herbein1, I wouldn't want the entirety of /tmp/herbein1 to be wiped out by default.

Yeah... I should have said the top level `rm -rf /tmp/herbein1/$prefix`. 

But as I see the code, it seems a fine grained approach would be a bit safer:

```diff
diff --git a/src/cmd/flux-tree b/src/cmd/flux-tree
index c3a9bb58..0cd50484 100755
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -635,6 +635,7 @@ submit() {
         then
             if [[ "x${FT_DRY_RUN}" != "xyes" ]]
             then
+                rm -rf "${rdir}/${prefix}.${rank}.pfs"
                 mkdir -p "${rdir}/${prefix}.${rank}.pfs"
             fi
             o="${o:+${o} }-o,-Srundir=${rdir}/${prefix}.${rank}.pfs"
```

dongahn:
@SteVwonder: I added the requested changes and also fixed Issue #635. I squashed all of the interim commits and forced a push. Thanks!

src/cmd/flux-tree
```diff
@@ -146,6 +146,9 @@ Options:\n\
                                with internal tree-node options like -T.\n\
                                    (default=${FT_LEAF})\n\
  -f, --flux-logs=DIR           Dump Flux logs for all instances into DIR\n\
+ -r, --flux-rundir=DIR         Set the rundir attribute of each Flux tree instance\n\
```
SteVwonder: One thought I had during my recent test run: if this is a command that we expect most people won't use, do we want to reserve a single-char for it (`-r`) or just use the double-dash long arg?  I don't have strong feelings one way or another, but wanted to at least bring it up.

dongahn:
> EDIT: to be more precise with my wording, it looks like the top-level flux-tree correctly submits instances with the rundir set, but I don't think flux-tree is recursively calling itself with the --flux-rundir argument because the leaves in a 1x1 topology revert to the flux default rundir (/var/tmp/herbein1).


Bad... Sorry. I was following the way I propagated --flux-log... Probably a silly mistake.

dongahn:
OK. I missed passing additional option to nested flux-trees as you suspected. Thanks a lot for catching this before the PR gets merged.

```diff
[gingerfoot:flux-sched/src/cmd] ahn1% git diff
diff --git a/src/cmd/flux-tree b/src/cmd/flux-tree
index 3e415dd5..ec6a476e 100755
--- a/src/cmd/flux-tree
+++ b/src/cmd/flux-tree
@@ -601,6 +601,8 @@ submit() {
     local M="${nx_match:+--match-policy=${nx_match}}"
     local F=''
     [[ "x${log}" != "x%^+_no" ]] && F="--flux-logs=${log}"
+    local R=''
+    [[ "x${rdir}" != "x%^+_no" ]] && R="--flux-rundir=${rdir}"
     local rank=0

     # Main Loop to Submit the Next-Level Flux Instances
@@ -655,10 +657,11 @@ submit() {
             dr_print ""
             continue
         fi
         jobid=$(\
 flux mini submit --job-name=${FT_JOB_NAME} -N${N} -n${N} -c${c} ${G} \
      flux start ${o} \
-     flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${X} ${J} \
+     flux tree -N${N} -c${c} ${G} ${T} ${Q} ${P} ${M} ${F} ${R} ${X} ${J} \
      -- "${FT_CL[@]}")
         jobids["${rank}"]="${jobid}"
     done
```  

```console
ahn1@edd7ea29bc50:/usr/src/rundir_test$ flux tree -f $PWD --flux-rundir $PWD -N 1 -c 1 -T 1x1 -- bash -c 'flux lsattr -v | grep rundir > test.out'
ahn1@edd7ea29bc50:/usr/src/rundir_test$ cat test.out
broker.rundir                           /usr/src/rundir_test/tree.1.1.pfs/0
content.backing-path                    /usr/src/rundir_test/tree.1.1.pfs/content.sqlite
local-uri                               local:///usr/src/rundir_test/tree.1.1.pfs/0/local
log-filename                            /usr/src/rundir_test/tree.1.1.log
parent-uri                              local:///usr/src/rundir_test/tree.1.pfs/0/local
rundir                                  /usr/src/rundir_test/tree.1.1.pf
```

```console
ahn1@edd7ea29bc50:/usr/src/rundir_test$ ll *.pfs
tree.1.1.pfs:
total 0
drwx------ 2 ahn1 ahn1 64 Mar 31 05:09 0

tree.1.pfs:
total 0
drwx------ 2 ahn1 ahn1 64 Mar 31 05:09 0
```

src/cmd/flux-tree-helper.py
```diff
@@ -194,7 +198,8 @@ def main():
     logger.debug("Combining data")
     combined_data = combine_data(this_data, child_data)
     logger.debug("Writing data to parent's KVS")
-    write_data_to_parent(flux_handle, args.kvs_key, combined_data)
+    if flux_handle != None:
```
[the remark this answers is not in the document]
dongahn: Good point. Will change.

src/cmd/flux-tree
dongahn: Let's leave it this way since -f is also there for --flux-logs. If we push short option letter space, we can drop the short options from both.

dongahn:
@SteVwonder: Ok. I forced a push. I squashed all of the interim changes. Unless I made late night mistakes, the only new review needed from would be:

https://github.com/flux-framework/flux-sched/pull/622/commits/7c40d0531fe6eef73aec51c6e980898eb52b3e6a

https://github.com/flux-framework/flux-sched/pull/622/commits/7d7040dd9926ffffa17c07617d966d16155a3ccd

https://github.com/flux-framework/flux-sched/pull/622/commits/3e4e0f3292e56a4c568cc7eb6514fd7b93be3902

SteVwonder:
Excellent cleanups @dongahn!  The quoted jobscripts, multi-level rundirs, and back-to-back executions all work flawlessly for me now!

Setting the label.

dongahn:
Thanks @SteVwonder for your thorough review! I will go ahead and install a version on LC systems for @jameshcorbett to start his experiments!

Issue #597, opened 2020-02-14
Replace sched-simple with qmanager as a test for state reconstruction


dongahn:
https://github.com/flux-framework/flux-sched/pull/593/commits/1a2ee8e7d7865c0bb60eca532d218618c8b57b43

When Issue #470 is implemented, use the `qmanager-reload` test as a verification.

dongahn:
I just realized this was a wrong trigging.

PR #625, opened 2020-03-19
Add basic state recovery support

### LINKED ISSUE #470: Add state reconstruction support to resource module

dongahn:
This PR implements the basic state recovery semantics.

Recover the state of both the running job queue and the resource graph data store for those running jobs. Meet the state-recovery requirement for a key near-term use case in which both `resource` and `qmanager` modules are shut down and reloaded in a coordinated fashion. Assume two conditions for full recovery:

-  The RV1 JSON object for each and every currently running job includes the `scheduling` key, therefore the JGF section.
- The `resource` module is reloaded before `qmanager`.

Lay some additional work to prepare for more comprehensive failure recovery in the future. If only the `qmanager` module is shut down and reloaded, still fully recover the state. If both qmanager and resource are reloaded, but the RV1 JSON  of a running job does not include the `scheduling` key, `qmanager` will fail to load because  it can't recover the resource state. However, if `resource-recovery-on-load=false` is given as a qmanager load option, make the following trade off:

- 😄 `qmanager` will load and schedule
- 😄 `qmanager` will allow clean FREE of the running jobs when they complete  (i.e., those jobs that were already  running on reload).
- 😬 But the resources that those running jobs are using can be scheduled to other jobs (i.e., overscheduling) because the resource module won't be able to reconstruct the state.

[Table 1] Current flux-sched state recovery semantics
| resource  | qmanager | JGF  | no-JGF |
|----------|-----------|------|--------|
|    reload   | reload          | full     | partial  |
|    running |  reload        | full     | full         |
|    reload    | running      | no      | no.         |

Resolve #470


dongahn:
This PR is built on PR #597 so until PR #597 is merged, it will contain the previous commits.

dongahn:
A new test failed on one of the Travis test instance:

```
FAIL: t4009-match-update.t 6 - update: resource.update works with R1
1636FAIL: t4009-match-update.t 7 - update: works on an existing jobid with same R
1637FAIL: t4009-match-update.t 8 - update: resource.update works with R2
1638FAIL: t4009-match-update.t 9 - update: resource.update works with R3
1639FAIL: t4009-match-update.t 10 - update: resource.update works with R4
```

Need to address this before taking off the WIP tag.

dongahn:
@Mergifyio rebase

dongahn:
OK. I believe the random CI failures were due to 841e3b4. I ran CI 3 times and all subtests pass. This should be ready for your review.

dongahn:
I will close this and post a new one that has been rebased with the current master.

PR #630, opened 2020-03-25
config: track flux-core developments in config file handling

garlick:
This PR should be merged after flux-framework/flux-core#2866.  It tracks changes proposed in that PR:

* Use new `flux_get_conf()` prototype.
* Don't call `flux_get_conf()` early with check for TOML parse errors (broker does that).
* Don't install a default `qmanager.toml` because doing so could overwrite a user's config file in an unpackaged install environment.
* Allow qmanager to load without a `qmanager.toml` (using default or command line settings).

Plus
* Fix `qmanager` config test so that it uses the in-tree module in all cases by using specialized rc1/rc3 files instead of the system ones.  (Otherwise the installed qmanager might fail on a config input before test script has a chance to reload the in-tree one).
* Fix `qmanager` handling of job manager handshake failure (which I ran into while trying to find the minimum set of qmanager/job-manager dependencies)


t/t1005-qmanager-conf.t
```diff
@@ -21,178 +21,197 @@ then
 fi
 conf_base=${SHARNESS_TEST_SRCDIR}/conf.d
 
+# Run broker with specified config file and qmanager options.
+# Usage: start_qmanager config-path [module options] >outfile
+start_qmanager () {
+    local config=$1; shift
+    QMANAGER_OPTIONS=$* flux broker \
+	-Sbroker.rc1_path=./rc1 -Sbroker.rc3_path=./rc3 \
```
SteVwonder: If we are specifying the rc paths at the launch line, do we still need the `FLUX_RC_EXTRA` line up above?  Or can that be tossed?

t/t1005-qmanager-conf.t
```diff
@@ -21,178 +21,197 @@ then
...
+	--config-path=${config} \
+	flux dmesg
+}
+start_qmanager_noconfig () {
+    QMANAGER_OPTIONS=$* flux broker \
+	-Sbroker.rc1_path=./rc1 -Sbroker.rc3_path=./rc3 \
+	flux dmesg
+}
+# Usage: check_enforced_policy outfile expected
+check_enforced_policy() {
```
SteVwonder: Very nice cleanup with this convenience function!

t/t1005-qmanager-conf.t
garlick: Er, I was not paying attention to this block at all:
```
unset FLUX_RESOURCE_RC_NOOP
unset FLUX_QMANAGER_RC_NOOP
unset FLUX_CONF_DIR &&
if test -z "${FLUX_SCHED_TEST_INSTALLED}" || test -z "${FLUX_SCHED_CO_INST}"
then
    export FLUX_RC_EXTRA="${SHARNESS_TEST_SRCDIR}/rc"
fi
```
It seems like all of it could go away maybe?

garlick: Just pushed a couple commits:
* undo FLUX_CONF_DIR setting in sched-sharness.sh, and the line above that unsets it
* fixup commit for 6d504e7 that removes the rest of it

(Good comment, I hadn't noticed the FLUX_CONF_DIR setting in the sharness script!)

grondo:
@garlick, want to squash the fixup commit and force a CI rebuild since flux-framework/flux-core#2866 is now in?

grondo:
Strangely, I'm not seeing flux-framework/flux-sched on either travis-ci.org or travis-ci.com anymore. I wonder if the update of flux-core to travis-ci.com actually updated the whole organization.

grondo:
> I'm not seeing flux-framework/flux-sched on either travis-ci.org or travis-ci.com anymore. 

Eh, nevermind, Travis just lost my organization memberships momentarily and was only showing my personal repos.

garlick:
Squashed.

grondo:
Travis failing due to a broken &&-chain:
```
rror: bug in the test script: broken &&-chain: 
    conf_name="15-no-qmanager" &&
    outfile=${conf_name}.out &&
    start_qmanager ${conf_base}/${conf_name} >${outfile}
    check_queue_params ${outfile} \
	"queue-depth=8192,max-queue-depth=1000000" &&
    check_policy_params ${outfile} \
	"reservation-depth=64,max-reservation-depth=100000"
```

grondo:
I set merge-when-passing, so once the chain-lint error is fixed up, this will go in.

garlick:
Whoopsie!  Fixed.  Thanks!

qmanager/modules/qmanager.cpp
```diff
@@ -366,11 +365,9 @@ static int enforce_queue_policy (std::shared_ptr<qmanager_ctx_t> &ctx)
 
     if (handshake_jobmanager (ctx) < 0) {
         flux_log_error (ctx->h, "%s: handshake_jobmanager", __FUNCTION__);
-        goto out;
+        return -1;
```
dongahn: Thanks for the fix! 

In terms of style, I sort of like one where a function returns from one site. The more complex a function, the more readable that way...

I see this change add `return -`1 to each error handling branch. Instead, can this commit only contains the following?

```c++
    if ( (rc = handshake_jobmanager (ctx)) < 0) {
         flux_log_error (ctx->h, "%s: handshake_jobmanager", __FUNCTION__);		         flux_log_error (ctx->h, "%s: handshake_jobmanager", __FUNCTION__);
         goto out;
```

dongahn:
Oops. I just see this is merged. I think I was too late to start reviewing this sorry about that. FWIW, I just added one thing that I caught. Not a big deal.

grondo:
Sorry @dongahn! We had just merged a breaking change in flux-core and didn't want flux-sched master to be broken for too long.

dongahn:
Absolutely. Not a problem on my part!

garlick:
Sorry about that @dongahn.

Issue #639, opened 2020-04-08
resource: JGF reader not adding elements to by_path map


milroy:
I've come across a problem when generating a resource graph with the JGF reader.  When the JGF reader is selected as the load format in the Flux resource module or in `resource-query` the `by_path` map in the resource graph metadata is not populated.  Here's an example where I modified `resource-query` to print the size of `by_path`:

```
$ resource-query -L ~/flux-sched/t/data/resource/jgfs/tiny.json -f jgf
Number of keys in by_path map: 0
INFO: Loading a matcher: CA
resource-query> quit
$ resource-query -L ~/flux-sched/t/data/resource/grugs/tiny.graphml -f grug
Number of keys in by_path map: 100
INFO: Loading a matcher: CA
resource-query> 
```

The cause is missing logic in `resource_reader_jgf_t::add_graph_metadata`:
 https://github.com/flux-framework/flux-sched/blob/280863295b335b2a005b315b78490fe8ff540b5d/resource/readers/resource_reader_jgf.cpp#L257
The loop should insert a key-value pair unless the second `if` condition is satisfied.  After making the modification and recompiling:

```
$ resource-query -L ~/flux-sched/t/data/resource/jgfs/tiny.json -f jgf
Number of keys in by_path map: 100
INFO: Loading a matcher: CA
resource-query>
```

milroy:
I should add that this behavior occurs with the latest flux-sched master.

dongahn:
Great catch! The patch sounds good to me.

PR #641, opened 2020-04-09
Add path, vertex pair to by_path map

milroy:
Problem: when loading a JGF graph, the JGF reader does not insert key-value pairs into the `by_path` map in `resource_graph_metadata_t` (issue #639).

This patch populates the map in the `add_graph_metadata` member function as intended.

### DIFF
--- a/t/t6000-graph-size.t
+++ b/t/t6000-graph-size.t
@@ -5,6 +5,9 @@ test_description='Test Correctness of Populating Graph Data Store'
 . $(dirname $0)/sharness.sh
 
 tiny_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+tiny_jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny.json"
+exclusive_001N_hwloc="${SHARNESS_TEST_SRCDIR}/data/hwloc-data/001N/exclusive\
+/04-brokers/0.xml"
 large_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/sierra.graphml"
 query="../../resource/utilities/resource-query"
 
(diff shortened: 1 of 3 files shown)


dongahn:
Thanks @milroy. LGTM. Is there anyway we can add a test case for this? If that's too difficult at this point, I will merge this in as is.

milroy:
I could add a test that checks for the `by_map` size on JGF load, but that would require additional console output.  I could add a debug flag to the `resource_query` initialization options so it doesn't print the info by default.

dongahn:
Thanks @milroy. Actually, the #entry of these by_* index tables would be a first augmentation under elapse time flag...

https://github.com/flux-framework/flux-sched/blob/master/resource/utilities/resource-query.cpp#L492

milroy:
> https://github.com/flux-framework/flux-sched/blob/master/resource/utilities/resource-query.cpp#L492

That looks like the right place for the output.  I'll add the `by_*` size outputs there and corresponding tests.

milroy:
I've added the printout and tests.  This is ready for you to look at again @dongahn.

t/t6000-graph-size.t
```diff
@@ -5,6 +5,8 @@ test_description='Test Correctness of Populating Graph Data Store'
 . $(dirname $0)/sharness.sh
 
 tiny_grug="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+tiny_jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny.json"
+exclusive_001N_hwloc="${SHARNESS_TEST_SRCDIR}/data/hwloc-data/001N/exclusive/04-brokers/0.xml"
```
dongahn, line 9: Would be good to wrap the long line at 80 chars if possible.

dongahn: Same comment for a few long lines below as well.

dongahn:
@milroy: Thank you for doing this. I know you are working on a your busy research schedule! I just made a few aesthetic comments. Once they are addressed, this should go in.

t/t6000-graph-size.t
milroy: The product of thoughtless pasting.  Sure, I'll fix these right away.

milroy:
No problem!  I just pushed a fix that addresses your request.  If that looks good let me know and I'll squash the commit.

dongahn:
LGTM. Please ping me once this is squashed. I will merge.

milroy:
I squashed the fixup- once the tests pass the PR's ready to be merged.  Thanks for the review @dongahn!

PR #644, opened 2020-04-19
Introduce generic exception handling wrapper support

dongahn:
This PR adds generic C++ exception handling wrapper support.

This is generic in that this will enable us to wrap functions of many types
and provide exception safety to those functions with minor efforts.
Specifically, this is needed to wrap `mod_main` of `qmanager`,
which is a part of this PR, and more importantly the schedutil
callbacks within `qmanager`, which will be a part of a future PR.

The wrappers uses perfect forwarding to wrap the target functor
of many different types and add an exception
handling semantics common for Flux. Essentially, map
certain C++ exceptions classes to `errno`s as they are caught.

Note that this won't be used to wrap "native" request handlers as
it doesn't do flux respond epilogues like what @trws's PR #542 does.
Wrapping native request handlers will need an interface like #542.
In fact, I will continue to grow `c++wrappers` directory to have
a more specialized wrapper like it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,6 +155,8 @@ AC_CONFIG_FILES([Makefile
   src/common/libtap/Makefile
   src/common/libutil/Makefile
   src/common/librbtree/Makefile
+  src/common/c++wrappers/Makefile
+  src/common/c++wrappers/test/Makefile
   src/cmd/Makefile
   resource/Makefile
   resource/planner/Makefile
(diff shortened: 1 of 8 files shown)


dongahn:
OK. Travis is happy. This should be ready for your review.

dongahn:
Since PR #646 (a superset of this PR) is being reviewed instead, I'm closing this one.

PR #646, opened 2020-04-25
qmanager: add exception safety for schedutil callbacks

dongahn:
This PR uses the `eh_wrapper_t` class to add
exception safety for schedutil callbacks.

- Refactor callbacks for `schedutil` into two classes
   in new header and source files:
   `qmanager_cb_t` contains the existing code;
   `qmanager_safe_cb_t` wraps it with `eh_wrapper_t`.

-  Avoid interdepedency between `qmanager` and
    the classes in `qmanager_callbacks.hpp` by
    introducing `qmanager_cb_ctx_t` to make
    it easy for the `qmanager` module to
    pass its state to these new classes.

- Update `qmanager` code and build rules to add
   the new files.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -31,6 +31,7 @@ extern "C" {
 #include <jansson.h>
 }
 
+#include "qmanager/modules/qmanager_callbacks.hpp"
 #include "qmanager/policies/base/queue_policy_base.hpp"
 #include "qmanager/policies/base/queue_policy_base_impl.hpp"
 #include "qmanager/policies/queue_policy_factory_impl.hpp"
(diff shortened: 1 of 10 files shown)


dongahn:
This was added on top of PR #644. So #644 needs to be reviewed and merged first before this one.

The other possibility is just to review and merge this if that's not too difficult and drop #644.

qmanager/modules/qmanager.cpp
```diff
@@ -62,155 +64,9 @@ struct qmanager_ctx_t {
     qmanager_conf_t conf;
     qmanager_args_t args;
     std::string queue_policy;
-    schedutil_t *schedutil;
-    std::shared_ptr<queue_policy_base_t> queue;
+    qmanager_cb_ctx_t c;
```
SteVwonder, line 67: nitpick, but it would be nice to have a longer variable name here.  Maybe `cb_ctx`?

src/common/c++wrappers/eh_wrapper.hpp
```diff
@@ -0,0 +1,121 @@
...
+     *                 forwarded to f.
+     */
+    template <typename Functor, typename ... Args>
+    auto operator() (Functor f, Args && ... args)
+        -> decltype(f (std::forward<Args> (args) ... )) {
+        int rc = -1;
+        exception_raised = false;
+        memset (err_message, '\0', 4096);
+        try {
+            return f (std::forward<Args> (args) ... );
```
SteVwonder, line 62: I had never heard of perfect forwarding before.  Cool stuff! Today I Learned :)

qmanager/modules/qmanager.cpp
dongahn: Yeah I was ambivalent :-). 

With a longer member variable name, I found some of the later expression becomes pretty long.  and decided to go with a short name. But I see your concern here. 

How about we actually make `qmanager_ctx_t` inherit from `qmanager_cb_ctx_t`, this way you can get rid of any references to `c` or `cb_ctrx` for good?

src/common/c++wrappers/eh_wrapper.hpp
dongahn: This was the first time I used this in real coding as well. Pretty cool stuff at the expense of brain bending if no simple walk-through tutorial is available -- so was the reference in the comment.

qmanager/modules/qmanager.cpp
SteVwonder: Sounds good to me :)

dongahn:
@SteVwonder: the changes requested have been committed and squashed. The CI is happy. This should go in.

SteVwonder:
Looks like the rebase had a merge conflict now that #644 is merged.  Sorry @dongahn but I think you will have to do a manual rebase onto master to resolve the conflict.

dongahn:
Ok. Resolved all the conflicts and forced a push. Thanks @SteVwonder!

PR #648, opened 2020-05-01
docker: add ENTRYPOINT to start munged by default

grondo:
Use the `entrypoint.sh` from flux-core by default in published flux-sched docker images, so that `docker run -ti fluxrm/flux-sched` works the same as `docker run -ti fluxrm/flux-core`, i.e. munged is started and user is dropped into an instance of size=1.

Also, pull in @trws' `--build-directory` option for `docker-run-checks.sh` since this has been useful.

### DIFF
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -143,8 +143,8 @@ if test -n "$TAG"; then
     docker commit \
 	--change 'ENTRYPOINT [ "/usr/local/sbin/entrypoint.sh" ]' \
 	--change 'CMD [ "/usr/bin/flux", "start", "/bin/bash" ]' \
-	--change 'USER flux' \
-	--change 'WORKDIR /home/flux' \
+	--change 'USER fluxuser' \
+	--change 'WORKDIR /home/fluxuser' \
 	tmp.$$ $TAG \
 	|| die "docker commit failed"
     docker rm tmp.$$
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -136,11 +136,13 @@ if test -n "$TAG"; then
         --volume=$TOP:/usr/src \
         --user="root" \
 	travis-builder:${IMAGE} \
-	sh -c "make install && \
+	sh -c "cd ${BUILD_DIR:-.} && \
+	       make install && \
                userdel $USER" \
 	|| (docker rm tmp.$$; die "docker run of 'make install' failed")
     docker commit \
-	--change 'CMD "/usr/bin/flux"' \
+	--change 'ENTRYPOINT [ "/usr/local/sbin/entrypoint.sh" ]' \
+	--change 'CMD [ "/usr/bin/flux", "start", "/bin/bash" ]' \
 	--change 'USER flux' \
 	--change 'WORKDIR /home/flux' \
 	tmp.$$ $TAG \
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -16,8 +16,8 @@ declare -r prog=${0##*/}
 die() { echo -e "$prog: $@"; exit 1; }
 
 #
-declare -r long_opts="help,quiet,interactive,image:,jobs:,no-cache,no-home,distcheck,tag:"
-declare -r short_opts="hqIdi:j:t:"
+declare -r long_opts="help,quiet,interactive,image:,jobs:,no-cache,no-home,distcheck,tag:,build-directory:"
+declare -r short_opts="hqIdi:j:t:D:"
 declare -r usage="
 Usage: $prog [OPTIONS] -- [CONFIGURE_ARGS...]\n\
 Build docker image for travis builds, then run tests inside the new\n\
@@ -34,6 +34,7 @@ Options:\n\
  -i, --image=NAME              Use base docker image NAME (default=$IMAGE)\n\
  -j, --jobs=N                  Value for make -j (default=$JOBS)\n
  -d, --distcheck               Run 'make distcheck' instead of 'make check'\n\
+ -D, --build-directory=DIRNAME Name of a subdir to build in, will be made\n\
  -I, --interactive             Instead of running travis build, run docker\n\
                                 image with interactive shell.\n\
 "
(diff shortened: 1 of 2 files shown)


dongahn:
`entrypoint.sh` has already merged into flux-core (in terms of merge coordination).

src/test/docker/docker-run-checks.sh
```diff
@@ -136,11 +136,13 @@ if test -n "$TAG"; then
...
 	travis-builder:${IMAGE} \
-	sh -c "make install && \
+	sh -c "cd ${BUILD_DIR:-.} && \
+	       make install && \
                userdel $USER" \
 	|| (docker rm tmp.$$; die "docker run of 'make install' failed")
     docker commit \
-	--change 'CMD "/usr/bin/flux"' \
+	--change 'ENTRYPOINT [ "/usr/local/sbin/entrypoint.sh" ]' \
+	--change 'CMD [ "/usr/bin/flux", "start", "/bin/bash" ]' \
```
dongahn, line 145: Nice addition as most users would type this as the first command! 

Do you know it will be obvious from the user's terminal that a Flux instance of size=1 is already running? E.g., would the flux start message appear in the user's console?

I ask because a handful of other type of users who use this image to do further development, they might not want to start flux at size=1 and any hint of it running would be a good reminder:  first "exit" and flux start with different options if needed.

dongahn:
Thanks @grondo. LGTM. Only a couple of comments above.

trws:
We could add a bit to set PS1 in the shell it starts, maybe even just 
`PS1=“(flux -s1) $PS1”` or something?

src/test/docker/docker-run-checks.sh
grondo: No, it will not be obvious that the current shell is running under `flux start`. Any arguments to `docker run`, however, will replace the default `CMD`, so to run without `flux start` would be as simple as

```
docker run -ti fluxrm/flux-sched bash
```

>  first "exit" and flux start with different options if needed.

Exiting from the shell started by `flux start` will exit the entire docker container unfortunately.

grondo:
> We could add a bit to set PS1 in the shell it starts, maybe even just
`PS1=“(flux -s1) $PS1”` or something?

This is a good idea. I have something like:
```
${FLUX_URI+(flux size=$(flux getattr size))}
```
In my `PS1`, so I know when I'm in a (non-system) flux instance.

```
(flux size=4) grondo@asp:~/git/f.git$ 
```
(For my own use, it would be nice if my prompt *also* showed the path to the current flux installation used by the instance, e.g. `/usr/bin` or `~/git/flux-core.git/src/cmd`... Maybe I'll play with that a bit.)

We could add this to the bashrc (and other shells?) in the docker image somehow. (This would need to be done in flux-core, then it would automatically be inherited by dependent docker images.

grondo:
Hm, or maybe instead of `flux` we adopt U+0192 as the flux unicode symbol?
```
ƒ(s=4) grondo@asp ~/git/flux-core.git$ 
```

trws:
I used to have my shell display the flux-uri and current slurm/bsub state, but never re-built it for my current setup.  I like the idea of having breadcrumbs for which one it is, maybe a reduced-path (first letter of non-basename levels or something)?

grondo:
I settled on adding the string `builddir` if `which flux` contains `src/cmd`, nothing if not (implying system flux)

```console
 grondo@asp:~/git/flux-core.git$ flux start -s 4
ƒ(s=4) grondo@asp:~/git/flux-core.git$ exit
exit
 grondo@asp:~/git/flux-core.git$ src/cmd/flux start -s 4
ƒ(s=4,builddir) grondo@asp:~/git/flux-core.git$ 

```

trws:
That looks great!

src/test/docker/docker-run-checks.sh
```diff
@@ -34,6 +34,7 @@ Options:\n\
  -i, --image=NAME              Use base docker image NAME (default=$IMAGE)\n\
  -j, --jobs=N                  Value for make -j (default=$JOBS)\n
  -d, --distcheck               Run 'make distcheck' instead of 'make check'\n\
+ -D, --build-directory=DIRNAME Name of a subdire to build in, will be made\n\
```
dongahn, line 37: Sorry I just caught this. Did you mean "Name of a subdir to build in"?

dongahn:
@grondo. Thanks. I just spotted one thing and left a comment. Other than that, is this PR ready to go in?

src/test/docker/docker-run-checks.sh
grondo: Yeah, I'll fix that in the morning. Good catch.

grondo:
> @grondo. Thanks. I just spotted one thing and left a comment. Other than that, is this PR ready to go in?

May want to wait until flux-framework/flux-core#2925 goes in, so that the docker images use the new `PS1`.

dongahn:
Got it. Thanks.

grondo:
While we're tweaking the docker image, should we change the default `USER` for containers from the present value of `flux`, which might be confused with proposed system instance owner of `flux`? Maybe `fluxuser` to match what is now in the Quickstart guide.

dongahn:
> While we're tweaking the docker image, should we change the default USER for containers from the present value of flux, which might be confused with proposed system instance owner of flux? Maybe fluxuser to match what is now in the Quickstart guide.

Sounds reasonable to me.

grondo:
Ok,  I've updated this PR to sync up with flux-core now that flux-framework/flux-core#2925 was merged.

grondo:
You might want this one to go in sooner rather than later. Because the default USER for flux-core changed to `fluxuser`, new docker images deployed by flux-sched will be a little broken until this is merged.

dongahn:
Done. Thanks.

grondo:
Thanks!

grondo:
Just to follow up, I did verify that the newly published flux-sched docker images work as expected:
```
grondo@flux-o-rama:~/git/flux-sched.git$ docker run -ti fluxrm/flux-sched
ƒ(s=1,d=0) fluxuser@af08fe3d073a:~$ flux mini run hostname
af08fe3d073a
ƒ(s=1,d=0) fluxuser@af08fe3d073a:~$ 
```

Issue #640, opened 2020-04-08
Add multiple queue support within qmanager


dongahn:
This is the first task to implement multiple queue support as discussed at https://github.com/flux-framework/flux-sched/issues/637.

1. Replace a single queue object (of `queue_policy_base_t` type) within `qmanager_ctx_t` with a `std::map` that contains a configurable number of queue objects keyed off of distinct queue names. 

2. Add ways to specify multiple queues both in configuration files and module load options while keeping the current conf/load option set backward compatible as much as possible. Each specified queue should be able to enforce a queueing policy, queuing parameters and queueing-policy parameters, independent of one another.

3. Bring in libjobspec to fetch the queue information from `system.attributes.queue` key of a jobspec and use it to identify the corresponding queue. 

4. Adjust biz logic in each and every job-manager callbacks like `jobmanager_alloc_cb`

5. Add the queue name into `R` (as part of `scheduling` key) so that it will help reconstruct multiple named queues on `qmanager` reload. 

6. Adjust the existing test cases (the backward compatibility from #2 should  make this easier) and add new cases.

Once this support is added, jobspecs with different queues will be enqueued into correspondingly named queues within `qmanager` upon submission. So these jobs will be bounded by the queuing policies etc, enforced on their belonging queues.

However, the actual resources will be selected from any resource in the `resource` module. This will be this way until the next set of tasks are implemented.

cmoussa1:
This is just for my own curiosity and learning, but how does `qmanager` perform a check to make sure the queue requested can fulfill the resources requested by a job? Here's a hypothetical example of what I am talking about:

A job is submitted to the **debug** queue; this job is requesting 5 nodes, but the **debug** queue only supports a max of 3 nodes per job.

Does `qmanager` have its own table (where each queue's limits are defined) that it checks? Or does it do something else?

Or maybe what I am asking about is already answered in your initial comment:

> Bring in libjobspec to fetch the queue information from system.attributes.queue key of a jobspec and use it to identify the corresponding queue.

dongahn:
@cmoussa1 and I chatted yesterday and this topic came up. I think the easiest way to put certain limits (e.g., max node count and wall time) would be through a configuration file (or module load option). We will not be able to enforce per-user limits this way, but I am not sure if that's necessary.

cmoussa1:
To add to this, from conversations with Ryan:

[...]

Issue #626, opened 2020-03-19
Pass starttime estimate to "note"


dongahn:
@chu11 and @garlick: 

We talked about utilizing the flux-core's front end as much as possible to pass scheduler information. I want to see what it takes to pass the starttime estimate to the "note" to each `alloc` request for backfill based policies within manager. What needs to happen within flux-core to make use of (e.g., display) this information?

garlick:
As we discussed earlier,  the response should be a JSON dict rather than a human readable message.  The job manager must be modified do something with this when it is received.  Originally the design was to attach it to the job hash entry, in memory only, and update it each time that response type is received for the job, discarding the old value.  (adding it to the persistent job event log seemed not to be useful and potentially high overhead if it could be updated many times per job)

I don't remember what we came up with for how to get the data from job manager to job-info and thus to the listing tool.  Do you remember @chu11?

garlick:
Ah actually we opened flux-framework/flux-core#2608 after that discussion.

dongahn:
> As we discussed earlier, the response should be a JSON dict rather than a human readable message. The job manager must be modified do something with this when it is received. Originally the design was to attach it to the job hash entry, in memory only, and update it each time that response type is received for the job, discarding the old value. (adding it to the persistent job event log seemed not to be useful and potentially high overhead if it could be updated many times per job)

Thanks. @garlick.

Since flux-core doesn't use "note" just yet, it seems I can make progress in flux-sched independently and pass a JSON object to the note instead of a plain text.  

Also too frequent of an update response will be bad as this can happen when there are lots of churns that lead to schedule loop invocations.I will see if I can add some heuristics to send a note response only when the state (starttime estimate and pending reason) change significantly.

dongahn:
At today's meeting, we agreed that it would be good for `qmanager` to add queue and start time to "note" in JSON. This can facilitate the development of properly displaying the scheduler info within flux-core's job listing tool.

dongahn:
Also would be good to add this soon.

dongahn:
Preliminary support has been added to https://github.com/dongahn/flux-sched/tree/starttime-estimate

dongahn:
```
update_annotations: job 313482280960 not found: Success
```

```console
ahn1@9322372c5ba3:/usr/src/t$ ./t1003-qmanager-policy.t
==29201== Memcheck, a memory error detector
==29201== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
[...]

Issue #485, opened 2019-07-05
Name changes for service module names


dongahn:
As part of a PR #481 review, @garlick suggested:

> You have two modules named qmanager and resource. Should they be named with a common prefix to indicate that they are part of a single scheduler implementation, like sched-full-qmanager and sched-full-resource? (substitute something more creative and cool sounding for "full" :-)

As I think about this, it may be good to have an actual cool name for the scheduler components and use that as the prefix for the services.

dongahn:
Our resource infrastructure starts to look pretty solid, and I think I should try to write up a paper on the core technologies sooner rather than later. It helps have a good name for that purpose as well as the issue described above. So here are two proposal. 

- GANTT (Graph’s Arbitrable Nodes and Time Traversers)
This will make our service modules look like: `gantt-resource` and `gantt-qmanager`. The prefix then can differentiate our full scheduler with `sched-simple` or test scheduler in flux-core. 

- GATES (Graph Aided Task Execution Scheduler or other backronym)
I just learned that "fluxgate" is actually a thing so this may work well with Flux although the initial feedback I got from @SteVwonder, Frank and Joe were not that great. https://en.wikipedia.org/wiki/Fluxgate_compass

Ultimately, we should refactor all the flux-core-agnostic components and package them into a separate repo so that it can be contributed and used by other projects beyond Flux  as a reusable general purpose componen. This will include the kubenetes scheduler plugin that @cmisale is currently working on. At the last call, she said she can do a `go` binding for our high level resource API. 

At that point, flux-sched will become a thin layer that integrates these general purpose components into `flux-core`.

Thoughts on the names?

grondo:
I respectfully submit:

GAUSS - Graph Abstraction for Universal System Scheduling

The U could also stand for Uncompromising, Undifferentiated, Unconfined, Unlimited, etc.

(I know you weren't really looking for new names, but I couldn't resist, sorry)

dongahn:
Thanks @grondo. New name proposals are appreciated since I still have time to decide the best one!

dongahn:
Maybe Graph-Aided... instead of Abstraction... Did you name this after a mathematician?

grondo:
Yes, mathematician, physicist, and also a "gauss" is a unit of measurement for magnetic flux density. Possibly there are already a lot of things called Gauss though...

dongahn:
[...]

PR #649, opened 2020-05-02
qmanager: add multi-queue support

dongahn:
This PR is work in progress to add multi-queue support within `qmanager`.

Build on top of a merge of two pending PRs: PR #645 and PR #646.

As described in Issue #640, it adds the following modifications:
 - Replace a single queue object (of `queue_policy_base_t` class) with
a `std::map` that contains a configurable number of queue objects
keyed off of distinct queue names.
- Add ways to specify multiple queues both through configuration files
and module load options while keeping the current conf/load option
set backward compatible as much as possible. 
Each specified queue can enforce a queueing policy, queuing
parameters and queueing-policy parameters, independent of one another.
- Bring in libjobspec to fetch the queue information from
`system.attributes.queue` key of a jobspec and use
it to identify the corresponding queue.
- Adjust the business logic in each and every job-manager callbacks
like `jobmanager_alloc_cb`
- Add a few configuration tests into `t1005-qmanager-conf.t`

I still need to add a bunch more test cases to finish this up.
As we discussed before, I'm not entirely happy with the current
way of configuring  queues. But I thought there is a benefit of getting
the runtime logic correct as soon as possible and then improve
upon user friendliness later.

If you want to test this:

```console
ahn1@6e8124a3b9bc:/usr/src$ flux start -s 1
ahn1@6e8124a3b9bc:/usr/src$ flux module remove qmanager
ahn1@6e8124a3b9bc:/usr/src$ flux module load qmanager queues="debug batch"
ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux mini run -n 2 --dry-run hostname | jq '.attributes.system.queue="batch"' | flux job submit -
10022792724480
ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux mini run -n 2 --dry-run hostname | jq '.attributes.system.queue="debug"' | flux job submit -
10185330393088
ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux mini run -n 2 --dry-run hostname | flux job submit -
10675594199040
ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux dmesg | grep alloc
2020-05-02T00:38:41.891482Z qmanager.debug[0]: alloc success (queue=batch id=10022792724480)
2020-05-02T00:38:51.581285Z qmanager.debug[0]: alloc success (queue=debug id=10185330393088)
2020-05-02T00:39:20.800212Z qmanager.debug[0]: alloc success (queue=batch id=10675594199040)
```

### DIFF
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -229,18 +229,44 @@ bool qmanager_opts_t::is_queue_params_set () const
 
 bool qmanager_opts_t::is_policy_params_set () const
 {
-    return m_queue_prop.policy_params != QMANAGER_OPTS_UNSET_STR;
+
+qmanager_opts_t &qmanager_opts_t::canonicalize ()
+{
+    if (m_per_queue_prop.empty ()) {
+        std::string qn = is_default_queue_set ()? m_default_queue
+                                                : m_default_queue_name;
+        auto ret = m_per_queue_prop.insert (
+                       std::pair<std::string, queue_prop_t> (qn,
+                                                             queue_prop_t ()));
+        if (!ret.second)
+            throw std::bad_alloc ();
+    }
+    for (auto &kv : m_per_queue_prop) {
+        if (!is_default_queue_set ())
+            m_default_queue = kv.first;
+        if (!kv.second.is_queue_policy_set ())
+            kv.second.queue_policy = m_queue_prop.queue_policy;
+        if (!kv.second.is_queue_params_set ())
+            kv.second.queue_params = m_queue_prop.queue_params;
+        if (!kv.second.is_policy_params_set ())
+            kv.second.policy_params = m_queue_prop.policy_params;
+    }
+    return *this;
 }
 
 qmanager_opts_t &qmanager_opts_t::operator+= (const qmanager_opts_t &src)
 {
     if (src.m_queue_prop.queue_policy != QMANAGER_OPTS_UNSET_STR)
         m_queue_prop.queue_policy = src.m_queue_prop.queue_policy;
+    if (src.m_default_queue != QMANAGER_OPTS_UNSET_STR)
+        m_default_queue = src.m_default_queue;
     if (src.m_queue_prop.queue_params != QMANAGER_OPTS_UNSET_STR)
         m_queue_prop.queue_params = src.m_queue_prop.queue_params;
     if (src.m_queue_prop.policy_params != QMANAGER_OPTS_UNSET_STR)
         m_queue_prop.policy_params = src.m_queue_prop.policy_params;
-    return canonicalize ();
+    if (!src.m_per_queue_prop.empty ())
+        m_per_queue_prop = src.get_per_queue_prop ();
+    return *this;
 }
 
 bool qmanager_opts_t::operator ()(const std::string &k1,
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -257,12 +257,28 @@ int qmanager_opts_t::parse (const std::string &k, const std::string &v,
 {
     int rc = 0;
     std::string dflt;
+    std::map<std::string, std::string> tmp_mp;
     int key = static_cast<int> (qmanager_opts_key_t::UNKNOWN);
 
     if (m_tab.find (k) != m_tab.end ())
         key = m_tab[k];
 
     switch (key) {
+    case static_cast<int> (qmanager_opts_key_t::QUEUES):
+        rc = parse_queues (v);
+        break;
+
+    case static_cast<int> (qmanager_opts_key_t::DEFAULT_QUEUE):
+        dflt = m_default_queue;
+        m_default_queue = v;
+        if (!m_per_queue_prop.empty ()
+            && m_per_queue_prop.find (v) == m_per_queue_prop.end ()) {
+            info += "Unknown default queue (" + v + ")! ";
+            info += "Use default.";
+            m_default_queue = dflt;
+        }
+        break;
+
     case static_cast<int> (qmanager_opts_key_t::QUEUE_POLICY):
         dflt = m_queue_prop.queue_policy;
         m_queue_prop.queue_policy = v;
@@ -279,6 +295,58 @@ int qmanager_opts_t::parse (const std::string &k, const std::string &v,
 
     case static_cast<int> (qmanager_opts_key_t::POLICY_PARAMS):
         m_queue_prop.policy_params = v;
+        break;
+
+    case static_cast<int> (qmanager_opts_key_t::QUEUE_POLICY_PER_QUEUE):
+        tmp_mp.clear ();
+        if ( (rc = parse_multi_options (v, ' ', ':', tmp_mp)) < 0)
+            break;
+        for (const auto &kv : tmp_mp) {
+            if (m_per_queue_prop.find (kv.first) == m_per_queue_prop.end ()) {
+                info += "Unknown queue (" + kv.first + ").";
+                errno = ENOENT;
+                rc = -1;
+                break;
+            }
+            dflt = m_per_queue_prop[kv.first].queue_policy;
+            m_per_queue_prop[kv.first].queue_policy = kv.second;
+            if (!known_queue_policy (kv.second)) {
+                info += "Unknown queuing policy (" + v + ") for queue ("
+                        + kv.second + ")! ";
+                info += "Use default. ";
+                m_per_queue_prop[kv.first].queue_policy = dflt;
+            }
+        }
+        break;
+
+    case static_cast<int> (qmanager_opts_key_t::QUEUE_PARAMS_PER_QUEUE):
+        tmp_mp.clear ();
+        if ( (rc = parse_multi_options (v, ' ', ':', tmp_mp)) < 0)
+            break;
+        for (const auto &kv : tmp_mp) {
+            if (m_per_queue_prop.find (kv.first) == m_per_queue_prop.end ()) {
+                info += "Unknown queue (" + kv.first + ").";
+                errno = ENOENT;
+                rc = -1;
+                break;
+            }
+            m_per_queue_prop[kv.first].queue_params = kv.second;
+        }
+        break;
+
+    case static_cast<int> (qmanager_opts_key_t::POLICY_PARAMS_PER_QUEUE):
+        tmp_mp.clear ();
+        if ( (rc = parse_multi_options (v, ' ', ':', tmp_mp)) < 0)
+            break;
+        for (const auto &kv : tmp_mp) {
+            if (m_per_queue_prop.find (kv.first) == m_per_queue_prop.end ()) {
+                info += "Unknown queue (" + kv.first + ").";
+                errno = ENOENT;
+                rc = -1;
+                break;
+            }
+            m_per_queue_prop[kv.first].policy_params = kv.second;
+        }
         break;
 
     default:
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -204,6 +204,15 @@ qmanager_opts_t &qmanager_opts_t::operator+= (const qmanager_opts_t &src)
     return canonicalize ();
 }
 
+bool qmanager_opts_t::operator ()(const std::string &k1,
+                                  const std::string &k2) const
+{
+    if (m_tab.find (k1) == m_tab.end ()
+        || m_tab.find (k2) == m_tab.end ())
+        return k1 < k2;
+    return m_tab.at (k1) < m_tab.at (k2);
+}
+
 int qmanager_opts_t::parse (const std::string &k, const std::string &v,
                             std::string &info)
 {
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -151,8 +151,16 @@ qmanager_opts_t::qmanager_opts_t ()
     bool inserted = true;
 
     auto ret = m_tab.insert (std::pair<std::string, int> (
-                   "queue-policy",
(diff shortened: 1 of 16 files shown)


grondo:
> ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux mini run -n 2 --dry-run hostname | jq '.attributes.system.queue="batch"' | flux job submit -
10022792724480

Just an FYI - this can more easily be done via:

```
$ flux mini submit -n2 --setattr system.queue=batch hostname
```

dongahn:
> $ flux mini submit -n2 --setattr system.queue=batch hostname

Yeah @SteVwonder and I discussed this as the way to enqueue a job into a non-default queue way back. But I completely forgot and used a much longer way to express it :-). 

Nice job on the flux-mini interface!

dongahn:
Regarding the config file specification form, please see https://github.com/flux-framework/flux-sched/issues/650

dongahn:
> ahn1@6e8124a3b9bc:/usr/src/t/multiqueue$ flux dmesg | grep alloc
2020-05-02T00:38:41.891482Z qmanager.debug[0]: alloc success (queue=batch id=10022792724480)
2020-05-02T00:38:51.581285Z qmanager.debug[0]: alloc success (queue=debug id=10185330393088)
2020-05-02T00:39:20.800212Z qmanager.debug[0]: alloc success (queue=batch id=10675594199040)

I need to add test cases to finish this up and pass/failure will be whether the job has been enqueued into the named queue correctly.

I wonder how I can validate this when we don't have a way to display scheduler info like queues through job listing tool like `flux jobs`, 

My expectation is some of the key scheduler info will ultimately be displayed using `flux jobs` by utilizing the "note" field -- https://github.com/flux-framework/flux-sched/issues/626

@chu11 or @garlick: do you know when we will able to add note support in job listing tools within flux-core?

dongahn:
Just a note: when PR #645 and #646 are merged into master, a merge conflict will need to be resolved.

dongahn:
OK. I have added some more test cases and split larger commits into smaller atomic commits. This is ready to be reviewed. Once this lands, I will do name changing PR per Issue #485.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -204,6 +204,15 @@ qmanager_opts_t &qmanager_opts_t::operator+= (const qmanager_opts_t &src)
     return canonicalize ();
 }
 
+bool qmanager_opts_t::operator ()(const std::string &k1,
+                                  const std::string &k2) const
+{
+    if (m_tab.find (k1) == m_tab.end ()
+        || m_tab.find (k2) == m_tab.end ())
+        return k1 < k2;
```
SteVwonder, line 303: If one of the keys is in `m_tab` and the other is not, would you want to handle the one that is in `m_tab` before handling the other?

qmanager/modules/qmanager_opts.cpp
```diff
@@ -79,94 +163,236 @@ qmanager_opts_t::qmanager_opts_t ()
...
+                                                : m_default_queue_name;
+        auto ret = m_per_queue_prop.insert (
+                       std::pair<std::string, queue_prop_t> (qn,
+                                                             queue_prop_t ()));
+        if (!ret.second)
+            throw std::bad_alloc ();
+    }
+    for (auto &kv : m_per_queue_prop) {
+        if (!is_default_queue_set ())
+            m_default_queue = kv.first;
```
SteVwonder, line 272: Will this ever be true due to the setting of `m_default_queue` to the first entry in `parse_queues`?

dongahn:
> Thanks @dongahn for putting this together! Little stuff is inline. Generally LGTM!

> The only big thing is that it is possible for the handshake with the job-manager to request single mode despite the user setting a queue policy of easy. After applying a patch to log the job-manager queue mode, I tried loading qmanager with the queue-policy-per-queue and got a surprising result of single. It looks like the handshake function is only checking the main queue property object for the policy and setting the job-manager queue mode based on that. If you don't specify queue-policy, but you do specify queue-policy-per-queue, you can get an incorrect state.

Ah... This is a great catch! 

With multi queue support, I am not even sure if "single" mode even makes sense any longer. If you have two FCFS queues at `qmanager`, you don't want "single" from `job-manager` reducing job throughput... The only mode where "single" makes sense is when you have a single queue within `qmanager` and it enforces FCFS. 

If there isn't much performance difference between `single` and `unlimited` for this case, we may want to deprecate "single" at `qmanager` for good...

qmanager/modules/qmanager_opts.hpp
```diff
@@ -95,10 +136,12 @@ class qmanager_opts_t : public optmgr_parse_t {
...
 private:
-    qmanager_opts_t &canonicalize ();
-    bool known_queue_policy (const std::string &policy);
+    int parse_queues (const std::string &queues);
 
+    const std::string m_default_queue_name = "default";
+    std::string m_default_queue = QMANAGER_OPTS_UNSET_STR;
     queue_prop_t m_queue_prop;
+    std::map<std::string, queue_prop_t> m_per_queue_prop;
     std::map<std::string, int> m_tab;
```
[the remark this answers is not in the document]
dongahn: This is used to quite often to indicate the variables are C++ class member variables. https://stackoverflow.com/questions/13018189/what-does-m-variable-prefix-mean.

Yes I can add some notes about these variables.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -204,6 +204,15 @@ qmanager_opts_t &qmanager_opts_t::operator+= (const qmanager_opts_t &src)
     return canonicalize ();
 }
 
+bool qmanager_opts_t::operator ()(const std::string &k1,
```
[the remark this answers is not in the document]
dongahn: Good catch. Will fix.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -79,94 +163,236 @@ qmanager_opts_t::qmanager_opts_t ()
...
+        rc = parse_queues (v);
+        break;
+
+    case static_cast<int> (qmanager_opts_key_t::DEFAULT_QUEUE):
+        dflt = m_default_queue;
+        m_default_queue = v;
+        if (!m_per_queue_prop.empty ()
+            && m_per_queue_prop.find (v) == m_per_queue_prop.end ()) {
+            info += "Unknown default queue (" + v + ")! ";
+            info += "Use default.";
```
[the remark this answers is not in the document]
dongahn: Agreed. Will change.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -79,94 +163,236 @@ qmanager_opts_t::qmanager_opts_t ()
...
+        }
+        break;
+
     case static_cast<int> (qmanager_opts_key_t::QUEUE_POLICY):
-        dflt = m_queue_prop.queue_policy;
-        m_queue_prop.queue_policy = v;
-        if (!known_queue_policy (m_queue_prop.queue_policy)) {
+        if (!m_queue_prop.set_queue_policy (v)) {
             info += "Unknown queuing policy (" + v + ")! ";
             info += "Use default.";
```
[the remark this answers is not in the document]
dongahn: Ditto.

t/t1006-qmanager-multiqueue.t
```diff
@@ -0,0 +1,152 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
dongahn: Will change. Feels like I can benefit from a spell checker for git commit messages.

qmanager/modules/qmanager_opts.cpp
dongahn: I think in this case this would be okay as the unknown keys will be just ignored. What we need to make sure is the known keys get to process in order. In the case of the `qmanager_opts_t` class, only the `queues=` option is required to be processed first.

qmanager/modules/qmanager_opts.cpp
dongahn: Yes, this will be true for example if `queues=qbatch qdebug` is given. In that case, the `m_default_queue` would be `0xdeadbeef` and set to the named queue that is lexicographically smallest. On the other hand, if `default-queue=qdebug` is passed also, `m_default_queue` is set to that so the default queue becomes `qdebug` instead of `qbatch.`

If this is not the case, there is a bug somewhere :-)

SteVwonder:
> If there isn't much performance difference between single and unlimited for this case, we may want to deprecate "single" at qmanager for good...

Sounds good to me.

qmanager/modules/qmanager_opts.cpp
SteVwonder: > and set to the named queue that is lexicographically smallest.

I guess what I'm getting at is that wouldn't that setting be done in the [`parse_queues` method](https://github.com/flux-framework/flux-sched/pull/649/files/db42eff456a9beb040cfce008b5b236d3bbd7731#diff-ecff1f5a8c8cc615ce9f7c7ca69362cfR47) causing this particular code to never run because `is_default_queue_set` will always return true?  For example, I applied the patch pasted below and could not seem to trigger this code to run, including with the example you provided.

Patch:
```diff
diff --git a/qmanager/modules/qmanager_opts.cpp b/qmanager/modules/qmanager_opts.cpp
index aaf7fdb2..7557f7e1 100644
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -22,6 +22,7 @@
  *  See also:  http://www.gnu.org/licenses/
 \*****************************************************************************/

+#include <iostream>
 #include "qmanager_opts.hpp"

 using namespace Flux;
@@ -43,8 +44,10 @@ int qmanager_opts_t::parse_queues (const std::string &queues)
             goto done;
         m_per_queue_prop.clear (); // clear the default queue entry
         for (const auto &entry: entries) {
-            if (m_default_queue == m_default_queue_name)
+            if (m_default_queue == m_default_queue_name) {
+                std::cerr << "Parse_queues setting m_default_queue to " << entry << std::endl;
                 m_default_queue = entry;
+            }
             auto ret = m_per_queue_prop.insert (
                            std::pair<std::string, queue_prop_t> (
                                entry, queue_prop_t ()));
@@ -268,8 +271,10 @@ qmanager_opts_t &qmanager_opts_t::canonicalize ()
             throw std::bad_alloc ();
     }
     for (auto &kv : m_per_queue_prop) {
-        if (!is_default_queue_set ())
+        if (!is_default_queue_set ()) {
+            std::cerr << "Canonicalize setting m_default_queue to " << kv.first << std::endl;
             m_default_queue = kv.first;
+        }
         if (!kv.second.is_queue_policy_set ())
             kv.second.set_queue_policy (m_queue_prop.get_queue_policy ());
         if (!kv.second.is_queue_params_set ())
```

Example:
```
sherbein .../hdd/Repositories/flux-framework/flux-sched (multi-queue !?S)
❯ flux module reload qmanager queues="debug batch"                                                                                                                                                                                                                   23:58:38 ()
Canonicalize setting m_default_queue to batch

sherbein .../hdd/Repositories/flux-framework/flux-sched (multi-queue !?S)
❯ flux module reload qmanager queues="debug batch" default-queue="batch"                                                                                                                                                                                             23:58:49 ()

sherbein .../hdd/Repositories/flux-framework/flux-sched (multi-queue !?S)
❯ flux module reload qmanager queues="debug batch" default-queue="debug"                                                                                                                                                                                             23:59:02 ()

sherbein .../hdd/Repositories/flux-framework/flux-sched (multi-queue !?S)
❯ flux module reload qmanager                                                                                                                                                                                                                                        23:59:06 ()
Canonicalize setting m_default_queue to default
```

dongahn:
I will work on it today to land it.

dongahn:
> If there isn't much performance difference between single and unlimited for this case, we may want to deprecate "single" at qmanager for good...

Looks to me like `qmanage` performs about the same whether it is `unlimited` or `single`. I will just use `unlimited`.

unlimited, fcfs
```console
ahn1@90127f184db7:/usr/src/t/perf_test$ ./bulksubmit.py jobs/*.json
bulksubmit: Starting...
bulksubmit: submitted 1024 jobs in 3.98s. 257.25job/s
bulksubmit: First job finished in about 3.983s

bulksubmit: Ran 1024 jobs in 61.4s. 16.7 job/s
```

single, fcfs
```console
ahn1@90127f184db7:/usr/src/t/perf_test$ ./bulksubmit.py jobs/*.json
bulksubmit: Starting...
bulksubmit: submitted 1024 jobs in 3.76s. 272.30job/s
bulksubmit: First job finished in about 3.763s

bulksubmit: Ran 1024 jobs in 64.3s. 15.9 job/s
```

qmanager/modules/qmanager_opts.cpp
dongahn: Ah. Sorry I misunderstood your comment. Yes, I think this is a dead code, that remained after a code refactoring. Will fix.

dongahn: Hmmmm. I will have to take it back. Some tests are failing without this code. Let me look more.

dongahn: My apology. I don't think I understand your comment any longer. You are not saying Line 272 will never be executed, are you? Your final example 

> sherbein .../hdd/Repositories/flux-framework/flux-sched (multi-queue !?S)
❯ flux module reload qmanager                                                                                                                                                                                                                                        23:59:06 ()
Canonicalize setting m_default_queue to default

shows it is being set to "default"

Maybe you are referring to Line 47 within `parse_queues` instead?

dongahn: > Maybe you are referring to Line 47 within parse_queues instead?

Ok. I think this is a dead code. I will eliminate that.

dongahn:
@SteVwonder: Ok. I addressed all of the issues above. If Travis CI turns green, this should be good to go.

dongahn:
All green now!

PR #655, opened 2020-05-09
Update the project name to Fluxion

dongahn:
This PR updates the project name to Fluxion.

Three commits update `resource`, `qmanager` and README.md,
respectively.

Code changes:
- Change the module name to use the `sched-fluxion` prefix
   to avoid a name conflict with a new generic module called `resource`
   into flux-core.
- Update resource's high level API so that it will use the updated
   topic strings for RPCs on `sched-fluxion-resource`.
- Change rc1,3 script names for the module
   name change and update them to use `FLUXION_`*
   environment variables instead of `FLUX_`*.
   Use `sched-fluxion-resource` and `sched-fluxion-qmanager`
   instead of `resource` and `qmanager`
   for `flux module load`, `reload` and `remove`.

Testsuite changes:
- Update the `flux-resource` command to use the new
   topic string prefix (`sched-fluxion-resource`)
   and change the name of this command to `flux-ion-resource`.
- Update sharness test scripts to use `FLUXION_`*
    instead of `FLUX`*
- Rename the testing rc scripts to have `fluxion-` prefix.
- Update sharness tests for fluxion name change.
   use `sched-fluxion-resource` and `-qmanager`
   for `flux module load`, `reload` and `remove`.
   use `flux-ion-resource` instead
   of `flux-resource` to drive the `sched-fluxion-resource`
   module. change `FLUX_RESOURCE_`* to `FLUXION_RESOURCE_`*
   and `FLUX_QMANAGER_`* to `FLUXION_QMANAGER_`*.

README.md change:
- flux-sched: repo name
- fluxion: the project name for the advanced scheduler
   that flux-sched contains.
- Update the main module names.


dongahn:
OK. This PR is ready to review and I suggested a few potential reviewers. Thanks. It will be good if we can land this before flux-core introduces the new generic `resource` module to avoid any breakage.

README.md
```diff
@@ -93,8 +109,9 @@ make install
 ##### Flux Instance
 
 The examples below walk through exercising functioning flux-sched modules (i.e.,
```
SteVwonder, line 111: Not sure if this was intentional or missed.  I could see it being written either way.  Just wanted to call it out in case it was missed.

```suggestion
The examples below walk through exercising functioning Fluxion modules (i.e.,
```

grondo:
There still seems to be one use of `flux resource` in `flux-tree`?
```console
 grondo@asp:~/git/flux-sched.git$ git grep 'flux resource'
src/cmd/flux-tree:        flux resource -h > /dev/null 2>&1 && avail="yes"
src/cmd/flux-tree:        avg=$(flux resource stat | grep "Avg" | awk '{print $4}')
```

garlick:
Just curious, in #485 you thumbs upped this suggestion:

> Could also go with sched-ion-* for the MOD_NAME and reserve the pun for the front end command...

which would make everything a little less verbose by eliminating "flux" from the names (a little redundant).  Decided not to go that way huh?

README.md
dongahn: It was a bit intentional because I wanted to have a few references to flux-sched as well and this was a reasonable place.

dongahn:
> There still seems to be one use of flux resource in flux-tree?

Great catch! Because if the command is not found, the test ignores it so I missed it. I will update.

dongahn:
> which would make everything a little less verbose by eliminating "flux" from the names (a little redundant). Decided not to go that way huh?

Thank you for asking. I actually considered `sched-ion-*` but that didn't read very well in my mind and seemed a bit ad hoc so I stuck with the full name. I looked up if schedion was a word but it was not (maybe Italian :-) sounds cool though). 

If MOAB or PPSPRO ports their schedulers for Flux and they go with a convention, it will be something like `sched-moab-component1` or `sched-ppspro-component1` and probably not dropping their names so I thought I would use the full up project name. 

In this PR, the places where I dropped flux from fluxion was where flux and fluxion appear one after another: environment variable names (e.g., FLUX_RESOURCE -> FLUXION_RESOURCE), configure variables and the test front end script as its commands appears (`flux-ion-resource`).

On a separate note, looking at the distant future, there will be non-flux projects that only want to use our scheduler components like IBM's kubernetes efforts. So I thought it would be good to have the "stand alone" project name like Fluxion. But still I wanted to have a good association with Flux. In this way, for example, a project wants to combine kubernetes with our scheduler only can be named  kube-fluxion (IBM) but a project that combine kubernetes with full up flux becomes kube-flux. Either way we will keep "flux" in it and that was sort of one reasoning of the name "Fluxion" compared to other cool names.

dongahn:
>> There still seems to be one use of flux resource in flux-tree?

> Great catch! Because if the command is not found, the test ignores it so I missed it. I will update.

OK. I fixed this and squashed the change. GIT force-pushed. 

Unless you want me to fix other things, this should be good to go (once Travis turns green of course).

milroy:
As a small addendum to the conversation, I really like the name Fluxion.  @dongahn I think your inspiration was _The Method of Fluxions and Infinite Series: With Its Application to the Geometry of Curve-lines_ by Isaac Newton.  It doesn't get much more fundamental than that!

README.md
```diff
@@ -4,22 +4,38 @@
 are not yet stable.* The github issue tracker is the primary way to
 communicate with the developers.
 
-### flux-sched
+### Fluxion: An Advanced Graph-Based Scheduler for HPC
 
-flux-sched contains an advanced job scheduling facility for the Flux resource
-manager framework.
+Welcome to Fluxion, an advanced job scheduling software
```
SteVwonder, line 9: Just a suggestion, but do you think it is worth linking to the wikipedia article on [Fluxion](https://en.wikipedia.org/wiki/Fluxion) as a sort of easter egg explaining the thought behind the name?

t/t4001-match-allocate.t
```diff
@@ -35,34 +35,35 @@ test_debug '
...
+load-file=${grug} prune-filters=ALL:core \
 load-format=grug subsystems=containment policy=high
 '
 
 test_expect_success 'match-allocate works with a 1-node, 1-socket jobspec' '
-    flux resource match allocate ${jobspec} &&
-    flux resource match allocate ${jobspec} &&
-    flux resource match allocate ${jobspec} &&
-    flux resource match allocate ${jobspec}
+    flux ion-resource match allocate ${jobspec} &&
```
garlick, line 44: Does it make sense for command line tool to be `flux ion` instead of `flux ion-resource`?

configure.ac
```diff
@@ -134,11 +134,11 @@ AC_SUBST(fluxionresourcerc1dir)
 AS_VAR_SET(fluxionresourcerc3dir, $sysconfdir/flux/rc3.d)
 AC_SUBST(fluxionresourcerc3dir)
 
-AS_VAR_SET(fluxqmanagerrc1dir, $sysconfdir/flux/rc1.d)
-AC_SUBST(fluxqmanagerrc1dir)
+AS_VAR_SET(fluxionqmanagerrc1dir, $sysconfdir/flux/rc1.d)
```
garlick, line 137: These variables could be eliminated if the generic "flux" names are used as suggested in an earlier comment

configure.ac
```diff
@@ -128,11 +128,11 @@ AC_SUBST(fluxlibdir)
 AS_VAR_SET(fluxmoddir, $libdir/flux/modules)
 AC_SUBST(fluxmoddir)
 
-AS_VAR_SET(fluxresourcerc1dir, $sysconfdir/flux/rc1.d)
```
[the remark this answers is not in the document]
dongahn: Thanks. This makes sense we won't have fluxion specific rc directories.

t/t4001-match-allocate.t
dongahn: Well, that will be the final fluxion front end command. `flux ion-resource` is a test script that interfaces only with `resource`. `flux ion` will later wrap some of its functionality and ways to communicate with `qmanager` as well.

configure.ac
dongahn: Agreed!

dongahn:
> One other general comment is there is a lot of repetition of long commands in the tests that could be abstracted to shell functions to reduce the amount of repetition. If some refactoring were committed first, then the name change commit might be a lot smaller.

Yeah I think this will be particularly useful for sharness tests. Let me add some shell functions.

t/t4001-match-allocate.t
garlick: Ah sorry that makes sense!

dongahn:
OK. I made a few more changes.

- Add a footnote to explain the name behind Fluxion
- Use generic fluxrc* config variables
- Refactor module load/remove into shell functions

I squashed the first two changes but introduced new commits for the third change (since we may want to go back to the original version for debugging and such).

Let me know if there is any thing else you want me to address.

garlick:
If the refactor commits came before the name change commits, wouldn't the name change commits be much smaller?

dongahn:
> If the refactor commits came before the name change commits, wouldn't the name change commits be much smaller?

Good point. Yes, I considered that. In fact, the refactor commits are such that I can easily squash those refactor commits to the name change commits. The main thing was whether we consider the name change commits as just "interim" commits or having some values to them.  My thought was there is a value of keeping those name change commits. 1) we already have them; 2) if in the future, if we have a case where these "one-for-all functions" wouldn't work well, we can use those name change commits as the starting points to play individualized load/remove. 

Anyway, the delta between refactor and name changes are minor; I'm okay with either way. Let me know if you feel strongly about one way or the other.

garlick:
No I think they should remain separate commits, I was just questioning the order.  If you are OK with the extra churn then I'm fine with this as is.  Thanks.

dongahn:
> No I think they should remain separate commits, I was just questioning the order. If you are OK with the extra churn then I'm fine with this as is. 

Sorry I misunderstood you. 

Splitting the shell function change (https://github.com/flux-framework/flux-sched/pull/655/commits/320734d40f1f3a2c05ef5e49ebcf1c0ef59e37b5) into a separate commit will do then? This is a good thing to do anyway.

garlick:
I think the commit you referenced does not need to be split (It's already one standalone change).
Sorry for the confusion!  In case I still haven't communicated, the (minor) point I was trying to make was that if the shell function refactor came before the module rename, then the module rename could take place in the shell function instead of all the places where it is called.  That would make that commit significantly smaller, thus reducing the amount of code change introduced in the history by this PR.

However, you may decide it's not worth the rebase work, and that is fine with me!

dongahn:
> In case I still haven't communicated, the (minor) point I was trying to make was that if the shell function refactor came before the module rename, then the module rename could take place in the shell function instead of all the places where it is called. That would make that commit significantly smaller, thus reducing the amount of code change introduced in the history by this PR.

Ah Thanks. I appreciate your perspective.I think my initial response still applies:

> The main thing was whether we consider the name change commits as just "interim" commits or having some values to them. My thought was there is a value of keeping those name change commits. 1) we already have them; 2) if in the future, if we have a case where these "one-for-all functions" wouldn't work well, we can use those name change commits as the starting points to play individualized load/remove.

Essentially, I see some values in having a commit that explicitly changes the module names in our testsuites as a future reference. Since this seems this is minor enough, let's keep this as is.

garlick:
p.s. you've been approved twice so I think it's fair game to set MWP!

PR #657, opened 2020-05-16
testsuite: update flux-sharness.sh from flux-core

grondo:
This PR updates `t/sharness.d/flux-sharness.sh` from flux-core, including

 * @trws's `run_timeout.py` test timeout improvements
 * Support for `TEST_CHECK_PREREQS` to validate `jq` usage protected by `HAVE_JQ`

Also adds `TEST_CHECK_PREREQS=t` to one travis builder, and fixes resulting fallout.

The test is based on #655, so that name changes would not conflict. This PR should go in after that one is merged.

PR #661, opened 2020-05-27
Fix t2002-qmanager-reload.t with new core resource module

grondo:
This PR fixes a hang that will come up once flux-framework/flux-core#2949 is merged. Since the `resource` module now shares resource configuration with sched-simple via the acquire interface, this module must be reloaded once hwloc data is manually overridden, or else sched-simple will reject free requests after it is reloaded at the end of the test.

I suppose since #657 has not been merged yet, I can tack this onto that PR if that would be better.


grondo:
Rebased on current master

grondo:
Thanks! Setting MWP.

Issue #662, opened 2020-05-31
[discussion] changes needed to integrate with resource interface 


dongahn:
@milroy and @SteVwonder: 

As we discussed by phone, I am creating this ticket to flesh out some details for our task breakdown for our next 3-week sprint. The goal is to test fluxion alone with flux-core on `fluke` in a more meaningful way. Towards the goal, we need to modify fluxion to integrate it with flux-core's `resource` interface.

The immediate capability of `resource` is its support for compute nodes that join the system instance in a staggered fashion. Its `acquire` interface uses RPC streaming to respond to the remote caller (`sched-fluxion-resource` in our case) with the info on the newly joined nodes and their execution targets (in the `idset` format).

As I understood, this module will ultimately provide an interface that allows the remote callers to get those nodes that are going up and down as well.

As is, fluxion is still operational even with the introduction of `resource`. `sched-fluxion-resource` uses the blocking watch on per-rank hwloc xml file from KVS so its initialization will block until the very last node will join the system instance and dump its hwloc file. 

But to keep abreast of `resource` development and be meaningful for fluke testing, we need to put in place a baseline integration. 

### I. 
The least time consuming solution would be:
`sched-fluxion-resource` sets up a streaming RPC with `resource.acquire.` On each response, fluxion can go to the corresponding ranks in the idset and join the hwloc xml in KVS for each rank into the resource graph. I believe we already have all of the code that can do "unpack" each hwloc into the graph so the new code needed will be set up for  `resource.acquire` and move resource graph construction code into "runtime" (as opposed to "module init" time.)

### II. 
However, the above in and of itself won't be so future proof. As nodes are going up and down, you want to be able to mark a subgraph accordingly so that those unavailable resources will be excluded from scheduling at "runtime." This would probably entail *at least* the following:

1. We need a class member variable for [`resource_pool_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/schema/resource_data.hpp#L41) to keep track of each resource's up/down state. (Let's decide whether we want the state to represent in this member just as "up" or "down" or we need to represent some other state.)
2. We then augment [the traverser's availability check](https://github.com/flux-framework/flux-sched/blob/master/resource/schema/resource_data.hpp#L47), not only based on the available resource count but also based on the up/down state.
3. We need an efficient way to fetch a subgraph corresponding to each execution target. We probably can model this after `by_path` kind index maps. 
[...]

Issue #618, opened 2020-03-07
fault tolerance: need error propagation analysis


dongahn:
This will be  likely to be broken into multiple issues but I wanted to open this to remember this important items as we will work on "stabilization" tasks towards a tape out. Within `resource` and `qmanager`, there are some RPCs that leave the internal states inconsistent when a failure occurs. We need to analyze this more closely and have a clearer error handling semantics.

dongahn:
There are several call sites where return codes are not checked:

https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L461

dongahn:
There are a few error paths where `errno` is not preserved. We need to save and restore `errno` for library calls (e.g., `json_decref`) being made on the error paths. 

https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.cpp#L261

SteVwonder:
From #679:
> There are some error paths within the sched-fluxion-resource module that are allowing errors to pass by silently and responding to the request RPC with a successful response. In particular, this happens when the writer fails to emit properly. We should make these errors loud and respond to the RPC with an error.

dongahn:
Here is an additional problem:

Our DFU traverser concatenate one or more error strings to its `err_message` string member so that the upper layer can use `get_err_message()` to print it. There are several place which err_message string added has the newline character in the end, which doesn't work well with `flux_log_error`. An example:

```console
ahn1@5b12c7ea7263:/usr/src$ python3 t/scripts/flux-ion-resource.py find status=adown
2020-07-15T17:03:56.268829Z sched-fluxion-resource.err[0]: run_find: find: invalid criteria: status=adown.
2020-07-15T17:03:56.268864Z sched-fluxion-resource.err[0]: : Invalid argument
``` 

Look at the extra colon in front of "Invalid argument". We will need a way for the upper layer to iterate each error string to print out properly. 

This may also give a better way to resolve one of the pending issues: #409.

dongahn:
I want to spend a bit more time for this. Targeting Sep release.

PR #665, opened 2020-06-08
Resource status

milroy:
This PR addresses issue #662 item II.1 by creating an enum class as part of [`resource_pool_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/schema/resource_data.hpp#L41) that tracks resource status (e.g. `UP` and `DOWN`).  To read, change status values, and enable testing, this PR makes additional contributions:

1. enable the JGF reader to unpack the optional `status` key
2. a new `resource-query` command (`dump`) to write the resource graph to a file according to status
2. a new `resource-query` command (`set-status`) to perform a DFV update of a resource subtree status

A follow-up PR will enable scheduling by resource status based on the additions here.

### DIFF
--- a/resource/schema/resource_data.hpp
+++ b/resource/schema/resource_data.hpp
@@ -27,6 +27,7 @@
 #include <string>
 #include <cstring>
 #include <map>
+#include <unordered_map>
 #include <set>
 #include "resource/schema/color.hpp"
 #include "resource/schema/data_std.hpp"
@@ -44,6 +45,14 @@ struct resource_pool_t {
     resource_pool_t &operator= (const resource_pool_t &o);
     ~resource_pool_t ();
 
+    enum class status_t : int { 
+        UP = 0,
+        DOWN = 1
+    };
+
+    typedef std::unordered_map<std::string, status_t> string_to_status;
+    static const string_to_status str_to_status;
+
     // Resource pool data
     std::string type;
     std::map<std::string, std::string> paths;
@@ -58,6 +67,7 @@ struct resource_pool_t {
 
     schedule_t schedule;    //!< schedule data
     pool_infra_t idata;     //!< scheduling infrastructure data
+    status_t status = status_t::UP;
 };
 
 /*! Resource relationship type.
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -286,8 +286,20 @@ int dfu_traverser_t::remove (int64_t jobid)
         return -1;
     }
 
-    vtx_t root = get_graph_db ()->metadata.roots.at(dom);
+    vtx_t root = get_graph_db ()->metadata.roots.at (dom);
     return detail::dfu_impl_t::remove (root, jobid);
+}
+
+int dfu_traverser_t::mark (const std::string &root_path, 
+                           resource_pool_t::status_t status)
+{
+    return detail::dfu_impl_t::mark (root_path, status);
+}
+
+int dfu_traverser_t::mark (std::set<int64_t> &ranks, 
+                           resource_pool_t::status_t status)
+{
+    return detail::dfu_impl_t::mark (ranks, status);
 }
 
 /*
--- a/resource/traversers/dfu.hpp
+++ b/resource/traversers/dfu.hpp
@@ -148,6 +148,24 @@ public:
      */
     int remove (int64_t jobid);
 
+    /*! Mark the resource status up|down|etc starting at subtree_root.
+     *
+     *  \param root_path     path to the root of the subtree to update.
+     *  \param status        new status value
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: roots or by_path not found.
+     */
+    int mark (const std::string &root_path, resource_pool_t::status_t status);
+
+    /*! Mark the resource status up|down|etc for subgraph represented by ranks.
+     *
+     *  \param ranks         set of ranks representing the subgraph to update.
+     *  \param status        new status value
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: roots or by_path not found.
+     */
+    int mark (std::set<int64_t> &ranks, resource_pool_t::status_t status);
+
 private:
     int schedule (Jobspec::Jobspec &jobspec, detail::jobmeta_t &meta,
                   bool x, match_op_t op, vtx_t root,
--- a/resource/traversers/dfu_impl.hpp
+++ b/resource/traversers/dfu_impl.hpp
@@ -243,6 +243,25 @@ public:
      */
     int remove (vtx_t root, int64_t jobid);
 
+    /*! Update the resource status to up|down|etc starting at subtree_root.
+     *
+     *  \param root_path     path to the root of the subtree to update.
+     *  \param status        new status value
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: graph, roots or match callback not set.
+     */
+    int mark (const std::string &root_path, resource_pool_t::status_t status);
+
+    /*! Update the resource status to up|down|etc for subgraph 
+     *  represented by ranks.
+     *
+     *  \param ranks         set of ranks representing the subgraphs to update.
+     *  \param status        new status value
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: roots or by_path not found.
+     */
+    int mark (std::set<int64_t> &ranks, resource_pool_t::status_t status);
+
 private:
 
     /************************************************************************
--- a/resource/traversers/dfu_impl_update.cpp
+++ b/resource/traversers/dfu_impl_update.cpp
@@ -565,6 +565,59 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
     return (root_has_jtag)? rem_dfv (root, jobid) : rem_exv (jobid);
 }
 
+int dfu_impl_t::mark (const std::string &root_path, 
+                      resource_pool_t::status_t status)
+{
+    std::map<std::string, vtx_t>::const_iterator vit_root =
+        m_graph_db->metadata.by_path.find (root_path);
+
+    if (vit_root == m_graph_db->metadata.by_path.end ()) {
+        errno = EINVAL;
+        m_err_msg += __FUNCTION__;                
+        m_err_msg += ": could not find subtree path ("
+                  + root_path + ") in resource graph.\n";
+        return -1;
+    }
+    (*m_graph)[vit_root->second].status = status;
+    
+    return 0;
+}
+
+int dfu_impl_t::mark (std::set<int64_t> &ranks, 
+                      resource_pool_t::status_t status)
+{
+    std::map<int64_t, std::vector <vtx_t>>::iterator vit;
+    std::string subtree_path = "", tmp_path = "";
+    const std::string &dom = m_match->dom_subsystem ();
+    vtx_t subtree_root;
+
+    for (auto &rank : ranks) {
+        // Now iterate through subgraphs keyed by rank and 
+        // set status appropriately
+        vit = m_graph_db->metadata.by_rank.find (rank);
+        if (vit == m_graph_db->metadata.by_rank.end ()) {
+            errno = EINVAL;
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": could not find rank path in by_rank map.\n";
+            return -1;
+        }
+
+        subtree_root = vit->second.front ();
+        subtree_path = (*m_graph)[subtree_root].paths.at (dom);
+        for (vtx_t v : vit->second) {
+            // The shortest path string is the subtree root. 
+            tmp_path = (*m_graph)[v].paths.at (dom);
+            if (tmp_path.length () < subtree_path.length ()) {
+                subtree_path = tmp_path;
+                subtree_root = v;
+            }
+        }
+        (*m_graph)[subtree_root].status = status;
+    }
+
+    return 0;
+}
+
 /*
  * vi:tabstop=4 shiftwidth=4 expandtab
  */
--- a/resource/utilities/command.cpp
+++ b/resource/utilities/command.cpp
@@ -56,6 +56,8 @@ command_t commands[] = {
 "resource-query> set-property resource PROPERTY=VALUE" },
 { "get-property", "g", cmd_get_property, "Get all properties of a resource: "
 "resource-query> get-property resource" },
+{ "get-status", "e", cmd_get_status, "Get the graph resource vertex status: "
+"resource-query> get-status PATH_TO_VERTEX" },
     { "list", "l", cmd_list, "List all jobs: resource-query> list" },
     { "info", "i", cmd_info,
 "Print info on a jobid: resource-query> info jobid" },
@@ -418,6 +420,43 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
                     out << p_it->first << "=" << p_it->second << std::endl;
         }
     }
+    return 0;
+}
+
+int cmd_get_status (std::shared_ptr<resource_context_t> &ctx,
+                      std::vector<std::string> &args)
+{
+    if (args.size () != 2) {
+        std::cerr << "ERROR: malformed command" << std::endl;
+        return 0;
+    }
+    std::string vtx_path = args[1];
+    std::map<std::string, vtx_t>::const_iterator it =
+        ctx->db->metadata.by_path.find (vtx_path);
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+    std::string status = "";
+
+    if (it == ctx->db->metadata.by_path.end ()) {
+        std::cout << "Could not find path " << vtx_path
+                     << " in resource graph." << std::endl;
+        return 0;
+    }
+
+    for (auto &status_it : sts) {
+        if (status_it.second == ctx->db->resource_graph[it->second].status) {
+            status = status_it.first;
+            break;
+        }
+    }
+
+    if (status == "") {
+        std::cerr << "ERROR: vertex " << vtx_path
+                     << " has unknown status." << std::endl;
+        return 0;     
+    }
+
+    std::cout << vtx_path << " is " << status << std::endl;
+
     return 0;
 }
 
--- a/resource/utilities/command.cpp
+++ b/resource/utilities/command.cpp
@@ -56,6 +56,8 @@ command_t commands[] = {
 "resource-query> set-property resource PROPERTY=VALUE" },
 { "get-property", "g", cmd_get_property, "Get all properties of a resource: "
 "resource-query> get-property resource" },
+{ "set-status", "t", cmd_set_status, "Set resource status on vertex: "
+"resource-query> set-status PATH_TO_VERTEX {up|down}" },
 { "get-status", "e", cmd_get_status, "Get the graph resource vertex status: "
 "resource-query> get-status PATH_TO_VERTEX" },
     { "list", "l", cmd_list, "List all jobs: resource-query> list" },
@@ -421,6 +423,34 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
         }
     }
     return 0;
+}
+
+int cmd_set_status (std::shared_ptr<resource_context_t> &ctx,
+                      std::vector<std::string> &args)
+{
+    if (args.size () != 3) {
+        std::cerr << "ERROR: malformed command" << std::endl;
+        return 0;
+    }
+    std::string vtx_path = args[1];
+    std::string status = args[2];
+    std::map<std::string, vtx_t>::const_iterator it =
+        ctx->db->metadata.by_path.find (vtx_path);
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+
+    if (it == ctx->db->metadata.by_path.end ()) {
+        std::cout << "Could not find path " << vtx_path
+                     << " in resource graph." << std::endl;
+        return 0;
+    }
+
+    auto status_it = sts.find (status);
+    if (status_it == sts.end ()) {
+        std::cerr << "ERROR: unrecognized status" << std::endl;
+        return 0;
+    }
+
+    return ctx->traverser->mark (vtx_path, status_it->second);
 }
 
 int cmd_get_status (std::shared_ptr<resource_context_t> &ctx,
--- a/t/data/resource/commands/status/cmds07
+++ b/t/data/resource/commands/status/cmds07
@@ -1,0 +1,5 @@
+set-status /tiny0/rack0/node0 down
+get-status /tiny0/rack0/node0
+set-status /tiny0/rack0/node0 up
+get-status /tiny0/rack0/node0
+quit
--- a/t/t3024-resource-status.t
+++ b/t/t3024-resource-status.t
@@ -1,0 +1,75 @@
+#!/bin/sh
+
+test_description='Test Resource Status On Tiny Machine Configuration'
+
+. $(dirname $0)/sharness.sh
+
+cmd_dir="${SHARNESS_TEST_SRCDIR}/data/resource/commands/status"
+exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/status"
+grugs="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/tiny.graphml"
+grug_aux="${SHARNESS_TEST_SRCDIR}/data/resource/grugs/coarse_iobw.graphml"
+query="../../resource/utilities/resource-query"
+
+#
+# Selection Policy -- High ID first (-P high)
+#     The resource vertex with higher ID is preferred among its kind
+#     (e.g., node1 is preferred over node0 if available)
+#
+
+test001_desc="error on malformed get-status "
+test_expect_success "${test001_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds01 \
+2> 001.R.out &&
+    test_cmp 001.R.out ${exp_dir}/001.R.out
+'
+
+test002_desc="exit on get-status of nonexistent resource "
+test_expect_success "${test002_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds02 \
+> 002.R.out &&
+    test_cmp 002.R.out ${exp_dir}/002.R.out
+'
+
+test003_desc="get resource vertex status "
+test_expect_success "${test003_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds03 \
+> 003.R.out &&
+    test_cmp 003.R.out ${exp_dir}/003.R.out
+'
+
+test004_desc="error on malformed get-status "
+test_expect_success "${test004_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds04 \
+2> 004.R.out &&
+    test_cmp 004.R.out ${exp_dir}/004.R.out
+'
+
+test005_desc="exit on set-status of nonexistent resource "
+test_expect_success "${test005_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds05 \
+> 005.R.out &&
+    test_cmp 005.R.out ${exp_dir}/005.R.out
+'
+
+test006_desc="error on set-status of nonexistent status "
+test_expect_success "${test006_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds06 \
+2> 006.R.out &&
+    test_cmp 006.R.out ${exp_dir}/006.R.out
+'
+
+test007_desc="change resource vertex status and get output "
+test_expect_success "${test007_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds07 \
+> 007.R.out &&
+    test_cmp 007.R.out ${exp_dir}/007.R.out
+'
+
+test008_desc="change aux hierarchy resource vertex status and get output  "
+test_expect_success "${test008_desc}" '
+    ${query} -L ${grug_aux} -F jgf -S CA -P high < ${cmd_dir}/cmds08 \
+> 008.R.out &&
+    test_cmp 008.R.out ${exp_dir}/008.R.out
+'
+
+test_done
--- a/resource/readers/resource_reader_hwloc.cpp
+++ b/resource/readers/resource_reader_hwloc.cpp
@@ -102,6 +102,7 @@ vtx_t resource_reader_hwloc_t::add_new_vertex (resource_graph_t &g,
     g[v].name = basename + istr;
     g[v].paths[subsys] = prefix + "/" + g[v].name;
     g[v].idata.member_of[subsys] = "*";
+    g[v].status = resource_pool_t::status_t::UP;
 
     // Indexing for fast look-up
     m.by_path[g[v].paths[subsys]] = v;
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -44,6 +44,7 @@ struct fetch_helper_t {
     int64_t size = 0;
     int64_t uniq_id = 0;
     int exclusive = 0;
+    resource_pool_t::status_t status = resource_pool_t::status_t::UP;
     const char *type = NULL;
     const char *name = NULL;
     const char *unit = NULL;
@@ -186,11 +187,11 @@ int resource_reader_jgf_t::unpack_vtx (json_t *element, fetch_helper_t &f)
         m_err_msg += std::string (f.vertex_id) + ".\n";
         goto done;
     }
-    if ( (json_unpack (metadata, "{ s:s s:s s:s s:I s:I s:I s:b s:s s:I }",
+    if ( (json_unpack (metadata, "{ s:s s:s s:s s:I s:I s:I s?:i s:b s:s s:I }",
                                  "type", &f.type, "basename", &f.basename,
                                  "name", &f.name, "id", &f.id,
                                  "uniq_id", &f.uniq_id, "rank", &f.rank,
-                                 "exclusive", &f.exclusive,
+                                 "status", &f.status, "exclusive", &f.exclusive,
                                  "unit", &f.unit, "size", &f.size)) < 0) {
         errno = EPROTO;
         m_err_msg += __FUNCTION__;
@@ -245,6 +246,7 @@ vtx_t resource_reader_jgf_t::create_vtx (resource_graph_t &g,
     g[v].size = fetcher.size;
     g[v].uniq_id = fetcher.uniq_id;
     g[v].rank = fetcher.rank;
+    g[v].status = fetcher.status;
     g[v].id = fetcher.id;
     g[v].name = fetcher.name;
     g[v].properties = fetcher.properties;
--- a/resource/schema/resource_data.cpp
+++ b/resource/schema/resource_data.cpp
@@ -106,6 +106,10 @@ resource_relation_t::~resource_relation_t ()
 
 }
 
+const resource_pool_t::string_to_status resource_pool_t::str_to_status = 
+      { { "up", resource_pool_t::status_t::UP }, 
+      { "down", resource_pool_t::status_t::DOWN } };
+
 
 } // Flux::resource_model
 } // Flux
--- a/resource/utilities/command.hpp
+++ b/resource/utilities/command.hpp
@@ -91,6 +91,8 @@ int cmd_set_property (std::shared_ptr<resource_context_t> &ctx,
                       std::vector<std::string> &args);
 int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
                       std::vector<std::string> &args);
+int cmd_get_status (std::shared_ptr<resource_context_t> &ctx,
+                      std::vector<std::string> &args);
 int cmd_list (std::shared_ptr<resource_context_t> &ctx,
               std::vector<std::string> &args);
 int cmd_info (std::shared_ptr<resource_context_t> &ctx,
--- a/resource/utilities/command.hpp
+++ b/resource/utilities/command.hpp
@@ -91,6 +91,8 @@ int cmd_set_property (std::shared_ptr<resource_context_t> &ctx,
                       std::vector<std::string> &args);
 int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
                       std::vector<std::string> &args);
+int cmd_set_status (std::shared_ptr<resource_context_t> &ctx,
+                      std::vector<std::string> &args);
 int cmd_get_status (std::shared_ptr<resource_context_t> &ctx,
                       std::vector<std::string> &args);
 int cmd_list (std::shared_ptr<resource_context_t> &ctx,
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -57,6 +57,7 @@ TESTS = \
     t3021-resource-mtl3.t \
     t3022-resource-update.t \
     t3023-resource-update2.t \
+    t3024-resource-status.t \
     t4000-match-params.t \
     t4001-match-allocate.t \
     t4002-match-reserve.t \
--- a/t/data/resource/commands/status/cmds01
+++ b/t/data/resource/commands/status/cmds01
@@ -1,0 +1,2 @@
+get-status
+quit
--- a/t/data/resource/commands/status/cmds02
+++ b/t/data/resource/commands/status/cmds02
@@ -1,0 +1,2 @@
+get-status /dne
+quit
--- a/t/data/resource/commands/status/cmds03
+++ b/t/data/resource/commands/status/cmds03
@@ -1,0 +1,2 @@
+get-status /tiny0/rack0/node0
+quit
--- a/t/data/resource/commands/status/cmds04
+++ b/t/data/resource/commands/status/cmds04
@@ -1,0 +1,2 @@
+set-status
+quit
--- a/t/data/resource/commands/status/cmds05
+++ b/t/data/resource/commands/status/cmds05
@@ -1,0 +1,2 @@
+set-status /dne down
+quit
--- a/t/data/resource/commands/status/cmds06
+++ b/t/data/resource/commands/status/cmds06
@@ -1,0 +1,2 @@
+set-status /tiny0/rack0/node0 dwn
+quit
--- a/t/data/resource/commands/status/cmds08
+++ b/t/data/resource/commands/status/cmds08
@@ -1,0 +1,6 @@
+get-status /cz0/pfs0/bandwidth0
+set-status /cz0/pfs0/bandwidth0 down
+get-status /cz0/pfs0/bandwidth0
+set-status /cz0/pfs0/bandwidth0 up
+get-status /cz0/pfs0/bandwidth0
+quit
--- a/t/data/resource/expected/status/001.R.out
+++ b/t/data/resource/expected/status/001.R.out
@@ -1,0 +1,1 @@
+ERROR: malformed command
--- a/t/data/resource/expected/status/002.R.out
+++ b/t/data/resource/expected/status/002.R.out
@@ -1,0 +1,4 @@
+INFO: Loading a matcher: CA
+resource-query> get-status /dne
+Could not find path /dne in resource graph.
+resource-query> quit
--- a/t/data/resource/expected/status/003.R.out
+++ b/t/data/resource/expected/status/003.R.out
@@ -1,0 +1,4 @@
+INFO: Loading a matcher: CA
+resource-query> get-status /tiny0/rack0/node0
+/tiny0/rack0/node0 is up
+resource-query> quit
--- a/t/data/resource/expected/status/004.R.out
+++ b/t/data/resource/expected/status/004.R.out
@@ -1,0 +1,1 @@
+ERROR: malformed command
--- a/t/data/resource/expected/status/005.R.out
+++ b/t/data/resource/expected/status/005.R.out
@@ -1,0 +1,4 @@
+INFO: Loading a matcher: CA
+resource-query> set-status /dne down
+Could not find path /dne in resource graph.
+resource-query> quit
--- a/t/data/resource/expected/status/006.R.out
+++ b/t/data/resource/expected/status/006.R.out
@@ -1,0 +1,1 @@
+ERROR: unrecognized status
--- a/t/data/resource/expected/status/007.R.out
+++ b/t/data/resource/expected/status/007.R.out
@@ -1,0 +1,8 @@
+INFO: Loading a matcher: CA
+resource-query> set-status /tiny0/rack0/node0 down
+resource-query> get-status /tiny0/rack0/node0
+/tiny0/rack0/node0 is down
+resource-query> set-status /tiny0/rack0/node0 up
+resource-query> get-status /tiny0/rack0/node0
+/tiny0/rack0/node0 is up
+resource-query> quit
--- a/t/data/resource/expected/status/008.R.out
+++ b/t/data/resource/expected/status/008.R.out
@@ -1,0 +1,10 @@
+INFO: Loading a matcher: CA
+resource-query> get-status /cz0/pfs0/bandwidth0
+/cz0/pfs0/bandwidth0 is up
+resource-query> set-status /cz0/pfs0/bandwidth0 down
+resource-query> get-status /cz0/pfs0/bandwidth0
+/cz0/pfs0/bandwidth0 is down
+resource-query> set-status /cz0/pfs0/bandwidth0 up
+resource-query> get-status /cz0/pfs0/bandwidth0
+/cz0/pfs0/bandwidth0 is up
+resource-query> quit


milroy:
I am working on tests that should mitigate the reported drop in code coverage.

dongahn:
@milroy: just let me know when you want me to give my review for this PR.

milroy:
All tests are green and I dropped the "WIP" tag.  This PR is ready for your review, @dongahn.

resource/schema/resource_data.hpp
```diff
@@ -37,6 +38,14 @@
 namespace Flux {
 namespace resource_model {
 
+enum class resource_status_t : int { 
+    up = 0,
```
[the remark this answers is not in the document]
milroy: That sounds reasonable.  I'll try it and report back.

resource/schema/resource_data.hpp
```diff
@@ -37,6 +38,14 @@
...
+    down = 1
+};
+
+static std::unordered_map<std::string, resource_status_t> str_to_status =
```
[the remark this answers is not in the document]
milroy: > When you move its definition to a source file, static wouldn't work if a function defined in another source file uses this variable.

That was my thought, but your idea making `str_to_status` a class static member of `resource_pool_t` sounds like the preferred approach.

resource/utilities/command.cpp
```diff
@@ -62,6 +63,7 @@ command_t commands[] = {
     { "stat", "s", cmd_stat,
  "Print overall stats: resource-query> stat jobid" },
     { "cat", "a", cmd_cat, "Print jobspec file: resource-query> cat jobspec" },
+    { "dump", "dm", cmd_dump_graph, "Dump the graph to file: resource-query> dump /outfile/path up|down|any" },
```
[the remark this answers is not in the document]
milroy: I will implement these suggestions.

resource/utilities/command.cpp
```diff
@@ -493,6 +495,72 @@ int cmd_cat (std::shared_ptr<resource_context_t> &ctx,
...
+    f_resource_graph_t fg = *(ctx->fgraph);
+    std::unordered_set<vtx_t> vtx_set;
+
+    std::string out_file = args[1];
+    std::string status = args[2];
+    std::ofstream out (out_file);
+
+    if (status == "any") {
+        for (tie (vi, v_end) = vertices (fg); vi != v_end; ++vi) {
+            if ( (rc = ctx->writers->emit_vtx ("", fg, *vi,
```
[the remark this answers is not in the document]
milroy: Thanks for the style guide; I've been trying to find something like this.

resource/utilities/command.cpp
```diff
@@ -493,6 +495,72 @@ int cmd_cat (std::shared_ptr<resource_context_t> &ctx,
...
+                        goto done;
+                    }
+                    vtx_set.insert (*vi);
+                }
+            }
+        }
+    }
+
+    for (tie (ei, e_end) = edges (fg); ei != e_end; ++ei) {
+        if ( (vtx_set.count (target (*ei, fg)) > 0) &&
```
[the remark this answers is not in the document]
milroy: [`count`](https://en.cppreference.com/w/cpp/container/unordered_map/count)'s complexity is "constant on average, worst case linear in the size of the container."  I don't know what constant it averages to, though.

I chose `count` because the `find` involves an additional check for equivalence with `.end ()`.

resource/utilities/command.cpp
```diff
@@ -423,6 +425,39 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
...
+        std::cerr << "ERROR: malformed command" << std::endl;
+        return 0;
+    }
+
+    std::string status = args[1];
+    std::string subtree_path = args[2];
+    std::string parent = "";
+    std::map<std::string, vtx_t>::const_iterator vit =
+        ctx->db->metadata.by_path.find (subtree_path);
+    vtx_t parent_vtx;
```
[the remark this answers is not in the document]
milroy: Oops, good catch.

resource/traversers/dfu.hpp
```diff
@@ -148,6 +148,18 @@ class dfu_traverser_t : protected detail::dfu_impl_t
...
 
+    /*! Update the resource status to up|down|etc starting at subtree_root.
+     *
+     *  \param subtree_root  root of the subtree to update.
+     *  \param status        new status value
+     *  \param patent        parent path of subtree_root to stop upward traversal
+     *  \return          0 on success; -1 on error.
+     *                       EINVAL: graph, roots or match callback not set.
+     */
+    int update_status (vtx_t subtree_root,
```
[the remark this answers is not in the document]
milroy: I will do that!

resource/traversers/dfu_impl_update.cpp
```diff
@@ -543,6 +572,16 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
     return (root_has_jtag)? rem_dfv (root, jobid) : rem_exv (jobid);
 }
 
+int dfu_impl_t::update_status (vtx_t subtree_root, vtx_t parent,
```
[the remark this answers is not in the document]
milroy: I agree and am working on a solution for this issue now.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -451,6 +457,29 @@ int dfu_impl_t::rem_exv (int64_t jobid)
...
 }
 
+int dfu_impl_t::status_dfv (vtx_t u, resource_status_t &status)
+{
+    int rc = 0;
+    const std::string &dom = m_match->dom_subsystem ();
+    f_out_edg_iterator_t ei, ei_end;
+
+    (*m_graph)[u].status = status;
+    (*m_graph)[u].idata.colors[dom] = m_color.black ();
```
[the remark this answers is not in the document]
milroy: It doesn't appear to be necessary.  `tick()` also does this: `m_best_k_cnt++` which I think is undesirable here.

resource/utilities/command.cpp
```diff
@@ -423,6 +425,39 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
...
+        return 0;
+    }
+
+    if (vit == ctx->db->metadata.by_path.end ()) {
+        std::cout << "Could not find subtree path " << subtree_path
+            << " in resource graph." << std::endl;
+        return 0;
+    }
+
+    parent = subtree_path.substr (0, subtree_path.length ()
```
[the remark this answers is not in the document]
milroy: Agreed.

milroy:
I was attempting to get this PR out quickly and it suffered for my haste.  @dongahn thank you for the feedback and direction; I'll address your changes ASAP!

resource/schema/resource_data.hpp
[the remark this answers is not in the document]
milroy: Ok, I've got this working and committed in my dev branch.

resource/utilities/command.cpp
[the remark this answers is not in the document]
milroy: I implemented the suggestions, but I think the `dump ' ' outfile` semantics is a bit odd.  If it's okay, I'd like to leave in the `ANY` option.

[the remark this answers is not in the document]
dongahn: Yes that's fine to me. I got that idea from `jq`. `jq ' ' file` would dump the whole thing with its predefined format.

milroy:
@dongahn I've made pretty substantial changes to this PR per your change requests, and have squashed my fixes.  This PR is ready for the second-round review.

Note that I am unsure if I've handled the parent pruning filter updates properly in `mark_accum_to_parent`.

resource/utilities/command.cpp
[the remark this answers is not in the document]
milroy: It turns out it doesn't actually need to change.

resource/utilities/command.cpp
```diff
@@ -62,6 +63,8 @@ command_t commands[] = {
     { "stat", "s", cmd_stat,
  "Print overall stats: resource-query> stat jobid" },
     { "cat", "a", cmd_cat, "Print jobspec file: resource-query> cat jobspec" },
+    { "dump", "d", cmd_dump_graph, "Dump the graph by status to file or stdout "
+"if file is not specified: resource-query> dump [UP|DOWN|ANY] /outfile/path" },
```
dongahn, line 69: Thank you for reordering @milroy. This is not a Unix utility, but it may be helpful to following its conversion. `[ ]` typically notates optional. So perhaps this can look a bit more conventional with:

`dump {UP|DOWN|ANY} [outfile_path]`?

resource/utilities/command.cpp
```diff
@@ -493,6 +556,49 @@ int cmd_cat (std::shared_ptr<resource_context_t> &ctx,
...
+
+    int rc = -1;
+    std::stringstream o;
+    std::ofstream out_stream;
+    std::streambuf *out_buf;
+
+    std::string status = args[1];
+
+    if (args.size () == 3) {
+        out_stream.open (args[2]);
```
dongahn, line 610: Need to check an error when `open` fails?

resource/utilities/command.cpp
```diff
@@ -286,6 +289,66 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+    f_vtx_iterator_t vi, v_end;
+    f_edg_iterator_t ei, e_end;
+    f_resource_graph_t fg = *(ctx->fgraph);
+    std::unordered_set<vtx_t> vtx_set;
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+
+    if (status == "ANY") {
+        for (tie (vi, v_end) = vertices (fg); vi != v_end; ++vi) {
+            if (ctx->writers->emit_vtx ("", fg, *vi, 
+                  fg[*vi].size, false) < 0) {
```
dongahn, line 307: Adjust this line to the first argument of the line above if this still fits within 80 characters?

resource/utilities/command.cpp
```diff
@@ -286,6 +289,66 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+                std::cerr << "ERROR: writer emit vertex: " 
+                            << strerror (errno) << std::endl;
+                goto done;
+            }
+            vtx_set.insert (*vi);
```
dongahn, line 312: You can add an error check. Though it is less likely you will get an error but in case BGL has a bug.

```c++
auto ret = vtx_set.insert (*vi);
if (!ret.second) {
    std::cerr << "ERROR: a duplicate vertex detected"
    goto done;
}

File: resource/utilities/command.cpp [hunk 15]
```diff
@@ -286,6 +289,66 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+        }
+    }
+    else {
+        auto status_it = sts.find (status);
```
dongahn, line 316: Neat way to make this future proof!

resource/utilities/command.cpp
```diff
@@ -286,6 +289,66 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+        if (status_it == sts.end ()) {
+            std::cerr << "ERROR: unrecognized status" << std::endl;
+            goto done;
+        } 
+        else {
+            for (tie (vi, v_end) = vertices (fg); vi != v_end; ++vi) {
+                if (fg[*vi].status == status_it->second) {
+                    if (ctx->writers->emit_vtx ("", fg, *vi, 
+                           fg[*vi].size, false) < 0) {
```
dongahn, line 325: Will aligning this to the first argument of the line above function still fit the whole ting to 80 characters? Otherwise, it is good as is.

resource/utilities/command.cpp
```diff
@@ -286,6 +289,66 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+                        std::cerr << "ERROR: writer emit vertex: " 
+                                    << strerror (errno) << std::endl;
+                        goto done;
+                    }
+                    vtx_set.insert (*vi);
```
dongahn, line 330: Same as above;

resource/utilities/command.cpp
```diff
@@ -57,6 +57,8 @@ command_t commands[] = {
 "resource-query> set-property resource PROPERTY=VALUE" },
 { "get-property", "g", cmd_get_property, "Get all properties of a resource: "
 "resource-query> get-property resource" },
+{ "set-status", "t", cmd_set_status, "Set resource status on subgraph: "
+"resource-query> set-status /path/to/subtree [UP|DOWN]" },
```
dongahn, line 61: `set-status /path/to/subtree {UP|DOWN}`?

resource/utilities/command.cpp
milroy: Good point.  I'll do that.

resource/utilities/command.cpp
milroy: I definitely should have!  I'll be sure to include a check.

resource/utilities/command.cpp
milroy: Unfortunately it doesn't quite fit: it's 81 characters.

resource/traversers/dfu.hpp
```diff
@@ -148,6 +148,28 @@ class dfu_traverser_t : protected detail::dfu_impl_t
...
+     *  \param subtree_root  root of the subtree to update.
+     *  \param status        new status value
+     *  \param patent        parent path of subtree_root to stop upward traversal
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: roots or by_path not found.
+     */
+    int mark (vtx_t subtree_root, 
+              const resource_pool_t::status_t &status,
+              std::string &parent);
+
```
dongahn, line 162: @milroy: I can see why you chose this method signature (easy of implementation). But I think it is asking too much from the client code as a method within the high level traversal class.

I am thinking it may still suffice with:

```c++
int mark (const std::string path, resource_pool_t::status_t status);
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +621,45 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
...
+    
+    m_color.reset ();
+    (*m_graph)[parent].idata.colors[dom] = m_color.black ();
+    if ( (mark_dfv (subtree_root, status, to_parent)) < 0) {
+         m_err_msg += __FUNCTION__;
+         m_err_msg += ": mark_dfv returned < 0.\n";
+         return -1;
+    }
+
+    while (parent_path != "") {
```
dongahn, line 640: This code essentially implement `propagation` operation which you see from an algorithm like red black tree. (https://www.kernel.org/doc/Documentation/rbtree.txt).

It would be great if this code is factored out as its down method called `propagate()`. For now this, we can use the path name based search as is. But some point, we want to think about how we can require our graph topology so that `propagate()` can be done on any subsystem. If the graph in a subsystem is bidirectional, we should be easily find the parent vertex of any vertex within that subsystem. But it is not clear whether we can do this w/ directional graph...

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +621,45 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
     return (root_has_jtag)? rem_dfv (root, jobid) : rem_exv (jobid);
 }
 
+int dfu_impl_t::mark (vtx_t subtree_root, vtx_t parent,
+                      std::string parent_path,
+                      const resource_pool_t::status_t &status)
+{
+    const std::string &dom = m_match->dom_subsystem ();
```
dongahn, line 628: Related to my comment at the `dfu.hpp`, is it possible to retrieve `subtree_root` and `parent` without having to require the client code pass that in? 

If we can, this would simplify the method signature.

resource/traversers/dfu.cpp
```diff
@@ -286,10 +286,49 @@ int dfu_traverser_t::remove (int64_t jobid)
         return -1;
     }
 
-    vtx_t root = get_graph_db ()->metadata.roots.at(dom);
+    vtx_t root = get_graph_db ()->metadata.roots.at (dom);
     return detail::dfu_impl_t::remove (root, jobid);
 }
 
+int dfu_traverser_t::mark (vtx_t subtree_root, 
+                           const resource_pool_t::status_t &status,
```
dongahn, line 294: [Since status is just an integer, I think it would be a bit better to pass status as `resource_pool_t::status_t status` instead of const reference.]( https://stackoverflow.com/questions/44749658/c-is-it-better-to-pass-a-enum-class-as-value-or-const-reference/44749716)

resource/traversers/dfu.cpp
```diff
@@ -286,10 +286,49 @@ int dfu_traverser_t::remove (int64_t jobid)
...
+        }
+        parent_vtx = vit->second;
+    }
+
+    return detail::dfu_impl_t::mark (subtree_root, parent_vtx, 
+                                     parent, status);
+}
+
+int dfu_traverser_t::mark (std::set<int64_t> ranks, 
+                           const resource_pool_t::status_t &status)
```
dongahn, line 326: Same for this `status`.

resource/traversers/dfu_impl.hpp
```diff
@@ -243,6 +243,29 @@ class dfu_impl_t {
...
 
+    /*! Update the resource status to up|down|etc starting at subtree_root.
+     *
+     *  \param subtree_root  root of the subtree to update.
+     *  \param patent        parent of subtree_root to stop upward traversal
+     *  \param status        new status value
+     *  \return              0 on success; -1 on error.
+     *                       EINVAL: graph, roots or match callback not set.
+     */
+    int mark (vtx_t subtree_root, vtx_t parent,
```
dongahn, line 254: Same comment as above. However, this is an implementation level API, I would worry less about the API complexity.

resource/traversers/dfu_impl_update.cpp
milroy: Yes, that will work really well because I need to do the same for the Not Yet Implemented `mark` overload.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -176,6 +176,26 @@ int dfu_impl_t::accum_to_parent (vtx_t u, const subsystem_t &subsystem,
     return 0;
 }
 
+int dfu_impl_t::mark_accum_to_parent (vtx_t u, const subsystem_t &subsystem,
+                                    const std::map<std::string, int64_t> &dfu,
+                                    std::map<std::string, int64_t> &to_parent)
+{
+    // Build up the new aggregates that will be used by subtree
+    // aggregate pruning filter. If exclusive, none of the vertex's resource
```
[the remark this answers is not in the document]
milroy: Oops, that's a remnant of copying the function from `accum_to_parent`

milroy:
> If you want to go through some of the involved suggestions, let's chat by phone.

Yes, let's do that.  I'll send you an email.

resource/traversers/dfu_impl_update.cpp
dongahn: Yes, that's sort of the idea. 

Just as a context:

From my perspective, one of the compelling scalable properties of a tree is the fact that a vertex can scalably keep track of certain invariants about the subtree rooted at that vertex. Each invariant accurately tracks a certain function of the subtree values (e.g., min, max, sum, integer multiplication... -- in fact any function that can be used as a "reduce" operator) and only `O(tree height)` update cost is required to keep the invariant consistent on each vertex update.  

Fluxion supports more than a tree, of course. But I restricted our graph topology such that, given a single subsystem, its belonging subgraph topology is a tree so I took advantage of this tree property and applied that to aggregate-based pruning filter update. (In the future, we can look at ways to add more invariant for more aggressive search pruning using the same infrastructure if needed). 

BTW, I didn't explicitly introduce `propagate()` either. But I realize that abstracting out invariant updates to the rightly named function can improve readability as well as our longer-term future directions.

As we more rigoriously abstract these graph algorithms into good abstractions, it will open up our door to make Fluxion resource service distributed (or parallel) if and when more scalability is needed. For example, one graph partitioning approach can be: Fluxion resource service at rank 0 only maintains higher level subgraph and other ranks takes care of lower level subgraphs. And the distributed match/update sequence will be very nicely defined.

This is a longer-term research topic. 

The more we defined graph operations as mathematically (and rigorously defined) now, the better to parallelize our service in the future IMHO. Example parallel trees would be MRNet and (not surprisingly) MPI. 

http://www.paradyn.org/mrnet/
https://mpitutorial.com/tutorials/mpi-reduce-and-allreduce/


This is a bit related to your cluster submission. But this is only for a single instance scalability.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -176,6 +176,26 @@ int dfu_impl_t::accum_to_parent (vtx_t u, const subsystem_t &subsystem,
...
+    // is available (0). If not, all will be available (needs).
+    if ((*m_graph)[u].status == resource_pool_t::status_t::DOWN)
```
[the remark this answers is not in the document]
milroy: That is a good plan.  I will make this change.

resource/traversers/dfu_impl_update.cpp
grondo: > As we more rigoriously abstract these graph algorithms into good abstractions, it will open up our door to make Fluxion resource service distributed (or parallel) if and when more scalability is needed.

This sounds great! It would be really neat to do a proof of concept of this feature, since it was one of our earliest ideas for Flux!

dongahn: @grondo: exactly! 

I just chatted with @milroy for this PR and he and I agree this should be an excellent research contribution -- there is nothing like this out there and can inform the future "reduction to practice." 

I have an idea about how this research should be conducted. I may try to get a good summer intern and do a proof of concept as well. This will be a very challenging task, but the payoff will be awesome in my mind if successful.

resource/utilities/command.cpp
milroy: Unfortunately, no.  Doing so extends the line to 89 characters

resource/traversers/dfu_impl_update.cpp
```diff
@@ -176,6 +176,26 @@ int dfu_impl_t::accum_to_parent (vtx_t u, const subsystem_t &subsystem,
...
+        accum_if (subsystem, (*m_graph)[u].type, (*m_graph)[u].size, 
+                    to_parent);
+
+    // Pass up the new subtree aggregates collected so far to the parent.
+    for (auto &kv : dfu)
+        accum_if (subsystem, kv.first, kv.second, to_parent);
+
+    return 0;
+}
+
```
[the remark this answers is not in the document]
milroy: I've implemented and tested this change, and it works as intended.

resource/traversers/dfu_impl_update.cpp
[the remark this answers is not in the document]
milroy: Obviated by the removal of `mark_accum_to_parent`.

resource/traversers/dfu_impl_update.cpp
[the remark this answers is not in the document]
milroy: This suggestion appears to conflict with your suggestion of eliminating the need for `mark_accum_to_parent` by modifying `accum_to_parent`.  If we only have `accum_to_parent`, then the `resource_pool_t::status_t` must be mapped to a `bool`.  

I can add a `switch` in the expectation that if there are multiple statuses that need to be mapped to values beyond `bool` in the future, then we will need a distinct `mark_accum_to_parent` to handle the cases.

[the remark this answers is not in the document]
dongahn: Yeah, this was a suggestion before my "merge" suggestion. Please disregard it.

milroy:
@dongahn this is ready for another pass.  After further consideration, I decided that an iteration through the vertices in a subgraph followed by `mark_dfv` was the cleanest approach for the `mark` overload.

@SteVwonder you can start your review when the Travis tests come back green.

milroy:
Note that the Codecov report reflects that the `mark` overload used by `sched-fluxion-resource` in PR #675 is not covered.  I can add test cases in this PR, but I'll probably need to add more functionality to `resource-query`.

resource/traversers/dfu_impl.hpp
```diff
@@ -368,6 +390,18 @@ class dfu_impl_t {
     int rem_upv (vtx_t u, int64_t jobid);
     int rem_dfv (vtx_t u, int64_t jobid);
     int rem_exv (int64_t jobid);
+    int mark_upv (vtx_t u, const resource_pool_t::status_t status,
+                  std::map<std::string, int64_t> &to_parent);
+    int mark_dfv (vtx_t u, const resource_pool_t::status_t status,
+                  std::map<std::string, int64_t> &to_parent);
+
+    int propagate (vtx_t parent, std::string &parent_path,
+                   const resource_pool_t::status_t status,
```
dongahn, line 399:
```suggestion
                   resource_pool_t::status_t status,
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -157,14 +157,15 @@ int dfu_impl_t::upd_plan (vtx_t u, const subsystem_t &s, unsigned int needs,
...
                                  std::map<std::string, int64_t> &to_parent)
 {
     // Build up the new aggregates that will be used by subtree
-    // aggregate pruning filter. If exclusive, none of the vertex's resource
-    // is available (0). If not, all will be available (needs).
-    if (excl)
+    // aggregate pruning filter. If exclusive or the status is DOWN, 
+    // none of the vertex's resource is available (0). 
+    // If not, all will be available (needs).
+    if (unavail)
```
dongahn, line 168: This is just my note:

Currently `accum_if` does not set errno when a failure happens and so this function always returns success. This needs to be improved. Can you open a ticket for this?

resource/traversers/dfu_impl_update.cpp
```diff
@@ -451,6 +459,123 @@ int dfu_impl_t::rem_exv (int64_t jobid)
...
+                rc += mark_dfv (tgt, status, dfu);
+            else
+                rc += mark_upv (tgt, status, dfu);
+        }
+    }
+
+    if (accum_to_parent (u, dom, needs, unavail, dfu, to_parent) < 0) {
+        m_err_msg += __FUNCTION__;
+        m_err_msg += ": accum_to_parent returned < 0.\n";
+        rc = -1;
```
dongahn, line 497: It would be a bit better to accumulate rc: `rc -= 1;`

Because `accum_to_parent` doesn't currently returns a neg (nor set errno), this is a somewhat moot point, though. Maybe we should revisit error propagation in a one big PR later...

resource/traversers/dfu_impl_update.cpp
```diff
@@ -451,6 +459,123 @@ int dfu_impl_t::rem_exv (int64_t jobid)
...
+        case resource_pool_t::status_t::UP:
+            needs = (*m_graph)[parent].size;
+        case resource_pool_t::status_t::DOWN:
+            unavail = true;
+    }
+
+    while (parent_path != "") {
+        if ( (accum_to_parent (parent, dom, needs, unavail, dfu, 
+                                 to_parent)) < 0) {
+             m_err_msg += __FUNCTION__;
```
dongahn, line 522: Same comment as above for my own note.

resource/utilities/command.cpp
```diff
@@ -57,6 +57,8 @@ command_t commands[] = {
 "resource-query> set-property resource PROPERTY=VALUE" },
 { "get-property", "g", cmd_get_property, "Get all properties of a resource: "
 "resource-query> get-property resource" },
+{ "set-status", "t", cmd_set_status, "Set resource status on subgraph: "
+"resource-query> set-status [path_to_subtree] {UP|DOWN}" },
```
dongahn, line 61: Since `path_to_subtree` not optional, maybe `PATH_TO_SUBTREE` with no `[]` is a bit better?

dongahn:
@milroy. Thanks. I left some comments that include method parameter types, safety of `[]` method and error propagation. 

Since you now have a functional code, I will rebase PR #675 on top of this and see if I find anything unusual as well.

resource/traversers/dfu.cpp
```diff
@@ -286,10 +286,23 @@ int dfu_traverser_t::remove (int64_t jobid)
         return -1;
     }
 
-    vtx_t root = get_graph_db ()->metadata.roots.at(dom);
+    vtx_t root = get_graph_db ()->metadata.roots.at (dom);
     return detail::dfu_impl_t::remove (root, jobid);
 }
 
+int dfu_traverser_t::mark (const std::string root_path,
```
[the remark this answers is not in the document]
milroy: Ugh, I completely missed this.  Thanks!

resource/traversers/dfu.cpp
```diff
@@ -286,10 +286,23 @@ int dfu_traverser_t::remove (int64_t jobid)
...
     return detail::dfu_impl_t::remove (root, jobid);
 }
 
+int dfu_traverser_t::mark (const std::string root_path, 
+                           const resource_pool_t::status_t status)
+{
+    return detail::dfu_impl_t::mark (root_path, status);
+}
+
+int dfu_traverser_t::mark (std::set<int64_t> ranks,
```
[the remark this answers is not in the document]
milroy: You're right; I don't want to do that.

resource/traversers/dfu_impl_update.cpp
milroy: No problem.  I'll get to it ASAP.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +690,76 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
...
+        // set status appropriately
+        vit = m_graph_db->metadata.by_rank.find (rank);
+        if (vit == m_graph_db->metadata.by_rank.end ()) {
+            std::cout << "Could not find rank path " << rank
+                         << " in by_rank map.\n";
+            return -1;
+        }
+
+        subtree_root = vit->second.front ();
+        subtree_path = (*m_graph)[subtree_root].paths.at (dom);
```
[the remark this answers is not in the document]
milroy: Unfortunately Fluxion won't compile with this change.  The error originates from Boost and amounts to this:

```
has no member named 'at'
subtree_path = (*m_graph).at (subtree_root).paths.at (dom);
```

There are many examples in `dfu_impl_t.cpp` with statements such as:

```x_checker = (*m_graph)[u].idata.x_checker;```

so I'm guessing people have run into this before.

**Comment by milroy** [comment]
@dongahn I think I've addressed your changes from the latest round, except for the `[]` versus `.at ()` changes.  As I noted in the comments, BGL doesn't appear to support `.at ()` for `adjacency_list`.

File: resource/traversers/dfu_impl_update.cpp (hunk 35 above)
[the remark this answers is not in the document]
Comment (dongahn) [reply resource/traversers/dfu_impl_update.cpp:748]: Using `[x]` is fine when there is a guarantee that `x` exists. For example, the source of `x` was from the container itself (in the case of graph, graph iteration). Or Wou already checked the presence of the element with `find`.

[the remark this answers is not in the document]
Comment (dongahn) [reply resource/traversers/dfu_impl_update.cpp:748]: > Unfortunately Fluxion won't compile with this change. The error originates from Boost Graph Library and amounts to this:

Ah... BGL probably overrode this operator. We will have to review the BGL reference or implementation. I will look.

File: resource/traversers/dfu_impl_update.cpp (hunk 29 above)
Comment (milroy) [reply resource/traversers/dfu_impl_update.cpp:168]: I created issue #680 from this improvement request.

File: t/t3024-resource-status.t [hunk 36]
```diff
@@ -0,0 +1,58 @@
...
+#
+# Selection Policy -- High ID first (-P high)
+#     The resource vertex with higher ID is preferred among its kind
+#     (e.g., node1 is preferred over node0 if available)
+#
+
+cmds001="${cmd_dir}/cmds01.in"
+test001_desc="dump subgraph with any resource status "
+test_expect_success "${test001_desc}" '
+    sed "s~@TEST_SRCDIR@~${SHARNESS_TEST_SRCDIR}~g" ${cmds001} > cmds001 &&
```
SteVwonder, line 21: Since these command files don't require the variable substitution for the `${SHARNESS_TEST_SRCDIR}` like other tests and can be used "as-is", my recommendation would be to drop the `.in` suffix on the command files checked into version control, and remove these sed lines from the tests.

resource/utilities/command.cpp
```diff
@@ -286,6 +291,75 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+                std::cerr << "ERROR: writer emit edge: " 
+                            << strerror (errno) << std::endl;
+                goto done;
+            }
+        }
+    }
+
+    rc = 0;
+done:
+    return 0;
```
SteVwonder, line 360:
```suggestion
    return rc;
```

When I compiled, my compiler warned that:
```
command.cpp: In function ‘int Flux::resource_model::emit_vtx_edg(std::shared_ptr<Flux::resource_model::resource_context_t>&, std::stringstream&, const string&)’:
command.cpp:297:9: warning: variable ‘rc’ set but not used [-Wunused-but-set-variable]
     int rc = -1;
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +695,79 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
...
+    if ( (mark_from_parent (subtree_root, parent_path, status)) < 0) {
+         m_err_msg += __FUNCTION__;
+         m_err_msg += ": mark_from_parent returned < 0.\n";
+         return -1;
+    }
+    
+    return 0;
+}
+
+int dfu_impl_t::mark (std::set<int64_t> &ranks,
```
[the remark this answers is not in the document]
milroy: Not currently.  @dongahn is adding the user in PR #675.  This function should be covered when that PR's tests are merged.

resource/utilities/command.cpp
```diff
@@ -62,6 +63,8 @@ command_t commands[] = {
     { "stat", "s", cmd_stat,
  "Print overall stats: resource-query> stat jobid" },
     { "cat", "a", cmd_cat, "Print jobspec file: resource-query> cat jobspec" },
+    { "dump", "d", cmd_dump_graph, "Dump the graph by status to file or stdout "
+"if file is not specified: resource-query> dump {UP|DOWN|ANY} [outfile_path]" },
```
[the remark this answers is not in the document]
milroy: Yes, the `match-format` is used; you raise a good point.  I'll update the help string.

resource/utilities/command.cpp
```diff
@@ -493,6 +495,28 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
...
+    std::string parent = "";
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+
+    auto status_it = sts.find (status);
+    if (status_it == sts.end ()) {
+        std::cerr << "ERROR: unrecognized status" << std::endl;
+        return 0;
+    }
+
+    return ctx->traverser->mark (subtree_path, status_it->second);
```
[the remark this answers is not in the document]
dongahn: @milroy asked me to comment:

Yes, if `mark` returns a negative return code, `resource-query` should report the error message by using `err_message()` and perhaps clear the message after the print.

`sched-fluxion-resource` will do the same thing but it will also make use of `errno` so all the error paths should set `errno` appropriately. If there are cases where `errno` isn't appropriately set, we should at least add that deficiency to PR #618.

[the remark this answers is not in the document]
milroy: > `sched-fluxion-resource` will do the same thing but it will also make use of `errno` so all the error paths should set `errno` appropriately. 

@dongahn are you referring to the case where `mark` handles resources not in the dominant hierarchy?  I agree with @SteVwonder that `mark` should return an error in this case at least for now.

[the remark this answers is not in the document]
dongahn: I was just referring to general errno handling semantics. What happens if the `subtree_path` is defined on an auxiliary subsystem at this point? Why can't this be supported?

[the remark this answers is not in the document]
milroy: Currently, 
```C++
int dfu_impl_t::mark (std::set<int64_t> &ranks, 
                      resource_pool_t::status_t status)
``` 
and `dfu_impl_t::mark_from_parent` set the dominant subsystem string via:

```C++ 
const std::string &dom = m_match->dom_subsystem ();
```
Now `dom` is used to find each ancestor of `subtree_root` in the `mark` overload, and is used to reset the graph coloring and to conditionally set `parent_vtx` in `mark_from_parent`.  Furthermore, `dfu_impl_t::mark_dfv` (in analogy with also sets `dom` via `m_match->dom_subsystem ()` and uses it for coloring.

In short, the functions assume that the subtree traversals will occur within the dominant hierarchy.  However, I can generalize the functions to use any subsystem by identifying the subsystem from the `subtree_path`.

What are your thoughts on that approach?

resource/utilities/command.cpp
```diff
@@ -286,6 +289,75 @@ static int update (std::shared_ptr<resource_context_t> &ctx,
...
+    }
+    else {
+        auto status_it = sts.find (status);
+        if (status_it == sts.end ()) {
+            std::cerr << "ERROR: unrecognized status" << std::endl;
+            goto done;
+        } 
+        else {
+            for (tie (vi, v_end) = vertices (fg); vi != v_end; ++vi) {
+                if (fg[*vi].status == status_it->second) {
```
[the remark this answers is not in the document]
milroy: No objection; I'll implement your change.

milroy:
@SteVwonder we've covered these items in our conference calls, but answer here, too.  

> As a point of clarification, this PR is not meant to prevent jobs from being scheduled on DOWN resources, right? It is to just mark resources as DOWN, and the effects of status on scheduling will be added later?

Yes to both questions.  Using the status to make scheduling decisions is PR #667.

> My concern here is that I think the accumulation to parent is happening even if the resource was already marked DOWN and is getting marked DOWN a second time (e.g., you mark a socket down, then you mark the whole node down, the socket and its subtree will be subtracted from the pruner twice, right?).

The subtree totals aren't subtracted when the state is set to `DOWN`.  The counts are set to zero in [dfu_impl_t::accum_if](https://github.com/flux-framework/flux-sched/blob/39e263f9a1febbc61d45df8d22607f051e6772da/resource/traversers/dfu_impl.cpp#L316) the same as if exclusivity is true.  The counts are accumulated if the status is set to `UP`.

> Is the marking idempotent?

I think setting resources `DOWN` will be idempotent, but I'm not sure what will happen with `UP`.  I will write some testsuite checks for idempotence.  @dongahn is there a better way to test pruning filter correctness?

resource/utilities/command.cpp
milroy: Good catch!

milroy: Actually I want to return 0 to return REPL control, so `return 0` is correct, but there's not need for `rc`.

resource/utilities/command.cpp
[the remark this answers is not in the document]
milroy: Unexpectedly, this change results in a segfault when `status == "ANY"` resolves to `true`.  Any ideas why that might be?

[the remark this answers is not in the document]
dongahn: Perhaps `status` is not string?

dongahn:
@milroy: Just FYI -- https://github.com/flux-framework/flux-sched/pull/684/commits/457493825f9060df6d5c37749af48af126a6868f adds status to string as well. It doesn't need to be moved to this PR though.

dongahn:
@SteVwonder makes really good point about pruning filter update. I was looking more at coding aspects of this PR and somehow glossed over the pruning filter implications. 

The pruning filter algorithm is:
1. Each vertex *can* have a pruning filter per each subsystem it belongs to. The filter is configurable as to which vertex will have pruning filters on what subtree resource types.
2. During the traverser initialization, we walk the entire graph and populate these pruning filters first. For example, how many cores does a vertex have at the subtree rooted at this vertex.
3. Now, every time you get an allocation (or reservation), you will update the state of each allocated vertex (and its ancestor vertices). And you will do this from the fringes of the dominant subsystem tree towards the root. As we update the resource graph from the bottom and up, we exploit this to update the pruning filters from the bottom up at the same time. This is probably the most efficient way to keep the pruning filters globally consistent -- we don't walk the graph all the time just to update the pruning filters (Computationally intractable). Code-wise, we do this within the [upd_agfilter method](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L82). In that method, we simply add a new multi-planner span object which will subtract available subtree resources from the initially populated filter for that span. BTW, we also maintain the [exclusivity filter](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L55) in a similar way. 
4. Every time you remove an allocation (or reservation), we reverse this operation.

Currently, I find that your PR actually doesn't update the pruning filters. Sorry I missed this over during my preview. While it accumulates aggregates to the ancestors, this isn't hook to pruning filter updates. Funnily, though, this seems still correct: Yes, the pruning filter can overestimate the subtree resources when some resources are down and doesn't prune your next search walk downward; But the nature of pruning filters is "trigging" and optimization anyway so you get your accurate answer just by doing a bit more search.

Now that I think about this, hooking `mark()` into the pruning filter can bring in a bag of technical challenges.
1. @SteVwonder already identified idempotency. If not careful, you can subtract (or add) things multiple times. Update for an allocation (or reservation) is called only once so there is a guarantee for that case. But we don't have a guarantee with how many times `mark()` will be called. 
2. Keeping track of a multi-planner object that was inserted from "down" marking and removing it when the resource becomes "up" marking can also be challenging.

I think 1) and 2) can be addressed. But this will require substantial changes from what you already have in this PR. 

Perhaps what we should do for this round is just NOT updating pruning filters for `mark()` at all. Filters will not prune your search based on the subtree aggregate, but this won't lead to incorrect behavior either. If we take this route, `mark()` can also be dead simple -- no propagate and path-string-based ancestor search etc. Further, I don't think the performance impact won't be huge either in production. When you have a "down" node vertex, the traversal will be pruned during the pre-order visit on that vertex. So you won't even rely on aggregate pruning filter much.

Thoughts, @SteVwonder and @milroy?

resource/utilities/command.cpp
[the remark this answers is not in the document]
milroy: It is a string: 
```C++ 
const std::string &status
```
I'm still not sure what's causing this, but I will continue investigating.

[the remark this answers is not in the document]
dongahn: I don't have the new code base so it is hard to see. Maybe you can print out the stack traces from the core? If you are using a docker container:

```console
% sudo apt-get install gdb
% ulimit -c unlimited
% gdb PATH_TO_EXECUTABLE core
> bt
```

dongahn:
Leaving these comments here as well.

https://github.com/flux-framework/flux-sched/pull/684#issuecomment-653241370

https://github.com/flux-framework/flux-sched/pull/684#issuecomment-653275361

milroy:
> Perhaps what we should do for this round is just NOT updating pruning filters for mark() at all. Filters will not prune your search based on the subtree aggregate, but this won't lead to incorrect behavior either. If we take this route, mark() can also be dead simple -- no propagate and path-string-based ancestor search etc.

For the benefit of other readers, this is the decision we made in our meeting today.

Note that my [comment](https://github.com/flux-framework/flux-sched/pull/684#issuecomment-653288424) in PR #684 also pertains to the discussion here.  In particular, if we need to `mark` all resources in a subtree (not merely the root) then we still must address traversing an auxiliary subsystem in: 

```c++
int dfu_impl_t::mark (const std::string &root_path, 
                      resource_pool_t::status_t status)
```
called from `resource-query`.

dongahn:
I still believe that the top-level only marking approoach can equally work well for both dominant and auxiliary hierarchies.

milroy:
> I still believe that the top-level only marking approach can equally work well for both dominant and auxiliary hierarchies.

I agree.  I'm just wondering if the top-level only marking approach can deal with the type of complexity I mentioned [here](https://github.com/flux-framework/flux-sched/pull/684#issuecomment-653315222).

dongahn:
> I agree. I'm just wondering if the top-level only marking approach can deal with the type of complexity I mentioned here.

Please see https://github.com/flux-framework/flux-sched/pull/684#issuecomment-653355701

milroy:
See [my comment](https://github.com/flux-framework/flux-sched/pull/684#issuecomment-654302650) on PR #684.  I'm at work on the `mark`-top-level only approach.

milroy:
@dongahn and @SteVwonder: the top-level only `mark` requires reworking the `resource-query` functions.  What should be the desired behavior of the `dump` command?  It will require a depth-first traversal of the resource graph; for `dump UP` should descent into a subtree be halted when arriving at a vertex with `DOWN` status (and the reverse for `dump DOWN`)?

dongahn:
> @dongahn and @SteVwonder: the top-level only mark requires reworking the resource-query functions. What should be the desired behavior of the dump command? It will require a depth-first traversal of the resource graph; for dump UP should descent into a subtree be halted when arriving at a vertex with DOWN status (and the reverse for dump DOWN)?

@milroy: since this PR only marks a single vertex, I think it makes sense to cover only simpler tests: "set-status" on a vertex_path and  "get-status" on vertex_path to make sure status of that vertex is set correctly.

Also please see https://github.com/flux-framework/flux-sched/pull/684#issuecomment-654618368. 

To implement `sched.resource-status` and position us for future frontend work, we will need a more generalized version of `dump`: `find()`. 

`dump` can be covered with find commands like:

```
find status=up
find status=down
find status=up or status=down
```

milroy:
> @milroy: since this PR only marks a single vertex, I think it makes sense to cover only simpler tests: "set-status" on a vertex_path and "get-status" on vertex_path to make sure status of that vertex is set correctly.

Ok, that makes sense.

> dump can be covered with find commands...

So I'll refactor the `dump` command to a more limited `set-status` and let `find` take care of the full functionality.

milroy:
I've refactored the PR; it's ready for a final review from @dongahn and @SteVwonder when Travis returns successfully.

milroy:
> I've refactored the PR; it's ready for a final review from @dongahn and @SteVwonder when Travis returns successfully.

Actually, please hold off.  I added a new test case that exposed a small problem.

milroy:
Ok, @dongahn and @SteVwonder, this is ready for a final review!

resource/utilities/command.cpp
```diff
@@ -421,6 +423,43 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
...
+                     << " in resource graph." << std::endl;
+        return 0;
+    }
+
+    for (auto &status_it : sts) {
+        if (status_it.second == ctx->db->resource_graph[it->second].status) {
+            status = status_it.first;
+            break;
+        }
+    }
```
SteVwonder, line 480: Minor nit, but a bidirectional map here would be nice.  I've never used it before, but I wonder if it would be worth it to use [boost::bimap](https://www.boost.org/doc/libs/1_73_0/libs/bimap/doc/html/index.html)

t/data/resource/commands/status/cmds07
```diff
@@ -0,0 +1,5 @@
+set-status /tiny0/rack0/node0 down
+get-status /tiny0/rack0/node0
+set-status /tiny0/rack0/node0 up
+get-status /tiny0/rack0/node0
```
SteVwonder, line 4: I fully expect that the new implementation is idempotent, but can we expand this test a little bit to set the node down twice and set it up twice?  Just want to make sure we have a simple test guaranteeing idempotence for the future.

t/t3024-resource-status.t
```diff
@@ -0,0 +1,67 @@
...
+    test_cmp 006.R.out ${exp_dir}/006.R.out
+'
+
+test007_desc="change resource vertex status and get output "
+test_expect_success "${test007_desc}" '
+    ${query} -L ${grugs} -F jgf -S CA -P high < ${cmd_dir}/cmds07 \
+> 007.R.out &&
+    test_cmp 007.R.out ${exp_dir}/007.R.out
+'
+
```
SteVwonder, line 66: Can we also add one test that sets a resource in a secondary hierarchy to down and then up again?

t/data/resource/commands/status/cmds07
SteVwonder: Nevermind.  Looks like this is actually an issue for #667.

resource/utilities/command.cpp
milroy: The `bimap` will obviate the loop at the expense of greater memory consumption.  Since I think this check will only ever be used by `resource-query` for a single vertex my opinion is the increased memory consumption is the greater cost.

t/t3024-resource-status.t
milroy: Will do!

milroy:
I've added the test on the auxiliary hierarchy, and the Travis check passes.  @SteVwonder let me know if it looks good and I'll squash the commit so we can merge this.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +565,59 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
...
+    vtx_t subtree_root;
+
+    for (auto &rank : ranks) {
+        // Now iterate through subgraphs keyed by rank and 
+        // set status appropriately
+        vit = m_graph_db->metadata.by_rank.find (rank);
+        if (vit == m_graph_db->metadata.by_rank.end ()) {
+            errno = EINVAL;
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": could not find rank path in by_rank map.\n";
```
dongahn, line 601: Can we have a whitespace after "."? If multiple error messages are accumulated this will ensure this line is separated with the next via a whitespace.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -565,6 +565,59 @@ int dfu_impl_t::remove (vtx_t root, int64_t jobid)
...
+                      resource_pool_t::status_t status)
+{
+    std::map<std::string, vtx_t>::const_iterator vit_root =
+        m_graph_db->metadata.by_path.find (root_path);
+
+    if (vit_root == m_graph_db->metadata.by_path.end ()) {
+        errno = EINVAL;
+        m_err_msg += __FUNCTION__;                
+        m_err_msg += ": could not find subtree path ("
+                  + root_path + ") in resource graph.\n";
```
dongahn, line 578: Can we have a whitespace after "."? If multiple error messages are accumulated this will ensure this line is separated with the next via a whitespace.

resource/utilities/command.cpp
```diff
@@ -423,6 +425,34 @@ int cmd_get_property (std::shared_ptr<resource_context_t> &ctx,
...
+        return 0;
+    }
+    std::string vtx_path = args[1];
+    std::string status = args[2];
+    std::map<std::string, vtx_t>::const_iterator it =
+        ctx->db->metadata.by_path.find (vtx_path);
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+
+    if (it == ctx->db->metadata.by_path.end ()) {
+        std::cout << "Could not find path " << vtx_path
```
dongahn, line 442:
```suggestion
        std::cerr << "ERROR: could not find path " << vtx_path
                     << " in resource graph." << std::endl;
````

dongahn:
Ah... FYI I didn't realize there was MWP label already added and the fact that I approved this merged this PR in. I will address the very minor nit in my PR. Thanks!

milroy:
> Ah... FYI I didn't realize there was MWP label already added and the fact that I approved this merged this PR in.

That's good to know for future reference.  Thanks again for the reviews @dongahn and @SteVwonder!

PR #674, opened 2020-06-17
Add by_rank map to graph store

milroy:
This PR addresses [item II.3](https://github.com/flux-framework/flux-sched/issues/662#issue-628053510) from issue #662 by introducing a `by_rank` map to `resource_graph_metadata_t`.  The map is keyed by the `int64_t` `rank` and takes a `std::vector` of `vtx_t` corresponding to the subgraph associated with the rank.  The changes include populating the map in the GRUG, JGF, and hwloc readers and the addtion of testsuite checks for correct subgraph sizes.

This PR should be merged before #665 and #667 as those depend on the features created here.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -33,6 +33,7 @@ extern "C" {
 #include "config.h"
 #endif
 #include <flux/core.h>
+#include <flux/idset.h>
 #include <jansson.h>
 #include "src/common/libutil/shortjansson.h"
 }
@@ -90,6 +91,7 @@ resource_ctx_t::~resource_ctx_t ()
 {
     flux_msg_handler_delvec (handlers);
 }
+
 
 /******************************************************************************
  *                                                                            *
@@ -1114,10 +1116,60 @@ error:
         flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
+static int get_stat_by_rank (std::shared_ptr<resource_ctx_t>& ctx, json_t *o)
+{
+    int rc = -1;
+    int saved_errno = 0;
+    char *str = nullptr;
+    struct idset *ids = nullptr;
+    std::map<size_t, struct idset *> s2r;
+
+    for (auto &kv : ctx->db->metadata.by_rank) {
+        if (kv.first == -1)
+            continue;
+        if (s2r.find (kv.second.size ()) == s2r.end ()) {
+            if ( !(ids = idset_create (0, IDSET_FLAG_AUTOGROW)))
+                goto done;
+            s2r[kv.second.size ()] = ids;
+        }
+        if ( (rc = idset_set (s2r[kv.second.size ()],
+                              static_cast<unsigned int> (kv.first))) < 0)
+            goto done;
+    }
+
+    for (auto &kv : s2r) {
+        if ( !(str = idset_encode (kv.second,
+                                   IDSET_FLAG_BRACKETS | IDSET_FLAG_RANGE))) {
+            rc = -1;
+            goto done;
+        }
+        if ( (rc = json_object_set_new (o, str,
+                      json_integer (static_cast<json_int_t> (kv.first)))) < 0) {
+            errno = ENOMEM;
+            goto done;
+        }
+        saved_errno = errno;
+        free (str);
+        errno = saved_errno;
+        str = nullptr;
+    }
+
+done:
+    for (auto &kv : s2r)
+        idset_destroy (kv.second);
+    saved_errno = errno;
+    s2r.clear ();
+    free (str);
+    errno = saved_errno;
+    return rc;
+}
+
 static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
                              const flux_msg_t *msg, void *arg)
 {
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
+    int saved_errno;
+    json_t *o = nullptr;
     double avg = 0.0f;
     double min = 0.0f;
 
@@ -1125,15 +1177,35 @@ static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
         avg = ctx->perf.accum / (double)ctx->perf.njobs;
         min = ctx->perf.min;
     }
-    if (flux_respond_pack (h, msg, "{s:I s:I s:f s:I s:f s:f s:f}",
+    if ( !(o = json_object ())) {
+        errno = ENOMEM;
+        goto error;
+    }
+    if (get_stat_by_rank (ctx, o) < 0) {
+        flux_log_error (h, "%s: get_stat_by_rank", __FUNCTION__);
+        goto error_free;
+    }
+    if (flux_respond_pack (h, msg, "{s:I s:I s:o s:f s:I s:f s:f s:f}",
                                    "V", num_vertices (ctx->db->resource_graph),
                                    "E", num_edges (ctx->db->resource_graph),
+                                   "by_rank", o,
                                    "load-time", ctx->perf.load,
                                    "njobs", ctx->perf.njobs,
                                    "min-match", min,
                                    "max-match", ctx->perf.max,
-                                   "avg-match", avg) < 0)
-        flux_log_error (h, "%s", __FUNCTION__);
+                                   "avg-match", avg) < 0) {
+        flux_log_error (h, "%s: flux_respond_pack", __FUNCTION__);
+    }
+
+    return;
+
+error_free:
+    saved_errno = errno;
+    json_decref (o);
+    errno = saved_errno;
+error:
+    if (flux_respond_error (h, msg, errno, NULL) < 0)
+        flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
 }
 
 static inline int64_t next_jobid (const std::map<uint64_t,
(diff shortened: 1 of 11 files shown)


milroy:
All checks are green and Codecov looks good.  @dongahn you can give this a pass when you're able, or I can add @SteVwonder as a reviewer.

resource/store/resource_graph_store.hpp
```diff
@@ -42,6 +42,7 @@ struct resource_graph_metadata_t {
     std::map<subsystem_t, relation_infra_t> v_rt_edges;
     std::map<std::string, std::vector <vtx_t>> by_type;
     std::map<std::string, std::vector <vtx_t>> by_name;
+    std::map<int64_t, std::vector <vtx_t>> by_rank;
```
dongahn, line 45: Very minor nit: on the commit message, "imperative" tense is typically recommended: so add instead of adds.

t/t6000-graph-size.t
```diff
@@ -30,13 +30,19 @@ Pairs: //") &&
...
     test ${by_name} -eq 52 &&
-    test ${by_path} -eq 100
+    test ${by_path} -eq 100 &&
+    test ${rank} -eq -1 &&
+    test ${nvertices} -eq 100
 '
 
-test_expect_success "by_type, by_name, and by_path map sizes are correct \
-for JGF" '
+test_expect_success "by_type, by_name, by_path, and by_rank map sizes are \
```
dongahn, line 44: In this case, `by_rank` is equal to the overall vertex size because by default every vertex will get -1. Maybe adding a short comment above this line to indicate that? This may be surprising for those who don't know the internals...

milroy:
> The current testing of using resource-query is limited as all vertices will be only mapped to one rank (-1). My suggestion would be to add one more simple test by modifying sched-fluxion-resource's stat interface and flux ion-resource.

I've been poking at this and am struggling to deal with the need to iterate over multiple ranks.  In the [stat interface](https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L1128) I could send multiple responses (one per rank to vector size association).  That approach seems ugly.  Another possibility would be to iterate over the map keys and flatten the key-value pairs to a single string, and transmit that string in the response.  What do you think @dongahn?

dongahn:
>I've been poking at this and am struggling to deal with the need to iterate over multiple ranks. In the stat interface I could send multiple responses (one per rank to vector size association). That approach seems ugly and complex.

I agreed. stat is to get the overall statistics and multiple responses would defeat the purpose.

> Another possibility would be to iterate over the map keys and flatten the key-value pairs to a single string, and transmit that string in the response. What do you think @dongahn?

This sounds good to me. Perhaps we can compactly flatten this using `struct idset`. For example, if the number of vertices are the same across all ranks, you can condense this to: `"by_rank": { "[0-1021]" : 100 }`. If there are two rank groups each with different vertex count, it would be something like: `"by_rank": {"[0-511]" : 100, "[512-1021]": 200}`.

Doc for `idset` is [here](https://github.com/flux-framework/flux-core/blob/master/doc/man3/idset_create.adoc).

milroy:
> If there are two rank groups each with different vertex count, it would be something like: "by_rank": {"[0-511]" : 100, "[512-1021]": 200}.

So the general way to do this is iterate through the ranks and populate an auxiliary map that's keyed by subgraph sizes, and takes vectors of ranks as values.  Then I can iterate through the size keys and decode the vector of ranks with `idset`.  Is that the idea?

milroy:
A more fundamental question is: how do I create a `resource_graph` with multiple ranks in `flux ion-resource`?

dongahn:
> A more fundamental question is: how do I create a resource_graph with multiple ranks 

You can use hwloc reader with sched-fluxoin-resource (I think this is the default). As each hwloc files are unpacked into the resource graph, a distinct rank will be assigned.

dongahn:
> So the general way to do this is iterate through the ranks and populate an auxiliary map that's keyed by subgraph sizes, and takes vectors of ranks as values. Then I can iterate through the size keys and decode the vector of ranks with idset. Is that the idea?

Yes. But instead of using  vectors of ranks as values, you can use an "essentially" idsets as values to directly encode the rank set into idset during the initial iteration. I say "essentially" because using a c++ wrapper around struct idset will make memory management a bit easier. 

```c++
class idset_wrap_t {
public:
    ~idset_wrap_t() {
        if (m_idset)
            idset_destroy (m_idset);
    }
    const get_idset () const {
        return m_idset;
    }
    void set_idset (struct idset *idset) {
        if (idset)
            m_idset = idset;
    }
private:
    struct idset *m_idset{nullptr};
};

std::map<size_t,  idset_wrap_t> s2r; // subgraph size to ranks

for (auto &kv : by_rank) {
    if (s2r.find (kv.second.size ()) == s2r.find.end ()) {
        idset_wrap_t iw;
        struct idset *ids = NULL;
        if ( !(ids = idset_create (0, IDSET_FLAG_AUTOGROW))) {
             error_check ()
        }
        iw.set_idset (ids);
        s2r[kv.second.size ()] = iw;
    }
    if (idset_set (s2r[kv.second.size ()].get_idset (), kv.first) < 0)
        error_check ();
}

// once the above iteration is done, each value of s2r
// has the fully encoded rank idset.
```

The code not tested at all, though.

dongahn:
> Yes. But instead of using vectors of ranks as values, you can use an "essentially" idsets as values to directly encode the rank set into idset during the initial iteration. I say "essentially" because using a c++ wrapper around struct idset will make memory management a bit easier.

@milroy: if you are busy working on other PRs, I can add this support. Let me know.

milroy:
@dongahn unfortunately the implementation is taking me longer than I expected.  Can you add the support?  

Also, I think you will need to make some adjustments to `idset_wrap_t`:

```c++
class idset_wrap_t {
private:
    struct idset *m_idset{nullptr};
public:
    ~idset_wrap_t() {
        if (m_idset)
            idset_destroy (m_idset);
    }
    const idset get_idset () {
        return m_idset;
    }
    void set_idset (struct idset *idset) {
        if (idset)
            m_idset = idset;
    }
};
```

dongahn:
> @dongahn unfortunately the implementation is taking me longer than I expected. Can you add the support?

Sure thing. I will get to this soon so that you can focus on the other PRs!

milroy:
I thoughtlessly pulled your new commits @dongahn and pushed some new ones of my own.  That added me as the committer to commits authored by you.  Is there some way to fix this?

grondo:
@milroy, IMO it is fine if you are committer, this happens quite often. As long as authorship is preserved it is somewhat standard practice when cherry-picking as you are doing.

If you really need to reset the committer, I think you can amend the commit with `GIT_COMMITTER_NAME` and `GIT_COMMITTER_EMAIL` set in your environment. You may also have to add `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL`, or use the `--author` option of git-commit.

dongahn:
> I thoughtlessly pulled your new commits @dongahn and pushed some new ones of my own. That added me as the committer to commits authored by you. Is there some way to fix this?

Don't worry about it. Also this didn't changed the author anyway.

dongahn:
@milroy: your new changed LGTM. @SteVwonder: if you can take a quick look at this PR, this should be ready to go in.

t/t6002-graph-hwloc.t
```diff
@@ -0,0 +1,41 @@
...
+
+test_under_flux 4
+
+test_expect_success 'qmanager: hwloc reload works' '
+    flux hwloc reload ${excl_4N4B}
+'
+
+test_expect_success 'qmanager: loading resource and qmanager modules works' '
+    flux module remove sched-simple &&
+    load_resource prune-filters=ALL:core subsystems=containment policy=low
```
SteVwonder, line 29: (Mainly a note to myself)

We may need to add an whitelist/allowlist here to make this test work under both hwloc1 and hwloc2, since the number of resources under each execution target may differ (e.g., `numanode` vs no `numanode`).  Since this PR is on the critical path, I wouldn't worry about it right now.  If it does turn out to be an issue, I can handle it in my upcoming hwloc2 PR.

milroy:
@dongahn thank you very much for jumping in and creating the additional functionality.  @SteVwonder thanks for the quick review!

I'm going to set MWP after flux-sched gets taggged.

resource/modules/resource_match.cpp
```diff
@@ -1114,26 +1116,95 @@ static void info_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+{
+    int rc = -1;
+    int saved_errno = 0;
+    char *str = nullptr;
+    struct idset *ids = nullptr;
+    std::map<size_t, struct idset *> s2r;
+
+    for (auto &kv : ctx->db->metadata.by_rank) {
+        if (kv.first == -1)
+            continue;
```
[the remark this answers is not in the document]
milroy: According to my brief investigation yesterday and @grondo, `idset` doesn't support negative integers.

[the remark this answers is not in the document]
dongahn: Yes, that was reason. We can add this as one special case. But for the sake of time, maybe we should go with what we have for this round?

t/t6002-graph-hwloc.t
dongahn: Thanks and this makes sense. Yes, let's target this work when hwloc2 PR is posted. We can either solve this w/ allowlist or by calculating number of ranks x per rank vertices = overall vertex count.

grondo:
@milroy, in case you missed the slack message, flux-sched v0.9.0 has been tagged.

PR #675, opened 2020-06-17
[WIP] Integrate Fluxion with flux-core's resource module

dongahn:
This PR is experimental and only for early reviews: It needs PR #674, #665 and #667 to be merged first before making further progress.

Integrate Fluxion with flux-core's resource module. 

- `sched-fluxion-resource`: perform the initial resource graph data store population when the first response of resource.acquire arrives. Update the resource graph whenever subsequence responses arrive. Its semantics and limitations are described in commit 1be7d51d2f05323e5df6caa87e2164eab5f890df

- `sched-fluxion-qmanager`: Send sched-fluxion-resource.notify streaming RPC to get synchronized with `sched-fluxion-resource` for resource acquision and state change events. Delay its handshaking protocol with job-manager until receiving the first response of `sched-fluxion-resource.notify`. Upon receiving each response, run schedule loop as it represents a scheduleable resource event. Stop the reactor and module exit when receiving the ECANCELED response of `sched-fluxion-resource.notify`, which occurs when sched-fluxion-resource is unloaded.

- Adjust test cases for newly created dependencies among flux-core's `resource`, `sched-simple` and fluxion's two service modules.

The high-level interaction diagram is available at https://github.com/flux-framework/flux-sched/issues/662#issuecomment-643496117.

PR #677, opened 2020-06-24
Add support for hwloc 2.0+

### LINKED ISSUE #656: Hwloc 2.0+ Support
One change between hwloc 1.0 and 2.0+ that affects flux-sched is:

> In hwloc v1.x, NUMA nodes were inside the tree, for instance Packages contained 2 NUMA nodes which contained a L3 and several cache.

> Starting with hwloc v2.0, NUMA nodes are not in the main tree anymore. They are attached under objects as Memory Children on the side of normal children. This memory children list starts at obj->memory_first_child and its size is obj->memory_arity. Hence there can now exist two local NUMA nodes, for instance on Intel Xeon Phi processors.

> if there are two NUMA nodes per package, a Group object may be added to keep cores together with their local NUMA node

[[Source]](https://www.open-mpi.org/projects/hwloc/doc/v2.2.0/a00339.php)

This means that the `NUMANode` resource type never appears in the resource graph when using an XML generated by hwloc 2.0+ (even when the V1 compatibility flag is provided at export time).  In the case where the `NUMANode` affects the structure of the tree, a `Group` resource appears instead.  When the `NUMANode` doesn't affect the structure, the tree goes straight from `Machine` to `Package` to `L3Cache` (or `Core` or whatever used to be the child of `NUMANode`).  

This creates issues for several of the tests in `t4004-match-hwloc.t` where one of several condititions causes an error:
- the expected output from `resource-query` contains the `numanode` vertex
- the requested jobspec explicitly requests a `numanode` resource
- the requested jobspec implicitly requires a common ancestor between two resources that used to be numanode (at least, I think that is what is happening, I need to dig further into it, the failing jobspec is `data/resource/jobspecs/basics/test013.yml`)

One option is to not try and make these two versions of hwloc produce the same resource graph or the same output.  This would affect jobspec portability.

Another option is to try and make the resource graph generated with a hwloc2 reader look similar/the same as the resource graph generated with a hwloc1 reader.  This will require more work and may not be possible in every situation, but I think at least attempting it is worth the potential portability gains.

One thought I had in the direction of the latter option was to rename any `Group`s with a `NUMANode` child to be a `NUMANode`.  This would put `numanode` back "in the tree", but it won't apply to every topology.  Specifically, if the `NUMANode` doesn't affect the topology structure, then hwloc2 elides the `Group` object entirely and makes `NUMANode` a child of `Machine`:

```
sherbein .../data/hwloc-data/001N/exclusive (hwloc2 !?S)
❯ lstopo-no-graphics -i 04-brokers/0.xml
Machine (16GB total)
  NUMANode L#0 (P#0 16GB)
  Package L#0 + L3 L#0 (20MB)
    L2 L#0 (256KB) + L1d L#0 (32KB) + Core L#0 + PU L#0 (P#0)
    L2 L#1 (256KB) + L1d L#1 (32KB) + Core L#1 + PU L#1 (P#1)
    L2 L#2 (256KB) + L1d L#2 (32KB) + Core L#2 + PU L#2 (P#2)
    L2 L#3 (256KB) + L1d L#3 (32KB) + Core L#3 + PU L#3 (P#3)
```

Maybe as we are traversing the resource tree, whenever we detect a `NUMANode` at a given level, we insert the `numanode` back into the tree and then make the resources that used to be its siblings into its children.  I'm worried about the potential corner cases with this approach.  I don't have a good feel for all the possible topologies to be confident to say this wouldn't break anything.  At a high-level, it sounds like the inverse operation of what was done during the v1->v2 transition.

SteVwonder:
This PR adds support for hwloc versions 2.0+.  Specifically, it:
- Adds an `#ifdef` block similar to the one in flux-core to do filtering of resources
- Replaces how children are accessed with a more forward-compatible method
- Uses the `resource-query` whitelist feature to normalize the resource set tested across hwloc 1 & 2
  - Updates the test input/output accordingly
- Adds a few tweaks to various tests and test scripts

I have it marked as a WIP because `t4004-match-hwloc.t` still has a single failing test at the end.  Oddly enough, the same resource data and jobspec works perfectly fine with `resource-query` but not with `flux-ion-resource` and the full `sched-fluxion-resource` module.  I also haven't verified that things still work 100% in hwloc 1.

EDIT: Closes #656


resource/readers/resource_reader_hwloc.cpp
```diff
@@ -301,11 +301,32 @@ int resource_reader_hwloc_t::unpack_internal (resource_graph_t &g,
...
     if ( hwloc_topology_set_flags (topo, HWLOC_TOPOLOGY_FLAG_IO_DEVICES) != 0) {
         errno = EINVAL;
         m_err_msg += "Error setting hwloc topology flag; ";
         goto done;
     }
+#else
+    if (hwloc_topology_set_io_types_filter(topo,
+                                           HWLOC_TYPE_FILTER_KEEP_IMPORTANT)
+        < 0) {
+      m_err_msg += "hwloc_topology_set_io_types_filter";
```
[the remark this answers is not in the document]
dongahn: Somehow my other suggestion didn't make it in. We will need `errno` support for these new error branches before the jump: 
```c++
    errno = EINVAL;
```

dongahn:
> I have it marked as a WIP because t4004-match-hwloc.t still has a single failing test at the end. 

@SteVwonder: I think I manually reproduced the issue and this was a problem I wanted to look into further. So this came out in an opportune time. 

Before this test is hung, I get the following error message:

> 2020-06-24T18:30:21.552622Z sched-fluxion-resource.err[0]: match_request_cb: match failed due to match error (id=0): Success
2020-06-24T18:30:21.552668Z sched-fluxion-resource.err[0]: match_request_cb: flux_respond_error: Invalid argument

This probably means that the graph created with hwloc2 support doesn't match w/ the underlying job spec. I will look but this could be the case where more hierarchical resource types than the jobspec supported (e.g., socket).

The bigger issue is, though: when an match error is detected, `flux mini run` (or equivalent) should handle it more gracefully. I actually hit this while I was testing PR #675. 

Let me look at a couple of things and make a specific suggestion.

dongahn:
The following patch should work for this PR

```diff
diff --git a/t/t4004-match-hwloc.t b/t/t4004-match-hwloc.t
index 3a994ee4..7aca3868 100755
--- a/t/t4004-match-hwloc.t
+++ b/t/t4004-match-hwloc.t
@@ -119,7 +119,8 @@ test_expect_success 'reloading session/hwloc information with test data' '
 '

 test_expect_success 'loading resource module with default resource info source' '
-    load_resource subsystems=containment policy=high
+    load_resource subsystems=containment policy=high \
+load-whitelist=node,socket,core
 '

 test_expect_success 'match-allocate works with four two-socket jobspecs' '
```

My guess is without this option the resulting graph contains other intermediate resource types between `node` and `socket` such a way that the given jobspec won't match.

```yaml
version: 1
resources:
  - type: node
    count: 1
    with:
      - type: slot
        count: 2
        label: default
        with:
          - type: socket
            count: 1
```

I will push this change as a "fixup" commit to your branch to see if that makes CI happy.

For the "match error" leading to a front end hang issue, I will open up a Issue ticket to address in a separate PR.

dongahn:
FYI -- pushed this change.

dongahn:
@SteVwonder: the CI is happy now. Once you address the review comments and merge my fixup commit, this PR should be good to be merged.

SteVwonder:
Thanks @dongahn for figuring that out!  I appreciate the help!  That reminds me that I should add the latest ubuntu image that was added to flux-core so that we have CI coverage of hwloc 2.X (the current testers only have 1.X).

I'll address your review comments today as well.

dongahn:
Awesome! Thanks @SteVwonder.

SteVwonder:
I just restarted the Ubuntu 20.04 builder now that the flux-core image is being published (thanks @grondo !!!!).  Looks like both issues occur there (the failed test and the empty `R` in match responses):

```
ok 11 - loading resource module with default resource info source
loading resource module with default resource info source
expecting success: 
    flux ion-resource match allocate ${jobspec_2socket} &&
    flux ion-resource match allocate ${jobspec_2socket} &&
    flux ion-resource match allocate ${jobspec_2socket} &&
    flux ion-resource match allocate ${jobspec_2socket}
JOBID                STATUS               AT                   OVERHEAD (Secs)     
0                    ALLOCATED            2020-06-26T04:22:41  0.0005309581756591797
================================================================================
MATCHED RESOURCES:
JOBID                STATUS               AT                   OVERHEAD (Secs)     
1                    ALLOCATED            2020-06-26T04:22:41  0.00048804283142089844
================================================================================
MATCHED RESOURCES:
JOBID                STATUS               AT                   OVERHEAD (Secs)     
2                    ALLOCATED            2020-06-26T04:22:41  0.00046181678771972656
================================================================================
MATCHED RESOURCES:
JOBID                STATUS               AT                   OVERHEAD (Secs)     
3                    ALLOCATED            2020-06-26T04:22:41  0.0004439353942871094
================================================================================
MATCHED RESOURCES:

```

```
expecting success: 
    remove_resource &&
    flux hwloc reload ${excl_4N4B_sierra2} &&
    load_resource subsystems=containment \
policy=high load-whitelist=node,socket,core,gpu &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    flux ion-resource match allocate ${jobspec_1socket_2gpu} &&
    test_expect_code 16 flux ion-resource match allocate ${jobspec_1socket_2gpu}
OSError: error(16): Device or resource busy
not ok 14 - reloading sierra xml and match allocate
```

I'll begin poking at that tomorrow and next week.

dongahn:
@SteVwonder: LMK if you need any help to finalize this. This should def. go into our end-of-July freeze.

SteVwonder:
Travis is finally green! :partying_face: https://travis-ci.org/github/flux-framework/flux-sched/builds/708137964

@dongahn, I addressed your comments as a fixup commit.  Let me know if the changes look good, and then I'll squash and this PR should be good to go.

resource/readers/resource_reader_hwloc.cpp
```diff
@@ -309,23 +309,26 @@ int resource_reader_hwloc_t::unpack_internal (resource_graph_t &g,
         goto done;
     }
 #else
-    if (hwloc_topology_set_io_types_filter(topo,
+    if (hwloc_topology_set_io_types_filter (topo,
                                            HWLOC_TYPE_FILTER_KEEP_IMPORTANT)
```
dongahn: Super minor knit: after adding the whitespace, the two lines are not aligned.

resource/readers/resource_reader_hwloc.cpp
```diff
@@ -309,23 +309,26 @@ int resource_reader_hwloc_t::unpack_internal (resource_graph_t &g,
...
         < 0) {
-      m_err_msg += "hwloc_topology_set_io_types_filter";
-      goto done;
+        errno = EINVAL;
+        m_err_msg += "hwloc_topology_set_io_types_filter; ";
+        goto done;
     }
-    if (hwloc_topology_set_cache_types_filter(topo,
+    if (hwloc_topology_set_cache_types_filter (topo,
                                               HWLOC_TYPE_FILTER_KEEP_STRUCTURE)
```
dongahn: Same

resource/readers/resource_reader_hwloc.cpp
```diff
@@ -309,23 +309,26 @@ int resource_reader_hwloc_t::unpack_internal (resource_graph_t &g,
...
         < 0) {
-      m_err_msg += "hwloc_topology_set_cache_types_filter";
-      goto done;
+        errno = EINVAL;
+        m_err_msg += "hwloc_topology_set_cache_types_filter; ";
+        goto done;
     }
-    if (hwloc_topology_set_icache_types_filter(topo,
+    if (hwloc_topology_set_icache_types_filter (topo,
                                                HWLOC_TYPE_FILTER_KEEP_STRUCTURE)
```
dongahn: Same.

SteVwonder:
Thanks @dongahn! I'll address those right away.

I forgot to call out that I did have to add a new commit: https://github.com/flux-framework/flux-sched/pull/677/commits/e350c5b6ad67c7724d24a898270a9af96a1e795d.  If you haven't already, that's probably worth looking over.

dongahn:
> I forgot to call out that I did have to add a new commit: e350c5b. If you haven't already, that's probably worth looking over.

Yap that commit looks perfect to me and the detailed commit message is awesome! Great findings!

SteVwonder:
Thanks @dongahn for the quick reviews!  Just pushed the fixes and squashed.

PR #681, opened 2020-06-28
qmanager: Update for flux-core annotation support changes

chu11:
@dongahn and flux-sched team

I know you were going to handle the changes, but since https://github.com/flux-framework/flux-core/pull/2960  has a lot of changes to `libschedutil`, I decided to do a quick run to get `flux-sched` to compile to show you the high level changes.  This patch is mostly informational.  You can just close the PR if you'd like or take the patch as a starting point or if you think it's close to what you'd want, LMK and I can clean it up and finish it.

I did not look at anything beyond just "make it compile" and did not look at any fallout from these changes.  And (somewhat naturally) it doesn't take advantage of anything new (which should probably be later, once support is in `flux-jobs`).

The 1 logical change made was to the `exception_cb` functions, which should call `alloc_cancel()` instead of `alloc_deny()` now.  I didn't rename the functions to `cancel_cb`, but they should probably be renamed for consistency.

I did see this test fail once in awhile when running `make -j16 check`.

```
FAIL: t1001-qmanager-basic.t 4 - qmanager: basic job runs in simulated mode
```

not sure why its racy (or maybe it was racy to begin with) or was to due to the `exception_cb` logic change.

### DIFF
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -67,10 +67,10 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
         const std::string &queue_name = kv.first;
         std::shared_ptr<queue_policy_base_t> &queue = kv.second;
         while ( (job = queue->alloced_pop ()) != nullptr) {
-            if (schedutil_alloc_respond_R (schedutil, job->msg,
-                                           job->schedule.R.c_str (),
-                                           NULL) < 0) {
-                flux_log_error (h, "%s: schedutil_alloc_respond_R (queue=%s)",
+            if (schedutil_alloc_respond_success_pack (schedutil, job->msg,
+                                                      job->schedule.R.c_str (),
+                                                      NULL) < 0) {
+                flux_log_error (h, "%s: schedutil_alloc_respond_pack (queue=%s)",
                                 __FUNCTION__, queue_name.c_str ());
                 goto out;
             }
@@ -79,11 +79,11 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
         }
         while ( (job = queue->rejected_pop ()) != nullptr) {
             std::string note = "alloc denied due to type=\"" + job->note + "\"";
-            if (schedutil_alloc_respond_denied (schedutil,
-                                                job->msg,
-                                                note.c_str ()) < 0) {
+            if (schedutil_alloc_respond_deny (schedutil,
+                                              job->msg,
+                                              note.c_str ()) < 0) {
                 flux_log_error (h,
-                                "%s: schedutil_alloc_respond_denied (queue=%s)",
+                                "%s: schedutil_alloc_respond_deny (queue=%s)",
                                 __FUNCTION__, queue_name.c_str ());
                 goto out;
             }
@@ -167,8 +167,8 @@ void qmanager_cb_t::jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
         return;
     }
     if (ctx->queues.find (queue_name) == ctx->queues.end ()) {
-        if (schedutil_alloc_respond_denied (ctx->schedutil, msg, NULL) < 0)
-            flux_log_error (h, "%s: schedutil_alloc_respond_denied",
+        if (schedutil_alloc_respond_deny (ctx->schedutil, msg, NULL) < 0)
+            flux_log_error (h, "%s: schedutil_alloc_respond_deny",
                             __FUNCTION__);
         errno = ENOENT;
         flux_log (h, LOG_DEBUG, "%s: queue (%s) doesn't exist",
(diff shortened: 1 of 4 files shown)


chu11:
Now that #687 appears to be independent of this PR, I'm going to clean up this patch a bit more.  Try to make it more like a merge-able PR.

chu11:
I've updated the PR to be a legitimate contender for merging.  Travis is failing and I assume its b/c the flux-core its building against is the wrong one.

grondo:
Minor suggestion: don't reference PR or issue numbers in commit message or PR titles (since PR title becomes merge commit title). These should be used primarily as extra reference material since they are not easily at hand when browsing history via commandline.

Instead describe the issue being addressed as if there wasn't a PR/Issue to reference. (my preference anyway)

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -67,10 +67,10 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
         const std::string &queue_name = kv.first;
         std::shared_ptr<queue_policy_base_t> &queue = kv.second;
         while ( (job = queue->alloced_pop ()) != nullptr) {
-            if (schedutil_alloc_respond_R (schedutil, job->msg,
-                                           job->schedule.R.c_str (),
-                                           NULL) < 0) {
-                flux_log_error (h, "%s: schedutil_alloc_respond_R (queue=%s)",
+            if (schedutil_alloc_respond_success_pack (schedutil, job->msg,
+                                                      job->schedule.R.c_str (),
+                                                      NULL) < 0) {
```
dongahn, line 72: @chu11: I am not asking you to do this necessarily as part of this PR. This are just questions for my own edification: 

If you want to pass scheduler-specific data such as queue name, how would you suggest to pack it? The C string for queue name should be available as `queue_name.c_str ()`. Key-value I was thinking about is `queue=batch' or `queue=debug`. 

Looking ahead, backfilling scheduler will send multiple annotation response messages with "start time estimate". If that's the case, should this call do something special to remove those annotation keys to keep them from going into a more permanent place? Or would those keys be removed automatically? 

In general, if I happened to send annotation responses w/ scheduler-specific keys and want to remove those keys on Alloc Success, how do you do that?

dongahn:
@chu11: This looks good to me and can go in from my perspective. Thank you for putting this together. I add one inline comment with some questions. And yes, I like the callback name change from exception to cancel. The code base wasn't up to date with that protocol change.

qmanager/modules/qmanager_callbacks.cpp
chu11: > If you want to pass scheduler-specific data such as queue name, how would you suggest to pack it? The C string for queue name should be available as queue_name.c_str (). Key-value I was thinking about is queue=batch' or queue=debug`.

So we haven't defined a queue as one of the standard `sched` annotations in RFC27.  But maybe we should?  I think this currently falls into the camp of additional keys the scheduler can define.  I think `sched.queue` would be fine enough.

> In general, if I happened to send annotation responses w/ scheduler-specific keys and want to remove those keys on Alloc Success, how do you do that?

Right now the scheduler should delete any annotations they no longer want available to the user after an alloc success.  A json `null` is used to identify that a key should be removed.  In testing I've been doing things like:

```
    if (schedutil_alloc_respond_success_pack (sc->schedutil_ctx,
                                              job->msg,
                                              "1core",
                                              "{ s:n }",
                                              "sched.reason_pending") < 0)
        flux_log_error (sc->h, "schedutil_alloc_respond_success_pack");
```

chu11:
> @chu11: This looks good to me and can go in from my perspective. Thank you for putting this together. I add one inline comment with some questions. And yes, I like the callback name change from exception to cancel. The code base wasn't up to date with that protocol change.

Awesome.  Just to double check, you are signing off with https://github.com/flux-framework/flux-core/pull/2960 going in?

qmanager/modules/qmanager_callbacks.cpp
dongahn: > I think this currently falls into the camp of additional keys the scheduler can define. I think sched.queue would be fine enough.

As far as "additional keys" can be supported and kept, I'd be happy. This is an area were we can't defined all possible keys that a scheduler want to annotate an allocation with.

dongahn:
> Awesome. Just to double check, you are signing off with flux-framework/flux-core#2960 going in?

Unless you have anything else you want to add, this can go in from my perspective. Since this will need some coordination with your flux-core PR, let's not add MWP just yet.

chu11:
> Unless you have anything else you want to add, this can go in from my perspective. Since this will need some coordination with your flux-core PR, let's not add MWP just yet.

Waiting for #690 & #692 to get in since merging https://github.com/flux-framework/flux-core/pull/2960 will break the build.  Then we can merge that, then come back to this.

chu11:
rebased & re-pushed, hopefully can pass travis now that https://github.com/flux-framework/flux-core/pull/2960 is merged

grondo:
You may have to wait until the new flux-core docker image is pushed...

chu11:
hmm coverage is sort of bad, mostly due to missed alloc DENY paths.  They probably weren't covered to begin with?

dongahn:
> hmm coverage is sort of bad, mostly due to missed alloc DENY paths. They probably weren't covered to begin with?

I think so. Our testing strategy was to do the main matching/unmatching testing using resource-query and deny/unsat test probably didn't make into testing w/ qmanager. 
If you are willing, you can add a simple test in https://github.com/flux-framework/flux-sched/blob/master/t/t1001-qmanager-basic.t

You can create a jobspec which cannot be satisfied and do flux job submit on it.

If you are busy w/ something else, I'm okay to merge this in without it. But in that case, I will appreciate if you can open a ticket for this.

chu11:
> You can create a jobspec which cannot be satisfied and do flux job submit on it.

Ok, threw up a commit with an extra test. Hopefully we can get > 54% coverage this time :P

dongahn:
LGTM! Adding MWP.

chu11:
@dongahn you need to approve :-)

dongahn:
Ugg...

PR #684, opened 2020-07-02
Integrate Fluxion with core's resource module

dongahn:
Note that #665 and #667 need to be merged first before this PR.

This PR integrates Fluxion with flux-core's resource module. 

- `sched-fluxion-resource`: perform the initial resource graph data store population when the first response of resource.acquire arrives. Update the resource graph whenever subsequence responses arrive. Its semantics and limitations are described in commit 1be7d51d2f05323e5df6caa87e2164eab5f890df

- `sched-fluxion-qmanager`: Send sched-fluxion-resource.notify streaming RPC to get synchronized with `sched-fluxion-resource` for resource acquision and state change events. Delay its handshaking protocol with job-manager until receiving the first response of `sched-fluxion-resource.notify`. Upon receiving each response, run schedule loop as it represents a scheduleable resource event. Stop the reactor and module exit when receiving the `ECANCELED` response of `sched-fluxion-resource.notify`, which occurs when sched-fluxion-resource is unloaded.

- Adjust test cases for newly created dependencies among flux-core's `resource`, `sched-simple` and fluxion's two service modules.

- Add cases to test correct handling of resource status changes (`t1011-dynstate-change.t`)

The high-level interaction diagram is available at https://github.com/flux-framework/flux-sched/issues/662#issuecomment-643496117.


dongahn:
The only missing piece in this PR is support for `sched.resource-status`. I will work on this next.

dongahn:
> The only missing piece in this PR is support for sched.resource-status. I will work on this next.

My current thinking is to introduce `find ()` method within the top-level dfu traverser class. 

It will take in a match criteria (of `std::string` type) and a base writer object and will walk the resource graph and perform matches on this criteria. It will emit the matched vertices and edges. Different from the `run()` methods, `find ()` won't result in an allocation or reservation in the end. 

This way, we can be future proof -- when RV2 will require a new writer (based on JGF writer), we can simply work on the new writer and pass that writer to this `find ()` method with no modification in the future.

The initial set of match criteria should be minimal and cover the criteria that jobspec doesn't support. 
- resource vertex status (up or down)
- sched (allocated or reserved)

```c++
  int find (std::shared_ptr<match_writers_t> &writers,
            const std::string &criteria);
```

The matching criteria examples can be:

1) "--status up" 
2) "--status up --sched allocated"
3) "--sched alloc --status down"

Having a command line look and feel for the criteria parameter would make it easy to expose this to resource-query command line interface for testing.

Initially, the interpretation of a multi-criteria expression can be "and", but we can later extend this to support more complex compound expression including "or" and parentheses as well.

Since this can blow up the scope of this PR, I propose this as a separate PR and then make this PR build on top of that new dependency.

dongahn:
> Since this can blow up the scope of this PR, I propose this as a separate PR and then make this PR build on top of that new dependency.

I will be prototyping this concept a bit to quickly find if marking "down" status only on the top-level resource (node) would be good enough for `find()` and inform @milroy's PR #665. 

If so, we should try to land PR #665 as soon as possible to reduce moving pieces. I start to feel uneasy to have too many PRs in flight.

dongahn:
> I will be prototyping this concept a bit to quickly find if marking "down" status only on the top-level resource (node) would be good enough for find() and inform @milroy's PR #665.

I looked at the code a bit more closely. It looks like it will take some time even to implement a prototype. 

But so that I am not in the way of @milroy, I'd say I'm reasonably confident that marking only top-level resources "down" should work for `find()` as well. Nothing demonstrates better than having a proof-of-concept code though...

milroy:
> I will be prototyping this concept a bit to quickly find if marking "down" status only on the top-level resource (node) would be good enough for find() and inform @milroy's PR #665.

@SteVwonder and I discussed this at coffee time today.  One of the primary concerns we thought of when only marking the subtree root as `DOWN` is the resulting status inconsistency between the root and the other subtree vertices.  This may result in problems with multiple `mark` calls with different subtree roots within a larger subtree.  Such inconsistency will also cause problems if we want to emit the status key in JGF.

dongahn:
> This may result in problems with multiple mark calls with different subtree roots within a larger subtree. Such inconsistency will also cause problems if we want to emit the status key in JGF.

Hmmm... I don't quite understand why this would be a problem though. Shouldn't the "down" state of the highest level vertex implies the entire subtree is also "down"?

dongahn:
> Such inconsistency will also cause problems if we want to emit the status key in JGF.

I think we should table this for now and pick this up when we make more progress with RV2 spec.

At first glance though:

For RV2 used for resource allocation, I think there would be no "down" resources selected. So we are covered for that case.

But for RV2 used in `sched.resource-status`,  having the status exported in the resource set will make our protocol efficient as @grondo said at today's meeting.

But for that, I have to think that what matters would not be how we mark vertices internally in the resource graph itself but be how `find()` or similar will serialize the matching resources into this format. At this point, I have to believe `find()` can effect our writers such a way that "status" field can be exported consistently with top-level resource only "down" marking.

milroy:
> Hmmm... I don't quite understand why this would be a problem though. Shouldn't the "down" state of the highest level vertex implies the entire subtree is also "down"?

That's true, but I'm thinking of a scenario where `/rack0/node0` is marked `DOWN`, then `/rack0` is marked `DOWN`.  If `/rack0` is later marked back `UP`, this should mean all child resources are to be considered `UP`.  But here, `/rack0/node0` will still be `DOWN` and thus unschedulable.  So I'm thinking the subtree traversal might still be desirable.

milroy:
> For RV2 used for resource allocation, I think there would be no "down" resources selected. So we are covered for that case.

I guess I'm wondering if we should protect against potential future statuses that can allow selection and thus will be part of RV2?

dongahn:
> That's true, but I'm thinking of a scenario where `/rack0/node0` is marked `DOWN`, then `/rack0` is marked `DOWN`. 

A good example and I see your point. But I have to think this is really a semantics question. Either semantics seems to have a merit. But my preference would be top-level only.

> If /rack0 is later marked back UP, this should mean all child resources are to be considered UP. But here, /rack0/node0 will still be DOWN and thus unschedulable.

In this case, I can see why you might want `/rack0/node0` to be UP but at the same time why you don't want  `/rack0/node0` to be UP.  

A use case for the latter would be: 

1. `/rack0/node0` was administratively drained so that the node can be worked on when the job using that node completes. So it becomes DOWN.
2. Then, for rack maintenance work, `/rack0` gets marked down. 
3. When the maintenance work at step 2 is done, we mark `/rack0` back UP.

Now, with the mark-all-subtree semantics, step 3 will also clear up the administrative drain of `/rack0/node0`. So if that node still has not been worked on, one needs to replay the drain RPC.

With the top-level only semantics,  step 3 will not clear up the administrative drain of `/rack0/node0` so the node work can be still performed with no redundant drain action. 

Similarly, if the node work is done before the rack maintenance work, `/rack0/node0` will be marked UP first. But because `/rack0` still has "DOWN", this node will not be scheduled until `/rack0` is turned back "UP". Desired behavior.

I can see other use cases where mark-all-subtree semantics make sense: Mostly, improving user convenience.  

Given the use cases are divided, however, I think a more explicit semantics with a simpler implementation would be better. 

Besides, w/ the current drain/undrain at rank granularity, the issues around differently scoped marking won't affect us much for a foreseeable future.

dongahn:
> I guess I'm wondering if we should protect against potential future statuses that can allow selection and thus will be part of RV2.

Thinking future proof is definitely a good practice! I just don't yet see convincing benefits with "mark-all-subtree" semantics. If we extend the status types, we can surely revisit this semantics.

I am sorry that this work had to change so much from the initial posting. But I feel I learned a lot from all the review discussions.

milroy:
> Now, with the mark-all-subtree semantics, step 3 will also clear up the administrative drain of /rack0/node0. So if that node still has not been worked on, one needs to replay the drain RPC.

Good point!

> Besides, w/ the current drain/undrain at rank granularity, the issues around differently scoped marking won't affect us much for a foreseeable future.

Ok, that's good to know.

> I just don't yet see convincing benefits with "mark-all-subtree" semantics. If we extend the status types, we can surely revisit this semantics.

That sounds good to me; I'm at work on the top-level only approach.

milroy:
> I am sorry that this work had to change so much from the initial posting. But I feel I learned a lot from all the review discussions.

I certainly didn't understand the extent of the problem and I'm glad we're converging on the right unified approach!

dongahn:
> Initially, the interpretation of a multi-criteria expression can be "and", but we can later extend this to support more complex compound expression including "or" and parentheses as well.

FYI - It felt like in a foreseeable future, we will need to extend `find`'s matching expression system to query qualified resources for fluxion front end. So I bit the bullet and implemented matching multi-criteria expression evaluators: b0fb0aa, 25da2fd, and f2a7454.

> "--status up"
"--status up --sched allocated"
"--sched alloc --status down"

For leaf expressions, I actually ended up settling with `key=value`-based expression:
```
"status=up"
"status=up and sched=allocated"
"sched=alloc or (status=up and sched=reserved)"
```
I will use this in DFU's `find()` and integrate it w/ `resource-query` first. I am using this PR for now until I will grow this to the point where it can go into its own PR.

dongahn:
>> I will be prototyping this concept a bit to quickly find if marking "down" status only on the top-level resource (node) would be good enough for find() and inform @milroy's PR #665.
I looked at the code a bit more closely. It looks like it will take some time even to implement a prototype.

> But so that I am not in the way of @milroy, I'd say I'm reasonably confident that marking only top-level resources "down" should work for find() as well. Nothing demonstrates better than having a proof-of-concept code though...

@milroy: I now have a proof of concept that the top-level marking will work great! At the time of our call this afternoon, I was struggling a bit to make compound expression work. But prototyping helps a lot and I believe I have an algorithm for that one as well!

Issue #612, opened 2020-03-03
Change pending queue order by `t_submit`


dongahn:
This is from today's coffee hour discussion.

Currently, the pending job queue within`qmanager` is ordered by enqueue time as seen by `qmanager` not by the enqueue time as seen by `job-manager`.

https://github.com/flux-framework/flux-sched/blob/master/qmanager/policies/base/queue_policy_base_impl.hpp#L206

One problem this will lead to is when a pending job is modified. When modified, the existing ALLOC request will be cancelled and a new ALLOC request will be issued with an altered job spec. However, this will put the newly modified job at the end of the pending queue as is, which isn't correct.

Need to change the pending job order of `qmanager` sorted by a job's `t_submit` instead, which will not be changed when the job is modified.

dongahn:
This should be fixed before July code freeze.

dongahn:
I think it has be: order by `priority` first and then order by  `t_submit`.

dongahn:
PR #695 addresses this.

PR #695, opened 2020-07-17
qmanager: support for ordering pending jobs by priority/submit time

dongahn:
This is a bug fix PR fixing Issue #612.

Order pending jobs by priority and submit time:

Problem: Currently the pending jobs within
fluxion's `qmanager` is ordered by enqueue time
as seenby qmanager, not by the enqueue time as seen
by job-manager.  Further, it doesn't take
into account the job's priority either, which
can lead to various side-effects. 

Change the "key" of the pending queue map from
`uint64_t` to be `std::vector<double>` such that
pending jobs are now ordered lexicographically
according to this vector.

Use <priority, t_submit> as the current key.

This change can easily accomodate other
order-affecting changes that are expected
to be introduced in the future including 
fair share priorities.

Fix Issue #612. 

This PR is written on top of PR #694 
and only consists of three commits.


dongahn:
LGTM! 

The only question for my own edification is if it is an established practice to include an issue number in shareness test titles. This works for me. But I think I have seen a recommendation not to include Issue number into the commit message as well.

W/ respect to a related PR https://github.com/flux-framework/flux-sched/pull/695, did I get it right that there isn't an easy way for `qmanager` to see a pending job order change once jobs are enqueued into its queue? I'm using `unlimited` for all cases.

dongahn:
Once PR #694 goes int, I will rebase this. Too much conflict at this point.

dongahn:
I think it would be best to close this and do this in a new PR.

PR #698, opened 2020-07-18
resource: errno support for std::ifstream open

dongahn:
This PR fixes Issue #561 

Problem:
We currently set `errno=EIO` when
`ifstreamm::good()` returns false after
`ifstream::open()`. Errno from
`std::ifstream` operations is undefined
in C++ standards so we manually set it
this way to describe "all IO errorrs."
However, `EIO` is often too broad for this
purpose. For example, if `open` fails because
the file does not exist, the most common case,
you would rather want to see `errno=ENOENT`

Solution:
Leverage the fact that most of platforms will
actually set `errno` correctly as they are set
from the underlying system libraries and we
only set `errno=EIO` when it is not the case.

Augment the testing of `errno` behavior for
a few failure modes.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -476,11 +476,8 @@ static int populate_resource_db (std::shared_ptr<resource_ctx_t> &ctx)
 
     gettimeofday (&st, NULL);
     if (ctx->args.load_file != "") {
-        if (populate_resource_db_file (ctx, rd) < 0) {
-            flux_log (ctx->h, LOG_ERR, "%s: error loading resources from file",
-                      __FUNCTION__);
+        if (populate_resource_db_file (ctx, rd) < 0)
             goto done;
-        }
         flux_log (ctx->h, LOG_INFO, "%s: loaded resources from %s",
                   __FUNCTION__,  ctx->args.load_file.c_str ());
     } else {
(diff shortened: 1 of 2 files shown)


resource/modules/resource_match.cpp
```diff
@@ -357,16 +357,25 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx,
...
     std::stringstream buffer{};
 
+    saved_errno = errno;
+    errno = 0;
     in_file.open (ctx->args.load_file.c_str (), std::ifstream::in);
     if (!in_file.good ()) {
-        errno = EIO;
-        flux_log (ctx->h, LOG_ERR, "%s: opening %s",
+        if (errno == 0) {
+            // C++ standard doens't guarantted to set errno but
```
SteVwonder, line 369:
```suggestion
            // C++ standard doesn't guarantee to set errno but
```

resource/modules/resource_match.cpp
```diff
@@ -357,16 +357,25 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx,
...
+            // many of the underlying system call set it appropriately.
+            // we manually set errno only when it is not set all all.
```
SteVwonder, line 371:
```suggestion
            // we manually set errno only when it is not set at all.
```

dongahn:
@Mergifyio rebase.

dongahn:
@SteVwonder: I believe this can go in. Mind if I add MWP label?

SteVwonder:
Yep! Sounds good. Just added the MWP label.

PR #700, opened 2020-07-23
qmanager: use proposed new schedutil interface

garlick:
Update qmanager to perform its job manager handshake asynchronously, using schedutil interface proposed in flux-framework/flux-core#3063.

One test that expected `flux module load qmanager` to fail due to a failure in the handshake had to be tweaked to look for the failure in the broker log instead.

Travis will fail until the flux-core PR is merged.


garlick:
Closing per discussion at meeting.  We will revisit these changes after next tag.

Issue #506, opened 2019-07-30
Add Valgrind tests


dongahn:
From @grondo's review on PR #500:

> Now that running a workload is possible again, it might be useful to turn the valgrind tests back on. (Of course, this could be saved for a future PR)

PR #703, opened 2020-07-25
testsuite: enable valgrind testing

garlick:
Just some small updates to get valgrind running again in the test suite.

### DIFF
--- a/t/valgrind/valgrind-workload.sh
+++ b/t/valgrind/valgrind-workload.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 echo FLUX_URI=$FLUX_URI
-TESTDIR=${SHARNESS_TEST_SRCDIR:-.}/valgrind/workload.d/*
 exitcode=0
 
-# Check for no workload:
-test -d ${TESTDIR} || exit $exitcode
+flux module remove -f sched-simple
+flux module load sched-fluxion-resource load-allowlist=node,core,gpu
+flux module load sched-fluxion-qmanager
 
-for file in ${TESTDIR}/*; do
+for file in ${SHARNESS_TEST_SRCDIR:-..}/valgrind/workload.d/*; do
 	echo Running $(basename $file)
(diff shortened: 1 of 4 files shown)


t/valgrind/valgrind.supp
```diff
@@ -80,3 +80,13 @@
...
+{
+   <issue_2223>
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:dl_open_worker
+   fun:_dl_catch_*
+   fun:_dl_open
+   ...
+}
```
dongahn, line 92: Can we add a new line at this end? The GitHub render marks it. 

It is not an error at all but at the same time we won't need to get this flagged by GitHub either :-)

t/valgrind/valgrind-workload.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 echo FLUX_URI=$FLUX_URI
-TESTDIR=${SHARNESS_TEST_SRCDIR:-.}/valgrind/workload.d/*
 exitcode=0
 
-# Check for no workload:
-test -d ${TESTDIR} || exit $exitcode
+flux module remove -f sched-simple
+flux module load sched-fluxion-resource
```
dongahn, line 7: I think we need to `load-allowlist` to produce a resource graph conforming to Jobspect V1. 

```suggestion
flux module load sched-fluxion-resource
flux module load sched-fluxion-resource load-allowlist=node,core,gpu
```

t/valgrind/valgrind.supp
garlick: Interesting, I guess the one in core doesn't have one either.  OK I'll add one :-)

garlick:
I went ahead and forced a push with those changes.  Thanks @dongahn.

garlick:
One more force push to fix the issue reference (should have been #506)

PR #705, opened 2020-07-28
NEWS.md: add release notes for v0.10.0

dongahn:
The release note includes the pending PR #694. So don't merge this before that PR. I don't know if PR #695 can also be merged in time. If it does, we should add the following to our note as well.

### Fixes
* qmanager: support for ordering pending jobs by priority/submit time

PR #706, opened 2020-07-28
qmanager: order pending jobs by priority/submit time

dongahn:
This is a bug fix PR fixing Issue #612.

Order pending jobs by priority and submit time:

Problem: Currently the pending jobs within
fluxion's qmanager is ordered by enqueue time
as seenby qmanager, not by the enqueue time as seen
by job-manager. Further, it doesn't take
into account the job's priority either, which
can lead to various side-effects.

Change the "key" of the pending queue map from
uint64_t to be std::vector<double> such that
pending jobs are now ordered lexicographically
according to this vector.

Use <priority, t_submit> as the current key.

This change can easily accomodate other
order-affecting changes that are expected
to be introduced in the future including
fair share priorities.

Fixes Issue #612.

This PR is written on top of PR #694
and only consists of three commits.

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -33,6 +33,7 @@ extern "C" {
 }
 
 #include <map>
+#include <vector>
 #include <unordered_map>
 #include <string>
 #include <memory>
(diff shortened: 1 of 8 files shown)


dongahn:
@cmoussa1 and @SteVwonder:  If you can review this PR by tomorrow morning, this can go into our v0.10 tag. 

@cmoussa1: I selected you as a reviewer because this can show a glimpse of a way that your muti-factor priorities including fair share can effect the job ordering. 

If you can't get to this by tomorrow morning, that's fine. I know you are busy with your own tasks. This can wait for the next release.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -103,6 +104,7 @@ class job_t {
     std::string note = "";
     t_stamps_t t_stamps;
     schedule_t schedule;
+    const int max_priority = 32;
```
SteVwonder, line 107: Nit: Seems a bit odd to me that this is a per-job const and not a const defined either in this header or in the implementation file.  I don't have a strong objection to it being per-job either.

dongahn: Yes I agree. Because this layer should be flux-core agnostic (so that the queue policy can be carved out and used by other RMs if needed), I should just get rid of `max_priority` field from this class and only use `priority`. Then, we can have the flux-core-aware layer (qmanager) to make this conversion (32 - priority) and assign the converted value to job->priority. Would that work?

SteVwonder: 👍 Sounds good to me!

t/t1013-qmanager-priority.t
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+test_description='Fluxion takes into account pririty and t_submit'
```
cmoussa1, line 3: nit: I think this should be "priority"

dongahn: Oh.. nice catch. Thanks I will fix it.

dongahn:
I made those changes. If/When Travis CI turns green, this is good to be merged. I will update PR #705 as well.

dongahn:
Ugg.. CI fails on the new test. Let me look.

dongahn:
Everything is green. Setting MWP.

Issue #711, opened 2020-07-30
testsuite fails in resource/evaluators/test in TOSS4 buildfarm


grondo:
The TOSS 4 rpmbuild for flux-sched is failing in `make check` with a SIGABRT from `expr_eval_test01`.

I will try to get more info out of the build, including the full logs.

```
make[4]: Entering directory '/builddir/build/BUILD/flux-sched-0.10.0/resource/evaluators/test'
  CXX      expr_eval_test01-expr_eval_test01.o
  CXXLD    expr_eval_test01
make[4]: Leaving directory '/builddir/build/BUILD/flux-sched-0.10.0/resource/evaluators/test'
make  check-TESTS
make[4]: Entering directory '/builddir/build/BUILD/flux-sched-0.10.0/resource/evaluators/test'
make[5]: Entering directory '/builddir/build/BUILD/flux-sched-0.10.0/resource/evaluators/test'
../../../config/test-driver: line 107: 10649 Aborted                 (core dumped) "$@" > $log_file 2>&1
FAIL: expr_eval_test01
```

grondo:
full test output:
```
================================================================
   flux-sched 0.10.0: resource/evaluators/test/test-suite.log
================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
.. contents:: :depth: 2
FAIL: expr_eval_test01
======================
1..82
ok 1 - validates simple expr: ^  status=up      $
ok 2 - validates simple expr: ^  status=up          sched-now=allocated         
$
ok 3 - validates simple expr: ^status=up and sched-now=allocated$
ok 4 - validates simple expr: ^status=up or sched-now=allocated$
ok 5 - validates simple expr: ^status=up sched-now=free$
ok 6 - validates simple expr: ^status=up   and sched-now=free$
ok 7 - validates simple expr: ^status=up or sched-now=free$
ok 8 - validates simple expr: ^status=down   sched-now=allocated$
ok 9 - validates simple expr: ^status=down and sched-now=allocated$
ok 10 - validates simple expr: ^status=down or sched-now=allocated$
ok 11 - validates simple expr: ^status=down sched-now=free$
ok 12 - validates simple expr: ^status=down and sched-now=free$
ok 13 - validates simple expr: ^status=down or sched-now=free$
ok 14 - validates simple expr: ^sched-now=allocated status=up$
[...]

PR #715, opened 2020-07-31
resource: check return value of `find_first_not_of`

SteVwonder:
Problem: `find_first_not_of` can return `string::npos`, which if passed
as an index to an element acess, can cause an out-of-bounds access

Solution: check for this possible return value, and handle it
appropriately

Related: #711

### DIFF
--- a/resource/evaluators/expr_eval_api.cpp
+++ b/resource/evaluators/expr_eval_api.cpp
@@ -37,6 +37,9 @@ namespace resource_model {
 bool expr_eval_api_t::is_paren (const std::string &e, std::size_t at) const
 {
     std::size_t fnws = e.find_first_not_of (" \t", at);
+    if (fnws == std::string::npos) {
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


dongahn:
LGTM!

As we chatted the other day, https://github.com/flux-framework/flux-sched/blob/master/resource/evaluators/test/expr_eval_test01.cpp#L123 test case actually covers this case so I will put MWP. It just that we didn't have a compiler that could trigger it in our Travis CI matrix. This test case actually worked one our build farm :-).

Let's revisit the libc++ debug mode sooner rather than later. As stabilization should be one of the highest priories for flux-shed.

Thanks @SteVwonder!

grondo:
Sorry I meant to try this fix in the buildfarm but didn't get to it yet. I can try it before the next tag though.

dongahn:
Not a problem @grondo. With our monthly release cadence, it seems to make sense to address issues in "forward looking" manner...

Issue #713, opened 2020-07-30
resource: let flux module load return before resource acquisition


garlick:
As noted in #712, fluxion-resource needs to acquire its resources asynchronously, after the reactor starts so that `flux module load` can return while the cluster is brought up incrementally.

The libschedutil API can help if you want, or you can ignore it and continue to send the RPC directly.

If libschedutil is used, then fluxion-resource should populate a `struct schedutil_ops` struct with only the `resource_*` functions non-NULL.  Call `schedutil_create()` with the ops pointer, then call `schedutil_start()` and enter the reactor.

The `ops->resource_acquire()` callback provides the initial resource set (initially down).
The `ops->resource_up()` and `ops->resource_down()` callbacks modify the up/down state.

Alternatively, send the `resource.acquire` request, register a continuation for responses, and enter reactor without waiting for the first one.

In `ops->resource_acquire()` or in your continuation function on the first response, respond to a pending `sched-fluxion-resource.notify` request.  If it comes before acquire has responded, it will need to be cached until it the response is received.

garlick:
Renaming issue to reflect the real goal, which can be achieved much more simply by calling `flux_module_set_running()`, once flux-framework/flux-core#3104 is merged.

PR #716, opened 2020-08-02
WIP:  let flux module load return early

### LINKED ISSUE #712: qmanager: let flux module load return before synchronous fluxion resource RPC
flux-framework/flux-core#3063 proposes to change the libschedutil API to allow the scheduler to start up asynchronously.  Currently, synchronous calls are made
1) fluxion-resource module to resource.acquire
2) qmanager to fluxion-resource
3) qmanager to job manager

The end goal is to allow `flux module load` of the scheduler module(s) to return before resources are acquired so that we can boot the cluster in an incremental fashion.

There is a small overview of the new API usage in the description of flux-framework/flux-core#3063

Let's focus this issue on qmanager initialization.  It should go something like this:
* call `scheudtil_create()` with `ops->resource_acquire` set to NULL
* send request to `sched-fluxion-resource.notify`
* enter reactor
* on first notify response, call `schedutil_start()`
* `ops->hello()` will be called zero or more times to get _R_ for running jobs
* `ops->ready()` will be called once all of those jobs are processed.  Call `schedutil_ready()`
* `ops->alloc()`, `ops->free()`, and `ops->cancel()` requests will be received as before

The main change is that what used to happen before entering the reactor now happens afterwards, and the sequence is triggered by the first notify response from sched-fluxion-resource.  After that libschedutil takes care of the synchronization in the reactor.

Edit: #700 was a PR (withdrawn) to implement this in qmanager, before #694 was merged.
Note that since then I noticed the `ready_cb()` in that PR had the wrong type for its "context", and I added a `flux_t *` argument to the ready prototype, which makes it look like all the other callbacks, and helps in this situation.

garlick:
Call `flux_module_set_running()` before beginning synchronous RPCs in fluxion-qmanager and fluxion-resource so that `flux module load` doesn't block until the initialization completes.  This resolves a deadlock introduced by broker initialization changes introduced in flux-framework/flux-core#3057.

Fixes #712 and #713.

Marking as a WIP until this can be tested with the broker initialization changes.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -394,6 +394,14 @@ static int enforce_options (std::shared_ptr<qmanager_ctx_t> &ctx)
         flux_log_error (ctx->h, "%s: enforce_queues", __FUNCTION__);
         return rc;
     }
+    /* Before beginning synchronous handshakes with fluxion-resource
+     * and job-manager, set module status to 'running' to let flux module load
+     * return success.
+     */
+    if ( (rc = flux_module_set_running (ctx->h)) < 0) {
+        flux_log_error (ctx->h, "%s: flux_module_set_running", __FUNCTION__);
+        return rc;
+    }
     if ( (rc = handshake_resource (ctx)) < 0) {
         flux_log_error (ctx->h, "%s: handshake_resource", __FUNCTION__);
         return rc;
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -869,6 +869,13 @@ static int populate_resource_db (std::shared_ptr<resource_ctx_t> &ctx)
         flux_log_error (ctx->h, "%s: gettimeofday", __FUNCTION__);
         goto done;
     }
+    /* Before beginning synchronous resource.acquire RPC, set module status
+     * to 'running' to let flux module load return success.
+     */
+    if ( (rc = flux_module_set_running (ctx->h)) < 0) {
+        flux_log_error (ctx->h, "%s: flux_module_set_running", __FUNCTION__);
+        goto done;
+    }
     if (ctx->args.load_file != "") {
         if (populate_resource_db_file (ctx) < 0)
             goto done;


dongahn:
> FAIL: t1008-recovery-none.t 5 - recovery: qmanager w/o an option must fail (rv1_nosched)
1740     t1008-recovery-none.t: FAIL: N=7   PASS=6   FAIL=1 SKIP=0 XPASS=0 XFAIL=0
1741ERROR: t1008-recovery-none.t - exited with status 1
1742t1009-recovery-multiqueue.t: PASS: N=9   PASS=9   FAIL=0 SKIP=0 XPASS=0 XFAIL=0
1743FAIL: t1010-sync-modules.t 1 - sync: fluxion-resource cannot load w/ sched-simple
1744FAIL: t1010-sync-modules.t 2 - sync: qmanager does not load w/o fluxion-resource
1745      t1010-sync-modules.t: FAIL: N=9   PASS=7   FAIL=2 SKIP=0 XPASS=0 XFAIL=0
1746ERROR: t1010-sync-modules.t - exited with 

> FAIL: t4000-match-params.t 4 - loading resource module with a nonexistent GRUG fails
1780FAIL: t4000-match-params.t 5 - loading resource module with a nonexistent XML fails
1781FAIL: t4000-match-params.t 6 - loading resource module with incorrect reader fails
1782     t3025-resource-find.t: PASS: N=20  PASS=20  FAIL=0 SKIP=0 XPASS=0 XFAIL=0
1783      t4000-match-params.t: FAIL: N=9   PASS=6   FAIL=3 SKIP=0 XPASS=0 XFAIL=0
1784ERROR: t4000-match-params.t - exited with status 1   1      1   49.05s 0

I think the failures are pretty minimal and probably as expected... At first glance, I have to believe these are manageable. If you want, I can take a look early next week so that you can continue to focus on the system instance (and your vac).

garlick:
That'd be great @dongahn.  I was just trying to get a build environment working for sched up here - was hoping travis would pass and I would get lucky, but no such luck.  I'll post changes if I run anything down but your help would be most appreciated when you have time.

dongahn:
@garlick: I can look at this today. @garlick: Mind if I close this and work on a PR? Pushing mine to your repo etc can be a bit cumbersome.

garlick:
Sure, whatever works for you @dongahn!

dongahn:
One case where this leads to a problem is that a test expects `flux module load` failure yet it returns success. This would be a case when the failure happens after the module calls `flux_module_set_running()`. 

Instead of `test_must_fail flux module load ...`, I can try `flux module info` or `flux module stat`?

dongahn:
`flux module stats` seems to work.

PR #718, opened 2020-08-05
Update for non-integer jobid encodings

grondo:
This small PR updates various parts of flux-sched to avoid the assumption of integer-only jobids.

Most of the changes are straightforward, and should work whether flux-core commands are emitting integer or F58 encoded jobids. So this paves the way for an update in flux-core to allow more F58 encoded jobids.

### DIFF
--- a/t/scripts/flux-ion-resource.py
+++ b/t/scripts/flux-ion-resource.py
@@ -12,6 +12,8 @@ import json
 import flux
 import time
 
+from flux.job import JobID
+
 def heading ():
     return '{:20} {:20} {:20} {:20}'.format ('JOBID', 'STATUS',
                                              'AT', 'OVERHEAD (Secs)')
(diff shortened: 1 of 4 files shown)


dongahn:
I just had a cursory look and this looks good to me. I will take a look at this more closely as soon as practical.

t/scripts/flux-ion-resource.py
```diff
@@ -277,13 +279,13 @@ def main ():
     # Positional argument for update sub-command
     #
     parser_u.add_argument ('RV1', metavar='RV1', type=str, help='RV1 file name')
-    parser_u.add_argument ('jobid', metavar='Jobid', type=int, help='Jobid')
+    parser_u.add_argument ('jobid', metavar='Jobid', type=JobID, help='Jobid')
```
dongahn, line 282: @grondo: I don't know JobID class well enough. How does this work if you reference `jobid` (of class `JobID`) in the code. Will the class automatically convert this into an integer encoding? If so, this code looks great.

grondo: It should, `JobID` is a subclass of `int`. However, if you use JobID in a context where it is converted to a string, then it will encode as whatever the `__str__` method encodes to (currently decimal integer, but will become F58 encoded jobid after flux-framework/flux-core#3111 is merged)

grondo:
Thanks!

PR #719, opened 2020-08-05
Let flux module load return early

### LINKED ISSUE #712: qmanager: let flux module load return before synchronous fluxion resource RPC

dongahn:
This PR is the rework of @garlick's closed PR #716.

- Call flux_module_set_running() before beginning synchronous RPCs in `sched-fluxion-qmanager` and `sched-fluxion-resource` so that flux module load doesn't block until the initialization completes. This resolves a deadlock introduced by broker initialization changes introduced in flux-framework/flux-core#3057.

- Adjust the test cases that require `flux module load` of a fluxion module to fail. When `flux_module_set_running()` was called, the error hasn't yet happened so `flux module load` returned successfully, producing some test failures. Use `flux module stats` to test the module initialization failure, instead.

Fixes #712 and #713.

### DIFF
--- a/t/t1008-recovery-none.t
+++ b/t/t1008-recovery-none.t
@@ -37,7 +37,8 @@ test_expect_success 'recovery: submit a job (rv1_nosched)' '
 
 test_expect_success 'recovery: qmanager w/o an option must fail (rv1_nosched)' '
     reload_resource load-allowlist=node,core,gpu match-format=rv1_nosched &&
-    test_must_fail reload_qmanager
+    reload_qmanager &&
+    test_must_fail flux module stats sched-fluxion-qmanager
 '
 
 test_expect_success 'removing resource and qmanager modules' '
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -394,6 +394,13 @@ static int enforce_options (std::shared_ptr<qmanager_ctx_t> &ctx)
         flux_log_error (ctx->h, "%s: enforce_queues", __FUNCTION__);
         return rc;
     }
+    return rc;
+}
+
+static int handshake (std::shared_ptr<qmanager_ctx_t> &ctx)
+{
+    int rc = 0;
+
     if ( (rc = handshake_resource (ctx)) < 0) {
         flux_log_error (ctx->h, "%s: handshake_resource", __FUNCTION__);
         return rc;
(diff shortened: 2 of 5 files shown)


dongahn:
@Mergifyio rebase

t/t1008-recovery-none.t
```diff
@@ -37,7 +37,8 @@ test_expect_success 'recovery: submit a job (rv1_nosched)' '
 
 test_expect_success 'recovery: qmanager w/o an option must fail (rv1_nosched)' '
     reload_resource load-allowlist=node,core,gpu match-format=rv1_nosched &&
-    test_must_fail reload_qmanager
+    reload_qmanager &&
+    test_must_fail flux module stats sched-fluxion-qmanager
```
garlick, line 41: I think this is OK but just to double check that we are on the same page, this is not racy because the stats request will not be handled during the synchronous RPCs.  Ultimately the module will exit and the stats request (still in the queue) will get an ENOSYS response from the module teardown code?

dongahn: Exactly!

dongahn:
> Only possible change I would suggest is the title of the second commit. Maybe it should be something like qmanager,resource: let module load return early? "fluxion/testsuite" seems like the commit just pertains to the test suite since fluxion is the name of the project.

Yes this make sense. I will make this change and set MWP unless @SteVwonder has other quick comments.

Issue #508, opened 2019-08-09
qmanager: preserve reservations across scheduler loops (and optimize reservation clearing)


SteVwonder:
Right now (in #504), at the end of every scheduler loop, reservations are cleared.

One minor change would be to clear reservations at the start of each scheduler loop.  This would mean that reservation information is available for querying in between scheduler loops (e.g., to estimate when a job will start)

As an optimization in backfill schedulers, we should only clear reservations when necessary.  As a first cut, we can keep reservations across scheduler loops and only clear them when a job completes or when a new job is submitted whose priority is higher than a job with an existing reservation.  As a second cut, we can reduce the number of reservations cleared in that latter case by only clearing reservations of jobs whose priority is lower than the new job.

dongahn:
I am looking at this and I have to believe making the following one-line change can give this functionality. 

```diff
diff --git a/qmanager/policies/queue_policy_bf_base_impl.hpp b/qmanager/policies/queue_policy_bf_base_impl.hpp
index da3fadab..027e7432 100644
--- a/qmanager/policies/queue_policy_bf_base_impl.hpp
+++ b/qmanager/policies/queue_policy_bf_base_impl.hpp
@@ -185,8 +185,8 @@ int queue_policy_bf_base_t<reapi_type>::run_sched_loop (void *h,
 {
     int rc = 0;
     rc = cancel_completed_jobs (h);
-    rc += allocate_orelse_reserve_jobs (h, use_alloced_queue);
     rc += cancel_reserved_jobs (h);
+    rc += allocate_orelse_reserve_jobs (h, use_alloced_queue);
     return rc;
 }
```

This seems to mostly work. But once time on my Mac OS/X docker, I got the following error message. This is as if the reservation cancel request from `qmanager` arrive after the match request on the same jobid. 

This shouldn't be the case because of the RPC non-overtaking rule (a later RPC cannot overtake an earlier RPC). 


```console
ahn1@81451f60e736:/usr/src$ flux start -s 4

ahn1@81451f60e736:/usr/src$ flux mini submit -N 2 -n 8 sleep 360

ahn1@81451f60e736:/usr/src$ flux mini submit -N 2 -n 8 sleep 360
699173699584
ahn1@81451f60e736:/usr/src$ flux mini submit -N 2 -n 8 sleep 360
711672725504
ahn1@81451f60e736:/usr/src$ flux resource list
     STATE NNODES   NCORES    NGPUS
      free      0        0        0
[...]

PR #735, opened 2020-08-27
qmanager: Add support for job annotation

dongahn:
This PR adds support for annotating jobs with their belonging queue name and start-time estimate.

- Modify the backfill scheduler queue policy base class code such that all reservations are removed in the beginning, instead of at the end, of each scheduler loop iteration. This way, the reserved time for each high priority job is preserved until the next scheduler loop is called and can be used as its start-time estimate during this window.

- As part of our scheduler loop iteration epilogue, send a job annotation for each of the pending jobs that have reservations. Only send the annotation when the reserved time changes from its previously calculated reserved time.

- Add tests to ensure annotating jobs with queue and start-time estimate (`t_estimate`) work as expected. Test covers all three backfilling policies. Note that this introduced a test with a known breakage where flux-jobs fail when there are jobs that miss
the t_estimate annotation (which would generally be the case with non conservative backfilling policy): Issue https://github.com/flux-framework/flux-core/issues/3161.

Note that this test would be affected by this effect (https://github.com/flux-framework/flux-core/issues/3165) when run under Valgrind (`FLUX_TEST_VALGRIND=t`).

Fix Issue #626, #508.

### DIFF
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -69,7 +69,10 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
         while ( (job = queue->alloced_pop ()) != nullptr) {
             if (schedutil_alloc_respond_success_pack (schedutil, job->msg,
                                                       job->schedule.R.c_str (),
-                                                      NULL) < 0) {
+                                                      "{ s:{s:s s:f} }",
+                                                      "sched",
+                                                          "queue", queue_name.c_str (),
+                                                          "t_estimate", 0.0) < 0) {
                 flux_log_error (h, "%s: schedutil_alloc_respond_pack (queue=%s)",
                                 __FUNCTION__, queue_name.c_str ());
                 goto out;
@@ -89,6 +92,22 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
             }
             flux_log (h, LOG_DEBUG, "%s (id=%jd queue=%s)", note.c_str (),
                      static_cast<intmax_t> (job->id), queue_name.c_str ());
+        }
+        for (job = queue->pending_begin (); job != nullptr;
+             job = queue->pending_next ()) {
+            // if old_at == at, then no reason to send this annotation again.
+            if (job->schedule.at == job->schedule.old_at)
+                continue;
+            if (schedutil_alloc_respond_annotate_pack (
+                    schedutil, job->msg,
+                    "{ s:{s:s s:f} }",
+                    "sched",
+                        "queue", queue_name.c_str (),
+                        "t_estimate", static_cast<double> (job->schedule.at))) {
+                flux_log_error (h, "%s: schedutil_alloc_respond_annotate_pack",
+                                __FUNCTION__);
+                goto out;
+            }
         }
     }
     rc = 0;
--- a/t/t1014-annotation.t
+++ b/t/t1014-annotation.t
@@ -1,0 +1,182 @@
+#!/bin/sh
+
+test_description='Test job annotation'
+
+. `dirname $0`/sharness.sh
+
+hwloc_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/hwloc-data`
+# 4 brokers, each (exclusively) have: 1 node, 2 sockets, 16 cores (8 per socket)
+excl_4N4B="${hwloc_basepath}/004N/exclusive/04-brokers"
+
+skip_all_unless_have jq
+
+test_under_flux 1
+
+nonexistent_annotation(){
+    jobid=$(flux job id ${1}) &&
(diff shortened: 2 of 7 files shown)


dongahn:
@Mergifyio rebase

dongahn:
@Mergifyio rebase

dongahn:
OK. This PR is also ready for your review. 

@chu11: if you can take a pick at how I used annotation API, that'd be useful. Also I left some notes on known breakage due to a race in annotation API and an issue with flux-jobs. Reviewing this might be useful when you want to test your fix on a reproducer.

@SteVwonder: If you can review the preservation of reservation etc, that'd be appreciated.

dongahn:
@Mergifyio rebase

dongahn:
If this can be reviewed soon, this can go into our Aug release.

dongahn:
Great! Thanks.

dongahn:
@mergifyio rebase

dongahn:
@SteVwonder or @milroy: Just one very quick look at this PR from either of you is all I need to make our fluxion v0.11 release in Aug!

SteVwonder:
Ok.  Starting my pass now.

t/t1014-annotation.t
```diff
@@ -0,0 +1,177 @@
...
+}
+
+validate_sched_annotation(){
+    jobid=$(flux job id ${1}) &&
+    queue_name=${2} &&
+    start_time=${3} &&
+    ann=$(flux job list -A | grep ${jobid} | jq -c '.annotations') &&
+    queue=$(echo ${ann} | jq '.sched.queue') &&
+    t_est=$(echo ${ann} | jq '.sched.t_estimate') &&
+    test "\"${queue_name}\"" = "${queue}" &&
```
SteVwonder, line 28: Out of curiosity, why the escaped quotes `"\"` and `\""` just here and not elsewhere?

t/t1014-annotation.t
```diff
@@ -0,0 +1,177 @@
...
+    test "${start_time}" != "${t_est}"
```
SteVwonder, line 29: Why not test for equality?
```suggestion
    test "${start_time}" == "${t_est}"
```

The tests below are confusing to read since I was expecting the final argument to be the expected start time estimate.

t/t1014-annotation.t
```diff
@@ -0,0 +1,177 @@
...
+    print_queue
+'
+
+test_expect_success 'annotation: cancel all active jobs 1' '
+    active_jobs=$(flux job list --state=active | jq .id) &&
+    for job in ${active_jobs}; do flux job cancel ${job}; done &&
+    for job in ${active_jobs}; do flux job wait-event -t 10 ${job} clean; done
+'
+
+test_expect_failure 'annotation: flux-jobs fail when some jobs miss t_est' '
```
SteVwonder, line 73: Is the idea to update this test once we fix the `flux jobs` failure in flux-core?

t/t1014-annotation.t
dongahn: Yes, this is because this is a JSON string which has enclosed with double quotes.

t/t1014-annotation.t
dongahn: OK. This is a good idea to increase readability. I will make this change.

t/t1014-annotation.t
dongahn: Yes, when flux-jobs is fixed, we will turn this to `test_expect_success`.

dongahn:
So based on @chu11 and @SteVwonder's reviews, I will make one minor update without interim commit and add MWP label. Thanks Al and Steven!

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -69,7 +69,10 @@ int qmanager_cb_t::post_sched_loop (flux_t *h, schedutil_t *schedutil,
         while ( (job = queue->alloced_pop ()) != nullptr) {
             if (schedutil_alloc_respond_success_pack (schedutil, job->msg,
                                                       job->schedule.R.c_str (),
-                                                      NULL) < 0) {
+                                                      "{ s:{s:s s:f} }",
+                                                      "sched",
+                                                          "queue", queue_name.c_str (),
+                                                          "t_estimate", 0.0) < 0) {
```
chu11, line 75: my internal cache of the annotations API had been flushed and I just realized.  This is ok b/c of https://github.com/flux-framework/flux-core/issues/3161, but after that is fixed, this should presumably be changed to `"{ s:{s:s s:n} }`, b/c you presumably want to wipe out the `t_estimate` field once a job has been allocated?

But you don't have to.  Simply setting `t_estimate` to 0.0 when it has started running is also fine.  Just depends on what you want to do.

dongahn: Thanks @chu11. I think I kept it that way for testing because of other failures with flux-jobs (https://github.com/flux-framework/flux-core/issues/3161). Once that's fixed, I will go with `n` and pass none to remove that from the job record.

Issue #709, opened 2020-07-30
resource: segfault during match of multi-socket jobspec on Corona


SteVwonder:
Using flux-sched v0.10.0 built from source with the docker-run-checks.sh script based on the `focal` image.

I encounter a segfault when querying a multi-socket jobspec using resource-query against a single-node hwloc of Corona:
```
erbein1@94f9c9953c87:/usr/src/t$ gdb --args ../resource/utilities/resource-query -L /usr/src/t/data/hwloc-data/001N/amd_gpu/corona11.xml -f hwloc -W node,socket,core -S CA -P low
<snip>
(gdb) r
Starting program: /usr/src/resource/utilities/resource-query -L /usr/src/t/data/hwloc-data/001N/amd_gpu/corona11.xml -f hwloc -W node,socket,core -S CA -P low
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
INFO: Loading a matcher: CA
resource-query> match allocate /usr/src/t/data/resource/jobspecs/node_local_storage/corona.yaml

Program received signal SIGSEGV, Segmentation fault.
0x000055de0d731e25 in Flux::resource_model::detail::dfu_impl_t::dom_slot (this=<optimized out>, meta=..., u=<optimized out>, slot_shape=..., pristine=<optimized out>, excl=<optimized out>, dfu=...) at traversers/dfu_impl.cpp:513
513                     eval_edg_t ev_edg ((*egroup_i).edges[0].count,
(gdb) bt
#0  0x000055de0d731e25 in Flux::resource_model::detail::dfu_impl_t::dom_slot (this=<optimized out>, meta=..., u=<optimized out>, slot_shape=..., pristine=<optimized out>, excl=<optimized out>, dfu=...) at traversers/dfu_impl.cpp:513
#1  0x000055de0d730c7e in Flux::resource_model::detail::dfu_impl_t::dom_dfv (this=<optimized out>, meta=..., u=<optimized out>, resources=std::vector of length 1, capacity 1 = {...}, pristine=<optimized out>, excl=0x7ffe9575363f,
    to_parent=...) at traversers/dfu_impl.cpp:554
[...]

PR #737, opened 2020-08-29
Handle invalid sibling requests in jobspec more gracefully 

dongahn:
This PR adds support for detecting and more gracefully handling a class of invalid sibling requests within jobspecs. While our DFU traverser has in-line documents and README to describe its fundamental inability to match a jobspec containing sibling requests with the same resource type, for example,

```
    node[1]->socket[1]
           ->socekt[1]
```

the current code does not detect and handle this condition gracefully.

- Detect this condition at the `match()` method and bubble this up to the upper calls so that they can deem this as `match_kind_t::NONE_MATCH`;

- Related, avoid a potential buffer overflow that can arise when we iterate through the slots discovered by the DFU walk from the visiting resource vertex. This condition can occur if `cnt_slot()` incorrectly counts the number of available slots. Because `cnt_slot()` depends on the user-provided jobspec input, this can be exploited or at least lead to a less robust or undefined behavior;

- Add sharess test cases to check if `resource-query` can handle this class of jobspecs for three different job specs.

Fix Issue #709.

### DIFF
--- a/resource/evaluators/edge_eval_api.cpp
+++ b/resource/evaluators/edge_eval_api.cpp
@@ -195,10 +195,28 @@ int evals_t::merge (evals_t &o)
     return 0;
 }
 
-void evals_t::rewind_iter_cur ()
+void evals_t::eval_egroups_iter_reset ()
 {
-    iter_cur = m_eval_egroups.begin ();
+    m_iter_cur_reset = true;
 }
+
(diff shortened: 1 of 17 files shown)


dongahn:
Thanks @SteVwonder! I just resolved the conflict with the latest merge and forced a push. I'm putting MWP as well.

dongahn:
@Mergifyio rebase

PR #741, opened 2020-09-01
rc: only load/unload scheduler on rank 0

garlick:
Problem: flux-core is about to transition to rc1/rc3 scripts executing
on all ranks, which would cause the fluxion scheduler to be loaded
on all ranks.

Check the rank at the beginning of the rc scripts and exit 0 if running
on a rank other than 0.

This change has no effect when run with the current 0.19.0 flux-core, but it will allow flux-sched to continue to work into 0.20.0 development.

### DIFF
--- a/etc/01-sched-fluxion-qmanager-stop
+++ b/etc/01-sched-fluxion-qmanager-stop
@@ -1,5 +1,7 @@
 #!/bin/bash -e
 
+test $(flux getattr rank) -eq 0 || exit 0
+
 if [ -z ${FLUXION_QMANAGER_RC_NOOP} ]; then
     flux module remove -f sched-fluxion-qmanager
 fi
--- a/etc/01-sched-fluxion-resource-start
+++ b/etc/01-sched-fluxion-resource-start
@@ -18,6 +18,8 @@
 # We now use "flux reload -f resource" so that when these failing tests
 # load it via FLUX_RC_EXTRA, the in-tree version can be loaded.
 
+test $(flux getattr rank) -eq 0 || exit 0
+
 if [ -z ${FLUXION_RESOURCE_RC_NOOP} ]; then
     # Unloading sched-simple as sched-fluxion-resource requires resource.acquire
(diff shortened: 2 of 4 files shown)


garlick:
Would be nice to get this in ASAP, so that we can get results from sched build in flux-framework/flux-core#3168, if someone has a moment for a review.

grondo:
I'm not a "fluxioneer", but this LGTM.

The use of
```sh
test $(flux getattr rank) -eq 0 || exit 0
```

might be common enough that it made me wonder if we should provide a simple(r) command to check if the current rank is 0. Not sure what the command would be, but I thought I'd throw the idea out into the ether.

dongahn:
I'm on it now.

dongahn:
Perhaps we should do this at the source code level so that this constraint will be applied not only for initial setup but also for manual module load?

To be explicit, having this conditional in the RC scripts is okay as well though. So I'm adding MWP label.

dongahn:
@Mergifyio rebase

garlick:
Thanks!  I think we also need a review to merge?

> Perhaps we should do this at the source code level so that this constraint will be applied not only for initial setup but also for manual module load?

You mean have the module refuse to load on rank > 0?   That'd be fine, although we would likely still want to make the rc script conditional to avoid the extra overhead.  I think the net result is already that a load on another rank will fail, since neither job manager nor resource module will allow more than one scheduler to handshake.

dongahn:
> You mean have the module refuse to load on rank > 0? That'd be fine, although we would likely still want to make the rc script conditional to avoid the extra overhead.

Yes, the conditional is fine by me and hence MWP. 

>  I think the net result is already that a load on another rank will fail, since neither job manager nor resource module will allow more than one scheduler to handshake.

Well, if fluxion modules are manually loaded into non root rank where there is no scheduler being loaded, this would still succeed? This shouldn't be mode to support just yet? 

I will take care of review approval as well.

garlick:
Thanks!

Issue #736, opened 2020-08-28
resource: support alternate writer format for `find`


SteVwonder:
Currently, `find` only supports RV1_nosched as the writer format. It would be nice to be able to customize the writer format (potentially even per-request).

dongahn:
Yes, this can be done fairly easily by allowing the predefined writer name as an optional argument.

The first real use case for `find` was for `flux resource list` which requires `rv1_nosched` but the code is there to emit this in other formats.

My guess is your use case is to emit this in `JGF` for your multi storage work?

SteVwonder:
> My guess is your use case is to emit this in JGF for your multi storage work?

Yeah, it's been really useful for debugging unsatisfiable jobspecs.  I have a patch in my current commit stack that sets the writer for `find` to the same as the one for `match`, but I think it could easily be extended to be an argument in the `find` RPC.  So once the L2 is done, I can take a shot at that.

PR #743, opened 2020-09-02
Add job shell staging plugin

SteVwonder:
This PR demonstrates the end-to-end capability of staging to multiple storage tiers within a single jobspec.  It includes sharness tests of the job-shell in standalone mode as well as end-to-end job submission tests.  These tests verify that you can stage to node-local storage, cluster-local storage, and to both simultaneously.  In the node-local case, every job shell rank that has a storage unit assigned to it stages into the storage.  In the cluster-local case, only the rank 0 job shell stages the data.

Other additions:

- Adds the nlohmann json library for more native C++ json handling
- Adds support for alternate writer formats for `find` (#736)


dongahn:
Awesome! This is a significant addition to our L2 milestone this year, @SteVwonder! We will surely circle back to this next fiscal year to land this.

SteVwonder:
Now that I've completed the refactor I wanted to do, and Travis is happy, I think this PR is finally ready for review.

dongahn:
@SteVwonder: so that you know. This is my next priority.

t/data/resource/jobspecs/advanced/test006.yaml
```diff
@@ -0,0 +1,39 @@
+version: 1
```
dongahn, line 1: This would not be version 1.

t/t3006-resource-advanced.t
```diff
@@ -64,4 +64,15 @@ test_expect_success LONGTEST "${test004_desc}" '
     test_cmp 004.R.out ${exp_dir}/004.R.out
 '
 
+cmds005="${cmd_dir}/cmds05.in"
+test005_desc="Resource labels in jobspec carrythrough to JGF R"
```
dongahn, line 68:
```suggestion
test005_desc="Resource labels in jobspec carry through to JGF R"
```

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+        json label = storage_attr["label"];
+        if (!label.is_string()) {
+            shell_log_error ("Storage label must be a string");
+            return JobspecState::invalid;
+        }
+
+        std::string granularity = storage_attr["granularity"].get<std::string>();
+        if ((granularity != "node") && (granularity != "job")) {
+            shell_log_error ("Storage entry contains a granularity not 'job' or 'node' (%s)", granularity.c_str());
```
dongahn, line 57: line too long?

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+        shell_log_error ("Error retrieving shell info string");
+        errno = EINVAL;
+        return -1;
+    }
+
+    json info;
+    try {
+        info = json::parse(info_cstr);
+    } catch (json::parse_error& e) {
+        shell_log_error ("Error parsing shell \"info\" json at byte %zu. %d: %s", e.byte, e.id, e.what());
```
dongahn, line 202: line too long?

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+        shell_log_error ("Error extracting keys from \"info\" json. %d: %s", e.id, e.what());
+        errno = EINVAL;
+        return -1;
+    }
+
+    JobspecState jobspec_state = validate_jobspec (jobspec);
+    switch (jobspec_state) {
+    case JobspecState::invalid: errno=EINVAL; return -1;
+    case JobspecState::valid_no_staging: return 0;
+    case JobspecState::valid_staging: break;
```
dongahn, line 223: Probably more conventional and readable to have each statement written on its own line.

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include <string>
+#include <iostream>
+#include <boost/optional.hpp>
+#include <boost/filesystem.hpp>
+#include "src/common/libutil/json.hpp"
+
+namespace fs = boost::filesystem;
+using json = nlohmann::json;
+
+enum JobspecState { invalid, valid_no_staging, valid_staging};
```
dongahn, line 26: We have been using `enum class` instead of plain `enum`:
https://github.com/flux-framework/flux-sched/blob/master/resource/schema/resource_data.hpp#L48

We have not been using camel back style. So my suggestion would be `enum class job_spec_state_t`

The fields would be better with capital: e.g., `INVALID`.

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+static JobspecState validate_jobspec (json jobspec)
```
dongahn, line 28: Did you mean to pass jobspec by value?

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+            }
+        }
+
+        json label = storage_attr["label"];
+        if (!label.is_string()) {
+            shell_log_error ("Storage label must be a string");
+            return JobspecState::invalid;
+        }
+
+        std::string granularity = storage_attr["granularity"].get<std::string>();
```
dongahn, line 55: Did you mean to assign by value?

src/shell/data-staging.cpp
```diff
@@ -0,0 +1,243 @@
...
+        return boost::none;
+    }
+}
+
+static std::map<std::string, json> find_storage_resources (json R, int rank)
+{
+    std::map<std::string, json> storage_resources;
+    json jgf;
+    try {
+        jgf = R["scheduling"]["graph"];
```
dongahn, line 101: There are lots of uses of `operator[]`. I don't know what the underlying container this new json library is using. But there are various side effects with `operator[]` if this is `std::map` type. For example, if the key is non-existent, this puts a temporary value object under that key. 

The rule of thumb has been to use this method only when the key is guaranteed to exist. Just wanted to leave a note in case this has been your assumption too...

t/Makefile.am
```diff
@@ -81,7 +81,8 @@ TESTS = \
     t5000-valgrind.t \
     t6000-graph-size.t \
     t6001-match-formats.t \
-    t6002-graph-hwloc.t
+    t6002-graph-hwloc.t \
+	t7000-shell-datastaging.t
```
dongahn, line 85: Seems a tab is used instead of whitespaces or similar. The indentation looks incorrect with Github renderer.

t/data/shell/cluster-local/jobspec.yaml
```diff
@@ -0,0 +1,38 @@
+version: 1
```
dongahn, line 1: Version 9999 instead

t/data/shell/combined/jobspec.yaml
```diff
@@ -0,0 +1,46 @@
+version: 1
```
dongahn, line 1: Version

t/data/shell/cluster-local/jobspec.yaml
```diff
@@ -0,0 +1,38 @@
...
+# a comment
+attributes:
+  system:
+    duration: 57600
+    shell:
+      options:
+        verbose: 1
+  user:
+    data-staging:
+      - test: true
```
dongahn, line 29: For my own edification, what is the `test` key?

t/data/shell/node-local/jobspec.yaml
```diff
@@ -0,0 +1,37 @@
+version: 1
```
dongahn, line 1: Version

t/t7000-shell-datastaging.t
```diff
@@ -0,0 +1,141 @@
+#!/bin/sh
+
+test_description='Test data-staging job shell plugin'
+
+. `dirname $0`/sharness.sh
+
+shell_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/shell`
+shell_plugin_path="$(readlink -e ${SHARNESS_BUILD_DIRECTORY}/src/shell/.libs)"
+grug_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/resource/grugs/`
```
dongahn, line 9: This variable is unused.

dongahn (changes requested):
@SteVwonder: Thank you for the work! This is a great way to be prepare for future multi-tired storage support. 

I just had my first pass. As you requested, I looked more closely on your C++ codes added to the existing resource infrastructure. There are things like memory copy semantics I want you to look at. 

I had some cursory look at the jobshell plugin src codes and added some comments. There are some cleanup that would be good to have.

Once that's done, I will have another pass.

resource/traversers/dfu_impl.cpp
```diff
@@ -602,6 +602,13 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta, vtx_t u,
         goto done;
     to_parent.set_avail (avail);
     to_parent.set_overall_score (dfu.overall_score ());
+
+    for (auto &resource: resources) {
+        if ((resource.type == (*m_graph)[u].type) &&
+            (!resource.label.empty())) {
+            (*m_graph)[u].idata.ephemeral.insert (m_best_k_cnt, "label", resource.label);
```
dongahn, line 609: Oops. Meant to leave a note. This would need an error check.

SteVwonder:
Thanks @dongahn! All your suggestions seem doable in the next couple of days, so I will attempt to have a new version ready for review by the end of the week.

Excellent catch on the missing comparison in the tests.  I hope I just accidentally lost that line in a rebase somewhere.  🤦 

And thanks for taking an extra look at the C++ copy-relevant stuff.  That was exactly the kind of thing I was worried I was doing incorrectly without realizing it.

t/t1015-find-format.t
```diff
@@ -0,0 +1,105 @@
...
+test_under_flux 1
+
+test_expect_success 'find format: find query simple works' '
+    cat > find.cmd <<-EOF &&
+	find status=up
+	quit
+EOF
+    ${query} -d -L ${excl_1N1B}/0.xml -f hwloc -F simple -S CA -P high \
+             -W node,core -t simple.query.raw < find.cmd &&
+    strip_info_lines < simple.query.raw > simple.query.out
```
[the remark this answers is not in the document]
SteVwonder: Yeah, this is confusing.

Rather than add a bunch of new json/txt files to compare the output of `flux ion-resource find` against, I instead opted to generate the "ground truth" find files using `resource-query`, which is what is happening in these first few tests.

Depending on what you think is best, I could:

- just add a comment to the tests outlining what is happening and why the first set of tests don't have a comparison being performed
- add explicit "ground truth" files to the repo to compare against (in which case the first few tests could serve to test that `resource-query` outputs the correct find format too)

resource/writers/match_writers.cpp
```diff
@@ -215,11 +215,13 @@ int jgf_match_writers_t::emit (std::stringstream &out)
...
 int jgf_match_writers_t::emit_vtx (const std::string &prefix,
                                    const f_resource_graph_t &g, const vtx_t &u,
-                                   unsigned int needs, bool exclusive)
+                                   unsigned int needs, bool exclusive, uint64_t epoch)
 {
     int rc = 0;
     json_t *o = NULL;
     json_t *b = NULL;
+    auto ephemeral = g[u].idata.ephemeral;
+    auto eph_map = ephemeral.to_map (epoch);
```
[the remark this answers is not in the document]
SteVwonder: Good catch.  It looks like I'm bumping up against the fact that `g` is passed in as `const` to `emit_vtx`:

```
writers/match_writers.cpp: In member function ‘virtual int Flux::resource_model::jgf_match_writers_t::emit_vtx(const string&, const f_resource_graph_t&, const vtx_t&, unsigned int, bool, uint64_t)’:
writers/match_writers.cpp:223:55: error: passing ‘const Flux::resource_model::ephemeral_t’ as ‘this’ argument discards qualifiers [-fpermissive]
     auto &eph_map = g[u].idata.ephemeral.to_map (epoch);
                                                       ^
In file included from ../resource/schema/infra_data.hpp:29:0,
                 from ../resource/schema/resource_data.hpp:35,
                 from writers/match_writers.cpp:27:
../resource/schema/ephemeral.hpp:39:41: note:   in call to ‘std::map<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >& Flux::resource_model::ephemeral_t::to_map(uint64_t)’
     std::map<std::string, std::string>& to_map (uint64_t epoch);
```

If we were just attempting to convert the ephemeral container into a map with non-stale data, then const would be fine (although require some extra copying).  The sticking point is the call to `ephemeral_t::clear` further down in this function.

@dongahn: how would you recommend proceeding?

One option that I see would be to remove the `const` qualifier from `g` in `emit_vtx`, but that doesn't seem like a great option.  Are we OK with `emit_vtx` no longer be idempotent?

Another might be a `post_emit` or `vtx_cleanup` method within `dfu_impl.cpp` which is called after the `emit_vtx` and wouldn't have the `const` qualifier on `g` (or `needs`, `exclusive`, or `epoch`).  It would also simplify the cleanup of the vertices for other match writers (which in writing this, I just realized I did not attempt any `clear`ing of the ephemeral_t for non-jqf writers like rv1).

src/shell/data-staging.cpp
SteVwonder: Good call.  It looks like that behavior is indeed used in this json library to make it STL-like.  I switched the accesses over to [`.at()`](https://nlohmann.github.io/json/api/basic_json/at/)

resource/writers/match_writers.cpp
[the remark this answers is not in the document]
dongahn: I think `emit_vtx` remains to be `const` is preferred since a `emit` method typically implies read only. Is it possible to update `g` after `emit` as part of the infrastructure code (`dfu_impl.cpp`)? Or do you really need another callback?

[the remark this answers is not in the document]
dongahn: I don't know if it is possible,  but if "epoch" and such can also be hidden into the infrastructure code without having to expose it to the callback, that'd be even better. At least that's how I hide such implementation complex for other similar logic like coloring and state cleaning.

src/shell/data-staging.cpp
dongahn: Yeah, that sounds good. `at` can raise an exception (out of range) so please make sure you catch this exception. I don't know if this applies here, but having constness to methods that don't change the object can also be helpful to get compile time checks...

resource/schema/ephemeral.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+const std::map<std::string, std::string>& ephemeral_t::to_map (uint64_t epoch)
+{
+    check_and_clear_if_stale (epoch);
+    return this->store;
+}
+
+bool ephemeral_t::check_and_clear_if_stale (uint64_t epoch)
+{
+    if (this->epoch < epoch) {
```
milroy, line 64: Is it possible for the current epoch passed to `check_and_clear_if_stale` to be less than one second after `this->epoch` is set (i.e. the data is actually stale and should fail this check)?

milroy (commented):
This looks really good overall!  Most of my comments are nits, but I do have a question about a potential corner case for testing for stale data with `epoch`.

t/data/shell/cluster-local/jobspec.yaml
SteVwonder: The test key signals to the data-staging shell plugin that it shouldn't actually stage any files and instead just [log a message with the file source and destination](https://github.com/SteVwonder/flux-sched/blob/job-shell-staging-plugin/src/shell/data-staging.cpp#L170).  The tests in `t700-shell-datastaging.t` then verify that the log contains the correct source and destination.

resource/Makefile.am
```diff
@@ -25,6 +25,7 @@ libresource_la_SOURCES = \
     schema/infra_data.cpp \
```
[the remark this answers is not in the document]
SteVwonder: I used the backticks with the thought that Github would render them as literal strings via its markdown parser (see below for how it renders on GH).  I'm happy to change them to `""` if you think that is the better format outside of GH:

Screen Shot 2020-11-05 at 12 09 05 PM

t/data/shell/cluster-local/jobspec.yaml
SteVwonder: Unfortunately, the job shell still requires that `version` == `1`.  I missed that change in my other PRs to flux-core that enabled the end-to-end support.  I opened an issue (https://github.com/flux-framework/flux-core/issues/3310).  If there are no objections from the flux-core team to the suggestion, I'll allow `version` == `2` and change these to `version: 2`.

resource/writers/match_writers.cpp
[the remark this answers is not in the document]
SteVwonder: I refactored things in [`0820b6c` (#743)](https://github.com/flux-framework/flux-sched/pull/743/commits/0820b6cf001d03005af3357c6671ff4e4b4afcfd).  Most of the commit is undoing what was done in [`3dc3989` (#743)](https://github.com/flux-framework/flux-sched/pull/743/commits/3dc3989df79fae0c0ed022ee553cce874c993152). I dropped the `epoch` argument from the `emit_vtx` callbacks, and just wrapped the one call to `emit_vtx` in `dom_find_dfv` with calls to `ephemeral.check_and_clear_if_stale` and `ephemeral.clear` (plus lots of comments).  I think it actually ends up being nicer/cleaner than what was there before due to what you mention:

> if "epoch" and such can also be hidden into the infrastructure code without having to expose it to the callback, that'd be even better. At least that's how I hide such implementation complex for other similar logic like coloring and state cleaning.

Let me know if you think this refactor is ultimately the right way to go.  I'm happy to redo it a different way.

SteVwonder:
Thanks @dongahn and @milroy for the great comments!  This is ready for another round of reviews.  There is a stack of `fixup!` commits at the top of the commit stack that contain the changes since the last review round.

t/data/shell/cluster-local/jobspec.yaml
SteVwonder: On the team call, we agreed to let this go as is, and bump the version once the issue in flux-core is closed.  I opened an issue in flux-sched to track this: https://github.com/flux-framework/flux-sched/issues/770

resource/writers/match_writers.cpp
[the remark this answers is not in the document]
SteVwonder: @milroy: in your next review, I'd like to call your attention in particular to the refactor described in the above comment.

resource/schema/ephemeral.cpp
milroy: I'm answering my own question based on a conversation with @SteVwonder.  I interpreted `epoch` to be a timestamp (i.e. Unix epoch).  However, `epoch` is a counter, which will not cause the issue in my question.

resource/traversers/dfu_impl.cpp
```diff
@@ -606,7 +606,15 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta, vtx_t u,
...
         if ((resource.type == (*m_graph)[u].type) &&
             (!resource.label.empty())) {
-            (*m_graph)[u].idata.ephemeral.insert (m_best_k_cnt, "label", resource.label);
+            if ((*m_graph)[u].idata.ephemeral.insert (m_best_k_cnt,
+                                                      "label",
+                                                      resource.label) < 0) {
+                m_err_msg += "dom_dfv: inserting label into ephemeral failed.\n";
+                m_err_msg += strerror (errno);
+                m_err_msg += ".\n";
+                rc = -1;
```
milroy, line 615: The pattern throughout `dfu_impl.cpp` is to set the return code in the condition.  Unfortunately, my suggestion extends beyond 80 characters on one line, and I'm not sure the best way to split such a short line. 
```suggestion
            if ( (rc = (*m_graph)[u].idata.ephemeral.insert (m_best_k_cnt,
                                                             "label",
                                                             resource.label)) < 0) {
                m_err_msg += "dom_dfv: inserting label into ephemeral failed.\n";
                m_err_msg += strerror (errno);
                m_err_msg += ".\n";
```

src/shell/data-staging.cpp
```diff
@@ -23,97 +23,116 @@ extern "C" {
...
-    json storage_attrs = jobspec["attributes"]["user"]["data-staging"];
     std::string req_attrs[] = {"label", "granularity", "stage-in"};
     for (const json &storage_attr : storage_attrs) {
         for (const auto &req_attr: req_attrs) {
-            if (!storage_attr.contains(req_attr)) {
-                shell_log_error ("Storage entry missing %s key", req_attr.c_str());
-                return JobspecState::invalid;
+            if (!storage_attr.contains (req_attr)) {
+                shell_log_error ("Storage entry missing %s key",
+                                 req_attr.c_str());
```
milroy, line 49: Nit: space before parenthesis.
```suggestion
                                 req_attr.c_str ());
```

src/shell/data-staging.cpp
```diff
@@ -23,97 +23,116 @@ extern "C" {
...
 
-        std::string granularity = storage_attr["granularity"].get<std::string>();
+        auto granularity = storage_attr.at ("granularity").get_ref<
+            const std::string &>();
         if ((granularity != "node") && (granularity != "job")) {
-            shell_log_error ("Storage entry contains a granularity not 'job' or 'node' (%s)", granularity.c_str());
-            return JobspecState::invalid;
+            shell_log_error ("Storage entry contains a granularity not "
+                             "equal to 'job' or 'node' (%s)",
+                             granularity.c_str());
```
milroy, line 65: Nit: space before parenthesis.
```suggestion
                             granularity.c_str ());
```

src/shell/data-staging.cpp
```diff
@@ -23,97 +23,116 @@ extern "C" {
...
 }
 
-static std::map<std::string, json> find_storage_resources (json R, int rank)
+static std::map<std::string, json> find_storage_resources (json &R, int rank)
 {
     std::map<std::string, json> storage_resources;
     json jgf;
     try {
-        jgf = R["scheduling"]["graph"];
+        jgf = R.at("/scheduling/graph"_json_pointer);
```
milroy, line 115: Nit: space before parenthesis.
```suggestion
        jgf = R.at ("/scheduling/graph"_json_pointer);
```

src/shell/data-staging.cpp
```diff
@@ -197,30 +220,36 @@ static int main_init (flux_plugin_t *p)
...
 
     int rank;
     json jobspec, R;
     try {
-        rank = info["rank"].get<int>();
-        jobspec = info["jobspec"];
-        R = info["R"];
+        rank = info.at ("rank").get<int>();
+        jobspec = info.at ("jobspec");
+        R = info.at("R");
```
milroy, line 236: Nit: space before parenthesis.
```suggestion
        R = info.at ("R");
```

resource/schema/ephemeral.cpp
```diff
@@ -71,6 +71,11 @@ boost::optional<std::string> ephemeral_t::get (uint64_t epoch, const std::string
 const std::map<std::string, std::string>& ephemeral_t::to_map (uint64_t epoch)
 {
     check_and_clear_if_stale (epoch);
+    return this->to_map();
```
milroy, line 74: Nit: spacing
```suggestion
    return this->to_map ();
```

milroy (changes requested):
This looks great @SteVwonder and just needs a few minor style changes.

resource/traversers/dfu_impl.cpp
SteVwonder: I ended up just splitting is across two lines (and dropping the rc = -1):

```
           rc = (*m_graph)[u].idata.ephemeral.insert (m_best_k_cnt,
                                                       "label",
                                                       resource.label);
            if (rc < 0) {
```

dongahn:
FWIW, @SteVwonder and I had a little on-line code review for stale data deallocation scheme (as this was one that he need me to take a look) and all look great to me.

dongahn:
This is the last PR that needs to land before we can tag our next release, correct?

SteVwonder:
Yeah. This PR and then the release notes PR, which is already up.  I'm going to grab some dinner and then I'll be back to wrap this PR up. Should be relatively painless. 

Thanks @milroy for the thorough review!

dongahn:
Cool! I have to say this release cycle was pretty productive! Thank you everyone!

SteVwonder:
Just integrated @milroy's formatting changes and squashed down the `fixup!` commits.

@milroy or @dongahn: if you are happy with the PR, could you approve and toss the MWP label on?  Thanks again for all of the feedback.

dongahn (approved):

SteVwonder:
Darn.  Mergify won't go if there is still a reviewer requesting changes. @milroy, could you change your review to `accept` to appease mergify?  Sorry for bothering you on the weekend.

dongahn:
In this case, I believe it makes sense to manually merge this.

milroy:
Sorry @SteVwonder to forget about the changing my review to `accept`.  I'll remember next time!

PR #744, opened 2020-09-02
t/t1014-annotation: skip expected flux-core fail

chu11:
Skip test expected to fail in flux-core, so that flux-core fix can pass travis.

I didn't know if there was a convention for how to skip a test in this instance, so just added a prerequisite FLUXCORESKIP, which is not defined anywhere.

### DIFF
--- a/t/t1014-annotation.t
+++ b/t/t1014-annotation.t
@@ -75,7 +75,7 @@ test_expect_success 'annotation: cancel all active jobs 1' '
     for job in ${active_jobs}; do flux job wait-event -t 10 ${job} clean; done
 '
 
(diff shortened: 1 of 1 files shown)


grondo:
> I didn't know if there was a convention for how to skip a test in this instance, so just added a prerequisite FLUXCORESKIP, which is not defined anywhere.

Yeah, it probably doesn't matter since we'll re-enable the tests right after your PR is merged.

I was idly wondering if there was a way to more generically handle this case, but I could not think of anything ATM.

grondo (approved):
LGTM

dongahn:
Sounds good to me. Merging.

PR #745, opened 2020-09-03
t/t1014-annotation: Remove expected test fail

chu11:
Remove expected test failure now that flux jobs is fixed in flux-core.

Add a successful test check into the prior test instead.


chu11:
@dongahn didn't know if this is quite how you wanted to run the test, hope its ok

dongahn:
Yes, this works for me. Thanks!

dongahn (approved):

PR #746, opened 2020-09-03
qmanager: remove t_estimate when job alloc success

chu11:
When a job allocation is successful, rather than set t_estimate to
0.0, remove the annotation.

Update tests in t/t1014-annotation.t accordingly.


dongahn (approved):
LGTM! Thanks @chu11.

PR #750, opened 2020-09-14
testsuite:  use monitor-force-up in load_test_resources()

garlick:
This PR changes the `load_test_reosurces()` shell function to load the core resource module wiith the `monitor-force-up` option to ensure that all broker ranks are marked UP in the first response to `resource.acquire`.

The ensures that the change proposed in flux-framework/flux-core#3207 does not break scheduler tests.

Also, fix a couple of tests that were still calling `flux hwloc reload` directly to use `load_test_resources()`.

There is no harm in merging this PR before the core one, as the new option will simply be ignored by the resource module.

### DIFF
--- a/t/sharness.d/sched-sharness.sh
+++ b/t/sharness.d/sched-sharness.sh
@@ -131,7 +131,7 @@ load_test_resources () {
         flux hwloc reload $1 &&
         flux kvs eventlog append resource.eventlog \
             hwloc-discover-finish "{\"loaded\":true}" &&
-        flux module load resource
+        flux module load resource monitor-force-up
 }
 
 # N.B. this assumes that a scheduler is loaded
--- a/t/t4004-match-hwloc.t
+++ b/t/t4004-match-hwloc.t
@@ -115,8 +115,8 @@ test_expect_success 'removing resource works' '
     remove_resource
 '
 
-test_expect_success 'reloading session/hwloc information with test data' '
-    flux hwloc reload ${excl_4N4B}
+test_expect_success 'load test resources (4N4B)' '
+    load_test_resources ${excl_4N4B}
 '
 
 test_expect_success 'loading resource module with default resource info source' '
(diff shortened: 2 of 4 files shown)


dongahn:
@garlick: Thanks. This shouldn't be merged before https://github.com/flux-framework/flux-core/pull/3207/commits/af52a1b8cd5a7dcb7fe97dc39fc8802192f1eb2c gets merged to flux-core, correct?

garlick:
It should be fine to merge now.  The new option will just be ignored until the other PR is merged (and then it is needed).

dongahn:
What is the behavior of the core resource's `monitor-force-up` option when one or more compute nodes are actually down?

garlick:
> What is the behavior of the core resource's monitor-force-up option when one or more compute nodes are actually down?

It lies to the scheduler, so one wouldn't want to use that option unless all ranks are actually up.

When starting a non-system instance, all ranks are required to be up before the initial program starts so it should be fine in that situation (true for the sharness tests that are running within test_under_flux).

dongahn:
> It lies to the scheduler, so one wouldn't want to use that option unless all ranks are actually up.

> When starting a non-system instance, all ranks are required to be up before the initial program starts so it should be fine in that situation (true for the sharness tests that are running within test_under_flux).

Now I understand. Thanks.

For system instance work, we will want to test Fluxion to see if it can handle the resources that are down at "acquire" time. Looking at the code, as far as the initial `resource.acquire` response encodes the up ranks to `up` key,  this should work in theory work. BTW, t shouldn't have `down` key -- https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L687 and your RFC draft seems to confirm this (https://github.com/flux-framework/rfc/pull/253/commits/cbabcc3d4c14328b38bd8a98ac4558bdae222fd3).

I thought I had a test case for which we shouldn't use this option. -- Draining a rank and reload so that only a subset of ranks become available. I just can't find it and a meeting is coming up now. Will circle back to this.

garlick:
Just noticed I missed another test that manually reloads core resource.  I'll fix that and force a push.

garlick:
The aforementioned PR flux-framework/flux-core#3207 was merged.

To recap, what changed is that when the core resource module is reloaded, it may not mark all ranks as up in the initial `resource.acquire` response.  They may come up very shortly after that in an `up` response, once they respond to the rank 0 module publishing an event telling them to say hello again.

By adding a test-only module option to the core resource module when reloading, all broker ranks are marked up in the initial response.  This closes a race that could impact fluxion tests.  In fact on my laptop, `t6002-match-hwloc` is reliably failing now against flux-core master.

With the changes proposed here, it no longer fails.  So I think it'd be good to get this PR merged ASAP.

grondo:
> By adding a test-only module option to the core resource module when reloading, all broker ranks are marked up in the initial response. 

Sorry! I lost track of this PR.

This won't cause a different race in testing if a job is scheduled before ranks are actually available to run jobs?
(Didn't look at the changes yet, so sorry if that is an obvious one)

garlick:
> This won't cause a different race in testing if a job is scheduled before ranks are actually available to run jobs?

The presumption when using this option is that all ranks are known to be available.  That is the case when the resource module is reloaded from a script run under `test_under_flux`, so shouldn't cause any problems here.

grondo (approved):
FWIW, LGTM! Setting my approval in case flux-sched developers are busy.

garlick:
Someone on sched team want to give thumbs up to MWP?

SteVwonder (approved):
LGTM! Thanks @garlick!

garlick:
Thanks!

PR #763, opened 2020-10-15
testsuite: change method of setting test resources

garlick:
Note: This PR should not be merged until flux-framework/flux-core#3265 is merged.

Problem: verification of fake test resources fail with
latest changes to the core resource module.

The core resource module now checks that probed resources
match configured resources.  This check must be disabled
by loading the resource module with the 'noverify' option
in test.

Also, it is no longer necessary to post a 'hwloc-discover-finish'
event in the resource.eventlog when setting up fake resources.

Finally, the fake resource object (still in by_rank format at this time)
needs to be located in the KVS at resource.R.


garlick:
I think we've got a cleaner way to do this so I'll close this PR and resubmit a new one.

PR #774, opened 2020-11-24
codecov: ignore vendored json.hpp

SteVwonder:


dongahn:
Hmmm... It is still 68.04%...

SteVwonder:
It looks like codecov changed the yaml format out from underneath us.  So it has just been reverting to the default configuration and not respecting our ignored paths.  I wonder if that happened recently and is part of the reason why our coverage fell.

Either way, I sync'd up with what is being used over in flux-core.  Side benefit that the codecov bot will now delete any old comments before posting new ones.  So our PR discussions should look much cleaner going forward.

SteVwonder:
That seems to have brought it back up to 77.5%

dongahn:
72.2%?

dongahn:
Actually 73.5%. Much better than before...

SteVwonder:
Oh yeah. Sorry. I was looking at the wrong percentage.

Good to merge?

dongahn:
Yup. Adding MWP.

dongahn (approved):

PR #784, opened 2020-12-05
t0000-sharness.sh: fix test with debug/verbose set in env

SteVwonder:
Problem:  The sharness tests fail if debug=t or verbose=t is set
in the environment since sub-sharness tests cannot tolerate the
extra output triggered by these variales.

Unset 'debug' and 'verbose' in sharness sub-tests to ensure they
do not add extra output in the subtests.

Copied from @grondo's commit in https://github.com/flux-framework/flux-core/pull/3379


grondo (approved):
LGTM, thanks!

dongahn:
Thanks. Setting MWP.

dongahn (approved):

PR #787, opened 2020-12-14
Add support to use RV1+JGF to populate graph data store

dongahn:
This PR adds support to use RV1+JGF to populate graph data store.

- Extend the JGF reader to make use of namespace remapper in order
to remap the rank field of all of the resources as well as the Id fields
of the compute core resources.

- Delay creating the reader object within `sched-fluxion-resource` so
that this service can pick the right resource reader type depending
on the condition (e.g., if JGF is embedded in RV1, use the JGF reader.)

- Fetch the job RV1 from the parent if exists and use it in combination
with the RV1 received from core's resource.acquire to remap the
JGF information.

- Add test cases based on nested Flux instances. 

- Add `flux-ion-R.py` which takes in a RV1 JSON as input, generates JGF
and adds it to the `scheduling` key. This is currently a test-aiding internal
command. (Vendor the jsongraph package as this command depends
on it). 

- Use `flux-ion-R.py` to test the new logic based on a generated
RV1+JGF.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -472,12 +472,35 @@ error:
  *                                                                            *
  ******************************************************************************/
 
+static int create_reader (std::shared_ptr<resource_ctx_t> &ctx,
+                          const std::string &format,
+                          const std::string &allowlist)
+{
+    if ( (ctx->reader = create_resource_reader (format)) == nullptr)
+        return -1;
+    if (allowlist != "") {
+        if (ctx->reader->set_allowlist (ctx->args.load_allowlist) < 0)
+            flux_log (ctx->h, LOG_ERR, "%s: setting allowlist", __FUNCTION__);
+        if (!ctx->reader->is_allowlist_supported ())
+            flux_log (ctx->h, LOG_WARNING, "%s: allowlist unsupported",
+                      __FUNCTION__);
+    }
+    return 0;
+}
+
 static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = -1;
     int saved_errno;
     std::ifstream in_file;
     std::stringstream buffer{};
+
+    if (ctx->reader == nullptr
+        && create_reader (ctx, ctx->args.load_format,
+                          ctx->args.load_allowlist) < 0) {
+        flux_log (ctx->h, LOG_ERR, "%s: can't create reader", __FUNCTION__);
+        goto done;
+    }
 
     saved_errno = errno;
     errno = 0;
@@ -662,7 +685,12 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
         goto done;
     if (flux_rpc_get_unpack (f, "{s:o}", "xml", &xml_array) < 0)
         goto done;
-
+    if (ctx->reader == nullptr
+        && create_reader (ctx, "hwloc", ctx->args.load_allowlist) < 0) {
+        flux_log (ctx->h, LOG_ERR, "%s: can't create hwloc reader",
+                  __FUNCTION__);
+        goto done;
+    }
     if (db.metadata.roots.find ("containment") == db.metadata.roots.end ()) {
         if (rank != IDSET_INVALID_ID) {
             if (!(hwloc_xml = get_array_string (xml_array, rank)))
@@ -954,19 +982,6 @@ static int populate_resource_db (std::shared_ptr<resource_ctx_t> &ctx)
 
     if (ctx->args.reserve_vtx_vec != 0)
         ctx->db->resource_graph.m_vertices.reserve (ctx->args.reserve_vtx_vec);
-    if ( (ctx->reader = create_resource_reader (
-                            ctx->args.load_format)) == nullptr) {
-        flux_log (ctx->h, LOG_ERR, "%s: can't create load reader",
-                  __FUNCTION__);
-        goto done;
-    }
-    if (ctx->args.load_allowlist != "") {
-        if (ctx->reader->set_allowlist (ctx->args.load_allowlist) < 0)
-            flux_log (ctx->h, LOG_ERR, "%s: setting allowlist", __FUNCTION__);
-        if (!ctx->reader->is_allowlist_supported ())
-            flux_log (ctx->h, LOG_WARNING, "%s: allowlist unsupported",
-                      __FUNCTION__);
-    }
     if ( (rc = gettimeofday (&st, NULL)) < 0) {
         flux_log_error (ctx->h, "%s: gettimeofday", __FUNCTION__);
         goto done;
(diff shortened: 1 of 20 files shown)


dongahn:
@SteVwonder and @milroy: since this is almost COB today, this is a good time to cut this PR. This should be ready for your review. (I will work on two extra items as separate PRs: 1) further clean up `flux-ion-R` to become an external command; 2) add module load option for `sched-fluxion-resource` to override JGF writer to hwloc even if the JGF section exists from `resource.acquire`.

resource/readers/resource_namespace_remapper.cpp
```diff
@@ -42,58 +42,50 @@ using namespace Flux::resource_model;
...
-    ::distinct_range_t::distinct_range_t (uint64_t point)
+distinct_range_t::distinct_range_t (uint64_t point)
 {
     m_low = point;
     m_high = point;
 }
 
-resource_namespace_remapper_t
-    ::distinct_range_t::distinct_range_t (uint64_t lo, uint64_t hi)
+distinct_range_t::distinct_range_t (uint64_t lo, uint64_t hi)
```
SteVwonder, line 55: Minor: Since this is being made public, it might be worth validating that `lo < hi` here (and producing an error if not).

Before this was only used in one place (which did the check itself), so it wasn't really an issue.

t/t1018-rv1-bootstrap2.t
```diff
@@ -0,0 +1,191 @@
...
+	nested_jobid=\$(flux mini submit -n\$2 -N\$3 -c\$4 -g\$5 sleep 0)
+	flux job info \${nested_jobid} R > \$6
+	sleep inf
+EOF
+    chmod u+x nest.sh
+'
+
+# Idempotency check
+test_expect_success 'rv1-bootstrap: resource idempotency preserved' '
+    JOBID=$(flux mini batch -n4 -N4 -c44 -g4 \
```
SteVwonder, line 46: I am wondering how is this job satisfiable when only 43 cores are in the R.out file?

SteVwonder (commented):
Thanks @dongahn for putting this together. This generally LGTM.  I just want to give the code a quick spin tomorrow to make sure I understand the tests

resource/readers/resource_namespace_remapper.cpp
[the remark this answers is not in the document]
dongahn: Thank you for the late night! 

Yes this is a good point. It was trying to avoid throwing an exception from a constructor. It should be okay to throw `std::invalid_argument` standard exception and make sure the client code handle this correctly when this is used as keys to std containers like `std::map`. I will get to this tomorrow.

t/t1017-rv1-bootstrap.t
```diff
@@ -0,0 +1,213 @@
...
+    test_cmp job.norm.json nest.norm.json
+'
+
+# Cancel jobs
+test_expect_success 'rv1-bootstrap: killing a nested job works' '
+    flux job cancel ${JOBID} &&
+    flux job wait-event ${JOBID} release
+'
+
+# 2 full nodes are scheduled for a nest flux instnace -- requiring no remap
```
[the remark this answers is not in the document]
dongahn: Oops.

t/t1018-rv1-bootstrap2.t
dongahn: Hmmm. It should be 44 core with --cores=0-43 option to flux R encode.

dongahn:
> Thanks @dongahn for putting this together. This generally LGTM. I just want to give the code a quick spin tomorrow to make sure I understand the tests

Yeah, a quick spin in action would be good! This is relative foundational change. I plan to run some more manual testing myself tomorrow as well by feeding different `R` using `flux R` combined with `flux-ion-R`. 

I did run valgrind on the nested instance and didn't find any obvious memory problem though.

SteVwonder (approved):
Thanks @dongahn! I gave it a spin, and it all looks good to me!

dongahn:
I will need to circle back to this later tonight.

 I will address all the issues by then and put the MWP. Then do the same for PR #791 and tag a release. The rest of the release workflow should be handled automatically!

resource/readers/resource_namespace_remapper.cpp
```diff
@@ -90,9 +90,8 @@ bool distinct_range_t::operator!= (const distinct_range_t &o) const {
     return m_high != o.m_high || m_low != o.m_low;
```
milroy, line 90: Super nits: Thill -> This, simply -> simplify in the commit message.

resource/modules/resource_match.cpp
```diff
@@ -649,29 +726,83 @@ static int unpack_resobj (json_t *resobj, std::vector<resobj_t> &out)
...
+    // needs to remap them (e.g., GPU uses Id in global scope).
+    // This loop iterates through distict rank ranges in ascending order.
+    // We remap the ascending rank range sequence to contiguous rank range
+    // sequence starting from 0 (cur_rank): e.g., remapping <"3-4", "5", "20-21">
+    // to <"0-1", "2", "3-4">.
+    // Per core resource.acquire's remapping rule, for each range found
+    // in the parent namespace's R_lite, you should be able to find a
+    // corresponding range in the current namespace's R_lite. IOW,
+    // the current namespace's R_lite must contain a range which is
+    // a superset of the range in the parent namespace.
```
milroy, line 769: I'm not sure I understand this.  Do you mean that <"3-4", "5", "20-21"> -> <"0-1", "2", "3-4"> which becomes <"0-4">, and the 0-4 interval is longer than the sub-intervals in the parent's namespace?

milroy (approved):
@dongahn this looks great to go in as is!  I found a couple of typos and had one question, but that's it.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -38,12 +38,36 @@ extern "C" {
 using namespace Flux;
```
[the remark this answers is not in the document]
dongahn: Ah. Thanks.

resource/modules/resource_match.cpp
dongahn: No. <"3-4", "5", "20-21"> is first translated to the following sequence <"0-1", "2", "3-4">. And then, we find the corresponding sequence element from the current R_lite for each of these translated sequence element. In this particular example, it is likely the current R_lite has exactly <"0-1", "2", "3-4">. So one to one mapping. 

But generally speaking, the current R_lite can have longer ranges. Like

A: <"3", "5", "6", "20-21"> --> B: <"1", "2", "3", "4-5"> and the current R_lite might have C: <"1-2", "3", "4-5">. So "1" and "2" in B should resolve to to C's "1-2", for example. 

Hope this makes sense. I will try to clarify this.

PR #788, opened 2020-12-16
ci: drop unnecessary flux keygen

garlick:
Problem: flux-core no longer requires user CURVE keys to run Flux,
but flux-sched still generates them in ci scripts.

Drop flux keygen from Dockerfiles and travis_run.sh.


SteVwonder (approved):
LGTM! Thanks @garlick!

garlick:
Thanks!  I'll  go ahead and set MWP.

PR #794, opened 2020-12-21
ci: migrate to github actions

grondo:
This PR migrates flux-sched from Travis-CI to GitHub Actions/Workflows for all CI tests.

Most of the work was already done in flux-core. I generalized `docker-run-checks.sh` and `checks_run.sh` in flux-core first (PR pending), so that those files can be copied into flux-sched (and hopefully other flux-framework projects) with minimal changes (just update the PROJECT name at the top).

I tried to keep the checks run the same as were in Travis-CI.

One thing I didn't do here was add the python format and pylint checks. This is just a straight conversion at this time.


SteVwonder (approved):
LGTM! Should we wait for the generalization in flux-core to go in first?  I'm assuming this is OK to go in whenever?

grondo:
As long as they both go in eventually then future synchronization will be easier (order of PRs merged doesn't matter)

The coverage dropped here. I wonder if it is because python coverage is now supported by virtue of updating to `checks_run.sh` from flux-core...

grondo:
Looks like the drop is due to [`hlapi/bindings`](https://codecov.io/gh/flux-framework/flux-sched/pull/794/tree/resource/hlapi/bindings) (perhaps?) Was this source ignored before?

grondo:
Feel free to set MWP when you are ready, if you want to leave the coverage investigation for later.

grondo:
I fixed a typo in `src/test/checks_run.sh`. This will need to manually merged.

grondo:
I think the coverage change here can be explained by the fact that the old coverage method seemed to ignore sources that had 0 coverage, i.e. those that aren't run at all, while (somehow) the coverage in this PR finds the following "new" files:
```
 New file resource/hlapi/bindings/c/reapi_module.cpp
 New file resource/hlapi/bindings/c/reapi_cli.cpp
 New file resource/utilities/grug2dot.cpp
 New file src/cmd/flux-tree-helper.py
```

`flux-tree-helper.py` can be explained because updating `docker-run-checks.sh` from flux-core adds support for Python coverage. I'm not sure why the other files (which seem to be unused) are ignored on the main branch but found by codecov here.

SteVwonder:
> while (somehow) the coverage in this PR finds the following "new" files:

Thanks for digging into that @grondo!

> This will need to manually merged.

Pushing the button.

grondo:
Thanks!

Issue #799, opened 2021-01-08
t1018-rv1-bootstrap2.t: failure in flux-shed CI test over in flux-core


SteVwonder:
For the flux-core PR (https://github.com/flux-framework/flux-core/pull/3458), the flux-sched CI test failed with:

```
2021-01-08T05:11:00.9564100Z FAIL: t1018-rv1-bootstrap2.t 5 - rv1-bootstrap: resource idempotency preserved
```

I'm assuming this is just an intermittent failure, but I wanted to log it in case it starts to pop up more.  The full build log: https://gist.github.com/SteVwonder/05920bc917aac17deaee1dc63bd18605

dongahn:
Probably not the cause of this error but I see:

> 2021-01-08T05:06:05.8826545Z ./src/test/checks_run.sh: line 180: test: too many arguments

There is a minor bug in `checks_run.sh`:

```diff
runner@fv-az128-15:~/work/flux-sched/flux-sched/src/test$ diff checks_run.sh checks_run.sh.new 
180c180
< if test "$POISON" = "t" -a test "$PROJECT" = "flux-core"; then
---
> if test "$POISON" = "t" -a "$PROJECT" = "flux-core"; then
```

dongahn:
I believe this will be solved once PR #800 is merged.

chu11:
appeared to hit this in the centos 7 builder in https://github.com/flux-framework/flux-sched/pull/803

```
   rv1-bootstrap: creating a nested batch script
  
  expecting success: 
      JOBID=$(flux mini batch -n4 -N4 -c44 -g4 \
  	./nest.sh high 4 4 44 4 nest.json) &&
      $WAITFILE -t 600 -v -p \"R_lite\" nest.json &&
      jq " del(.execution.starttime, .execution.expiration) " \
  	nest.json > nest.norm.json &&
      flux job info ${JOBID} R | \
  	jq " del(.execution.starttime, .execution.expiration) " \
  	> job.norm.json &&
      test_cmp job.norm.json nest.norm.json
  
  waitfile: nest.json: 0.000s: No file yet
  waitfile: nest.json: 1.751s: wakeup
  waitfile: nest.json: 1.751s: File appeared with size 0
  waitfile: nest.json: 1.751s: back to sleep
  waitfile: nest.json: 600.001s: Timeout after 600s
  -rw-r--r-- 1 runner runner 0 Jan 20 18:12 nest.json
[...]

PR #800, opened 2021-01-09
fluxion-R: add flux-ion-R.py as a public command 

### LINKED ISSUE #792: Make flux-ion-R.py a public command
https://github.com/flux-framework/flux-sched/pull/787#issuecomment-747917754

dongahn:
This PR exposes `flux-ion-R.py` as a public command.

The main use case of `flux-ion-R` is to work with `flux-R` through Unix pipes to generate a new RV1 with the actual scheduling key populated. For example,

```console
% flux R encode --hosts=sierra[3682,3179,3683,3178]  --cores=0-43 --gpu=0-3 | flux ion-R encode
```

will take the RV1 produced from `flux-R encode` and generate a new RV 1 JSON document with the corresponding JGF graph embedded under the scheduling key.

Closes #792


grondo:
I wonder if we also want to bring in the `pylint` and `black` formatter ci checks with this PR, since it introduces and installs a new Python script.

dongahn:
Yeah sounds good. I will work on it shortly

dongahn:
> Yeah sounds good. I will work on it shortly

`pylint` and `black` support went into the GitHub action. 

I think adding a few `sharness` tests for `fluxion-ion-R` (e.g., passing unaccessible path etc) could be done. But otherwise, this PR is ready for your review @grondo and @SteVwonder.

grondo (approved):
Looks good to me after a quick pass through. (just one minor question inline)

I'll try to actually run some tests later on this evening.

src/cmd/flux-ion-R.py
```diff
@@ -0,0 +1,287 @@
+#!/bin/false
```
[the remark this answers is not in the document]
dongahn: Good point. I will add that.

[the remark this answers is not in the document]
dongahn: Added.

dongahn:
@mergefyio rebase

src/cmd/flux-ion-R.py
```diff
@@ -0,0 +1,287 @@
...
+            rank,
+            True,
+            "",
+            1,
+            hPath,
+        )
+        edg = FluxionResourceRelationshipV1(ppid, vtx.get_id())
+        self.__add_and_tick_uniq_id(vtx, edg)
+        for key, val in children.items():
+            for i in list(IDset(val)):
```
SteVwonder, line 175: nit: `IDset` implements `__iter__`, so you can iterate over it directly without the conversion to a list first:

```
>>> import flux
>>> from flux.idset import IDset
>>> for x in IDset("0-3"):
...     print(x)
...
0
1
2
3
```

src/cmd/flux-ion-R.py
```diff
@@ -0,0 +1,287 @@
...
+#
+# 'encode' subcommand routines
+#
+def encode_init(args, infile, outfile):
+    if args.ifn:
+        try:
+            infile = open(args.ifn, "r")
+        except OSError:
+            LOGGER.error("Could not open %s", args.ifn)
+            sys.exit(1)
```
SteVwonder, line 214: I think it might be best to just re-raise the original exception and let the `CLIMain` catch it (and print a "pretty" error message without a backtrace by default).  That way, if the user sets `FLUX_PYCLI_LOGLEVEL=10` when running this script, they will get the full backtrace.  Currently, this code will gobble the exception and throw away the backtrace.

Same for the exceptions with `sys.exit(1)` below.

src/cmd/flux-ion-R.py
```diff
@@ -0,0 +1,287 @@
...
+        help="redirect stdin from FILENAME",
+        metavar="FILENAME",
+    )
+
+    args = parser.parse_args()
+    args.func(args)
+
+    if not args.subcommand:
+        parser.error("subcommand is required")
+        sys.exit(1)
```
SteVwonder, line 290: Nit: it doesn't hurt to have this here, but I think it is redundant due to `subparsers.required = True`.

SteVwonder (commented):
LGTM! Thanks @dongahn! Two minor nits and one comment about exception handling in the flux-ion-R command:

src/cmd/flux-ion-R.py
dongahn: Didn't know. Awesome!

src/cmd/flux-ion-R.py
dongahn: I'm not familiar enough with what CLIMain does. So just re-raising a new OSError w/ a new message will do?

src/cmd/flux-ion-R.py
dongahn: ah. Good point. Will fix

src/cmd/flux-ion-R.py
SteVwonder: You can just do:

```
try:
    # code goes here
except:
    LOGGER.error("Error message")
    raise
```

The standalone `raise` automatically re-raises the caught exception. Then `CLIMain` will catch the exception, print the exception message, print the backtrace (if the LOGLEVEL is high enough), then exit with exit code 1: https://github.com/flux-framework/flux-core/blob/master/src/bindings/python/flux/util.py#L167

Depending on how useful the default error message within the OSError is, you might even be able to get away with not catching the `OSError` at all or logging the filename (that might already be in the OSError message)

dongahn: If that contains the file name, that's golden. Otherwise, will augment

dongahn:
@SteVwonder: does fixup commits look good to you? 

By the way, I saw one runner hung up on a test for 600 seconds. I reran the job and it is now okay. There may still be a lingering issue -- https://github.com/flux-framework/flux-sched/issues/799, though. If this keep impacting our PR, I may redesign that sharness tests.

dongahn:
Wait.... I didn't do:

```
>>> for x in IDset("0-3"):
```

dongahn:
@SteVwonder: FYI -- all fixup commits have been added. Let me know if there is anything else you want me to do.

SteVwonder (approved):
LGTM! Thanks @dongahn!

dongahn:
Thanks. Squashed the fixup commits and pushed a force. Adding MWP.

SteVwonder:
Looks like mergify won't merge this automatically because it is looking for a test called `bionic - clang6.0` but the test is actually called `bionic - clang-6.0` (note this extra dash).  @dongahn: I'm happy to create a separate, simple PR if you want, or would you prefer to add a new commit to this PR?  Either way, changing the mergify config will require manual merging.

grondo:
> because it is looking for a test called bionic - clang6.0 but the test is actually called bionic - clang-6.0

Ah, sorry about that typo!

dongahn:
> @dongahn: I'm happy to create a separate, simple PR if you want, or would you prefer to add a new commit to this PR? Either way, changing the mergify config will require manual merging.

Works for me. Thanks.

SteVwonder:
Thanks @grondo for helping get #804 in.  This should be good now.  I'll give mergify a quick kick to see if it'll merge now.

SteVwonder:
@mergifyio rebase

dongahn:
Screen Shot 2021-01-14 at 3 00 15 PM

This needs to be manually merged because it changes the GitHub action workflow spec. Anyone bother to click "merge"?

Issue #798, opened 2021-01-06
qmanager: urgency -> priority


chu11:
While working on a `flux-sched` fixup for https://github.com/flux-framework/flux-core/issues/3450, I noticed that in some locations the variable `urgency` is used when it should be `priority`.

for example

```
    if (schedutil_alloc_request_decode (msg, &job->id, &job->urgency,
                                        &job->userid, &job->t_submit) < 0) {
        flux_log_error (h, "%s: schedutil_alloc_request_decode", __FUNCTION__);
        return;
    }
```

The function `schedutil_alloc_request_decode` returns a job priority, not a job urgency.

I was going to make a change through, but the maximum urgency appears to be hard coded in for "priority" calculations.

```
    // Note that RFC27 defines 31 as the max urgency. Because our queue policy
    // layer sorts the pending jobs in lexicographical order
    // (<urgency, t_submit, ...> and lower the better, we adjust the urgency.
    job->urgency = 31 - job->urgency;
```

so I was a little reluctant to make this change at the moment.

perhaps this will all go away once a priority plugin is developed, but thought I should atleast document this.

dongahn:
The term changed from priority to urgency actually recently ( https://github.com/flux-framework/flux-sched/pull/781). If the range of the priority changed and documented in a RFC, it would be okay to revert the name to priority and increase the hard coded value.

BTW, the main reason to do this smallest to largest convergence is so that we could just use default ordering of `std::map`. We could also add a custom comparator to reverse or customize the ordering logic.

chu11:
ok, just didn't know if the hard coded 31 had meaning or not.

dongahn:
Yeah it is the MAX priority previously defined in an RFC.

PR #801, opened 2021-01-13
qmanager: update per libschedutil API changes

chu11:
Changes per `libschedutil` changes in https://github.com/flux-framework/flux-core/pull/3447

We should wait for changes upstream to be accepted before merging this of course.

These commits could be collapsed into fewer ones if desired.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -231,10 +231,7 @@ static int handshake_jobmanager (std::shared_ptr<qmanager_ctx_t> &ctx)
     int rc = -1;
     int queue_depth = 0;  /* Not implemented in job-manager */
 
-    if (schedutil_hello (ctx->schedutil,
-                         &qmanager_safe_cb_t::jobmanager_hello_cb,
-                         std::static_pointer_cast<
-			     qmanager_cb_ctx_t> (ctx).get ()) < 0) {
+    if (schedutil_hello (ctx->schedutil) < 0) {
         flux_log_error (ctx->h, "%s: schedutil_hello", __FUNCTION__);
         goto out;
     }
@@ -408,6 +405,13 @@ static int handshake (std::shared_ptr<qmanager_ctx_t> &ctx)
     return rc;
 }
 
+const struct schedutil_ops ops = {
+    .hello  = &qmanager_safe_cb_t::jobmanager_hello_cb,
+    .alloc  = &qmanager_safe_cb_t::jobmanager_alloc_cb,
+    .free   = &qmanager_safe_cb_t::jobmanager_free_cb,
+    .cancel = &qmanager_safe_cb_t::jobmanager_cancel_cb,
+};
+
 static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
 {
     std::shared_ptr<qmanager_ctx_t> ctx = nullptr;
@@ -416,9 +420,7 @@ static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
         ctx->h = h;
         set_default (ctx);
         if (!(ctx->schedutil = schedutil_create (ctx->h,
-                                   &qmanager_safe_cb_t::jobmanager_alloc_cb,
-                                   &qmanager_safe_cb_t::jobmanager_free_cb,
-                                   &qmanager_safe_cb_t::jobmanager_cancel_cb,
+                                   &ops,
                                    std::static_pointer_cast<
                                        qmanager_cb_ctx_t> (ctx).get ()))) {
             flux_log_error (ctx->h, "%s: schedutil_create", __FUNCTION__);
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -410,6 +410,7 @@ const struct schedutil_ops ops = {
     .alloc  = &qmanager_safe_cb_t::jobmanager_alloc_cb,
     .free   = &qmanager_safe_cb_t::jobmanager_free_cb,
     .cancel = &qmanager_safe_cb_t::jobmanager_cancel_cb,
+    .prioritize = NULL,
 };
 
 static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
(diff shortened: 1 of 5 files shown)


qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -99,7 +99,7 @@ class job_t {
     job_state_kind_t state = job_state_kind_t::INIT;
     flux_jobid_t id = 0;
     uint32_t userid = 0;
-    int urgency = 0;
+    unsigned int urgency = 0;
```
dongahn, line 102: Is what passed to this layer through schedutil "priority" or "urgency"? If "priority", we should change the field name to priority?

qmanager/modules/qmanager.cpp
```diff
@@ -410,6 +410,7 @@ const struct schedutil_ops ops = {
     .alloc  = &qmanager_safe_cb_t::jobmanager_alloc_cb,
     .free   = &qmanager_safe_cb_t::jobmanager_free_cb,
     .cancel = &qmanager_safe_cb_t::jobmanager_cancel_cb,
+    .prioritize = NULL,
```
dongahn, line 413: This is more of a question.

So this callback is called when there is a priority change of a pending job? Will this be currently called back (if installed) if a user change the urgency at all?

qmanager/policies/base/queue_policy_base.hpp
chu11: yes it's priority.  The urgency -> priority change is what i described in issue #798.  I have a branch with it done that goes on top of this.  If you'd like, I could just add it on top of this PR?  Or perhaps it should go in before this PR.

qmanager/modules/qmanager.cpp
chu11: > So this callback is called when there is a priority change of a pending job?

yes

> Will this be currently called back (if installed) if a user change the urgency at all?

By "currently", I'm not sure if you mean the current `flux-core` master?  If so yes, b/c currently `priority` == `urgency`, so an urgency change would lead to this being called.   Longer out, when the priority plugin is created, it'll be called if the priority plugin calculates a different priority.

chu11: per our offline discussion, trying to implement `prioritize_cb`, but it's not obvious to me how to do this given the current `queue` implementation.  I feel I should be doing something like this:

```
job = queue->lookup (id);
queue->take_off_queue (job);
job->priority = 4294967295 - priority;
queue->insert (job);
```

but I can't find how to do `take_off_queue`.  `remove` seems to alter some internal state of the `queue` so don't want to do that.  `pending_pop()` only seems to pop the front of the queue.  Not sure if I'm missing something.

chu11: Ahhh I figured out the trick.  `remove()` does in fact do more than just `remove` from the queue (which initially made me think it wasn't safe to use), but as long as we `insert` it back, everything will be reset back to the way it should be.

chu11:
rebased and re=pushed per discussion above

- Added fix for #798 
- Added a `prioritize_cb` and tests
-

qmanager/modules/qmanager.cpp
dongahn: chu11@: sched's queue uses `std::map` so `insert()` and other command operations are `O(N)`

dongahn: The current abstraction for `queue` base is that it manages jobs through multiple queues, pending queues, running queues, complete queues as each job is going through its lifecycle. So use of `remove()` might be a bit too error prone. Perhaps we should add a new method specifically targeting modifying an existing job if certain conditions are met...

dongahn: maybe `pending_modify()` or `pending_repriotize()` somewhere around: [here](https://github.com/flux-framework/flux-sched/blob/master/qmanager/policies/base/queue_policy_base.hpp#L121)

chu11: sounds good, i'll add a new member function of that nature.

chu11: since we're now adding a `pending_reprioritize()` to the queue data structure, it's probably best to spin the initial prioritize_cb into a new PR, as the work in clearly increasing beyond this PR.

chu11:
> Added a prioritize_cb and tests (I could squash this into an earlier commit that simply set .prioritize to NULL)

Since we're now adding new member functions into the queue policy implementation, the initial `prioritize_cb` implementation should be split off into a new PR.  Commits related to that prior work have been removed.

qmanager/modules/qmanager.cpp
dongahn: Sounds good as agreed at our meeting!

dongahn:
@Mergifyio rebase

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -174,25 +174,36 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h,
...
+                         "id", &job->id,
+                         "priority", &job->urgency,
+                         "userid", &job->userid,
+                         "t_submit", &job->t_submit,
+                         "jobspec", &jobspec) < 0) {
+        flux_log_error (h, "%s: flux_msg_unpack", __FUNCTION__);
+        return;
+    }
+    if (!(jobspec_str = json_dumps (jobspec, JSON_COMPACT))) {
+        errno = EPROTO;
```
dongahn, line 212:
```suggestion
        errno = ENOMEM;
```

dongahn (changes requested):
This looks great to me. Just one minor nit. The CI is failing because the corresponding flux-core hasn't landed, correct?

chu11:
> The CI is failing because the corresponding flux-core hasn't landed, correct?

Yeah, compilation should fail at this point.

chu11:
rebased & re-pushed per your comments above @dongahn.

If you think this looks good, then we can merge https://github.com/flux-framework/flux-core/pull/3447 and then merge this, as @garlick has approved the flux-core PR.

dongahn:
@chu11: sounds good to me. Go ahead and merge flux-core PR and I will merge this in.

dongahn:
@mergefyio rebase

dongahn:
@Mergifyio rebase

chu11:
@mergefyio rebase

chu11:
@Mergifyio rebase

dongahn (approved):

PR #803, opened 2021-01-14
qmanager: support re-prioritization of jobs

chu11:
Support re-prioritization of pending jobs by adding a prioritize callback through `libschedutil`.

Built on top of PR #801

### DIFF
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -40,6 +40,14 @@ using namespace Flux::queue_manager::detail;
 using namespace Flux::opts_manager;
 using namespace Flux::cplusplus_wrappers;
 
+static unsigned int calc_priority (unsigned int priority)
+{
+    // RFC27 defines 4294967295 as the max priority. Because
+    // our queue policy layer sorts the pending jobs in
+    // lexicographical order (<priority, t_submit, ...>) and lower the
+    // better, we need to calculate an adjusted priority.
+    return (4294967295 - priority);
+}
 
 int qmanager_cb_ctx_t::find_queue (flux_jobid_t id,
                                    std::string &queue_name,
@@ -166,12 +174,8 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg,
     queue_name = qn_attr? qn_attr : ctx->opts.get_opt ().get_default_queue ();
     json_decref (o);
     queue = ctx->queues.at (queue_name);
-    // Note that RFC27 defines 4294967295 as the max priority. Because
-    // our queue policy layer sorts the pending jobs in
-    // lexicographical order (<priority, t_submit, ...>) and lower the
-    // better, we adjust priority.
     running_job = std::make_shared<job_t> (job_state_kind_t::RUNNING,
-                                                   id, uid, 4294967295 - prio,
+                                                   id, uid, calc_priority (prio),
                                                    ts, R);
 
     if ( (rc = queue->reconstruct (static_cast<void *> (h),
@@ -220,11 +224,7 @@ void qmanager_cb_t::jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
     job->id = id;
     job->userid = userid;
     job->t_submit = t_submit;
-    // Note that RFC27 defines 4294967295 as the max priority. Because
-    // our queue policy layer sorts the pending jobs in
-    // lexicographical order (<priority, t_submit, ...> and lower the
-    // better, we adjust the priority.
-    job->priority = 4294967295 - priority;
+    job->priority = calc_priority (priority);
     jobspec_obj = Flux::Jobspec::Jobspec (jobspec_str);
     if (jobspec_obj.attributes.system.queue != "")
         queue_name = jobspec_obj.attributes.system.queue;
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -195,15 +195,19 @@ void qmanager_cb_t::jobmanager_alloc_cb (flux_t *h, const flux_msg_t *msg,
     Flux::Jobspec::Jobspec jobspec_obj;
     std::string queue_name = ctx->opts.get_opt ().get_default_queue ();
     std::shared_ptr<job_t> job = std::make_shared<job_t> ();
+    flux_jobid_t id;
+    unsigned int priority;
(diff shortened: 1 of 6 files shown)


qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -325,6 +325,56 @@ void qmanager_cb_t::jobmanager_cancel_cb (flux_t *h, const flux_msg_t *msg,
...
+                            __FUNCTION__, static_cast<intmax_t> (id));
+            continue;
+        }
+
+        // Note that RFC27 defines 4294967295 as the max priority. Because
+        // our queue policy layer sorts the pending jobs in
+        // lexicographical order (<priority, t_submit, ...> and lower the
+        // better, we adjust the priority.
+
+        if (queue->pending_reprioritize (id, 4294967295 - priority) < 0) {
```
dongahn, line 365: `4294967295 - priority` patterns seem to appear over and over again. Perhaps this should be refactored  somewhere so we only need to change one line if the max priority change once again for some reason. Any idea for refactoring? This shouldn't keep this PR from being merged though. If we don't want to address this now, we should open an issue ticket so that this can be refactored later.

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -325,6 +325,56 @@ void qmanager_cb_t::jobmanager_cancel_cb (flux_t *h, const flux_msg_t *msg,
...
+            if (errno == ENOENT || errno == EINVAL)
+                continue;
+            flux_log_error (h, "%s: queue pending_reprioritize (id=%jd)",
+                            __FUNCTION__, static_cast<intmax_t> (id));
+            return;
+        }
+
+        flux_log (h, LOG_DEBUG, "%s (id=%jd)", "priority changed",
```
dongahn, line 373: This might generate too many debug output when we add a fairshare priority plug in since it print one line per updated job?

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -325,6 +325,56 @@ void qmanager_cb_t::jobmanager_cancel_cb (flux_t *h, const flux_msg_t *msg,
...
+        }
+
+        // Note that RFC27 defines 4294967295 as the max priority. Because
+        // our queue policy layer sorts the pending jobs in
+        // lexicographical order (<priority, t_submit, ...> and lower the
+        // better, we adjust the priority.
+
+        if (queue->pending_reprioritize (id, 4294967295 - priority) < 0) {
+            if (errno == ENOENT || errno == EINVAL)
+                continue;
```
dongahn, line 367: It would be good to handle either errno with distinct error message?

`errno=ENOENT` would typically mean that there was a race condition: when a priority change was requested right before fluxion responds to alloc request. In this case, users might be interested in know why, we may want to log a message. 

`errno=EINVAL` means the upper layer requested a invalid priority and this suggests a bug somewhere. We will want to log an error message.

dongahn (changes requested):
Generally LGTM. I left a few minor change suggestions.

chu11:
re-pushed with fixes per comments above.  I added a new `calc_priority()` function in `qmanager_callbacks.cpp` to do the common priority calculation.  I considered if it should be in the `queue_policy` data structure, but since `job->priority` is public, there's probably no point to trying to abstract the setting of the priority internally to `queue_policy`.

dongahn:
> re-pushed with fixes per comments above. I added a new calc_priority() function in qmanager_callbacks.cpp to do the common priority calculation. I considered if it should be in the queue_policy data structure, but since job->priority is public, there's probably no point to trying to abstract the setting of the priority internally to queue_policy.

Yes, that is exactly what I was thinking about. The architecture is such that you put flux specific things to the module code and generic code in elsewhere. The queue layer is designed so that it can be layered with 3rd party (e.g., Kubernetes)  if they want.

chu11:
oops, noticed some English errors on some comments, re-pushed

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -410,6 +416,40 @@ std::map<uint64_t, flux_jobid_t>::iterator queue_policy_base_impl_t::
...
+    m_pending.erase ({static_cast<double> (job->priority),
+                      static_cast<double> (job->t_submit),
+                      static_cast<double> (job->t_stamps.pending_ts)});
+
+    job->priority = priority;
+
+    m_pending.insert (std::pair<std::vector<double>, flux_jobid_t> (
+        {static_cast<double> (job->priority),
+         static_cast<double> (job->t_submit),
+         static_cast<double> (job->t_stamps.pending_ts)}, job->id));
```
dongahn, line 446:
```suggestion
    auto res = m_pending.insert (std::pair<std::vector<double>, flux_jobid_t> (
                   {static_cast<double> (job->priority),
                    static_cast<double> (job->t_submit),
                    static_cast<double> (job->t_stamps.pending_ts)}, job->id));
    if (!res.second) {
        errno = EEXIST;
        goto out;
    }
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -286,6 +287,16 @@ class queue_policy_base_t : public detail::queue_policy_base_impl_t
      */
     int reconstruct (void *h, std::shared_ptr<job_t> job, std::string &R_out);
 
+    /*! Reprioritize a job with a new priority.
+     *
+     *  \param id        jobid of flux_jobid_t type.
+     *  \param priority  new job priority
+     *  \return          0 on success; -1 on error.
+     *                       ENOENT: unknown id.
+     *                       EINVAL: job not pending
```
dongahn, line 296: It will be a very unusual error but you would need to document:
`errno=EEXIST` if you follow one of my error handling suggestion for `map::insert`.

dongahn (approved):
@chu11: Is this ready to be merged? I've added a few minor comments. But since those are minor, I approve this PR. When you are ready, feel free to add the MWP label.

chu11:
yup, ready for merging.  Added your tweaks.  Using `auto res` messed up the use of `goto out` (the `crosses initialization` error), but since we don't have to do any cleanups, just replacing w/ `return -1` was an easy fix.

chu11:
appeared to hit #800, didn't want to restart builder if someone wanted to take look at it?

dongahn:
It is on my list. Feel free to restart.

chu11:
setting MWP per @dongahn conversation

dongahn:
Thanks @chu11!

PR #809, opened 2021-01-29
Add minor C++ safety improvements

dongahn:
This PR adds minor safety enhancements that can work around some issues caused by potentially buggy C++ compilers and/or other unexpected faults. 

- Add pre-check on the input strings that are passed to `std::stoi`
- Add catch-all exception handling for the `sched-fluxion-resource` module scope.

### DIFF
--- a/qmanager/policies/base/queue_policy_base_impl.hpp
+++ b/qmanager/policies/base/queue_policy_base_impl.hpp
@@ -93,6 +93,15 @@ done:
     return rc;
 }
 
+bool queue_policy_base_t::is_number (const std::string &num_str)
+{
+    if (num_str.empty ())
+        return false;
+    auto i = std::find_if (num_str.begin (), num_str.end (),
+                           [] (unsigned char c) { return !std::isdigit (c); });
+    return i == num_str.end ();
+}
+
 int queue_policy_base_t::set_queue_params (const std::string &params)
 {
     return set_params (params, m_qparams);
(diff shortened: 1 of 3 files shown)


SteVwonder (approved):
LGTM!

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -93,6 +93,15 @@ int queue_policy_base_t::set_params (const std::string &params,
     return rc;
 }
 
+bool queue_policy_base_t::is_number (const std::string &num_str)
```
SteVwonder, line 96: Might be worth adding a comment reference to the github issue that documents why this workaround is necessary.

dongahn: will do.

grondo (approved):
I'm not really capable of reviewing the code, but applied the patches in this PR to the TOSS4 RPM build in the buildfarm and verified that the qmanager no longer generates a corefile (we only have the corefile from the resource module)

dongahn:
> we only have the corefile from the resource module

Since a core file is generated even with `catch (...)`, clearly there is something wrong with how compiler generated code. Once I have TOSS4 Power LE access, I will try to get to the bottom of that. Thanks.

dongahn:
Hmmm @grondo: do you have any idea why ci validator fails? The fixup commit was squashed before this PR is merged...

Screen Shot 2021-01-29 at 10 17 01 PM

grondo:
I think you are just viewing the CI run before the fixup commit was squashed. Mergify would not merge the PR if the commit validator failed. See [here](https://github.com/flux-framework/flux-sched/actions)

PR #810, opened 2021-01-30
testsuite: fix shell path discovery in datastaging test

grondo:
Use a more foolproof method of discovering the default flux-shell path from the enclosing flux instance in `t7000-datastaging.t`. Without this, `make check` for flux-sched built against a side-installed flux-core may fail since the test could use the flux-shell from the system instance unless PKG_CONFIG_PATH happens to be set appropriately.

### DIFF
--- a/t/t7000-shell-datastaging.t
+++ b/t/t7000-shell-datastaging.t
@@ -7,9 +7,12 @@ test_description='Test data-staging job shell plugin'
 shell_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/shell`
 shell_plugin_path="$(readlink -e ${SHARNESS_BUILD_DIRECTORY}/src/shell/.libs)"
 jobspec_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/resource/jobspecs/`
-flux_shell="$(pkg-config --variable=prefix flux-core)/libexec/flux/flux-shell"
+
 
 test_under_flux 2
+
+flux_shell=$(flux builtin-conf shell_path)
+test_debug "echo flux_shell=$flux_shell"
 
 test_expect_success 'node-local storage allocation staging rank 0' '
 	cat <<-EOT >test-initrc.lua &&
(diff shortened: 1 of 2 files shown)


t/t7000-shell-datastaging.t
```diff
@@ -7,10 +7,13 @@ test_description='Test data-staging job shell plugin'
 shell_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/shell`
 shell_plugin_path="$(readlink -e ${SHARNESS_BUILD_DIRECTORY}/src/shell/.libs)"
 jobspec_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/resource/jobspecs/`
-flux_shell="$(pkg-config --variable=prefix flux-core)/libexec/flux/flux-shell"
+
```
dongahn, line 10: Super minor nit. The commit msg has a typo: 

< to the flux-shell will most likely pick up te system installed flux
> to the flux-shell will most likely pick up the system installed flux

t/t7000-shell-datastaging.t
```diff
@@ -7,10 +7,13 @@ test_description='Test data-staging job shell plugin'
...
 
 test_under_flux 2
 
+flux_shell=$(flux builtin-conf shell_path)
+echo "flux_shell=$flux_shell"
```
dongahn, line 15: Maybe this can go under `test_debug`. Your call though,

dongahn (approved):
Nice enhancement @grondo !

t/t7000-shell-datastaging.t
grondo: Yes, you're right. Fixed.

grondo:
Thanks for the quick review @dongahn. I've fixed the two issues you discovered and squashed the result.

SteVwonder (approved):
Awesome! Thanks @grondo!  LGTM.  Feel free to set MWP when you are ready.

grondo:
Thanks!

PR #812, opened 2021-02-26
[WIP] match: add uniform storage match policy plugin

SteVwonder:
This PR currently is very narrow in scope.  It adds a policy plugin that spreads the IO load across mtl1units based on their currently allocated number of storage slices.

During the team call today, we discussed that it would be good to evolve this into a more generic match policy plugin that would let you select which types to spread across versus pack.  For example, you might want to spread jobs across nodes (rather than packing nodes) when your apps are bandwidth-limited.  You might also want to spread jobs across sockets or caches for similar reasons.


dongahn:
My guess is this can be more enabled  in a more simplified fashion using our `dfu_match_multilevel_id.hpp` work (though I will need to take a look at the implemented policy more closely. Also tagging @jameshcorbett.

garlick:
No activity for over a year - closing.  Please reopen if this should be moved forward.

PR #817, opened 2021-03-24
Add bug fix for by_subplan pruning

dongahn:
This PR adds a bug fix for `by_subplan` pruning. The problem was discovered as I analyzed the performance of Fluxion against the MuMMI workflow @ ORNL Summit.

- Fix the main bug where the pruning filter state was miscalculated, which led to higher numbers of graph visits than necessary: For aggregate-based pruning, `accum_to_parent()` should propagate to the parent resource vertex the amount of resources that will become "unavailable" after each allocation. But code was propagating the new (reduced) available resource amount after the allocation.
- Improve performance by enabling `by_subplan` pruning filter for "headless" jobspecs, the class of jobspecs whose first resource type in the resource section is `slot`. 
- Add graph visit count reporting support into our traverser
- Add tests that use the visit counts to check the correctness of pruning filters.

Note that this is just a precursor to ultimate match performance important via introducing ultrafast first-match policy later.

### DIFF
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -57,8 +57,11 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
     planner_multi_t *p = NULL;
     const subsystem_t &dom = get_match_cb ()->dom_subsystem ();
 
-    if ((rc = detail::dfu_impl_t::select (jobspec, root, meta, x)) == 0)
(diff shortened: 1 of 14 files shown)


dongahn:
@SteVwonder: ok. Everything looks green and this is good for your review.

SteVwonder (approved):
LGTM! Thanks @dongahn!

dongahn:
Thanks. Setting MWP then.

PR #820, opened 2021-04-06
Add "ultrafast" first-match policy support

dongahn:
This PR adds a new match policy which selects the first matching resources and stops the search. This is called ultrafast first-match policy or simply first-match policy. With this policy, our match server does not evaluate the resource graph or `O(G)` in its entirety. Instead, the match performance of each job should scale with respect to the size of the jobspec or `O(J)`. Thus, this can be well-suited for high throughput-oriented workflows, which feature lots of small jobs running across a large scale allocation.

- Include "weight" support into resource graph edges.
- Add out-edge metadata support into our graph store class where we keep the outgoing edges for each resource vertex.
   Edges are essentially kept in their weight order such that traversing thru these edges will result in heavier weight first walk.
- Add out-edge metadata support into all of our resource reader classes (GRUG, HWLOC and JGF).
- Add/use an alternative traversal order within our traverser class that uses this metadata for first-match policy. For this policy, we also  stop traversing the subgraph of the visiting vertex as soon as the minimum resource requirements of the jobspec are met.
- Add test coverage on top of some existing test cases
- Misc. fix and improvement (b0ed7a9 and 6d61961)


dongahn:
The CI is green and this should be ready for your review.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -119,6 +119,47 @@ int dfu_impl_t::upd_idata (vtx_t u, const subsystem_t &s,
...
+    planner_multi_t *subplan = (*m_graph)[tgt].idata.subplans[subsystem];
+    if (subplan) {
+        if ( (len = planner_multi_resources_len (subplan)) == 0)
+            return -1;
+
+        // Set dynamic traversing order based on the following heuristics:
+        //     1. Current-time (jobmeta.now) resource availability
+        //     2. Last pruning filter resource type (if additional
+        //        pruning filter type was given, that's a good
+        //        indication that it is the scarcest resource
```
SteVwonder, line 136: Ultra Minor Nit: un-balanced `(` in the comment

resource/traversers/dfu_impl.hpp
```diff
@@ -429,6 +429,8 @@ class dfu_impl_t {
                       const std::map<std::string, int64_t> &dfu);
```
SteVwonder, line 429: Nit: in the commit message `scarest` -> `scarcest`

t/data/resource/expected/basics/041.R.out
```diff
@@ -0,0 +1,40 @@
+      ---------------core35[1:x]
```
SteVwonder, line 1: I'm confused why the high ID resources are being selected here since the `first` policy is technically lowest ID first under the hood (with the dynamic_explore enabled and select_best_k == 1).

SteVwonder (commented):
Thanks @dongahn for putting this together.  I looked through all of the code changes and the first set of testsuite changes.  The code changes look good to me.  With the testsuite, I'm confused by why the high ID resources are being selected by the `first` policy (see comment below).  That seems like a pretty critical thing that I'm misunderstanding, so before I review the remainder of the testsuite changes, maybe we can resolve that particular comment?

dongahn:
> Thanks @dongahn for putting this together. I looked through all of the code changes and the first set of testsuite changes. The code changes look good to me. With the testsuite, I'm confused by why the high ID resources are being selected by the first policy (see comment below). That seems like a pretty critical thing that I'm misunderstanding, so before I review the remainder of the testsuite changes, maybe we can resolve that particular comment?

This is just because how I break ties.

https://github.com/flux-framework/flux-sched/blob/7a46908a21c6c39167684eae11c71e45485481b9/resource/traversers/dfu_impl_update.cpp#L143

When the weight is the same, which should be the case in the intitial state, we use the `uniq_id` field of the vertex to break the tie. It happens that the higher DIs tend to get higher uniq_id from the way we construct  our resource graph.

resource/readers/resource_reader_grug.cpp
```diff
@@ -153,29 +152,59 @@ edg_t dfs_emitter_t::raw_edge (vtx_t src_v, vtx_t tgt_v)
...
     if (!inserted) {
-        m_err_msg += "error inserting a new edge:"
+        errno = ENOMEM;
+        m_err_msg += "error inserting a new edge: "
                         + g[src_v].name + " -> " + g[tgt_v].name + "; ";
+        return -1;
+    }
+    // add this edge to by_outedges metadata
+    auto iter = m_gm_p->by_outedges.find (src_v);
+    if (m_gm_p->by_outedges.find (src_v) == m_gm_p->by_outedges.end ()) {
```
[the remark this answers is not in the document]
dongahn: Ah good catch. I will fix.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -917,6 +917,36 @@ int resource_reader_jgf_t::unpack_edges (resource_graph_t &g,
                 g[e].idata.member_of[std::string (key)]
                     = std::string (json_string_value (value));
             }
+            // add this edge to by_outedges metadata
```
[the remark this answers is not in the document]
dongahn: Yeah this is a reasonable request. I think the thought was this should be just considered as "one liner" which isn't worthy of a refactor.

resource/policies/dfu_match_policy_factory.hpp
```diff
@@ -36,7 +36,7 @@
 namespace Flux {
 namespace resource_model {
 
-const std::string FIRST_MATCH = "first"
+const std::string FIRST_MATCH = "first";
```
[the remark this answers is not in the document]
dongahn: Good catch. I didn't notice this.

resource/traversers/dfu_impl_update.cpp
dongahn: Great.

t/t1003-qmanager-policy.t
```diff
@@ -35,7 +35,7 @@ test_expect_success 'load test resources' '
 
 test_expect_success 'qmanager: loading qmanager (queue-policy=easy)' '
     load_resource prune-filters=ALL:core \
-subsystems=containment policy=low load-allowlist=cluster,node,core &&
+subsystems=containment policy=first load-allowlist=cluster,node,core &&
```
SteVwonder, line 38: Nit: in commit title `polcies` -> `policies`

t/t3029-resource-prune.t
```diff
@@ -117,5 +117,47 @@ test_expect_success 'prune: default core-level pruning works on w/ headless spec
     test_cmp cmp04 visit_count.out04
 '
 
+#
+# First match only visits min number of vertices
+cmds05="${cmd_dir}/cmds01.in"
+test_expect_success 'prune: default core-level pruning works with first match' '
+    cat <<-EOF >cmp05 &&
+	INFO: PREORDER VISIT COUNT=27
```
SteVwonder, line 125: Just to confirm my understanding, the postorder visit count is 22 since that is the minimal number of resources in the final concrete resource set (1 cluster, 1 rack, 1 node, 1 socket, 18 cores: 22 total resources).  And the pre-order visit count is 5 vertices larger because of the 4 memory vertices (each of size 2) and the 1 gpu vertex under the socket that need to be visited in the pre-order traversal since they are under the socket (but they do not match so no post-order traversal).  Is that right (in particular the reason for the extra 5 visits in the preorder)?

SteVwonder: PS - This is an awesome demonstration of the power of this feature!

SteVwonder (approved):
Thanks for the clarification on the test output.  That makes sense.

Just finished going through the remaining test commits. LGTM @dongahn!

t/t3029-resource-prune.t
dongahn: Yes exactly!

resource/readers/resource_reader_jgf.cpp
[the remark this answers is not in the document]
dongahn: For this time, I decided to just keep the original code since these have become one liners after your suggested change.

dongahn:
OK. @SteVwonder: I addressed all of your requests and took liberty squashing all the changes. If CI turns green, this  is good to go. Since you already approved, I will add MWP.

PR #822, opened 2021-04-17
libczmqcontainers: introduce new internal library of czmq zhashx library

chu11:
This is an identical fix to what was done in https://github.com/flux-framework/flux-core/pull/3596.  We want to avoid excess `zhashx` memory growth due to a bug in the zhashx implementation.  Someday this can be removed when the bug fix gets sent down to OS distros.

Before merging, please double check that your reading of

https://www.mozilla.org/en-US/MPL/2.0/combining-mpl-and-gpl/

matches my reading, since `flux-sched` is GPL instead of LGPL.


garlick (approved):
Looks good to me.

Should we add this to CODE_COVERAGE_IGNORE_PATTERN in the top level Makefile?   (I think we missed that in flux-core, actually).

chu11:
> Should we add this to CODE_COVERAGE_IGNORE_PATTERN in the top level Makefile? (I think we missed that in flux-core, actually).

Oh whoops, wasn't aware of that.  I'll add it here and add a PR for flux-core.

chu11:
re-pushed, adding exclusion of libczmqcontainers in code coverage.

garlick:
Sounds like this is good to go based on @grondo's commet in flux-framework/flux-core#3598.  I'll set MWP - hope sched people don't mind!

Issue #836, opened 2021-05-11
intermittent(?) failure in t1009-recovery-multiqueue.t on aarch64


grondo:
I'm seeing intemittent (>50%) failure in `t1009-recovery-multiqueue.t` on aarch64 in the buildfarm when trying to build RPMs for flux-sched-0.16.0:
```
ok 1 - recovery: generate test jobspecs
PASS: t1009-recovery-multiqueue.t 1 - recovery: generate test jobspecs
ok 2 - load test resources
PASS: t1009-recovery-multiqueue.t 2 - load test resources
ok 3 - recovery: loading flux-sched modules with two queues
PASS: t1009-recovery-multiqueue.t 3 - recovery: loading flux-sched modules with 
two queues
2021-05-11T15:38:43.359529Z sched-fluxion-qmanager.err[0]: update_on_resource_re
sponse: exiting due to sched-fluxion-resource.notify failure: Operation canceled
ok 4 - recovery: submit to occupy resources fully (rv1)
PASS: t1009-recovery-multiqueue.t 4 - recovery: submit to occupy resources fully
 (rv1)
ok 5 - recovery: works when both modules restart (rv1)
PASS: t1009-recovery-multiqueue.t 5 - recovery: works when both modules restart 
(rv1)
ok 6 - recovery: a cancel leads to a job schedule (rv1)
PASS: t1009-recovery-multiqueue.t 6 - recovery: a cancel leads to a job schedule
 (rv1)
flux-start: 0 (pid 65395) Segmentation fault
ok 7 - recovery: cancel all jobs (rv1_nosched)
PASS: t1009-recovery-multiqueue.t 7 - recovery: cancel all jobs (rv1_nosched)
not ok 8 - cleanup active jobs
FAIL: t1009-recovery-multiqueue.t 8 - cleanup active jobs
139
#       
#           cleanup_active_jobs
#       
not ok 9 - removing resource and qmanager modules
FAIL: t1009-recovery-multiqueue.t 9 - removing resource and qmanager modules
#       
#           remove_qmanager &&
#           remove_resource
#       
# failed 2 among 9 test(s)
``` 

I happen to still have the hook to backtrace all corefiles in the specfile, so here's as much of the backtrace I got:

```
found corefile ./t/core.65395
+ gdb --command=cmds /usr/libexec/flux/cmd/flux-broker ./t/core.65395
[snip]
Thread 1 (Thread 0xffff5d7aee20 (LWP 65826)):
[...]

PR #838, opened 2021-05-13
Fix cancel race

dongahn:
This PR adds provisional cancel support into the base queue policy layer to fix a cancel race introduced by the FCFS async support PR (PR #826).

Problem: The `sched.cancel` request handling semantics used to be that when a request is received by `qmanager`, the pending job state was deterministic. The job is either pending or not. It is no longer true because there is a *maybe-pending* state: a match request could have been sent out to `sched-fluxion-resource` and the corresponding stream RPC has not been received. Because the mishandling, an intermittent crash occurs. 

Introduce a provisional cancel request queue and handle a request received during the *maybe pending* state when the schedule loop invocation completes.

Fixes Issue #836. I ran `t1009-recovery-multiqueue.t` 30 times on the docker on my Mac laptop and didn't see the crash.

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -77,6 +77,7 @@ struct t_stamps_t {
     uint64_t running_ts = 0;
     uint64_t rejected_ts = 0;
     uint64_t complete_ts = 0;
+    uint64_t canceled_ts = 0;
 };
 
 /*! Type to store a job's attributes.
@@ -121,11 +122,12 @@ public:
     bool is_scheduled ();
     void reset_scheduled ();
     bool is_sched_loop_active ();
-    void set_sched_loop_active (bool active);
+    int set_sched_loop_active (bool active);
 
 protected:
     int reconstruct_queue (std::shared_ptr<job_t> running_job);
     int pending_reprioritize (flux_jobid_t id, unsigned int priority);
+    int process_provisional_cancel ();
     int insert_pending_job (std::shared_ptr<job_t> &job, bool into_provisional);
     int erase_pending_job (std::shared_ptr<job_t> &job, bool &found_in_prov);
     std::shared_ptr<job_t> pending_pop ();
@@ -133,6 +135,7 @@ protected:
     std::shared_ptr<job_t> rejected_pop ();
     std::shared_ptr<job_t> complete_pop ();
     std::shared_ptr<job_t> reserved_pop ();
+    std::shared_ptr<job_t> canceled_pop ();
     std::map<std::vector<double>, flux_jobid_t>::iterator to_running (
         std::map<std::vector<double>,
                  flux_jobid_t>::iterator pending_iter,
@@ -152,14 +155,17 @@ protected:
     uint64_t m_dq_cnt = 0;
     uint64_t m_cq_cnt = 0;
     uint64_t m_oq_cnt = 0;
+    uint64_t m_cancel_cnt = 0;
     unsigned int m_queue_depth = DEFAULT_QUEUE_DEPTH;
     unsigned int m_max_queue_depth = MAX_QUEUE_DEPTH;
     std::map<std::vector<double>, flux_jobid_t> m_pending;
     std::map<std::vector<double>, flux_jobid_t> m_pending_provisional;
+    std::map<uint64_t, flux_jobid_t> m_pending_cancel_provisional;
     std::map<uint64_t, flux_jobid_t> m_running;
     std::map<uint64_t, flux_jobid_t> m_alloced;
     std::map<uint64_t, flux_jobid_t> m_complete;
     std::map<uint64_t, flux_jobid_t> m_rejected;
+    std::map<uint64_t, flux_jobid_t> m_canceled;
     std::map<flux_jobid_t, std::shared_ptr<job_t>> m_jobs;
     std::unordered_map<std::string, std::string> m_qparams;
     std::unordered_map<std::string, std::string> m_pparams;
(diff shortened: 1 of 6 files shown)


grondo:
I tried this PR out in the buildfarm where the test was failing quite regularly and could not reproduce the crash :+1:.

It does look like the compile is failing in the [centos7 build](https://github.com/flux-framework/flux-sched/pull/838/checks?check_run_id=2573912159#step:5:1055) though.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -404,7 +417,7 @@ class queue_policy_base_t : public detail::queue_policy_base_impl_t,
      *  so that this queue can be adapted for use within high-level
      *  resource API. Set the state of the scheduling loop.
      */
-    virtual void set_sched_loop_active (bool active);
+    virtual int set_sched_loop_active (bool active);
```
SteVwonder, line 420: It's not clear to me from this function name or the comment that the return value is the success/failure of process_provisional_cancel. Can you add a comment about the return value?

resource/hlapi/bindings/c++/reapi.hpp
```diff
@@ -89,7 +89,7 @@ class queue_adapter_base_t {
 
     /*! Set the state of the scheduling loop.
      */
-    virtual void set_sched_loop_active (bool active) = 0;
+    virtual int set_sched_loop_active (bool active) = 0;
```
SteVwonder, line 92: Same comment here about the return value.

SteVwonder (commented):
Thanks @dongahn for putting this together.  One edge case that I can't quite work through in my head that I want to make sure is covered:

What happens to the allocated resources in `resource` if a job that is provisionally canceled has a successful match reponse returned for it by `resource`?  I'm probably just missing it in all the async control flow, but I don't see where the `qmanager` is de-allocating the resources in `resource` during the processing of the provisional cancellation.  I'm mainly worried about a "leak" of allocated resources.

dongahn:
> What happens to the allocated resources in resource if a job that is provisionally canceled has a successful match reponse returned for it by resource? I'm probably just missing it in all the async control flow, but I don't see where the qmanager is de-allocating the resources in resource during the processing of the provisional cancellation. I'm mainly worried about a "leak" of allocated resources.

From https://github.com/flux-framework/rfc/blob/master/spec_27.rst, `sched.cancel` should be NOOP in this case.

> If the specified job does not have a pending sched.alloc request, the request SHALL be ignored by the scheduler.

A provision cancel is effected only for non pending job on the completion of the sched loop invocation as in  https://github.com/flux-framework/flux-sched/blob/9acca9eb18737f6a9edc6ec5e9375d8d23d0eb4f/qmanager/policies/base/queue_policy_base_impl.hpp#L611. So this should be handled correctly. 

Let me check the job states are also correctly set in this case.

qmanager/policies/base/queue_policy_base.hpp
dongahn: Yes! good point.

SteVwonder:
I think I'm still missing something.  The RFC you referenced describes the job-manager <---> qmanager interaction.  The edge case I'm worried about is the qmanager <---> resource one.  More explicitly:

```
job-manager                  qmanager---(multi match A,B)---->resource
job-manager-(cancel job A)-->qmanager                         resource
job-manager                  qmanager<---(match success A)----resource                            
job-manager<-(job A cancel)--qmanager                         resource
job-manager                  qmanager<----(match failure B)---resource     
job-manager                  qmanager-(release A resources)-->resource
```

It's that last interaction that I'm not seeing: qmanager releasing resources from `resource`.  Presumably because of the `match success` response from resource to qmanager, the `resource` module has the resources it allocated for Job A marked off as allocated.  Unless it is informed of the job cancellation, those resources will still be marked as allocated, right?

dongahn:
Yes this is a bit complex under the async model. I believe the corner case is resolved through `job-manager` and `qmanager` allocation protocol. I attempted to revise your diagram below. 

```
job-manager                  qmanager---(multi match A,B)---->resource
job-manager-(cancel job A)-->qmanager                         resource
job-manager                  qmanager<--(match success A)----resource      
job-manager<-(sched.alloc A)-qmanager                        resource                            
job-manager                  qmanager<----(match failure B)---resource   
# this ends the sched loop invocation and cancel job A request is ignored because A's sched.alloc has already responded. 
# then, I have to believe job-manager will send sched.free A
job-manager-(sched.free  A)->qmanager
job-manager                  qmanager-(release A resources)-->resource
```

SteVwonder:
>  I attempted to revise your diagram below.

Thanks @dongahn.  In this line:

```
job-manager<-(sched.alloc A)-qmanager                        resource
```

Is that a sched.alloc that contains resources, or is that a "alloc cancel" sched.alloc (https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_27.html#alloc-cancel)?  If the former, then I agree with your diagram that the job-manager will send a followup free request and everything should be fine.

dongahn:
> Is that a sched.alloc that contains resources, or is that a "alloc cancel" sched.alloc (https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_27.html#alloc-cancel)? If the former, then I agree with your diagram that the job-manager will send a followup free request and everything should be fine.

Former.

SteVwonder (approved):
Pending the addition of the comments, this LGTM! Thanks @dongahn!

dongahn:
> Pending the addition of the comments, this LGTM! Thanks @dongahn!

Thanks. BTW, I'm not super happy with coverage. I will try to add some tests to increase the coverage. Things are highly time dependent so my mileage will vary to cover non-deterministic testing space though.

dongahn:
> Pending the addition of the comments, this LGTM! Thanks @dongahn!

I've added those comments and squash the fixup commits.

>  I'm not super happy with coverage. I will try to add some tests to increase the coverage. 

I've also added a test to get a bit more coverage. Not perfect, but probably good enough for now. More coverage will comeas more policies and operations will be transition to async mode later.


@SteVwonder: If you want you can take a quick look at f195671 and if you are happy, feel free to add MWP.

SteVwonder:
LGTM! Thanks @dongahn!

dongahn:
Thanks!

Issue #830, opened 2021-05-04
Change default queue-depth to 32


dongahn:
Per https://github.com/flux-framework/flux-sched/pull/826#discussion_r625917462 with the new `delay scheduling` semantics and match-multi, we should consider changing the default queue depth small. It is currently 8192.

garlick:
Just curious, what is the difference between setting the fluxion queue-depth to 32 and setting the job manager scheduler mode to limited=32?  In the latter case, the scheduler is only presented with alloc requests for the 32 highest priority jobs.

dongahn:
Yeah this would primarily for `mode=unlimited`.

dongahn:
It is on my TODO list to evaluate the queue-depth support provided by `job-manager`. My guess is this can help with this problem: https://github.com/flux-framework/flux-core/issues/3630. 

I vaguely remember there could be some potential implication of this when we hook in multi-factor priority plugin. Somehow I can't find the discussion thread...

PR #843, opened 2021-05-19
Default policy changes

### LINKED ISSUE #827: Change the default match policy to "first match"?
Since the first match is a lot faster and most of our current users require high throughput mode, perhaps we should consider changing the default match policy to first match? Any thoughts?

dongahn:
This PR changes the default match policy to "first match" and the default queue-depth to 32. Given that many of the current users require high throughput scheduling and it would make sense the default values are tuned for them. For system-instance, the policies will be configured and set differently through configuration and rc files.

Fixes #827, #830.

### DIFF
--- a/qmanager/config/queue_system_defaults.hpp
+++ b/qmanager/config/queue_system_defaults.hpp
@@ -28,7 +28,7 @@
 namespace Flux {
 namespace queue_manager {
     const unsigned int MAX_QUEUE_DEPTH = 1000000;
-    const unsigned int DEFAULT_QUEUE_DEPTH = 8192;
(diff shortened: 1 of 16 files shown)


dongahn:
This should be ready for a review.

SteVwonder (approved):
LGTM! Weird that the valgrind test takes longer under first, but a bump in timeout is good by me.

dongahn:
Thanks @SteVwonder. 

> Weird that the valgrind test takes longer under first, but a bump in timeout is good by me.

Probably this can be attributed to more extensive use of C++ containers.

PR #845, opened 2021-05-27
python: export FluxionResource classes

### LINKED ISSUE #844: flux-ion-R: make FluxionGraphResourceV1 and other classes public
I've been leveraging these classes in my flux-coral2 work (by copying and pasting them out of the command) to construct JGF files from the DWS API.

It would be nice if these classes were installed and exported as part of the `fluxion` python module.  Then I could create subclasses in my flux-coral2 command that inherit from the classes exported by Fluxion.

SteVwonder:
Problem: the ResourceGraphV1 and other related class are useful to other
flux projects, like flux-coral2, but they are encapsulated in an
unimportable Flux sub-command `flux-ion-R`

Solution: spin out those classes into a `resourcegraph` sub-module of
the `fluxion` python module, making them importable by both the
`flux-ion-R` command and the subprojects. Also change their dunder
(`__`) methods/attr to just single under (`_`) so they can be more
effectively leveraged by subclasses.

Related: tacked on a few changes to .gitignores to produce a cleaner
`git status` after `make -j` and `make 

Closes #844


dongahn:
FYI -- reviewing this is one of my next priorities.

t/.gitignore
```diff
@@ -0,0 +1,3 @@
+/t9990-python-tests.t
```
dongahn, line 1: We can customize `.gitignore` by dropping this. But if the purpose is to keep it sync with `flux-core`'s file, I'm fine with keep it.

.gitignore
```diff
@@ -23,6 +23,20 @@ libtool
...
+# libtool pull-ins
+/config/libtool.m4
+/config/ltoptions.m4
+/config/ltsugar.m4
+/config/ltversion.m4
+/config/lt~obsolete.m4
+
+# docs intermediate files
+/doc/man*/*.xml
+/doc/_build
```
dongahn, line 40: We can customize `.gitignore` by dropping this. But if the purpose is to keep it sync with `flux-core`'s file, I'm fine with keep these lines..

.gitignore
```diff
@@ -23,6 +23,20 @@ libtool
 .libs/
 libltdl/
 
+# generated library version header
+/src/common/libflux/version.h
```
dongahn, line 27: We can customize `.gitignore` by dropping this line. But if the purpose is to keep it sync with `flux-core`'s file, I'm fine with keep.

dongahn (approved):
@SteVwonder: this PR looks great to me. I don't know why I delayed this review when it is so straightforward to review :-). 

I have added a few comments about whether we want to customize `.gitignore`s for fluxion differently from `flux-core`'s or keep them in sync with `flux-core`'s. 

Let me know your preference. I can go either way, but I thought we need to have a discussion at least for our future reference.

t/.gitignore
SteVwonder: Yeah, my thought was to copy it in sync so future copy overs are easy.  I can also go either way with a preference for keeping them in sync.

dongahn: OK. This can go in as is then. Adding MWP.

PR #846, opened 2021-06-17
planner: port to Yggdrasil's augmented red-black library

dongahn:
This PR ports planner to Yggdrasil's augmented red-black
package: https://github.com/tinloaf/ygg, which is MIT licensed. 

It replaces the planner’s dependency on Linux Kernel’s
red-black tree package (GPL-licensed) with Yggdrasil,
a most significant step towards migrating Fluxion
to a more permission licensing.

The Yggdrasil port has been done in multiple steps
and the commits closely follow these steps:
1) Port the current planner implementation to C++
 while keeping its C API;
2) Abstract out scheduled-point and resource search trees
into distinct C++ classes and port the planner to
the abstract interface;
3) Use the LK RB tree code to implement this abstraction
first as its reference implementation;
4) Separate out the required bits from Yggdrasil and vendor
them into the `src/common/yggdrasil` directory;
5) Aggressively clean up and backport these bits so that
they can be compiled with C++11 without a performance loss;
6) Port the search tree abstractions to a Yggdrasil-based
implementation and adjust the build environments as well
as README.

Finally, this PR deprecates the use of vendored czmq container
library and of the most of libutil code. The planner was
the last layer that required them and we migrated that to
C++ constructs such as STL containers and `new`/`delete`.


dongahn:
xzmalloc coverage change is impacting the codecov. There is still a very small footprint of `xzmalloc` within `planner_multi` so twecouldn't wean off of xzmalloc. Let me see if I can use some trick to remove this.

dongahn:
Everything looks good now. This is ready for you review.

SteVwonder (commented):
Thanks @dongahn for putting this together!  The biggest hurdle to the LGPL re-license is almost done :)  I want to take another look at 61bad932245ff4d833456903147af18c761b5641 with fresh eyes tomorrow morning, but I figured I'd leave the comments I have so far from my pass tonight.  Overall, looks good to me.  Thanks for structuring the commits the way you did; it made the review much easier.  The move to C++ resulted in some really nice improvements in various places.

resource/planner/mintime_resource_tree.cpp
```diff
@@ -0,0 +1,272 @@
...
+        if (min > left->subtree_min)
+            min = left->subtree_min;
+    }
+    mt_resource_rb_node_t *right = node->get_right ();
+    if (right) {
+        if (min > right->subtree_min)
+            min = right->subtree_min;
+    }
+    node->subtree_min = min;
+    fix_propagate (node);
```
SteVwonder, line 152: Nit, but I don't think `fix_propagate` is strictly necessary. I think you could pull this off in one function with less code duplication (totally optionatl):

```suggestion
while(node != nullptr) {
    int64_t min = node->at;
    mt_resource_rb_node_t *left = node->get_left ();
    if (left) {
        if (min > left->subtree_min)
            min = left->subtree_min;
    }
    mt_resource_rb_node_t *right = node->get_right ();
    if (right) {
        if (min > right->subtree_min)
            min = right->subtree_min;
    }
    node->subtree_min = min;
    node = node->get_parent();
}
```

resource/planner/mintime_resource_tree.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+template <class mt_resource_rb_node_t, class NodeTraits>
+template <class BaseTree>
+void mt_resource_node_traits<mt_resource_rb_node_t, NodeTraits>::rotated_left (
+         mt_resource_rb_node_t &node, BaseTree &tree)
+{
+    fix (&node);
+    if (node.get_parent ())
+       fix (node.get_parent ());
```
SteVwonder, line 182: I thought `fix` handled this already via the walk up the ancestor list via `fix_propagate`?

resource/planner/mintime_resource_tree.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+template <class mt_resource_rb_node_t, class NodeTraits>
+template <class BaseTree>
+void mt_resource_node_traits<mt_resource_rb_node_t, NodeTraits>::rotated_right (
+         mt_resource_rb_node_t &node, BaseTree &tree)
+{
+    fix (&node);
+    if (node.get_parent ())
+       fix (node.get_parent ());
```
SteVwonder, line 192: Same question as above

resource/planner/mintime_resource_tree.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+template <class mt_resource_rb_node_t, class NodeTraits>
+template <class BaseTree>
+void mt_resource_node_traits<mt_resource_rb_node_t, NodeTraits>::swapped (
+         mt_resource_rb_node_t &node1, mt_resource_rb_node_t &node2, BaseTree &t)
+{
+    fix (&node1);
+    if (node1.get_parent ())
+       fix (node1.get_parent ());
```
SteVwonder, line 210: Same question as above

SteVwonder (commented):
Thanks again @dongahn for putting this together.  I gave that commit a look today with fresh eyes and had one question (that applies in a few different places).  I think it's just me missing something, but for my own learning I'm hoping to grok it.

resource/planner/planner.cpp
```diff
@@ -653,26 +639,36 @@ static int span_input_check (planner_t *ctx, int64_t start_time,
...
 {
-    char key[32];
-    span_t *span = NULL;
-    if (span_input_check (ctx, start_time, duration, (int64_t)request) == -1)
-        goto done;
+    std::shared_ptr<span_t> span = nullptr;
+    try {
+        if (span_input_check (ctx, start_time, duration, (int64_t)request) == -1)
+            goto done;
+        ctx->span_counter++;
```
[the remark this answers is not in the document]
dongahn: This is intended. `span_counter` should hold the last span ID that was used. So it is incremented first and make sure the ID map doesn't already have this. This is a very unlikely condition but could happen if the lookup table was updated incorrectly during a code churn.

resource/planner/planner.cpp
```diff
@@ -368,93 +367,94 @@ static scheduled_point_t *mintime_resource_mintime (int64_t request,
...
+    scheduled_point_t *point = nullptr;
     struct rb_root *root = &(ctx->mt_resource_tree);
-    zlistx_t *keys = zhashx_keys (ctx->avail_time_iter);
-    const char *k = NULL;
-    for (k = reinterpret_cast<const char *> (zlistx_first (keys));
-         k; k = reinterpret_cast<const char *> (zlistx_next (keys))) {
-        point = reinterpret_cast<scheduled_point_t *> (
-                    zhashx_lookup (ctx->avail_time_iter, k));
+    for (auto &kv : ctx->avail_time_iter) {
+        point = kv.second;
```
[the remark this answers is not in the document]
dongahn: Yes, I love this syntax, too. This makes C++ code as simple as dynamic programming language like Python! At the same time, modern C++ adds so much into its "standard", sometime I feel overwhelmed to keep track of all the changes. I had to do that to backport Yggdrasil into C++11 from C++17...

resource/planner/mintime_resource_tree.cpp
dongahn: > I think you could pull this off in one function with less code duplication (totally optionatl):

BTW, code duplication is a part of initial performance optimization. This is the hottest spot of mintime resource search tree. Although modern C++ compilers do a good job of inlining, I found non negligible overheads with an old abstraction like `mintime_resource_subtree_min` in a separate function.

dongahn: I know where you are coming from. But the loop body of the `fix_propagate` and the node `fix` code is slightly different. The `fix_propagate` has one conditional statement where terminates the loop when the visiting node's existing `subtree_min` is the same as the new min. This way, you don't have to propagate the fixup to your ancestor nodes unless you have to.

How I see this is, there is one special case -- the initial node where the fixup is applied and then general case where the fixup + a conditional is performed. So you can consider this a loop unrolling: a bit better performance and a bit better readability,

Hopefully this makes sense.

resource/planner/planner_multi.cpp
```diff
@@ -169,13 +176,13 @@ extern "C" int64_t planner_multi_resource_total_by_type (
     int64_t rc = -1;
     if (!ctx || !resource_type)
         goto done;
-    for (i = 0; i < ctx->size; i++) {
-        if (!strcmp (ctx->resource_types[i], resource_type)) {
+    for (i = 0; i < ctx->planners.size (); i++) {
+        if ( !(strcmp (ctx->resource_types[i], resource_type))) {
```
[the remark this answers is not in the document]
dongahn: Yes, I agree the use of `std::find` is a more modern C++ pattern. 

But I think two things make the use of `std::find` a bit awkward in this case. 1) To be evolutionary with the original C code, the data (`planner_multi_t` context) is organized as vectors of objects as opposed to a single vector of a struct that contains all the related objects under the same index; 2) the data type of `resource_type` is a raw C char string. 

Because of 2), `std::find` will still have to use `strcmp` in a comparator function (e.g., lambda function) anyway. Because 1), somehow you should still calculate the index of the found element to retrieve the corresponding planner object from another vector. At this point, this will probably reduce readability such that this will outweigh any benefits of using `std::find`.

I could reorganize the `planner_multi_t` later to be a vector of struct. But that should probably come from a future PR.

resource/planner/mintime_resource_tree.cpp
dongahn: Actually, after looking at your downstream comments, I have a second thought. Perhaps implement the whole thing in a single loop makes more sense for readability. I will mark it resolved here and describe my thinking in one of your other comments below.

resource/planner/mintime_resource_tree.cpp
dongahn: Great question! Thank you for your thoughtful review; I know this code/algorithm is pretty involved to reason about though the code footprint is small. 

I believe a proper algorithm here is to do the full fixup for both of these nodes for rotate-left and -right. And four of the involved nodes in the swapped case as well. 

Let's use a right rotation as an example. When a red black tree performs a right rotation, the `node` object that is passed to the callback function is the old parent that just became the right child. Similarly, `node->get_parent()` is the old left child node that just became the new parent. So both of these nodes require a full fixup as their state changed. 

Since the proper fixup sequence is bottom first and then up in tree, we fix the `node` first and then fix `node->get_parent()`. The reason for the parent fixup here is this new parent just hasn't had its full fixup applied yet after its state changed. It did get a propagation applied but it in general is not a full fixup.  

Having said that, I should note that this is the conceptual algorithm. But I can see your great review points here. 

Currently, I see `fix_propagate` does its propagation too excessively such that the propagation nearly performs the full fixup on the parent nodes, making the actually parent node fixup nearly redundant. 

So, I think a more proper change would actually to be make `fix_propagation` less excessive and use the same double fixup sequence for rotates and swaps. If nothing else, this will improve readability with a slightly better separation of concerns.

Let me do a intermediate commit for your further review. I plan to do this in a single loop body as you suggested in another post.

resource/planner/mintime_resource_tree.cpp
dongahn: See above.

resource/planner/mintime_resource_tree.cpp
dongahn: Ok. 0c5a056 adds better pruning support into propagation.

SteVwonder (approved):
LGTM! Thanks @dongahn!

resource/planner/mintime_resource_tree.cpp
trws: Is this discussion still correct? It looks like `fix(&node)` propagates to the parent on line 136.  Is the second necessary because of the early break on `subtree_min` matching `old`, or is this a vestige of an older split fixup implementation?

dongahn: The discussion is correct and the https://github.com/flux-framework/flux-sched/pull/846/commits/bcd952ee15e1e5f7e1b0666a0b7d32a12f92c8de addresses it.

trws: Ok, I am probably misunderstanding something then.  The commit you linked looks good, and makes sense, but it looks like `fix(node)` runs `if(node.get_parent()) fix(node.get_parent())` internally, added in that commit in fact starting at line 136? The code seems correct, it just looks like the check for and fix of parent is run twice.

dongahn: Yeah, this little algorithm can be pretty deceiving with many corner cases. Generally, this algorithm needs the double fixup. For example, consider this case. 



The graph "After" is after you performed a right rotation. In this case, node `10` will be fixed up first. And when we fix it, we found its `subtree_min` is identical to its `at` so you will stop propagating. So without the explicit fixup for node `9`, the mintime resource search tree believes that time `at=1` is where a resource request of quantity 9 or less could be made. Clearly this is wrong.

dongahn:
OK. I squash the fixup commit. As far as I'm concerned, this can go in.

PR #849, opened 2021-06-24
Replace GNU readline with editline

dongahn:
This PR adds support for the editline library
and deprecates the use of GNU readline within
Fluxion.

Editline uses a BSD style license which is
much more compatible with more permissive
licensing including LGPL.

This PR does not yet include new LICENSE
term changes for Fluxion yet because PR #846 
still needs to be merged in before we can do that.

Terms will change in the near future at once
after both this PR and PR #846 will be merged.

### DIFF
--- a/src/test/docker/centos7/Dockerfile
+++ b/src/test/docker/centos7/Dockerfile
@@ -10,9 +10,9 @@ RUN sudo yum -y install \
 	libboost-filesystem-devel \
 	libboost-regex-devel \
 	libxml2-devel \
-	readline-devel \
 	python-yaml \
-	yaml-cpp-devel
+	yaml-cpp-devel \
+	libedit-devel
 
 # Add configured user to image with sudo access:
 #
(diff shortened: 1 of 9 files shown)


src/test/docker/centos7/Dockerfile
```diff
@@ -12,7 +12,8 @@ RUN sudo yum -y install \
 	libxml2-devel \
 	readline-devel \
```
SteVwonder, line 13: Looks like readline is still installed in some of the dockerfiles.  Should we remove that in this PR too?

SteVwonder (commented):
Darn. That was really straightforward.  Awesome! Just one comment:

src/test/docker/centos7/Dockerfile
trws: Probably best.  It will probably be installed anyway as a dependency of something else in there (gdb uses it, as do many other packages), but having it out of the dep list should make it more clear.

dongahn: Yap, good catch! I will remove it and squash the change.

dongahn:
OK. Done.

SteVwonder (approved):
Thanks @dongahn! LGTM!

Issue #856, opened 2021-09-01
Compilation error from unsigned int to signed int comparison


ksuarez1423:
Installing flux-sched version 0.16.0 via Spack with GCC 9.2.0 results in failure, due to an unsigned int and signed int comparison. While this would normally just be a warning, -Werror being in use in the compilation options of flux-sched when using GCC ensures it is taken as an error. 

While 0.16.0 is an past version, the lines of code and relevant datatypes causing the issue are identical in the repository as it currently stands.  The exact error occurs in a couple of the readers, with links here for the instances I get. Do I need to invoke Spack differently, or is this something that needs to be handled on the Flux side of things?

https://github.com/flux-framework/flux-sched/blob/edea1077c5b539aaf6f8e710265222c24f51dbd6/resource/readers/resource_reader_jgf.cpp#L300

https://github.com/flux-framework/flux-sched/blob/edea1077c5b539aaf6f8e710265222c24f51dbd6/resource/readers/resource_reader_hwloc.cpp#L285

dongahn:
Ugg. Yes this depends on how strictly a C++ compiler and version checks those things.

dongahn:
```c++ 
if ( remap_id > static_cast<uint64_t> (std::numeric_limits<int>::max ())) {
```

If you can patch the code as part of your Spack deployment, could you try the following?

ksuarez1423:
> 
> 
> ```c++
> if ( remap_id > static_cast<uint64_t> (std::numeric_limits<int>::max ())) {
> ```
> 
> If you can patch the code as part of your Spack deployment, could you try the following?

I apologize for taking so long to respond -- after applying a patch like this, I was able to install flux-sched! Thank you!

dongahn:
Great! Just out of curious, would it be possible to do a quick PR with this patch?

ksuarez1423:
@dongahn I cannot compile master right now, but I've got a PR at #858 which should fix the issue if I wrote it correctly. A similar change definitely handled things on 0.16.0.

PR #858, opened 2021-09-07
Using static_cast to ensure type compatibility in readers.

ksuarez1423:
Potential fix for #856. Cannot currently compile master version of flux-sched, and instead have compiled 0.16.0 with this change. There is a slight difference between these versions in the readers, but the bug remains the same -- regardless, the change to this particular version is untested. However, to the best of my knowledge, this will ensure that type differences do not prevent compilation on platforms using GCC.

### DIFF
--- a/resource/readers/resource_reader_hwloc.cpp
+++ b/resource/readers/resource_reader_hwloc.cpp
@@ -282,7 +282,8 @@ int resource_reader_hwloc_t::walk_hwloc (resource_graph_t &g,
                     rc = -1;
                     break;
                 }
-                if (remap_id > std::numeric_limits<int>::max ()) {
+                if (remap_id 
+                        > static_cast<uint64_t> (std::numeric_limits<int>::max ())) {
                     errno = EOVERFLOW;
                     m_err_msg += "Remapped gpu id too large; ";
                     rc = -1;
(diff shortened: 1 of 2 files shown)


resource/readers/resource_reader_hwloc.cpp
```diff
@@ -282,7 +282,7 @@ int resource_reader_hwloc_t::walk_hwloc (resource_graph_t &g,
                     rc = -1;
                     break;
                 }
-                if (remap_id > std::numeric_limits<int>::max ()) {
+                if (remap_id > static_cast<uint64_t>(std::numeric_limits<int>::max ())) {
```
dongahn, line 285:
```suggestion
                if (remap_id 
                        > static_cast<uint64_t> (std::numeric_limits<int>::max ())) {
```

dongahn: A code style suggestion and super minor nit. Please notice the whitespace for `static_cast`. The same comment apply below lines as well.

dongahn (changes requested):

resource/readers/resource_reader_hwloc.cpp
ksuarez1423: Made suggested change in all locations -- although, one of the locations uses remap_rank, and I may have misunderstood the suggestion for that spacing.

dongahn: Looks great. Could you squash the last commit with the first and force a push since we don't want to keep interim commits?

ksuarez1423: Sorry for the wait, @dongahn -- I think I've gotten it right?

SteVwonder: The squash LGTM!

SteVwonder (approved):
LGTM! Thanks @ksuarez1423!

Issue #842, opened 2021-05-19
first policy has an issue with unsatisfiability


dongahn:
With the first-match policy when all ranks are excluded, the multi-match reports the following error. This doesn't show up when the match policy is "search all."

> 2021-05-19T00:58:26.496614Z sched-fluxion-resource.err[0]: match_multi_request_cb: match failed due to match error (id=34594619392): No such file or directory
> 2021-05-19T00:58:26.531823Z sched-fluxion-resource.err[0]: match_multi_request_cb: match failed due to match error (id=35181821952): No such file or directory

dongahn:
Reproducer:

```console
ahn1@docker-desktop:/usr/src/t/Issue842$ cat resource.toml
[resource]
exclude = "0-3"
ahn1@docker-desktop:/usr/src/t/Issue842$ mv resource.toml resource.toml.old
ahn1@docker-desktop:/usr/src/t/Issue842$ export FLUX_CONF_DIR=`pwd`
ahn1@docker-desktop:/usr/src/t/Issue842$ flux start -s 4
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

ahn1@docker-desktop:/usr/src/t/Issue842$ flux module unload sched-fluxion-resource
2021-09-15T05:17:27.330671Z sched-fluxion-qmanager.err[0]: update_on_resource_response: exiting due to sched-fluxion-resource.notify failure: Operation canceled
ahn1@docker-desktop:/usr/src/t/Issue842$ mv resource.toml.old resource.toml
ahn1@docker-desktop:/usr/src/t/Issue842$ flux config reload
ahn1@docker-desktop:/usr/src/t/Issue842$ flux resource reload --xml /usr/src/t/data/hwloc-data/004N/exclusive/04-brokers-sierra2
ahn1@docker-desktop:/usr/src/t/Issue842$ flux kvs get resource.R
{"version": 1, "execution": {"R_lite": [{"rank": "0-3", "children": {"gpu": "0-3", "core": "0-43"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["sierra[3682,3179,3683,3178]"]}}
ahn1@docker-desktop:/usr/src/t/Issue842$ flux module load sched-fluxion-resource load-allowlist=cluster,node,gpu,core match-format=rv1 policy=first
ahn1@docker-desktop:/usr/src/t/Issue842$ flux module load sched-fluxion-qmanager
[...]

PR #859, opened 2021-09-16
resource: fix a satisfiabilty bug with backfilling

dongahn:
This PR adds a fix for a satisfiability bug logged
in Issue #842.

With a backfilling-based queuing policy (e.g., EASY),
`schedule()` within traverser iterates through existing
scheduled points until it can make a reservation for the
given job request. During this iteration, however, if
the job request is greater than the total number of cores
available under the cluster vertex level, it exits this iteration
with `errno=ERANGE`. This creates an issue with the final
satisfiability check. 

Add `errno=ERANGE` as one more condition to check to
allow for invoking the final satisfiability check (checking
satisfiability assuming all resources are available).

Remove `errno=ERANGE` as an error condition to report
during the final satisfiability check to avoid confusing
error messages. 

Add a test case to one of our satisfiability sharness tests
and updates the expected output.

Finally, this PR adds a misc. change to dynamic exploration
used by `policy=first` to make the new code more consistent
with the static exploration used by globally optimal
match policies.

Fixes Issue #842


resource/traversers/dfu.cpp
```diff
@@ -83,8 +83,10 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
             m_total_preorder += detail::dfu_impl_t::get_preorder_count ();
             m_total_postorder += detail::dfu_impl_t::get_postorder_count ();
         }
-        // The planner layer returns ENOENT when no scheduleable point exists
-        if (rc < 0 && errno == ENOENT) {
+        // The planner layer returns
+        //     ENOENT when no scheduleable point exists
+        //     ERANGE when the total core count at the root < the request
```
milroy, line 88: This line should be: `ERANGE` when the total **_avaliable_** core count at the root < the request, correct?

dongahn:
@milroy: I don't know why adding MWP doesn't merge this in automatically. Could you manually merge this PR?

PR #870, opened 2021-10-08
[DEMO] [WIP] resource: add mtl4 tests for lustre constraints

SteVwonder:
Adds test for fluxion-resource's ability to satisfy multi-tier lustre constraints.  In particular, the ability to simultaneously satisfy multi-tier storage requirements for OSTs, MDTs, and MGTs.  The first test on an "oversized" system using the `first` match policy passes.  The second test is on a "minimally" sized system (i.e., I can create a concrete R by hand that satisfies the jobspec and most of the rabbit storage) using the `first` match policy.  The second test fails due to Fluxion's graph coloring, which prevents the re-visiting of vertices within jobspec/resource graph traversal.


dongahn:
For my own note and also tagging @jameshcorbett, this needs to be incorporated into the main branch sooner rather than later (before this becomes too diverged).

garlick:
No activity for over a year - closing.  Please reopen if this should be moved forward.

Issue #814, opened 2021-03-03
Add `sched.feasibility` service in support of new job validation plugin


SteVwonder:
https://github.com/flux-framework/flux-core/pull/3533 added a job validation plugin that pings a `sched.feasibility` service to determine if a job is feasible for scheduling at ingest time.

Support for this service was added to `sched-simple`, and it would be good to have similar functionality in Fluxion too.

grondo:
Thanks, @SteVwonder!

Note that the `feasibility` validator plugin has an option to override the service name for the feasibility check, `--feasibility-service=NAME` in case Fluxion needs to send the RPC to the resource module (or another service) instead.

dongahn:
FYI -- I will be looking into this as my next item.

grondo:
> Can the Fluxion feasibility validator add a "feasibility validated" stamp to the jobspec?

Currently there is no way for the validator to modify, amend, or annotate the jobspec - validators are restricted to returning success (an `errnum` of `0`) or an error number and an error string.

Even in principle, Flux components cannot currently add anything to the jobspec since that would invalidate the user's signature on the jobspec.

Edit: Oh, after reading back your response, you may be talking some kind of external annotation of a job which Flux components could add. We have indeed talked about this before, but there is nothing currently that supports this.

I think what we'll want to do is have the feasibility plugin enabled by configuration and the scheduler could check for that same configuration to determine if a feasibility check has already been accomplished.

dongahn:
> Even in principle, Flux components cannot currently add anything to the jobspec since that would invalidate the user's signature on the jobspec.

From rabbit support discussion, I somehow mistakenly thought this could be done. 

> Edit: Oh, after reading back your response, you may be talking some kind of external annotation of a job which Flux components could add. We have indeed talked about this before, but there is nothing currently that supports this.

From the scheduler point of view, if the fact that the job was already checked through the feasibility validator is somehow passed to `sched.alloc`, it can make use of it to avoid redundant feasibility...

> I think what we'll want to do is have the feasibility plugin enabled by configuration and the scheduler could check for that same configuration to determine if a feasibility check has already been accomplished.

Good point. This would work for the system instance. 

But my concern would be fluxion users might want to specialize the scheduling of the user level instance without the front-loaded feasibility since this can significantly lower their throughput...


Ok. It seems this would be the path to least resistance.

dongahn:
[...]

PR #871, opened 2021-10-11
Add `sched.feasibility` service for flux-core's jobspec feasibility validator 

dongahn:
This PR adds `sched.feasibility` service within Fluxion to enable the jobspec feasibility validator from flux-core.

Add `MATCH_SATISFIABILITY` into our traverser code as one of the supported match-operation-enum kinds and work our way upward to enable this operation within all of the upper layers.

Use exactly the same mechanism for its satisfiability check as that of `MATCH_ALLOCATE_W_SATISFIABILITY` except that it doesn't create an allocation or reservation at all.

Add three levels of testing:
1. Augment `resource-query` to test the traverser's match satisfiability support in isolation;
2. Augment `flux ion-resource` to test match satisfiability service solely in the `sched-fluxion-resource` context;
3. Add a new sharness script to test the actual `sched.feasibility` service with the vanilla `feasibility` validation plugin for `flux-job-validator`.

Add other misc. cleanup and fixes.

Fixes Issue #814.

### DIFF
--- a/t/data/resource/commands/pruning/cmds04.in
+++ b/t/data/resource/commands/pruning/cmds04.in
@@ -1,0 +1,6 @@
+# 4x node[1]->slot[1]->socket[1]->core[1]
+match satisfiability @TEST_SRCDIR@/data/resource/jobspecs/pruning/test001.yaml
+match satisfiability @TEST_SRCDIR@/data/resource/jobspecs/pruning/test001.yaml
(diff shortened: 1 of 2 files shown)


dongahn:
I have addressed the python format issue and forced a push. Our CI is green. 

@milroy: let me know if you have any comments.

Issue #867, opened 2021-09-29
fluxion marks nodes 'free' when they are offline at startup


garlick:
On fluke testing, many nodes in the config were offline when flux was started.  This was shown by `flux resource status` and `flux overlay status`.  However, `flux resource list`, which queries the scheduler, reports:
```
[root@fluke1:~]# flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free    101      404        0 fluke[1,3,108,6-103]
 allocated      0        0        0 
      down      0        0        0 
```

dongahn:
Thanks

garlick:
@grondo had a good suggestion to bring over some of the system instance test infrastructure (most of which is embedded in `flux-start` and the current flux-core version of `flux-sharness.sh`.  We could possibly then just bring over [this test](https://github.com/flux-framework/flux-core/blob/master/t/t3301-system-latestart.t) to cover this issue for fluxion.

dongahn:
Right before, fluxion restarts its reactor, it marks all resources "down" and mark those up ranks to "up": https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L1353. But perhaps no "up" rank was reported in this case and we may have missed this conditional statement that mark "all" down. I will see if I can add some online/offline testing coverage suggested in https://github.com/flux-framework/flux-sched/issues/867#issuecomment-930534975 to see if I can reproduce this from a testing environment and develop a fix as well as some testing coverage.

garlick:
To clarify: roughly half did not ever wire up.  there were still a good number of nodes that did come online.

dongahn:
> To clarify: roughly half did not ever wire up. there were still a good number of nodes that did come online.

Hmmm. Then, this doesn't add up. Will add some more debug print and use your suggested test case.

dongahn:
I guess the question is when the  first`up` key will be reported? During the first `resource.acquire` or subsequent response? If subsequent, this is after fluxion finishes its initialization so the problem I described above will still happen.

garlick:
Here's the `resource.eventlog` from that system:
```
flux kvs eventlog get resource.eventlog
1628614560.333701 resource-init {"restart":false,"drain":{},"online":"0","exclude":""}
1628614560.334498 resource-define {"method":"dynamic-discovery"}
1632944117.120696 resource-init {"restart":true,"drain":{},"online":"0","exclude":"0-1,100"}
[...]

Issue #866, opened 2021-09-29
fluke testing:  system is deadlocked


garlick:
Gathering some observations from fluke after our flux-core 0.29.0 release was installed as the system instance
* `flux mini run hostname` hangs
* `flux overlay status -v` shows 47 nodes offline
* `flux overlay status -vvv` shows 53 nodes online
* `flux ping <rank>` seems to work for  the online nodes
* for the offline nodes, we get an immediate "no route to host error"
* checking service health:

```
$ flux exec -r 0 flux module list
Could not change dir to /g/g0/garlick: Permission denied. Going to /tmp instead
Module                       Size Digest  Idle  S Service
barrier                   1889344 ac8714d idle  R
job-exec                  2235160 b853f75 idle  R
heartbeat                 1868736 adf8b5d    0  R
job-list                  2454968 25556be idle  R
sched-fluxion-qmanager    6274336 0554708 idle  R sched
job-ingest                2163312 1b511fa idle  R
kvs                       2512784 8375ef8 idle  R
connector-local           1875000 67a36bc    0  R
cron                      1999320 93a60a7 idle  R
resource                  2334768 0c24437 idle  R
sched-fluxion-resource   31928072 6e11bb9 idle  R
job-info                  2371512 4d08f3d idle  R
kvs-watch                 2139704 d823c8f idle  R
content-sqlite            1901304 4248ee1 idle  R content-backing,kvs-checkpoint
job-manager               2594952 e085f6a   74  R
```

These seem OK:
```
$ flux ping -c 1 '0!job-ingest'
0!job-ingest.ping pad=0 seq=0 time=0.848 ms (baba9d09!986ad594!051c0114!c2163d34!23a510d2)
$ flux ping -c 1 'job-ingest'
job-ingest.ping pad=0 seq=0 time=0.342 ms (77d7ca79!986ad594!051c0114!9dee1e15)
$ flux ping -c 1 'job-manager'
job-manager.ping pad=0 seq=0 time=0.833 ms (eab90b22!986ad594!051c0114!c2163d34!58f460d8)
$ flux ping -c 1 'resource'
resource.ping pad=0 seq=0 time=0.404 ms (6fbacc97!986ad594!051c0114!78da2f6b)
```

Ah fluxion modules do not appear to be responding
```
$ flux ping sched-fluxion-qmanager
^C
$ flux ping sched-fluxion-resource
^C
$
```

dongahn:
The hang is understood:

[...]

PR #872, opened 2021-10-12
Fix an assortment of issues discovered during recent system instance testing

dongahn:
This PR adds an assortment of fixes that were discovered during our recent system instance testing on Fluke, which have since been fixed by @grondo, @garlick and @dongahn.

Add a proposed fix for the "all UP" bug where Fluxion reports all nodes as UP when a significant number of
nodes were still down.

Add @garlick's fix for fluxion to log the error text getting from resource.get-xml when it cannot retrieve all of the hwloc xml files during the run level 1 and stops immediately.

Cherry-pick @grondo's contributions to introduce some of the system instance testing coverage
into Fluxion and adapt them sightly in two ways: 1) add PYTHONPATH envVar to sharness test scripts to avoid a test startup failure; 2) add a simple case to test Fluxion does not hang when it can't get hwloc xml objects
from resource.get-xml.

Fixes Issue #867 and #866.

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -77,6 +77,7 @@ TESTS = \
     t3028-resource-grow.t \
     t3029-resource-prune.t \
     t3030-resource-multi.t \
+    t3300-system-dontblock.t \
     t3301-system-latestart.t \
     t4000-match-params.t \
     t4001-match-allocate.t \


grondo:
Hm, unexpected error from `make distcheck`:
```
2021-10-12T07:34:11.2916970Z ERROR: files left in build directory after distclean:
2021-10-12T07:34:11.2918174Z ./t/t3300-system-dontblock.output
2021-10-12T07:34:11.2919206Z ./t/t3301-system-latestart.output
```

I don't understand why it is only those two tests that leave `.output` files, but the `*.output` files should probably be removed by `clean-local`, e.g. from flux-core `t/Makefile.am`:
```Makefile
clean-local:
        rm -fr trash-directory.* test-results .prove *.broker.log */*.broker.log *.output python/__pycache__
```

dongahn:
> I'm often (but not always) getting a hang in t1017-rv1-bootstrap.t here:

This is unfortunately still an mystery. I used to have issues with those nested scheduling test cases on my docker on laptop. Since I increase the resources that the docker can use, I don't see those issue any longer. Are you running this on a local docker environment? What is the resources you gave to docker in terms of cpu count and memory?

grondo:
> Maybe related, but last night I noticed that "trash..." directories are created not only at t/ level but also at t/trash... level. I meant to ask you about this. I assume it is intended?

Yes, I think this occurs because the creates the top-level trash directory, changes to that as its working dir, then reinvokes the test after setting up some configuration for the system instance. The second invocation of the test creates trash directory and changes working to it...

dongahn:
Ok. I believe I addressed all the issues. @grondo: does this look good to you?

Issue #875, opened 2021-10-21
First match policy doesn't comply with RFC14 on min/max 


dongahn:
See 
- https://github.com/flux-framework/flux-sched/issues/833#issuecomment-948116340
- https://github.com/flux-framework/flux-sched/issues/833#issuecomment-948127112

dongahn:
PR #878 resolves this.

Issue #833, opened 2021-05-06
Add exclusive node scheduling policy for system-instance day 1


dongahn:
https://github.com/flux-framework/flux-core/discussions/3143

dongahn:
It looks like with `min : 1`, Fluxion only match and emit one core.

```yaml
  "resources": [
    {
      "type": "node",
      "count": 2,
      "with": [
        {
          "type": "slot",
          "count": 1,
          "with": [
            {
              "type": "core",
              "count": {
                  "min": 1,
                  "max": 100,
                  operator: "+",
                  operand: 1
              }
            }
          ],
          "label": "task"
        }
      ]
    }
  ]
```

```console
ahn1@docker-desktop:/usr/src/NODE_X$ ../resource/utilities/resource-query -L tiny.graphml -F pretty_simple
INFO: Loading a matcher: CA
resource-query> match allocate batch.yaml
      ---tiny0[1:shared]
      ------node3[1:shared]
      ---------core17[1:exclusive]
      ------node2[1:shared]
      ---------core17[1:exclusive]
INFO: =============================
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
INFO: =============================
```

dongahn:
Given the RFC 14 says the following, this is non-compliant. 

```
The default value for max SHALL be infinite, therefore a count which specifies only the min key SHALL be considered a request for at least that number of a resource, and the scheduler SHALL generate the R that contains the maximum number of the resource that is available and subject to the operator and operand. 
```

dongahn:
Ah... there seems to be a bug with the `first` match policy. With other 'global' policies like `low id first` and `high id first`, the fluxion is compliant. 

```ahn1@docker-desktop:/usr/src/NODE_X$ ../resource/utilities/resource-query -L tiny.graphml -F pretty_simple -P low
INFO: Loading a matcher: CA
resource-query> match allocate batch.yaml
      ---tiny0[1:shared]
      ------node1[1:shared]
      ---------core17[1:exclusive]
      ---------core16[1:exclusive]
      ---------core15[1:exclusive]
      ---------core14[1:exclusive]
      ---------core13[1:exclusive]
      ---------core12[1:exclusive]
      ---------core11[1:exclusive]
      ---------core10[1:exclusive]
      ---------core9[1:exclusive]
      ---------core8[1:exclusive]
      ---------core7[1:exclusive]
      ---------core6[1:exclusive]
[...]

PR #878, opened 2021-10-25
resource: beefing up min/max count-based matching

dongahn:
This PR beefs up our support for min/max count-based Jobspec. 

- Add min/max support for our relatively new first-match policy. (Unlike other global search based match policies, our first-match policy currently stops searching when the min count request is satisfied and this keeps this policy to support min/max requests fully.)
- Add more test cases for min/max Jobspecs.

Fixes Issue #875. This also supports Issue #833 as we are pursuing ways to use min/max based jobspec to support node-exclusive scheduling.


resource/traversers/dfu_impl.cpp
```diff
@@ -238,7 +238,7 @@ int dfu_impl_t::match (vtx_t u, const std::vector<Resource> &resources,
                 for (auto &c_resource : resource.with)
                     if (c_resource.type == "slot") {
                         *slot_resource = &c_resource;
-                        *nslots = c_resource.count.min;
+                        *nslots = m_match->calc_effective_max (c_resource);
```
milroy, line 241: Should `resource.count.min` be updated in `dfu_impl_t::is_enough` as well?

milroy (approved):
This PR LGTM.  If you have time @dongahn you can take a look at my question.  You may have a quick answer.

resource/traversers/dfu_impl.cpp
dongahn: Thank you for the question. `resource.count.min` is already considered from `calc_effective_max()` method so we should be good here. If `min` == `max`, min is used. Only when `min` < `max`, max is used here.

milroy: I didn't notice that `calc_effective_max ()` was called in `dfu_impl_t::is_enough`.  I must have been looking at the wrong version of the code.

Issue #876, opened 2021-10-21
Fluxion's Jobspec library cannot handle partially specified min/max


dongahn:
```json
  "resources": [
    {
      "type": "node",
      "count": 2,
      "with": [
        {
          "type": "slot",
          "count": 1,
          "with": [
            {
              "type": "core",
              "count": {
                  "min": 1
              }
            },
            {
              "type": "gpu",
              "count": {
                  "min": 1
              }
            }
          ],
          "label": "task"
        }
      ]
    }
  ],
```

```console
ahn1@docker-desktop:/usr/src/NODE_X$ ../resource/utilities/resource-query -L tiny.graphml -F pretty_simple -P high
INFO: Loading a matcher: CA
resource-query> match allocate batch2.yaml
ERROR: Jobspec error for 1: Key "max" missing from count
```

dongahn:
It also doesn't support `min: 0` spec:

```console
ahn1@docker-desktop:/usr/src/NODE_X$ ../resource/utilities/resource-query -L tiny.graphml -F pretty_simple -P high
INFO: Loading a matcher: CA
resource-query> match allocate batch.yaml
ERROR: Jobspec error for 1: "min" must be greater than zero
```

dongahn:
PR #879 resolves this.

PR #879, opened 2021-10-29
jobspec: make max/operator/operand optional

dongahn:
This PR adds a quick fix for Issue #876. This was written on top of PR #878 so don't merge it before PR #878. 

Problem: libjobspec within Fluxion currently requires all
four keys for a "moldable" jobspec to be present under
the resources.count: min, max, operator, and operand.
This does not comply with RFC14, as it specifies:

"The default value for max SHALL be infinite, therefore
a count which specifies only the min key SHALL be considered
a request for at least that number of a resource, and
the scheduler SHALL generate the R that contains the maximum
number of the resource that is available and subject to the
operator and operand. By contrast, if a fixed count is given
to the count key, the scheduler SHALL match any resource that
contains at least count of the resource, but its R SHALL
contain exactly count of the resource (potentially leaving
excess resources unutilized).

Handle the optional keys (max, operator, operand) appropriately
and adjust the relevant test cases.

I wanted to do a bit more work but found that there are some issues we need to fix in flux-core before having an end-to-end demonstration. So I created an Issue at https://github.com/flux-framework/flux-core/issues/3928 and decide to have this reviewed and landed into fluxion first.


resource/libjobspec/jobspec.cpp
```diff
@@ -64,43 +64,40 @@ void parse_yaml_count (Resource& res, const YAML::Node &cnode)
     if (!cnode["min"].IsScalar()) {
         throw parse_error (cnode["min"], "Value of \"min\" must be a scalar");
     }
-    if (!cnode["max"]) {
-        throw parse_error (cnode, "Key \"max\" missing from count");
-    }
-    if (!cnode["max"].IsScalar()) {
+    if (cnode["max"] && !cnode["max"].IsScalar()) {
```
milroy, line 67: Nits: you may want to fix the "interger" typo in line 48 and "existance" in line 60 above.

milroy (approved):
Nice work @dongahn! I found two residual typos, but the PR itself looks great.

resource/libjobspec/jobspec.cpp
dongahn: :-). Great catch, human valgrind! They must have been in there from day 1. I will do a fix up commit.

PR #881, opened 2021-11-04
build: drop ax_python_devel.m4

### LINKED ISSUE #880: configure fails to accept Python 3.10.0
I see flux deployment fail in a conda env with the Python 3.10.0:

```
...
checking for python3... /home/developer/ce_tutorial/bin/python3
checking for a version of Python >= '2.1.0'... yes
checking for a version of Python >='3.6'... no
configure: error: this package requires Python >='3.6'.
If you have it installed, but it isn't the default Python
interpreter in your system path, please pass the PYTHON_VERSION
variable to configure. See ``configure --help'' for reference.
```

```sh
$ /home/developer/ce_tutorial/bin/python3 -V
Python 3.10.0
```

It seems like `configure` has an incorrect lexical check in one place:

```py
>>> ver '3.6.9'
>>> ver >='3.6'
True
>>> ver = '3.10.0'
>>> ver >='3.6'
False
```

which presumably comes from  `config/ax_python_devel.m4`:

```m4
                ac_supports_python_ver=`$PYTHON -c "import sys; \
                        ver = sys.version.split ()[0]; \
                        print (ver >='3.6')"`
```

FWIW, `aclocal.m4` defines an `AM_PYTHON_CHECK_VERSION` macro which does the right thing (comparing hexversion instead of version string).

Happy to prepare a PR but am not sure what your preferred way of resolving this would be.  I would suggest to use the hexversion check also in `config/ax_python_devel.m4` - that would result in a very small diff.

garlick:
Problem: ax_python_devel.m4 is not used.

Since flux-sched doesn't use PYTHON_CLAGS or PYTHON_LIBS,
it would seem that there is no need for ax_python_devel.m4.
Furthermore, it has a bug that causes it to reject python
3.10.0 as too old.

Just drop ax_python_devel.m4 and adjust configure.ac
accordingly.

Fixes #880


dongahn:
You are so fast @garlick, I can't keep up :-) I was handling other stuff *sigh*

Thanks! This LGTM.

cmoussa1:
I can copy this solution over to flux-accounting and post a PR there too, thanks @garlick!

One tiny nit in the commit message (I could be wrong though):

> Since flux-sched doesn't use PYTHON_CLAGS or PYTHON_LIBS,
it would seem that there is no need for ax_python_devel.m4.
Furthermore, it has a bug that causes it to reject python
3.10.0 as too old.

`PYTHON_CLAGS` --> `PYTHON_CFLAGS`?

garlick:
Thanks Chris, I"ll fix.  I did just submit a PR to flux-accounting too so hopefully we didn't duplicate effort!

cmoussa1:
You just beat me to it! Thanks for posting it there too. :-)

dongahn:
>You just beat me to it! Thanks for posting it there too. :-)

Now you know how I feel @cmoussa1 :-) Outstanding teamwork.

garlick:
I like this team.  I step on people's toes and they thank me for it :-)

dongahn:
Lol

dongahn (approved):

PR #882, opened 2021-11-05
resource: improve effective max calc performance for operator="+"

dongahn:
This PR improves effective max calculation performance for operator="+" within a jobspec.

Problem: When the range between `count.min` and `count.max` is large for a jobspec, calculating the effective max can add too much performance overhead.

- Use a close form expression for operator='+'.
- Add some unit tests


dongahn:
@grondo or @milroy: can either of you please give a quick review to merge this for our delayed release?

grondo (approved):
This LGTM (as far as I was capable of reviewing it)

If I had one suggestion it would be to fill in the commit body for 075464ce4bcaf51b556e455005b2751dff6002b2 with a `Problem:` type statement for the benefit of future code spelunkers (e.g. how was the infinite loop triggered and how does the code address it?)

dongahn:
> If I had one suggestion it would be to fill in the commit body for 075464c with a Problem: type statement for the benefit of future code spelunkers (e.g. how was the infinite loop triggered and how does the code address it?)

Great point. Will do that.

PR #884, opened 2021-11-10
testsuite: use flux mini not flux jobspec

garlick:
Problem: flux jobspec command is going away, but flux-sched tests are still using it.

Convert to flux mini run --dry-run.


dongahn (approved):
LGTM. Thanks.

PR #887, opened 2021-11-11
t5000-valgrind.t: do not run valgrind test by default

grondo:
Problem: The t5000-valgrind.t test is susceptible to false positives
when run on new systems and different distros due to the potential
for new false positives from system libraries which have not yet been
added to the suppressions file. This can confuse new users when the
test unnecessarily fails.

Disable the valgrind test by default to avoid these situations.
The test can be run with FLUX_ENABLE_VALGRIND_TEST (e.g. in CI) or
by passing the `-d, --debug` flag (since this is common when running
the test by hand).

### DIFF
--- a/src/test/checks_run.sh
+++ b/src/test/checks_run.sh
@@ -156,6 +156,8 @@ export FLUX_TEST_MPI=t
 export FLUX_TESTS_LOGFILE=t
 export DISTCHECK_CONFIGURE_FLAGS="${ARGS}"
 
+# Force enable valgrind test
+export FLUX_ENABLE_VALGRIND_TEST=t
 
 if test "$CPPCHECK" = "t"; then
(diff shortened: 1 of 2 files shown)


garlick:
> This can confuse new users when the test unnecessarily fails

Like when spack throws you a random dependency permutation?  :-)

grondo:
Yeah, sorry, I should mention that we hit a failure of `t5000-valgrind.t` when building under spack with `gcc@11.2.0`. I just copied the commit message direct from flux-core, thus the general error above (which still applies here)

grondo:
Oh, I should also update the CI to set the requisite environment variable.

garlick (approved):
LGTM!

dongahn (approved):
LGTM. Set MWP.

PR #888, opened 2021-11-18
HLAPI: add CLI initialization and implement match_allocate, cancel, and info

milroy:
This PR adds new capability to the Fluxion C/C++ CLI resource HLAPI so that clients such as KubeFlux can initialize a Fluxion context (and maintain it in an external binding).  This PR implements the `match_allocate`, `cancel`, and `info` functions which KubeFlux needs first.  Other HLAPI functions can be implemented in straightforward analogy to the changes in this PR.

Data structures such as `resource_context_t` and `resource_params_t` are defined as in `resource-query` and provide a foundation for the `reapi_cli`.  Note that the implementation of `reapi_cli_info` changes the public API to add a mode argument that returns the job status.

Since the `resource_ctx_t` is created as a raw pointer (as distinct from `resource_query` where it's a `std::shared_ptr`) , a threaded client must be careful how it consumes the API.  

~~This PR is WIP since the most appropriate pointer type and how to perform CI testing is still to be determined.~~

### DIFF
--- a/qmanager/modules/Makefile.am
+++ b/qmanager/modules/Makefile.am
@@ -25,8 +25,6 @@ sched_fluxion_qmanager_la_SOURCES = \
     $(top_srcdir)/src/common/liboptmgr/optmgr_impl.hpp \
     $(top_srcdir)/resource/libjobspec/jobspec.hpp \
     $(top_srcdir)/resource/hlapi/bindings/c++/reapi.hpp \
-    $(top_srcdir)/resource/hlapi/bindings/c++/reapi_cli.hpp \
-    $(top_srcdir)/resource/hlapi/bindings/c++/reapi_cli_impl.hpp \
     $(top_srcdir)/resource/hlapi/bindings/c++/reapi_module.hpp \
     $(top_srcdir)/resource/hlapi/bindings/c++/reapi_module_impl.hpp \
     $(top_srcdir)/qmanager/config/queue_system_defaults.hpp \
--- a/resource/hlapi/bindings/c++/reapi.hpp
+++ b/resource/hlapi/bindings/c++/reapi.hpp
@@ -116,8 +116,8 @@ public:
      *  \return          0 on success; -1 on error.
      */
     static int match_allocate (void *h, bool orelse_reserve,
-                               const std::string &jobspec, const uint64_t jobid,
-                               bool &reserved,
+                               const std::string &jobspec,
+                               const uint64_t jobid, bool &reserved,
                                std::string &R, int64_t &at, double &ov)
     {
         return -1;
--- a/resource/hlapi/bindings/c++/reapi.hpp
+++ b/resource/hlapi/bindings/c++/reapi.hpp
@@ -191,6 +191,7 @@ public:
      *                   service module, it is expected to be a pointer
      *                   to a flux_t object.
      *  \param jobid     const jobid of the uint64_t type.
+     *  \param mode      return string containing the job state.
      *  \param reserved  Boolean into which to return true if this job has been
      *                   reserved instead of allocated.
      *  \param at        If allocated, 0 is returned; if reserved, actual time
@@ -201,7 +202,8 @@ public:
      *  \return          0 on success; -1 on error.
      */
     static int info (void *h, const uint64_t jobid,
-                     bool &reserved, int64_t &at, double &ov)
+                     std::string &mode, bool &reserved, int64_t &at, 
+                     double &ov)
     {
         return -1;
     }
--- a/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
@@ -18,7 +18,11 @@ extern "C" {
 #include <flux/core.h>
 }
 
+#include <stdexcept>
+#include <jansson.h>
+#include <boost/algorithm/string.hpp>
 #include "resource/hlapi/bindings/c++/reapi_cli.hpp"
+#include "resource/readers/resource_reader_factory.hpp"
 
 namespace Flux {
 namespace resource_model {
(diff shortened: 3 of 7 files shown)


dongahn:
@milroy: Great start! I know it is a WIP PR. Just let me know when you want me to give the first round of review.

resource/hlapi/bindings/c/reapi_cli.h
```diff
@@ -89,6 +98,7 @@ int reapi_cli_cancel (reapi_cli_ctx_t *ctx,
  *
  *  \param ctx       reapi_cli_ctx_t context object
  *  \param jobid     const jobid of the uint64_t type.
+ *  \param mode      string containing the job state.
```
cmisale, line 101: What is the meaning of `mode`?

milroy: It means the job state (`job_lifecycle_t`: INIT, ALLOCATED, RESERVED, CANCELED, ERROR).  I think this is important information for `info` to return.

dongahn:
> Since the resource_ctx_t is created as a raw pointer (as distinct from resource_query where it's a std::shared_ptr) , a threaded client must be careful how it consumes the API.

Yes. And passing the thread-safety responsibility to the caller is a great practice. Doing mutex within the API itself can open up a new can of worms.

qmanager/modules/Makefile.am
```diff
@@ -52,7 +52,7 @@ sched_fluxion_qmanager_la_CXXFLAGS = \
     $(FLUX_SCHEDUTIL_CFLAGS)
 
 sched_fluxion_qmanager_la_LIBADD = \
-    $(top_builddir)/resource/libjobspec/libjobspec_conv.la \
+    $(top_builddir)/resource/libresource.la \
```
dongahn, line 55: I believe it would be best to maintain the minimum dependency between qmanager and resource layer. 

My suggestion would be just to keep the original jobspec library depedency and drop the following dependencies from `sched_fluxion_qmanager_la_SOURCES`:

```
    $(top_srcdir)/resource/hlapi/bindings/c++/reapi_cli.hpp \
    $(top_srcdir)/resource/hlapi/bindings/c++/reapi_cli_impl.hpp \
```

Or you can even remove that and create a library dependency to `libreapi_module.la` here and see that still resolves.

dongahn (changes requested):
@milroy: Thank you for putting this together. 

Here is the very first set of my review. 

I do want to spend some time to think about the context passing part of your PR a bit more. My suspicion is there would be a simpler scheme to achieve the context passing than the current scheme. 

But I am doing this while I am in transit, I don't know when  I can get to that and at the same time.

Stay tuned.

qmanager/modules/Makefile.am
milroy: Unfortunately, there are dependencies on `reapi_cli_t` in ```queue_policy_factory_impl.hpp```, e.g.:
```C++
        if (policy == "fcfs") {
            if (reapi == "module")
                p = std::make_shared<queue_policy_fcfs_t<reapi_module_t>> ();
            else if (reapi == "cli")
                p = std::make_shared<queue_policy_fcfs_t<reapi_cli_t>> ();
        }
```

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -28,23 +28,32 @@ using namespace Flux::resource_model::detail;
...
+    ctx->resource_ctx = NULL;
+
 out:
     return ctx;
 }
 
 extern "C" void reapi_cli_destroy (reapi_cli_ctx_t *ctx)
 {
-    free (ctx);
+    delete ctx;
```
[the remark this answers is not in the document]
milroy: Sure thing!

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -28,23 +28,32 @@ using namespace Flux::resource_model::detail;
...
 extern "C" reapi_cli_ctx_t *reapi_cli_new ()
 {
     reapi_cli_ctx_t *ctx = NULL;
-    if (!(ctx = (reapi_cli_ctx_t *)malloc (sizeof (*ctx)))) {
+
+    try {
+        ctx = new reapi_cli_ctx_t;
+    }
+    catch (const std::bad_alloc &e) {
+        std::cerr << "Error allocating memory: " << e.what () << "\n";
```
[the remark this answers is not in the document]
milroy: Good suggestion.  I'll get to work on improving the error logging, and if it proves to be a blocker for this PR I will include it in a new one.

Do you think the public API should have a structured error message as part of the return or as an argument?

milroy:
@dongahn thanks for the helpful initial review during your travels!  I'll get to work on the items you identified and will wait for updates on the context passing.

resource/hlapi/bindings/c/reapi_cli.cpp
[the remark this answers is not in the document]
dongahn: I have been happy with the following style. If you meant an error catalogue with "A structured error message", I believe that's a bit overkill. Providing the client code with an error string right after some method would be good enough, I believe.

https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu.hpp#L42

https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu.hpp#L49

qmanager/modules/Makefile.am
dongahn: Ah... You can get rid of that for now. The intent is to created a CLI test for `qmanager` as well but we don't have to worry about this for now.

dongahn:
> @dongahn thanks for the helpful initial review during your travels! I'll get to work on the items you identified and will wait for updates on the context passing.

We will need a more in-person discussion for context passing. But here is my observation so far. 

1. I find the current `reapi_cli_t` implementation slightly breaks the intended separation of concerns of `reapi_t`. The first implementation of `reapi_t` was based on `sched-fluxion-resource` module service. This meant that its methods (such as `match_allocate`) can be called without needing to instantiate a class object and thus the methods are all class static. Now I see your cli implementation of this class goes beyond this model: it also allocates and initialize san object (`resource_context_t`). 

2. I believe you can address this by introducing a new class: `resource_query_t` and instantiating a proper class object using that first and then passing the object into your `reapi_cli_t` as the first argument. In addition to tighter separation of concerns, another advantage  would be  that you can now avoid using file-scoped static functions like [this](https://github.com/flux-framework/flux-sched/pull/888/commits/1b49b41564a5ffcb971dd892fe2c2a4b83bd2403#diff-9c9eea1269c68f630ee4f9e098625e6499501abf7e4a570c55db642897027326R32). They now can be private methods of this new class. 

3. It was there in the beginning, but with the introduction of an additional derived class, I am concerned about the use of `void *` `reapi_t` : [void *](https://github.com/flux-framework/flux-sched/blob/master/resource/hlapi/bindings/c%2B%2B/reapi.hpp#L118). The intention was to pass an external arbitrary handle (e.g., `flux_t *`) into each API implementation. But `void *` can quickly become unwieldy with multiple class derivations. For this round, I believe we should at least introduce an abstract class support to avoid the use of `void *`. For example,

```c++

class match_service_handle_t {

};
```

The module api user will extend it:

```c++

class sched_fluxion_resource_t : public match_service_handle_t {
public:
    flux_t *h;
};
```

The cli api user will extend it:

```c++

class resource_query_t : public match_service_handle_t {

};
```

resource/hlapi/bindings/c/reapi_cli.cpp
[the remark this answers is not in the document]
milroy: What I meant by "structured error message" is similar to what the GitHub API returns, e.g., 
```json
{"msg":"User quota exceeded","max_allowed":10000,"administrator_id":1234}
```
where the message indicates the exception, value provided, and allowed values.

I'll implement the method used by the traverser for this PR. Thanks!

milroy:
@dongahn: I've made good progress implementing your suggestions regarding the context passing.  There is a wrinkle with your third item:

> It was there in the beginning, but with the introduction of an additional derived class, I am concerned about the use of void * reapi_t : void *. The intention was to pass an external arbitrary handle (e.g., flux_t *) into each API implementation. But void * can quickly become unwieldy with multiple class derivations. For this round, I believe we should at least introduce an abstract class support to avoid the use of void *. For example,

Implementing `match_service_handle_t` requires a fair number of modifications to qmanager policies, as it passes `flux_t` handles around as `void*` (which need to be converted to `match_service_handle_t`s to be passed to `queue_policy` calls).  Here's a compilation error demonstrating this problem:
```c++
qmanager/policies/queue_policy_bf_base_impl.hpp:197:40: error: invalid conversion from ‘void*’ to ‘Flux::resource_model::detail::sched_fluxion_resource_t*’ 
```

Should I go ahead and modify the various functions in qmanager policies, or should I pursue a different approach?

milroy:
> It was there in the beginning, but with the introduction of an additional derived class, I am concerned about the use of void * reapi_t : void *. The intention was to pass an external arbitrary handle (e.g., flux_t *) into each API implementation. But void * can quickly become unwieldy with multiple class derivations. For this round, I believe we should at least introduce an abstract class support to avoid the use of void *. For example,

During the discussion with @dongahn today, we decided to delay implementing this item to a future PR.  It requires more thought and potentially restructuring some of qmanager.

I've made the requested updates to this PR; it's ready for another review @dongahn.

milroy:
I dropped the WIP title to indicate that the PR has been tested pretty thoroughly by me and @cmisale and should be ready for a final review.

dongahn:
Awesome! I will take a look.

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -27,24 +27,37 @@ using namespace Flux::resource_model;
...
+
 out:
     return ctx;
 }
 
 extern "C" void reapi_cli_destroy (reapi_cli_ctx_t *ctx)
 {
-    free (ctx);
+    int saved_errno = errno;
+    delete ctx;
```
dongahn, line 59: Don't we now delete `ctx->rqt` as well?

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -48,7 +55,70 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
                                  const uint64_t jobid, bool &reserved,
                                  std::string &R, int64_t &at, double &ov)
 {
-    return NOT_YET_IMPLEMENTED;
+    resource_context_t *resource_ctx = 
+                    *(static_cast<resource_context_t * *>(h));
```
dongahn, line 74: At first glance, this can be unsafe as you are relying on the actual memory layout. Can the caller actually pass the resource context instead of its containing struct pointer?

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -135,9 +150,35 @@ int reapi_cli_t::match_allocate_multi (void *h, bool orelse_reserve,
     return NOT_YET_IMPLEMENTED;
 }
 
-int reapi_cli_t::cancel (void *h, const int64_t jobid, bool noent_ok)
+int reapi_cli_t::cancel (void *h, const uint64_t jobid, bool noent_ok)
 {
-    return NOT_YET_IMPLEMENTED;
+    resource_context_t *resource_ctx = 
+                    *(static_cast<resource_context_t * *>(h));
```
dongahn, line 156: Same as my comment for `match_allocate`

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -36,6 +36,21 @@ static double get_elapse_time (timeval &st, timeval &et)
     return ts2 - ts1;
 }
 
+static int remove_job (resource_context_t * &resource_ctx, uint64_t jobid)
```
dongahn, line 39: Can this be a private member of the containing class for better isolation?

resource/hlapi/bindings/c++/reapi.hpp
```diff
@@ -201,7 +202,8 @@ class reapi_t {
      *  \return          0 on success; -1 on error.
      */
     static int info (void *h, const uint64_t jobid,
-                     bool &reserved, int64_t &at, double &ov)
+                     std::string &mode, bool &reserved, int64_t &at,
```
dongahn, line 205: `mode` will hold a return value? If so, perhaps the inline doc can have that described.

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -166,13 +166,31 @@ extern "C" int reapi_cli_cancel (reapi_cli_ctx_t *ctx,
...
+    if (!ctx || !ctx->rqt || !ctx->rqt->resource_ctx) {
         errno = EINVAL;
         return -1;
     }
-    return reapi_cli_t::info (ctx->h, jobid, *reserved, *at, *ov);
+    if ((rc = reapi_cli_t::info (ctx->rqt, jobid, mode_buf, 
+                                 *reserved, *at, *ov)) < 0)
+        goto out;
+    if ( !(mode_buf_c = strdup (mode_buf.c_str ()))) {
+        ctx->err_msg = __FUNCTION__;
```
dongahn, line 184: `errno = ENOMEM` FWIW.

dongahn (changes requested):
Looks pretty good. But I think I have found some minor issues. Some of the issues are hard to catch without actual test cases (which I know you and Abhik are planning as part of a future PR). But at least, could you run valgrrid memcheck since I think I have found bunch of memory leaks and we want to make sure we don't have any apparent memory problems before this gets merged in.

resource/hlapi/bindings/c/reapi_cli.h
```diff
@@ -135,6 +135,19 @@ int reapi_cli_set_handle (reapi_cli_ctx_t *ctx, void *handle);
  */
 void *reapi_cli_get_handle (reapi_cli_ctx_t *ctx);
 
+/*! Get the reapi cli error message.
+ *
+ *  \param ctx       reapi_cli_ctx_t context object
+ *  \return          string containing the error message
+ */
+char *reapi_cli_get_err_msg (reapi_cli_ctx_t *ctx);
```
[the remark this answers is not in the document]
milroy: Good catch!

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -89,6 +92,375 @@ void reapi_cli_t::clear_err_message ()
...
+    }
+
+    // Override defaults if present in options argument
+    if ( (tmp_json = json_object_get (opt_json, "load_format"))) {
+        this->resource_ctx->params.load_format = json_string_value (tmp_json);
+        if (!this->resource_ctx->params.load_format.c_str ()) { 
+            errno = EINVAL;
+            c_err_msg += __FUNCTION__;
+            c_err_msg += ": Error loading load_format\n";
+            json_decref (tmp_json);
```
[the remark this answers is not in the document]
milroy: Thanks for catching this!

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -89,6 +92,375 @@ void reapi_cli_t::clear_err_message ()
...
+            errno = EINVAL;
+            c_err_msg += __FUNCTION__;
+            c_err_msg += ": Error loading prune_filters\n";
+            json_decref (tmp_json);
+            json_decref (opt_json);
+            goto out;
+        }
+    }
+    if ( (tmp_json = json_object_get (opt_json, "reserve_vtx_vec")))
+        // No need for check here; returns 0 on failure
```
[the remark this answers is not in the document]
milroy: `josn_integer_value` returns the integer value, or 0 if `tmp_json` is not a JSON integer: https://jansson.readthedocs.io/en/2.11/apiref.html#c.json_integer_value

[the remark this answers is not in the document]
milroy: of course, 0 can also be a valid JSON integer value...

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -89,6 +92,375 @@ void reapi_cli_t::clear_err_message ()
...
+        goto out;
+    }
+
+    try {
+        this->resource_ctx->db = std::make_shared<resource_graph_db_t> ();
+    } catch (std::bad_alloc &e) {
+        c_err_msg += __FUNCTION__;
+        c_err_msg += ": ERROR: can't allocate memory for resource_ctx db\n";
+        errno = ENOMEM;
+        this->resource_ctx = nullptr;
```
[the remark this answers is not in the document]
milroy: The intention behind this was to catch the `nullptr` after the failed initialization:  https://github.com/flux-framework/flux-sched/blob/d23bd11f16f9939e4ecc69911066ee7cbec40dbb/resource/hlapi/bindings/c/reapi_cli.cpp#L78
I can add a `delete` if the condition is caught. Then is probably no reason for these checks: https://github.com/flux-framework/flux-sched/blob/d23bd11f16f9939e4ecc69911066ee7cbec40dbb/resource/hlapi/bindings/c/reapi_cli.cpp#L100

Another approach would be to delete `resource_ctx` immediately (i.e., at lines like 357 above) and throw an exception that will be caught by the caller.  I'm not sure which is the better approach. Do you have a preference @dongahn?

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -27,24 +27,37 @@ using namespace Flux::resource_model;
...
-    reapi_cli_ctx_t *ctx = NULL;
-    if (!(ctx = (reapi_cli_ctx_t *)malloc (sizeof (*ctx)))) {
+    reapi_cli_ctx_t *ctx = nullptr;
+
+    try {
+        ctx = new reapi_cli_ctx_t;
+    }
+    catch (const std::bad_alloc &e) {
+        ctx->err_msg = __FUNCTION__;
+        ctx->err_msg += "Error allocating memory: " + std::string (e.what ())
```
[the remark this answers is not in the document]
milroy: Thanks for catching this!

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
milroy: Won't this require reworking the functions so they don't take `void *` as the first argument? That appears to get us back to another version of this discussion above: https://github.com/flux-framework/flux-sched/pull/888#issuecomment-989595644

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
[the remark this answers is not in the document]
dongahn: You may be familiar with RAII: https://en.cppreference.com/w/cpp/language/raii

My preference is an object gets all the resources in the constructor orelse an error. That typically leads to a much better design when the object is composed with other subsystems like STL containers.

If an exception raising class isn't working for this case, I would treat the resource allocation needed for both the parent (`resource_ctx`) and a child (`resource_ctx->db`) a transaction.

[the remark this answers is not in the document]
dongahn: Is this (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#e6-use-raii-to-prevent-leaks) useful at all?

resource/hlapi/bindings/c++/reapi.hpp
milroy: Yes, I'll add a description.

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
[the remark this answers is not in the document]
milroy: I was familiar with RAII and was considering how to implement it correctly here.  The link you provided is very helpful; it looks like throwing errors in the `resource_query_t` constructor and using a `std::unique_ptr` for `resource_ctx` is a good approach.  Thanks!

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
dongahn: Well, I guess my question is if it is safe to pass `ctx->rqt` (which is `resource_query_t *`) to `void *` and cast it back to `resource_context_t *`.  

```suggestion
    resource_query_t *query = static_cast<resource_query_t *> (h);
```

And then directly use `query` as a handle to get to other state?  

As is, I can't follow the code sorry.

milroy: Oh, I understand what you mean now.  

Yes, I confess I never understood `*(static_cast<resource_context_t * *>(h))`, but we somehow got that to work in an hour-long hackathon.

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
milroy: Yes, that will be much cleaner.

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -149,6 +149,21 @@ extern "C" void *reapi_cli_get_handle (reapi_cli_ctx_t *ctx)
     return ctx->h;
 }
 
+extern "C" char *reapi_cli_get_err_msg (reapi_cli_ctx_t *ctx)
+{
+    std::string err_buf = "";
+
+    err_buf = ctx->err_msg;
+
+    return strdup (err_buf.c_str ());
```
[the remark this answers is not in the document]
milroy: It turns out that if I don't use `strdup` the returned string is `""`.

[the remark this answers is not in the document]
dongahn: Hmmm, is that still the case when you have `const char *` as the return type of the function?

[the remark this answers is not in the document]
milroy: Yes, it's still the case when the return type is `const char *`.

resource/hlapi/bindings/c/reapi_cli.cpp
```diff
@@ -149,6 +149,21 @@ extern "C" void *reapi_cli_get_handle (reapi_cli_ctx_t *ctx)
...
+}
+
+extern "C" void reapi_cli_clear_err_msg (reapi_cli_ctx_t *ctx)
+{
+    reapi_cli_t::clear_err_message ();
+    ctx->err_msg = "";
```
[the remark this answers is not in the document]
milroy: There are some events in the C API that can lead to errors that won't be captured if we don't have an `err_msg` for the C API, though.  E.g.: https://github.com/flux-framework/flux-sched/blob/b6f1f41a67b4a0bbcc42ec7f9ef675e632aea73c/resource/hlapi/bindings/c/reapi_cli.cpp#L79
and the following code.

milroy:
@dongahn thank you for the thorough review of the code and your helpful edits and suggestions!

Based on our discussions, I've made what I think to be substantial improvements to the PR.  In particular, I've collected everything `resource_ctx`-related into the `resource_query_t` class since that seemed to be an unnecessary and complicating distinction.  Together with limiting access to private class variables through getters and setters this makes the class easier to initialize and use.

Apart from a couple of questions related to your last review, this PR is ready for another look @dongahn.

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -89,6 +93,423 @@ void reapi_cli_t::clear_err_message ()
...
+void resource_query_t::set_job (const uint64_t jobid,
+                                const std::shared_ptr<job_info_t> &job)
+{
+    jobs[jobid] = job;
+}
+
+int resource_query_t::remove_job (const uint64_t jobid)
+{
+    int rc = -1;
+    if ((rc = traverser->remove ((int64_t)jobid)) == 0) {
```
dongahn, line 596: Minor issue: traverser class uses `int64_t` for jobid and for a very rare case, this conversion can be a problem. I'd suggest:

```suggestion
    if (jobid > std::numeric_limits<int64_t>::max) {
        errno = EOVERFLOW;
        return rc;
    }
    if ((rc = traverser->remove (static_cast<int64_t>(jobid)) == 0) {

**Review by dongahn — APPROVED** [review]
@milroy: This looks really good! Thank you for this importatn PR. I just added a comment for one minor issue about a jobid type conversion. Please put MWP when you are ready with this PR.

File: resource/hlapi/bindings/c++/reapi_cli_impl.hpp (hunk 18 above)
Comment (milroy) [reply resource/hlapi/bindings/c++/reapi_cli_impl.hpp:596]: Great catch, thanks!

Comment (milroy) [reply resource/hlapi/bindings/c++/reapi_cli_impl.hpp:596]: I'm going to leave this line as is:
```C++
    if ((rc = traverser->remove ((int64_t)jobid)) == 0) {
```
since changing it to:
```C++
    if ((rc = traverser->remove (static_cast<int64_t>(jobid)) == 0) {
```
results in a compilation error:
```sh
error: init-statement in selection statements only available with ‘-std=c++17’ or ‘-std=gnu++17
```

dongahn: Ugg. Alternatively, you can

```c++
rc = traverser->remove (static_cast<int64_t>(jobid);
if ( rc == 0) {
```

milroy: Ok, it's fixed if I pull the `rc` assignment out of the `if` statement.

milroy:
> @milroy: This looks really good! Thank you for this importatn PR. I just added a comment for one minor issue about a jobid type conversion. Please put MWP when you are ready with this PR.

Thanks a lot @dongahn for the detailed reviews and the help restructuring the PR!  I made the final change you suggested and will set MWP.

PR #890, opened 2021-11-19
testsuite: drop flux overlay status from test

garlick:
Problem: the flux overlay status interface is changing in
flux core but flux sched is still using the old one
in t3301-system-latestart.t.

Drop the calls to flux overlay status from this test so it
can work with flux-core both before and after the change
proposed in flux-framework/flux-core#3974.

The test is no worse for the wear.


grondo (approved):
LGTM! Thanks!

garlick:
I'm going to set MWP so we can progress the other issue.  Hope that is ok with the fluxioneers!

Issue #877, opened 2021-10-22
flux-sched: minor changes required for clang-12 and gcc-11


grondo:
From my PR to update spack packages I got the following feedback: See https://github.com/spack/spack/pull/26840#issuecomment-949995031

> flux-sched needed a few changes to build with clang and gcc@11. I pushed this commit to this PR, ok for you?
> flux-sched: Fix build with clang and gcc-11
>
> clang doesn't have `-Wno-maybe-uninitialized`:
> use `-Wno-uninitialized` instead
> clang@12: needs `-Wno-defaulted-function-deleted`
> gcc@11: and clang@13 need changes which are good in general:
>   - resource/schema/sched_data.hpp: Need to use `nullptr` instead of `NULL`
>   - resource/planner/planner.h: Need to use `std::size_t`, not `size_t`

as reported by @bernhardkaindl

grondo:
Not fixing this is causing issues getting spack packages accepted, we should try to get these simple changes into the next release perhaps.

dongahn:
Ah... I will put this into our Nov milestone target. Should be straightforward to fix (though I haven't looked into details just yet).

dongahn:
> resource/planner/planner.h: Need to use std::size_t, not size_t

Hmmmm. This doesn't seem to make sense since `planner.h` should be C.

bernhardkaindl:
No, it should be C++:

It's a header file, only included from C++.

Just in case: The extension `.h` doesn't make the C++ to become a C compiler magically.
(BTW, there is only one .c file in flux-sched.)

Therefore it should be named `planner.hpp` and it should `#include <stdef>`, see below.

I'm guessing that you might have one or two computers around which will happily confirm it when you give them the job to verify:

This would even build a gcc-11.2.0 and then build the flux-sched master branch internal to spack, so can use the latest GNU or LLVM compilers without any side effects on your systems:
```
bin/spack install -v flux-sched@master %gcc@11.2.0

     426    In file included from /mnt/data/.bk/spack/.var/stage/kaindlb/spack-stage-flux-sched-0.19.0-wg6nd7xahjc5adcj4rdxeimmhabjvuyn/spack-src/resource/schema/sched_data.hpp:16,
     427                     from /mnt/data/.bk/spack/.var/stage/kaindlb/spack-stage-flux-sched-0.19.0-wg6nd7xahjc5adcj4rdxeimmhabjvuyn/spack-src/resource/schema/sched_data.cpp:11:
[...]

PR #891, opened 2021-11-29
Fix compilation issues with clang-12 and gcc-11

dongahn:
This PR adds a few compilation issues that were reported against clang-12, clang-13 and gcc-11.

- Use `nullptr` instead of `NULL` within `resource/schema/sched_data.hpp` (Issue with gcc-11 and clang-13);
- Include `stddef.h` to pick up `size_t` for `resource/planner/planner.h` (Issue with gcc-11 and clang-13);
- Drop `const` from one of the class members to use defaulted constructors within `qmanager/modules/qmanager_opts.hpp` (Issue with clang-12 and clang-13).

Fixes Issue #877

### DIFF
--- a/qmanager/modules/qmanager_opts.hpp
+++ b/qmanager/modules/qmanager_opts.hpp
@@ -124,7 +124,7 @@ public:
 private:
     int parse_queues (const std::string &queues);
 
-    const std::string m_default_queue_name = "default";
+    std::string m_default_queue_name = "default";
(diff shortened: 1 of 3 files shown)


bernhardkaindl (approved):

grondo (approved):

dongahn:
Thanks @bernhardkaindl and @grondo!

PR #902, opened 2022-01-29
Replace DBL_MAX with std::numeric_limits

jedbrown:
Fixed compilation error because <cfloat> was not included.

### DIFF
--- a/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
@@ -318,7 +318,7 @@ int resource_query_t::set_resource_ctx_params (const std::string &options)
     json_t *tmp_json = NULL, *opt_json = NULL;
 
     // Set default values
-    perf.min = DBL_MAX;
+    perf.min = std::numeric_limits<double>::max();
     perf.max = 0.0f;
     perf.accum = 0.0f;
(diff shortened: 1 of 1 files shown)


dongahn:
Thanks. Yes, this is the right fix since this is C++. Could you add the build environment where you saw this? This will be useful to augment our CI variants later. Thanks. Merging.

jedbrown:
gcc-11, glibc-2.33

dongahn:
Are you using a specific linux distro as well?

jedbrown:
Arch Linux, but it's vanilla so I'd expect any system with that gcc/glibc would behave the same. This isn't the first time this breakage had been encountered.
https://github.com/flux-framework/flux-sched/commit/18c39747386ebaf52b1191c4b67532f494096ba6

jedbrown:
Arch Linux, but it's vanilla so I'd expect any system with that gcc/glibc would behave the same. This isn't the first time this breakage had been encountered.
https://github.com/flux-framework/flux-sched/commit/18c39747386ebaf52b1191c4b67532f494096ba6

grondo:
CI seems to have gotten stuck, so I'm going to force a rebase
@mergifyio rebase

grondo:
Seems like @dongahn you meant to approve and set merge-when-passing label?

grondo (approved):

Issue #912, opened 2022-03-02
incorrect core counts from 'flux ion-R encode' for heterogeneous cluster


ryanday36:
In some cases, a cluster may have, for example, login nodes with fewer cores than compute nodes. Consider a cluster 'foo' with a management node 'foo2' with 2 cores, a login node 'foo3' with 1 core, and two compute nodes foo[1,4] with 2 cores that is encoded as:

```
[day36@corona81:flux]$ (flux R encode -r 0 -c 0-1 -H foo2 && flux R encode -r 1 -c 0 -H foo3 && flux R encode -r 2-3 -c 0-1 -H foo[1,4]) | flux R append | flux ion-R encode
[...]

PR #916, opened 2022-03-03
flux-ion-R encodes the nodelist key in strict rank order

dongahn:
This PR fixes Issue #912. `flux-ion-R` misinterprets the `nodelist` key specified
in RFC20 and does not encode node names in strict flux rank order. This can easily
lead to an incorrect mapping between ranks and hostnames.

- Encode node names in strict rank order for `nodelist`.
- Add a test case to ensure this

### DIFF
--- a/src/python/fluxion/resourcegraph/V1.py
+++ b/src/python/fluxion/resourcegraph/V1.py
@@ -141,14 +141,16 @@ class FluxionResourceGraphV1(Graph):
         edg = FluxionResourceRelationshipV1(ppid, vtx.get_id())
         self._add_and_tick_uniq_id(vtx, edg)
 
-    def _encode_rank(self, ppid, rank, children, hList, hIndex):
-        hPath = f"/cluster0/{hList[hIndex]}"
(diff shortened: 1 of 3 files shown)


grondo:
I ran this branch in a test instance on fluke and it does seem to be working AFAICS! Nice quick fixes @dongahn!

Now that this issue is fixed, it uncovered an issue in the rank assignment of the flux shell for jobs with heterogeneous allocations (different numbers of cores per rank). So that will have to get fixed on the flux-core side. Once I have a fix for _that_ issue, I'll test it together with this PR and ensure there are no remaining issues.

One suggestion: Update the PR title for release notes purposes, since this PR now fixes both `flux ion-R` and the RV1 writer.

dongahn:
Great! Thank you for quick test @grondo. Learning big time on heterogeneity (our future) -- feeling good that flux/fluxion proper essentially can easily be extensible! If I have some more time, I'd like to work on gpu type heterogeneity as we discussed before. Not sure if I can find time for this release cycle, though. (Probably next release)

dongahn:
Adding MWP.

PR #918, opened 2022-03-09
resource: support discovery of AMD RSMI GPUs in hwloc reader

grondo:
Some AMD ROCm SMI GPUs do not appear as type COPROC devices. Instead they are discovered as GPU OSDev type devices with name `rsmiX`. This PR extends the resource hwloc reader to allow these types of devices to be discovered as gpus.

This allows the sched-fluxion-resource module to discover GPUs on the corona system.

### DIFF
--- a/resource/readers/resource_reader_hwloc.cpp
+++ b/resource/readers/resource_reader_hwloc.cpp
@@ -291,9 +291,16 @@ int resource_reader_hwloc_t::walk_hwloc (resource_graph_t &g,
     }
     case HWLOC_OBJ_OS_DEVICE: {
         supported_resource = false;
-        if (!obj->attr) {
+        if (!obj->attr)
             break;
-        } else if (obj->attr->osdev.type == HWLOC_OBJ_OSDEV_COPROC) {
+
+        /*  Detect any COPROC as GPU device.
+         *  Also, some AMD GPUs are detected as type == GPU with
+         *   subtype RSMI (rsmiX), detect those also:
+         */
+        if (obj->attr->osdev.type == HWLOC_OBJ_OSDEV_COPROC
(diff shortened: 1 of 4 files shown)


grondo:
FYI - before this patch:
```console
$ flux start flux resource list -v
     STATE NNODES   NCORES    NGPUS LIST
      free      1       48        0 rank0/core[0-47]
      down      0        0        0 
```

After:
```console
$ FLUX_MODULE_PATH_PREPEND=$(pwd)/qmanager/modules/.libs:$(pwd)/resource/modules/.libs flux start flux resource list -v
     STATE NNODES   NCORES    NGPUS LIST
      free      1       48        8 rank0/core[0-47],gpu[0-7]
      down      0        0        0 
```

dongahn:
Great. I will take a look at it soonish.

Is LIST field new with recent versions? Off topic, but could you tell me where is the code which create the LIST output from R? That can be reused to improve fluxion logging.

grondo:
> Is LIST field new with recent versions? Off topic, but could you tell me where is the code which create the LIST output from R? That can be reused to improve fluxion logging.

The LIST field is not new, it is the way that librlist emits its internal representation in a compact manner.

The code is in librlist [here](https://github.com/flux-framework/flux-core/blob/master/src/common/librlist/rlist.c#L1099), and is unfortunately not exported, though you can generate it from _R_ with `flux R decode --short`, or via Python bindings via the `ResourceSet` `dumps()` method.

dongahn (approved):
LGTM and I approve it.

dongahn:
@grondo: let me know if you want me to add MWP.

grondo:
Sure thanks!

Issue #828, opened 2021-05-01
fake R fails grow_resource_db_jgf


dongahn:
The following bug was discovered while testing https://github.com/flux-framework/flux-core/issues/3625#issuecomment-830183266
```
flux R encode -r '0-1' -c '0-64' | flux ion-R encode > fake.R.json
flux start
./R-feeder
flux module remove sched-fluxion-resource
flux module reload resource noverify
flux module load sched-fluxion-resource load-allowlist=node,core,gpu policy=first
2021-05-01T02:05:48.963973Z sched-fluxion-resource.err[0]: grow_resource_db_jgf: db.load: error inserting an edge to outedge metadata map: cluster0 -> quartz15; : Cannot allocate memory
2021-05-01T02:05:48.964257Z sched-fluxion-resource.err[0]: update_resource_db: grow_resource_db: Cannot allocate memory
2021-05-01T02:05:48.964297Z sched-fluxion-resource.err[0]: update_resource: update_resource_db: Cannot allocate memory
2021-05-01T02:05:48.964562Z sched-fluxion-resource.err[0]: populate_resource_db_acquire: update_resource: Cannot allocate memory
2021-05-01T02:05:48.964611Z sched-fluxion-resource.err[0]: populate_resource_db: loading resources using resource.acquire
```

dongahn:
It turned out that the identical hostname in the fake R is a problem. 

As part of [graph editing for elasticity](https://github.com/flux-framework/flux-sched/blob/master/resource/readers/resource_reader_jgf.cpp#L478), we only add a new vertex when its path does not exist in the graph. So, with two identical hostname, the second vertex is evaluated as "existence" and that causes the growth failure. 

Secondly, we have been using `ENOMEM` for map insert failures. But it seems it is an existent key is a more common problem that leads to a failure. I will change that to `EEXIST`.

dongahn:
Using `-H` option for `flux R encode` should work around the issue. 

```
ahn1@docker-desktop:/usr/src$ flux R encode -r '0-1' -c '0-64' -H host[1-2]
{"version": 1, "execution": {"R_lite": [{"rank": "0-1", "children": {"core": "0-64"}}], "starttime": 0.0, "expiration": 0.0, "nodelist": ["host[1-2]"]}}
```

dongahn:
@milroy: could you comment whether this is worthwhile to fix?

dongahn:
[...]

PR #924, opened 2022-04-02
Don't assume each hierarchical path is unique

dongahn:
Problem: We currently assume that there exits one and only one unique
graph vertex that corresponds to a path. But in various emulation
modes where we run multiple brokers on a single node, this unique path
assumption does not hold: having more than one node vertex with
an identical hostname breaks this assumption. As a result, our reader
code balks at loading our resource graph data store under certain
conditions. The JGF reader will fail to initialize with:
"grow_resource_db_jgf: db.load: error inserting an edge to outedge
metadata map"

Add support so that we can drop the unique path assumption.

- Modify `by_path` graph metadata index to be a map of vector of vertices
  instead of a map of vertices. Since this map is keyed by paths, one 
  path can have multiple vertices corresponding to it. 
- Adjust various readers to use the modified data structure to break
  their unique path assumption.
- Modify both `sched-fluxion-resource` module and `resource-query` to use 
  the new structure to break the assumption.
- Change the format of returned data for both `sched-fluxion-resource`
  and `resource-query` on path-based queries such as get-property.
- Adjust the affected test cases.

Fixes Issue #828

### DIFF
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -455,15 +455,20 @@ done:
     return v;
 }
 
-vtx_t resource_reader_jgf_t::vtx_in_graph (const resource_graph_metadata_t &m,
+vtx_t resource_reader_jgf_t::vtx_in_graph (const resource_graph_t &g,
+                                           const resource_graph_metadata_t &m,
                                            const std::map<std::string, 
-                                                          std::string> &paths)
+                                                          std::string> &paths,
+                                           int rank)
 {
-    std::map<std::string, vtx_t>::const_iterator bp_it;
     for (auto const &paths_it : paths) {
-        bp_it = m.by_path.find (paths_it.second);
-        if (bp_it != m.by_path.end ()) {
-            return bp_it->second;
+        auto iter = m.by_path.find (paths_it.second);
+        if (iter != m.by_path.end ()) {
+            for (auto &v : iter->second) {
+                if (g[v].rank == rank) {
+                    return v;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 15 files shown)


dongahn:
@milroy: I don't know how easy it is for you to review this and/or you have time (given the current SI push). But this is a good to have PR for our next release.

grondo (approved):
@dongahn, I could not do a technical review of the code per se, but I did run through some testing with this PR branch and didn't find any issues, so I'll set my approval.

dongahn:
@grondo: sounds good. I know @milroy is extremely busy with multiple priorities. Your approval should be good enough to merge it in.

dongahn:
I will wait until COB today before putting MWP though.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -599,23 +630,16 @@ int resource_reader_jgf_t::find_vtx (resource_graph_t &g,
...
-                errno = EINVAL;
-                m_err_msg += __FUNCTION__;
-                m_err_msg += ": inconsistent input vertex for " + kv.second;
-                m_err_msg += " (id=" + std::string (fetcher.vertex_id) + ").\n";
-                m_err_msg += std::to_string (u) + " != " + std::to_string (v) + ".\n";
-                goto done;
-            }
+        if (exist (g, m, kv.second, fetcher.rank, fetcher.vertex_id, u) < 0)
+            goto done;
+        if (v == nullvtx) {
```
milroy, line 635: This line wasn't introduced as part of this PR, but it looks like this condition will always be satisfied given ```v = nullvtx;``` above.  Consider dropping the `if` check.

milroy (approved):
This PR looks good.  Basically, the unique identifier is now a `(path, rank)` pair.

I did add a comment about a check I think you can drop, but it isn't necessary to merge this PR.

resource/readers/resource_reader_jgf.cpp
dongahn: Thanks @milroy. This is actually a necessary check because it is in a loop. Basically this code block finds a graph node that matches with the path for each and every subsystem. So if `v == nullvtx`, we assign the matching u to `v`. And then for other subsystem paths, we take the `else` branch and report an inconsistency error if `v != u`. Hope this makes sense.

PR #926, opened 2022-04-03
testsuite: disable hwloc reader tests

### LINKED ISSUE #925: hwloc reader uses resource.get-xml RPC which is scheduled for removal
flux-framework/flux-core#4263 removes the `resource.get-xml` RPC since fluxion now has the rv1 reader, but this causes some tests to fail:
```
2022-04-02T23:16:28.4623930Z FAIL: t1016-nest-namespace.t 3 - namespace: gpu id remapping works with hwloc (pol=hi)
2022-04-02T23:16:28.4624445Z FAIL: t1016-nest-namespace.t 4 - namespace: parent CUDA_VISIBLE_DEVICES has no effect
2022-04-02T23:16:28.4625912Z FAIL: t1016-nest-namespace.t 7 - namespace: gpu id remapping works with hwloc (pol=low)
2022-04-02T23:16:28.4626401Z FAIL: t1016-nest-namespace.t 8 - namespace: parent CUDA_VISIBLE_DEVICES has no effect
2022-04-02T23:16:28.4626873Z FAIL: t1016-nest-namespace.t 9 - namespace: removing resource and qmanager modules
```

@dongahn could you advise what should happen with those tests?  I think it is exercising the hwloc reader, which depends on the removed RPC [here](https://github.com/flux-framework/flux-sched/blob/d4118327916d146e7b3b4111b9b7ca7de9e73901/resource/modules/resource_match.cpp#L823).

I'll propose a PR that disables them for now and someone can follw up with whatever the proper cleanup is (maybe remove the hwloc reader code?)

garlick:
Problem: resource.get-xml RPC is being deprecated in flux-core,
but the fluxion hwloc reader still depends on it.

Disable sharness tests that fail without this RPC as a stopgap
measure, so flux-core can proceed with removal.

Fixes #925

### DIFF
--- a/t/t1016-nest-namespace.t
+++ b/t/t1016-nest-namespace.t
@@ -16,12 +16,12 @@ test_expect_success 'namespace: load test resources' '
     load_test_resources ${excl_1N1B}
 '
 
-test_expect_success 'namespace: loading resource and qmanager modules works' '
+test_expect_success HAVE_GETXML 'namespace: loading resource and qmanager modules works' '
     load_resource load-allowlist=cluster,node,gpu,core policy=high &&
     load_qmanager
 '
 
-test_expect_success 'namespace: gpu id remapping works with hwloc (pol=hi)' '
+test_expect_success HAVE_GETXML 'namespace: gpu id remapping works with hwloc (pol=hi)' '
     cat >nest.sh <<-EOF &&
 	#!/bin/sh
 	flux module load sched-fluxion-resource load-format=hwloc load-allowlist=cluster,node,gpu,core policy=high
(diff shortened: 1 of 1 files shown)


dongahn:
Sounds reasonable and I approve so feel free to put MWP.

So this RPC has been completely deprecated even for a singe user instance? 

Could you open up a ticket about this, though? This is an import test case for real users behind it and we probably want to port this to rv1exec in a near future.

garlick:
> So this RPC has been completely deprecated even for a singe user instance?

Yes.

> Could you open up a ticket about this, though? This is an import test case for real users behind it and we probably want to port this to rv1exec in a near future.

Sure, will do.  I _think_ these tests are repeated lower down against the rv1 reader but I could be mistaken.

PR #933, opened 2022-04-20
Fix an integer overflow bug etc.

### LINKED ISSUE #932: Successful match is policy dependent
During KubeFlux testing, I discovered that for a given jobspec `lonode` and `hinode` policies do not return a successful match, while `low` and `high` do.  Furthermore, `hinode` and `lonode` match behavior is inconsistent with expectation.

Here's the zipped JGF from a 16 `hpc6a` node AWS EKS cluster: [cluster.json.zip](https://github.com/flux-framework/flux-sched/files/8499108/cluster.json.zip), and the test jobspec that requests all cluster cores (`test.yaml`): 
```YAML
version: 9999
resources:
- type: slot
  count: 16
  label: default
  with:
  - type: core
    count: 94
attributes:
  system:
    duration: 3600
tasks:
- command: [""]
  slot: default
  count:
    per_slot: 1
```

For example, with `resource-query` (latest flux-sched master):
```bash
resource/utilities$ ./resource-query -p ALL:core -F pretty_simple -S CA -f jgf -P high -L cluster.json
INFO: Loading a matcher: CA
resource-query> match allocate test.yaml 
      ---k8scluster0[1:shared]
      ------rack0[1:shared]
      ---------ip-192-168-62-188.us-east-2.compute.internal6647[1:shared]
      ------------socket0[1:shared]
      ---------------core93[1:exclusive]
...
      ---------------core2[1:exclusive]
      ---------------core1[1:exclusive]
      ---------------core0[1:exclusive]
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
resource-query>
```
However, with `hinode` and `lonode` (tested as well) the jobspec isn't matched:
```bash
$ ./resource-query -p ALL:core -F pretty_simple -S CA -f jgf -P hinode -L cluster.json
INFO: Loading a matcher: CA
resource-query> match allocate test.yaml 
INFO: No matching resources found
INFO: JOBID=1
resource-query> 
```

As a another wrinkle, a jobspec that requests 1 slot is matched 11 times successfully with `hinode`:
```bash
$ ./resource-query -p ALL:core -F pretty_simple -S CA -f jgf -P hinode -L cluster.json
INFO: Loading a matcher: CA
resource-query> match allocate test2.yaml
      ---k8scluster0[1:shared]
      ------rack0[1:shared]
      ---------ip-192-168-32-123.us-east-2.compute.internal2[1:shared]
      ------------socket0[1:shared]
      ---------------core93[1:exclusive]
      ---------------core92[1:exclusive]
...
      ---------------core2[1:exclusive]
      ---------------core1[1:exclusive]
      ---------------core0[1:exclusive]
INFO: JOBID=11
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
resource-query> match allocate test2.yaml
INFO: No matching resources found
INFO: JOBID=12
```
But matched 16 times (allocating all cores in the cluster) with `lonode`:
```bash
$ ./resource-query -p ALL:core -F pretty_simple -S CA -f jgf -P lonode -L cluster.json
resource-query> match allocate test2.yaml
...
INFO: JOBID=15
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
resource-query> m allocate test2.yaml
      ---k8scluster0[1:shared]
      ------rack0[1:shared]
      ---------ip-192-168-38-107.us-east-2.compute.internal2217[1:shared]
      ------------socket0[1:shared]
      ---------------core93[1:exclusive]
      ---------------core92[1:exclusive]
      ---------------core91[1:exclusive]
...
      ---------------core3[1:exclusive]
      ---------------core2[1:exclusive]
      ---------------core1[1:exclusive]
      ---------------core0[1:exclusive]
INFO: JOBID=16
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
resource-query> m allocate test2.yaml
INFO: No matching resources found
INFO: JOBID=17
resource-query>
```

To be explicit, here is the `test2.yaml` jobspec:
```YAML
version: 9999
resources:
- type: slot
  count: 1
  label: default
  with:
  - type: core
    count: 94
attributes:
  system:
    duration: 3600
tasks:
- command: [""]
  slot: default
  count:
    per_slot: 1
```

dongahn:
This PR adds fixes for the following problems.

With the recently added lonode[x]/hinode[x] match policies, each node's
numeric ID (e.g., node1231 where ID=1231) is used in our score calculation
of all of its containing resource vertices (e.g., cores, gpus...). When its ID
becomes high, however, a type coercion bug can result in an integer
overflow of the score, leading to a undefined behavior. In addition, when
a node's ID becomes even larger such that it can't fit into a signed
32-bit integer, our graph loading fails. While it is less likely we will have
such a huge node ID, it would be ideal to prevent that from happening at all.

-  Use `int64_t` to capure a score within our traverser to match the type
with the rest of the code that is involved in scoring;
- Change `std::atoi` to `std::atoll` to convert a large number string into a number
without raising an overflow exception within a hostname querying method
and adjust the callers within our readers accordingly; 
- Align the column of `t3027-resource-RV.t` test cases where HERE doc is
used such a way that a consistent tab-leading indentation helps with readability. 

Fixes #932.

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -622,7 +622,7 @@ int dfu_impl_t::dom_slot (const jobmeta_t &meta, vtx_t u,
     qual_num_slots = cnt_slot (slot_shape, dfu_slot);
     for (unsigned int i = 0; i < qual_num_slots; ++i) {
         eval_egroup_t edg_group;
-        int score = MATCH_MET;
+        int64_t score = MATCH_MET;
         for (auto &slot_elem : slot_shape) {
             unsigned int j = 0;
             unsigned int qc = dfu_slot.qualified_count (dom, slot_elem.type);
--- a/resource/readers/resource_reader_base.cpp
+++ b/resource/readers/resource_reader_base.cpp
@@ -93,7 +93,7 @@ void resource_reader_base_t::clear_err_message ()
 
 int resource_reader_base_t::split_hostname (const std::string &hn,
                                             std::string &basename,
-                                            int &id) const
+                                            int64_t &id) const
 {
     std::string suffix;
     std::size_t first;
@@ -121,7 +121,7 @@ int resource_reader_base_t::split_hostname (const std::string &hn,
 
     suffix = suffix.substr (first);
     try {
-        id = std::stoi (suffix);
+        id = std::stoll (suffix);
     } catch (std::invalid_argument &e) {
         errno = EINVAL;
         goto error;
@@ -136,7 +136,7 @@ error:
 }
 
 int resource_reader_base_t::get_hostname_suffix (const std::string &hn,
-                                                 int &id) const
+                                                 int64_t &id) const
 {
     if (hn == "") {
         errno = EINVAL;
@@ -153,7 +153,7 @@ int resource_reader_base_t::get_host_basename (const std::string &hn,
         errno = EINVAL;
         return -1;
     }
-    int id;
+    int64_t id;
     return split_hostname (hn, basename, id);
 }
 
--- a/resource/readers/resource_reader_base.hpp
+++ b/resource/readers/resource_reader_base.hpp
@@ -102,7 +102,7 @@ public:
      *
      * \return          0 if succeeds or -1 with errno if an error encountered
      */
-    int get_hostname_suffix (const std::string &hn, int &id) const;
+    int get_hostname_suffix (const std::string &hn, int64_t &id) const;
 
     /*! Query the host basename from hostname.
      *
@@ -124,7 +124,7 @@ public:
 protected:
     bool in_allowlist (const std::string &resource);
     int split_hostname (const std::string &hn,
-                        std::string &basename, int &id) const;
+                        std::string &basename, int64_t &id) const;
     std::set<std::string> allowlist;
     std::string m_err_msg = "";
 };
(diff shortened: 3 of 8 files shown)


dongahn:
@milroy: Ok, I added one more test to cover an error branch and Codecov is happy now. This is ready for your quick review. Thank.s

resource/traversers/dfu_impl.cpp
```diff
@@ -622,7 +622,7 @@ int dfu_impl_t::dom_slot (const jobmeta_t &meta, vtx_t u,
     qual_num_slots = cnt_slot (slot_shape, dfu_slot);
     for (unsigned int i = 0; i < qual_num_slots; ++i) {
         eval_egroup_t edg_group;
-        int score = MATCH_MET;
+        int64_t score = MATCH_MET;
```
jameshcorbett, line 625: General question not related to this particular change: is there a reason not to use `long long` instead of `int64_t`? Do you need it to be enforced as exactly 64 bits?

In this case I see that score is being incremented by ``eval_egroup_t.score`` which is int64_t so it makes sense here.

jameshcorbett (approved):
LGTM, I can't see any potential issues.

resource/traversers/dfu_impl.cpp
dongahn: Thanks! Yeah, I could have used 'long long' as well. I just wanted to match the type across the board. For me, using an explicit type reduces my cognitive load but I understand there are other benefits to use implementation dependent types like 'long long`.

PR #934, opened 2022-04-20
reader: fix index type

dmargala:
Fix error reported by gcc 11.2.0: "comparison of integer expressions of different signedness: 'int' and 'size_t'"

I verified jansson expects `size_t` here:

https://jansson.readthedocs.io/en/latest/apiref.html?highlight=json_array_foreach#c.json_array_foreach

### DIFF
--- a/resource/readers/resource_reader_rv1exec.cpp
+++ b/resource/readers/resource_reader_rv1exec.cpp
@@ -176,7 +176,7 @@ int resource_reader_rv1exec_t::build_rmap (json_t *rlite,
                                            std::map<unsigned, unsigned> &rmap)
 {
     int i;
-    int index;
+    size_t index;
     unsigned rank;
     json_t *entry = nullptr;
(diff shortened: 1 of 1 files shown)


dongahn:
Thanks @dmargala. 

Your `fix-sign-compare-error` seems slightly out of date against the master branch. Could you do `git fetch upstream` (if `upstream` is your name of the main upstream repo) and `git rebase upstream/master` before `git push -f origin fix-sign-compare-error`  ?

dongahn:
I would be also good to incorporate your comments into the commit message. 

```
reader: Use size_t for json_array_foreach instead of int

Problem:  jansson expects size_t for json_array_foreach, but the code is using int.
GCC 11.2.0 (g++) reports an error: "comparison of integer expressions of different
signedness: 'int' and 'size_t'"

Use size_t instead.
```

dongahn (approved):

Issue #936, opened 2022-05-05
spack user reports compiler error due to invalid integer comparison in reapi_cli_impl.hpp on gcc 9.4.x


grondo:
See https://github.com/spack/spack/issues/30480 for details.

The explicitly offending comparison is here:

https://github.com/flux-framework/flux-sched/blob/6935cd092cec2503fe03b498bef1ec818b78497e/resource/hlapi/bindings/c%2B%2B/reapi_cli_impl.hpp#L597-L599

where `jobid` is a `uint64_t` being compared to an `int64_t`. I actually don't understand this test. Since `flux_jobid_t` is defined as `uint64_t`, any `uint64_t` value is a valid jobid, except for the special value `FLUX_JOBID_ANY`, which is defined as 
```c
#define FLUX_JOBID_ANY 0xFFFFFFFFFFFFFFFF // ~(uint64_t)0
```
To me, it feels like this check could just be dropped, or should specifically target `FLUX_JOBID_ANY`, but I'm not sure if the reapi interface even uses or depends on flux-core, so I'm not positive of the correct fix here.

In the spack package I will just add a patch to do this for v0.22.0, but we should try to make a valid fix for the next version of flux-sched.

There appear to be several other places in the reapi code where a `jobid` is cast to `int64_t`, but I'm not sure of the repercussions, so I can't comment if they need to be fixed. 

cc; @milroy

grondo:
I did verify that this patch works too

```diff
diff --git a/resource/hlapi/bindings/c++/reapi_cli_impl.hpp b/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
index 0b01b341..e0a91458 100644
--- a/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
@@ -594,7 +594,7 @@ int resource_query_t::remove_job (const uint64_t jobid)
 {
     int rc = -1;
 
-    if (jobid > std::numeric_limits<int64_t>::max ()) {
+    if (jobid > (uint64_t) std::numeric_limits<int64_t>::max ()) {
         errno = EOVERFLOW;
         return rc;
     }
```

PR #937, opened 2022-05-05
hlapi: workaround integer compare error with gcc 9.4.0

grondo:
This is a simple workaround for the issue reported in #936. I've already proposed a spack PR which adds this same patch to the build for 0.22.0 and later, but it would be nice to not have to carry that forward to the next release as well.


dongahn:
This looks good to me. I thought about the use of `static_cast<uint_64> ()` to do C++ style casting, but in this case, the difference would be null. I will approve and put MWP.

dongahn (approved):

grondo:
> This looks good to me. I thought about the use of static_cast<uint_64> () to do C++ style casting, but in this case, the difference would be null. I will approve and put MWP.

Thanks!

Not sure if you want to open a separate issue on the use of `int64_t` for `jobid` rather than `uint64_t` (`flux_jobid_t` is a `unit64_t`) I'm not sure if there any actual repercussions because I'm unsure if this code is used with a Flux instance or not.

dongahn:
>Not sure if you want to open a separate issue on the use of int64_t for jobid rather than uint64_t (flux_jobid_t is a unit64_t) I'm not sure if there any actual repercussions because I'm unsure if this code is used with a Flux instance or not.

Sure. There are a few places in the fluxion code that return -1 or jobid to check error conditions and we have assumed jobid won't exceed the max of `int64_t.`

PR #943, opened 2022-05-19
qmanager: fix potential leak in alloc callback

grondo:
Problem: If an alloc request results in a jobspec parse error, the
jobspec string is leaked.

Free the newly created jobspec string in the error path to avoid
the leak.


dongahn:
Great catch! Thanks LGTM.

dongahn (approved):

grondo:
@mergifyio refresh

PR #945, opened 2022-06-01
WIP Initial support for PerfFlowAspect integration

slabasan:
Build with an optional --with-perfflowaspect=PATH flag with configure,
indicating path to PerfFlowAspect install (subdirs are lib/, share/, and test/)

@tpatki


garlick:
This like an abandoned early start?  I'll go ahead and close this PR since there's not much here.

Issue #947, opened 2022-07-11
`--exclusive` nodes do not have their full resources emitted in R


jameshcorbett:
When match policy is *not* `lonodex` or `hinodex` and nodes are requested exclusively with `--exclusive`, only the minimum matching resource set is included in the generated R, though the node is still allocated exclusively. See e.g. this output on Fluxion, where the system instance uses `lonodex` but single-user instances do not.

```
[corbett8@fluke108:~]$ flux mini alloc -N1 
rsmi_init() failed
[corbett8@fluke6:~]$ flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        4        0 fluke6
 allocated      0        0        0 
      down      0        0        0 
[corbett8@fluke6:~]$ flux mini run -N1 -n1 --exclusive flux start flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        1        0 fluke6
 allocated      0        0        0 
      down      0        0        0 
[corbett8@fluke6:~]$ flux mini alloc -N1 --exclusive flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        1        0 fluke6
 allocated      0        0        0 
      down      0        0        0 
```

I think as long as a node is being allocated exclusively, all the resources should be made available to the job.

See also #928.

milroy:
I can reproduce the same behavior in `resource-query`. 

After experimenting with the code, it looks like the issue is caused by the traverser. Unfortunately, the only way I can see so far to emit the "shadow" node resources along with those matched exclusively with the `--exclusive` option is to call  [`add_exclusive_resource_type ("node");`](https://github.com/flux-framework/flux-sched/blob/1f0a19ffe734c641675248bc6191b985b5e818da/resource/policies/dfu_match_policy_factory.cpp#L50) before traversing the graph. Fluxion could be modified to check if the jobspec's node request is `exclusive` before traversing, but that seems like a hack.

Does `--exclusive` apply to other resource types as well?

grondo:
> Does --exclusive apply to other resource types as well?

Technically [yes](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html#resources). Can the traverser check to see which resource type is marked as `exclusive` and set `add_exclusive_resource_type (type)` for each exclusive type?

For practical purposes, only the `node` type will be marked exclusive by our current Flux submission tools.

[...]

Issue #928, opened 2022-04-06
add test case for `flux mini --exclusive` flag


grondo:
Jobspec V1 ([RFC 25](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html)), now specifically allows an `exclusive` key to be set on a `node` resource vertex in order to request a exclusive node access. Also, the `flux mini` commands now include an `--exclusive` flag which the user may supply (with `-N`) to set this key. The `--exclusive` flag is also implied when `-N` is used without `-n`, so that we can get sane behavior from e.g. `flux mini batch -N2 script`.

As requested by @dongahn, this issue requests that some test cases be added to flux-sched which use the `flux mini --exclusive` (actual or implied) flag, and ensure the resulting R is expected.

One thing I noticed when testing this feature with Fluxion is that, with `--exclusive`, only the minimum matching resource set is included in the generated R, though the node is still allocated exclusively. We can discuss if instead emitting _all_ the node resources in R would be better, and the less surprising result. (I can open a separate ticket on this if desired)

PR #951, opened 2022-08-02
Emit full exclusive resources

milroy:
As identified in issues #947 and #928 if nodes are allocated with the `--exclusive` option via, e.g., `flux mini run`, the complete set of allocated resources (including the shadow resources) is not emitted.

This PR adds each resource requested exclusively to the matcher's set of exclusive resource types. Inserting these resources allows the emitter to emit the full set of allocated resources.

The PR requires modification to the behavior of the traverser and matcher components of Fluxion.


grondo:
@milroy, I spot checked several output file changes and they all look reasonable to me. That is, in each case where the output changed (additional resources included in emitted _R_), there was an `exclusive` flag set on resource type `node`. So I have fairly good confidence that all you've done here is fix the bug as intended :slightly_smiling_face:

Some minor comments on the commits though, so they comply with our [RFC 1](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements) "patch requirements"

 * commits should have a commit body. We've found that first adding a `Problem: [Description]` statement, then a description of how the problem is resolved by the commit is very helpful.
 * commits should be in imperative, e.g. "add expected outputs" not "adding expected outputs"

O/w, I'm pretty sure this looks good to me!

milroy:
> Some minor comments on the commits though, so they comply with our [RFC 1](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements) "patch requirements"

I'll make sure to do this. I intended to make the commits conform to the requirements but wanted to check that all CI tests passed first. I'll fix that up now.

@grondo on another note, can you think of a better place to put the `add_exclusive_resource_type` call in `dfu_impl.cpp`? It seems like the `prime_jobspec` function isn't quite the right conceptual fit. The other alternative I considered was to make a new `set_exclusive_resource` function, but that seemed excessive.

grondo:
> @grondo on another note, can you think of a better place to put the add_exclusive_resource_type call in dfu_impl.cpp? It seems like the prime_jobspec function isn't quite the right conceptual fit. The other alternative I considered was to make a new set_exclusive_resource function, but that seemed excessive.

What you have done seemed fine to me, but I'm possibly not the right person to ask since I know very little of this code. Maybe @trws should get a comment? It is also on my list to test @jameshcorbett's reproducer, but I have high confidence it will work (i.e. you've fixed the specific bug he discovered)

trws (approved):
Assuming I'm seeing correctly that this is the only traverser, this change looks good.  If that ever changes we might have to rethink where some of these things happen.

grondo (commented):
I tested this in a subinstance to verify it is working. The simple case of a single core with non-exclusive nodes vs exclusive works, but when I allocate two cores across two nodes, I seem to get exclusive nodes, which should not be the case

. (I added the following change to print the current match policy just to be sure)

```diff
diff --git a/resource/modules/resource_match.cpp b/resource/modules/resource_match.cpp
index 9911bc13..fe250fb2 100644
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -453,6 +453,10 @@ static std::shared_ptr<resource_ctx_t> init_module (flux_t *h,
         goto error;
     }
     ctx->opts.canonicalize ();
+    flux_log (h,
+              LOG_DEBUG,
+              "using match policy: %s",
+              ctx->opts.get_opt().get_match_policy().c_str());
     if (flux_msg_handler_addvec (h, htab, (void *)h, &ctx->handlers) < 0) {
         flux_log_error (h, "%s: error registering resource event handler",
                         __FUNCTION__);
```
```
ƒ(s=4,d=1) grondo@fluke6:~/git/flux-core.git$ flux dmesg | grep policy
2022-08-02T22:36:29.620016Z sched-fluxion-resource.debug[0]: using match policy: first
2022-08-02T22:36:29.800192Z sched-fluxion-qmanager.debug[0]: enforced policy (queue=default): fcfs
2022-08-02T22:36:29.800199Z sched-fluxion-qmanager.debug[0]: effective policy params (queue=default): default

ƒ(s=4,d=1) grondo@fluke6:~/git/flux-core.git$ flux mini alloc -n1 flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        1        0 fluke8
      down      0        0        0 
[detached: session exiting]

ƒ(s=4,d=1) grondo@fluke6:~/git/flux-core.git$ flux mini alloc -n1 -N1 --exclusive flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        4        0 fluke6
      down      0        0        0 
[detached: session exiting]

ƒ(s=4,d=1) grondo@fluke6:~/git/flux-core.git$ flux mini alloc -n2 -N2 flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      2        8        0 fluke[8-9]
      down      0        0        0 

```

With the existing version of Fluxion installed on fluke, I see the following for the `-n2 -N2` case:

```
ƒ(s=2,d=1) grondo@fluke10:~$ flux mini alloc -N2 -n2 flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      2        2        0 fluke[10-11]
      down      0        0        0 
[detached: session exiting]
```

I think the expected behavior should be 2 cores across 2 nodes for `-N2 -n2` without the `--exclusive` flag.

grondo:
Here's the jobspec generated by `flux mini alloc -N2 -n2`. It also occurs to me someone else should try to reproduce the result I'm getting in case it is something in my environment or test methodology:

Generated by `flux mini alloc -N2 -n2 --dry-run --env=-*  hostname`

```json
{
  "resources": [
    {
      "type": "node",
      "count": 2,
      "with": [
        {
          "type": "slot",
          "count": 1,
          "with": [
            {
              "type": "core",
              "count": 1
            }
          ],
          "label": "task"
        }
      ]
    }
  ],
  "tasks": [
    {
      "command": [
        "flux",
        "broker",
        "hostname"
      ],
      "slot": "task",
      "count": {
        "per_slot": 1
      }
    }
  ],
  "attributes": {
    "system": {
      "duration": 0,
      "shell": {
        "options": {
          "per-resource": {
            "type": "node"
          },
          "pty": {
            "interactive": 1
          }
        }
      },
      "cwd": "/g/g0/grondo/git/flux-core.git",
      "environment": {}
    }
  },
  "version": 1
}

```

trws:
That's interesting, somehow `resource.exclusive` must be true for the node resource despite it being outside the slot.  Having it allocate the full node for `-N1 --exclusive` seems reasonable, but `-n 2 -N 2` I think I agree should be one core each on two nodes unless exclusive is set for the instance or on the command line.

grondo:
@milroy discovered that the real reproducer here is
```
$ flux mini alloc -N2 -n2 flux resource list                   # works as expected 1 core on each of 2 nodes
$ flux mini alloc -N1 -n1 --exclusive flux resource list       # works as expected, 1 core on 1 node
$ flux mini alloc -N2 -n2 flux resource list                   # doesn't work, exclusively allocates 2 nodes
```
The exclusive flag for resources is permanent and affects all future traversals unless cleared.

trws:
That's definitely a problem, and probably means a refactor is in order.  I see the conversation on slack from a bit ago, if matchers are cheap to create then doing that sounds good for now.  If not, in the end it really should be a property of a single resource node in the jobspec, so it might be worth the pain to get it there from where it is.

grondo:
Yeah, even if we get a quick fix in, I agree we'll want to change this behavior. 

> If not, in the end it really should be a property of a single resource node in the jobspec, so it might be worth the pain to get it there from where it is.

Agree! Since we need, and already use, forced exclusive "node" resources as a configurable policy, if we can change the exclusive behavior to be a property of the jobspec instead of the matcher/traverser/whatever it is, then the policy could either modify the jobspec to force exclusivity before running the matcher, or internally it could check for either the jobspec exclusive flag, or a policy default.

milroy:
@grondo thanks for catching the problem with my fix. I discovered that it manifested in other tests as well (such as my updates to `t/data/resource/expected/reservation/001.R.out`, especially JOBID 8.

I think I have a working update. Please test my latest push to see if it behaves as you expect.

grondo:
A simple test seems to work as expected!
```
ƒ(s=4,d=0) grondo@fluke108:~/git/flux-core.git$ flux mini alloc -N1 -n1 flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        1        0 fluke108
      down      0        0        0 
[detached: session exiting]
ƒ(s=4,d=0) grondo@fluke108:~/git/flux-core.git$ flux mini alloc -N1 -n1 --exclusive flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1        4        0 fluke108
      down      0        0        0 
[detached: session exiting]
ƒ(s=4,d=0) grondo@fluke108:~/git/flux-core.git$ flux mini alloc -N2 -n2 --exclusive flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      2        8        0 fluke[108,108]
      down      0        0        0 
[detached: session exiting]
ƒ(s=4,d=0) grondo@fluke108:~/git/flux-core.git$ flux mini alloc -N2 -n2 flux resource list
rsmi_init() failed
     STATE NNODES   NCORES    NGPUS NODELIST
      free      2        2        0 fluke[108,108]
      down      0        0        0 
[detached: session exiting]

```

milroy:
I added two more commits that get and reset the matcher `m_x_resource_types` set via public member functions, and functionality in the traverser to call them. The traverser gets the set of exclusive resources before running, and then resets the matcher exclusive resource types to the values before traversal.

I've cleaned up the commits and dropped the [WIP] label. This PR should be good to merge. Thank you for the validation and reviews @grondo and @trws!

grondo (approved):
Nicely done! Tested working in my environment.

grondo:
If it is ok, I'll set MWP and we can get this fix into a new flux-sched release.

grondo:
Thanks, @milroy!

PR #952, opened 2022-08-03
add release notes for v0.24.0

grondo:
@milroy fixed an important issue in flux-sched recently, and it would be nice to get this into a release and packaged for installation. Though there's only one item, I propose we do a release with that fix, so here's the proposed `NEWS.md` update.


garlick (approved):
Excellent!!

PR #956, opened 2022-08-09
testsuite: move flux-tree to the test suite

garlick:
Problem: the flux-tree tool has an ambiguous name, has no
documentation, is not user friendly (requires jobspec),
performs a function that is not generally useful, and is not
actively maintained.  As such, it is more of an attractive
nuisance than a benefit to users at this point.

Since the two sharness tests that ostensibly test flux-tree are
probably covering other things in fluxion, leave the tests in place
but relocate flux-tree and its helper to t/scripts and skip the
installation.

If we develop a need for this tool at some point, we should come
up with a new name, find a maintainer, and move it to its own
flux-framework project since it's not fluxion specific.

Make trivial changes to flux-tree, its helper, and sharness scripts
so they can function within the test directory.


grondo:
`t/sharness.d/sched-sharness.sh` adds `t/scripts` to `FLUX_EXEC_PATH` so I'm surprised you couldn't just move the sources there without having to touch anything else.

garlick:
Restarted the ubuntu based builders - looks like apt-get was failing to phone home before.

garlick:
> t/sharness.d/sched-sharness.sh adds t/scripts to FLUX_EXEC_PATH so I'm surprised you couldn't just move the sources there without having to touch anything else.

Oh, does it? I can't say I tried it!

I'll give it a try and see if I can back off the sharness script changes.  The flux-tree changes probably should remain so that it can be run by hand without path manipulation.

garlick:
Thanks for that!  It reduced the diff considerably.

grondo:
> The flux-tree changes probably should remain so that it can be run by hand without path manipulation.

Sounds fine, though I'm not sure what is wrong with appending to `FLUX_EXEC_PATH` since that is why we provide it. It is also a little strange that internally flux-tree runs itself as `flux flux-tree` but the initial invocation will be `/path/to/flux-tree` (I guess I can see the benefit of not having `flux tree` be a thing though, so maybe that is worth it.)

Meh, doesn't really matter and since it is working... (My only worry is setting a bad precedent for creating flux extension scripts and commands)

grondo (approved):
LGTM!

garlick:
> Sounds fine, though I'm not sure what is wrong with appending to FLUX_EXEC_PATH since that is why we provide it. It is also a little strange that internally flux-tree runs itself as flux flux-tree but the initial invocation will be /path/to/flux-tree (I guess I can see the benefit of not having flux tree be a thing though, so maybe that is worth it.)

Yeah, point taken.  Sorry it's taking me a bit to get my head around this.  More shrinkage coming up!

grondo:
I didn't mean to cause you more work, I was just trying to understand if you had thought of a drawback to using `FLUX_EXEC_PATH`/`FLUX_EXEC_PATH_PREPEND` etc.

garlick:
I"lll go ahead and set MWP.

garlick:
@mergifyio-update

grondo:
@mergifyio refresh

PR #957, opened 2022-08-09
add an exception note to fatal queue exceptions

garlick:
Problem: submitting a job to an unknown queue or failure to insert a job into a queue (an unlikely internal error) cause a fatal job exception with no "note", so the user doesn't get any reason why their job failed.

Add a note to the exception.


grondo (approved):
LGTM! Optional: add a test that exception note is non-empty in this failure condition.

garlick:
@Mergifyio refresh

PR #958, opened 2022-08-12
rc: combine fluxion rc scripts

garlick:
This is some minor cleanup of the fluxion rc scripts, which I was finding to be a bit confusing.  The main change is that there are no longer separate scripts for the two fluxion broker modules.


grondo (approved):
Very nice cleanup IMO!

garlick:
Let's try again with the `flux module reload -f` back in the rc1 script.  I couldn't seem to get some tests (particularly the flux tree tests) to work right on a system that has fluxion already without that, and this is taking up too much time.

garlick:
OK, that did it.  Since there was no new content with that change, and the stack was previously approved, I'll go ahead and set MWP.  Thanks for the review @grondo.

PR #966, opened 2022-09-12
testsuite: use explicit duration units

garlick:
Problem: flux mini will fail when small durations are specified without units after flux-framework/flux-core#4565.

Add explicit units to flux mini CMD -t where necessary.


grondo (approved):
Thanks!

Issue #913, opened 2022-03-02
update handling of duration=0 to match RFC 14 specification


grondo:
A pending PR against RFC 14 (flux-framework/rfc#315) updates the required behavior of schedulers when jobspec `duration` is unset (`duration=0.`).

This new specification requires that schedulers propagate the expiration set in _R_ upon resource acquisition into Rv1 fragments allocated to jobs when duration is not set. That is, if the jobspec does not indicate a duration, then at a minimum the currently acquired resource expiration should be propagated to the job's time limit.

Ideally, jobs requesting an explicit duration that would definitely exceed the current instance expiration should be rejected by the scheduler `feasibility` RPC, so that these jobs fail at submission time when the `feasibility` validator is enabled.

grondo:
@dongahn - this is another issue we may want to address near-term, so any pointers on where to start an implementation would be much appreciated.

Given that flux-core now has a jobspec duration validator, the only thing left to implement in Fluxion is for the rv1 emitter to propagate any `expiration` in `resource.R` into emitted Rv1 when the jobspec has specified `duration = 0`.

dongahn:
Yup. I will take a look and add some thoughts.

dongahn:
I will look at this more carefully later but per our discussion today. This is the configuration header that I used as a stand-in:  https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L18

dongahn:
R will be acquired by these points within sched-fluxion-resource:
https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L1230
https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L1202

Then, the duration in R should be propagated to the resource infrastructure and get to 
https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl.hpp#L75


There are a few different ways you can pass the duration to that point. One way might be extend a public method with the traverser like `run`: https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu.cpp#L277

But there might be a simpler way that we can achieve this w/o changing the public method like that.

You may need a `std::map` somewhere keyed by the queue name, that holds the duration per each queue as well.

BTW, will there be a case where there is no instance duration set? (Probably the case of our system instance?)

grondo:
> BTW, will there be a case where there is no instance duration set? (Probably the case of our system instance?)

[...]

PR #969, opened 2022-09-16
Resource graph duration and job expiration to conform to RFC 14

milroy:
This PR fixes issue #913 by introducing a `graph_duration_t` struct to hold a `std::chrono::time_point` for start and end times into the resource graph metadata and then setting it based on a `resource.acquire` RPC. At schedule time, jobs that specify a `duration` greater than the resource graph duration are rejected. However, jobs with scheduled runtimes that extend beyond the resource graph duration are fit into the resource graph duration procrustean bed.

~~This PR is WIP.~~ First, we need to determine if the procrustean approach complies with updated RFC 14 (https://github.com/flux-framework/rfc/pull/315). Furthermore, we need to consider if the `reader` is the right place for setting the resource graph duration. Beyond that, several items need to be addressed before it can be merged:

- [x] Are the `starttime` and `expiration` received via `resource.acquire` always type `double`? Other segments of `resource_match` unpack them as `I`. More generally: Fluxion stores these times as `int64_t` and `uint64_t`. We need to determine if this mismatch causes undefined behavior.
- [x] Ensure the checks in `unpack_resources` handle corner cases for `starttime` and `expiration`
- [x] Refactor or use system defaults: https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L18
- [x] Create test cases via commits to `testsuite`

The PR also introduces `std::chrono` into Fluxion. We intend for it to replace the mixture of `double`, `uint64_t` and `int64_t` time representations in Fluxion in the future. For the present, careful checks for representability, validity, and overflow need to be introduced to interface with flux-core and the various components of Fluxion.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -14,6 +14,7 @@
 #include <cerrno>
 #include <map>
 #include <cinttypes>
+#include <chrono>
 
 extern "C" {
 #if HAVE_CONFIG_H
@@ -105,7 +106,6 @@ struct resource_ctx_t : public resource_interface_t {
     std::map<uint64_t, uint64_t> reservations; /* Reservation table */
     std::map<std::string, std::shared_ptr<msg_wrap_t>> notify_msgs;
 };
-
 
 msg_wrap_t::msg_wrap_t (const msg_wrap_t &o)
 {
@@ -493,6 +493,7 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
     int saved_errno;
     std::ifstream in_file;
     std::stringstream buffer{};
+    graph_duration_t duration;
 
     if (ctx->reader == nullptr
         && create_reader (ctx, ctx->opts.get_opt ().get_load_format ()) < 0) {
@@ -523,6 +524,8 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
                   __FUNCTION__, ctx->reader->err_message ().c_str ());
         goto done;
     }
+    ctx->db->metadata.set_graph_duration (duration);
+
     rc = 0;
 
 done:
@@ -608,13 +611,14 @@ ret:
 /* Given 'resobj' in Rv1 form, decode the set of execution target ranks
  * contained in it as well as r_lite key.
  */
-static int unpack_resources (json_t *resobj,
-                             struct idset **idset,
-                             json_t **r_lite_p, json_t **jgf_p)
+static int unpack_resources (json_t *resobj, struct idset **idset,
+                             json_t **r_lite_p, json_t **jgf_p,
+                             graph_duration_t &duration)
 {
     int rc = 0;
     struct idset *ids;
     int version;
+    double start = 0.0, end = 0.0;
     json_t *r_lite = NULL;
     json_t *jgf = NULL;
     size_t index;
@@ -624,14 +628,48 @@ static int unpack_resources (json_t *resobj,
         return -1;
     if (resobj) {
         if (json_unpack (resobj,
-                         "{s:i s:{s:o} s?:o}",
+                         "{s:i s:{s:o s?F s?F} s?:o}",
                          "version", &version,
                          "execution",
                            "R_lite", &r_lite,
+                           "starttime", &start,
+                           "expiration", &end,
                          "scheduling", &jgf) < 0)
             goto inval;
+        // flux-core validates these numbers, but checking here
+        // in case Fluxin is plugged into another resource manager
         if (version != 1)
             goto inval;
+        if (start != 0 && (start == end))
+            goto inval;
+        if (start < 0.0 || end < 0.0)
+            goto inval;
+        if (start > end)
+            goto inval;
+        // Greatest lower bound is int64_t in the rest of Fluxion
+        if (start > static_cast<double> (std::numeric_limits<int64_t>::max ())
+                  || end > static_cast<double>
+                                       (std::numeric_limits<int64_t>::max ()))
+            goto inval;
+        // Ensure start and end are representable in system clock
+        if (start > static_cast<double>
+            (std::chrono::duration_cast<std::chrono::seconds>
+                (std::chrono::time_point<std::chrono::system_clock>::max ().time_since_epoch ()).count ())
+              || end > static_cast<double>
+                (std::chrono::duration_cast<std::chrono::seconds>
+                    (std::chrono::time_point<std::chrono::system_clock>::max ().time_since_epoch ()).count ()))
+            goto inval;
+        // Expects int type argument
+        duration.graph_start = std::chrono::system_clock::from_time_t (static_cast<int64_t> (start));
+        duration.graph_end = std::chrono::system_clock::from_time_t (static_cast<int64_t> (end));
+        // Ensure there is no overflow in system clock representation
+        // (should be handled by previous check).
+        if (std::chrono::duration_cast<std::chrono::seconds>
+                        (duration.graph_start.time_since_epoch ()).count () < 0
+            || std::chrono::duration_cast<std::chrono::seconds>
+                       (duration.graph_end.time_since_epoch ()).count () < 0) {
+            goto inval;
+        }
         json_array_foreach (r_lite, index, val) {
             struct idset *r_ids;
             unsigned long id;
@@ -950,6 +988,8 @@ static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx,
 {
     int rc = 0;
     int saved_errno;
+    // Unused: necessary to avoid function overload and duplicated code
+    graph_duration_t duration;
     json_t *p_jgf = NULL;
     json_t *p_r_lite = NULL;
     json_t *p_resources = NULL;
@@ -957,8 +997,8 @@ static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx,
     if ( (rc = get_parent_job_resources (ctx, &p_resources)) < 0
          || !p_resources)
         goto done;
-    if ( (rc = unpack_resources (p_resources,
-                                 &p_grow_set, &p_r_lite, &p_jgf)) < 0)
+    if ( (rc = unpack_resources (p_resources, &p_grow_set, &p_r_lite,
+                                 &p_jgf, duration)) < 0)
         goto done;
     if (!p_grow_set || !p_r_lite || !p_jgf) {
         errno = EINVAL;
@@ -1027,12 +1067,13 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
                              json_t *resources)
 {
     int rc = 0;
+    graph_duration_t duration;
     struct idset *grow_set = NULL;
     json_t *r_lite = NULL;
     json_t *jgf = NULL;
 
     if ( (rc = unpack_resources (resources,
-                                 &grow_set, &r_lite, &jgf)) < 0) {
+                                 &grow_set, &r_lite, &jgf, duration)) < 0) {
         flux_log_error (ctx->h, "%s: unpack_resources", __FUNCTION__);
         goto done;
     }
@@ -1062,7 +1103,8 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
             errno = EINVAL;
             rc = -1;
         }
-   }
+    }
+    ctx->db->metadata.set_graph_duration (duration);
 
 done:
     idset_destroy (grow_set);
@@ -1175,6 +1217,8 @@ static int update_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
                                const char *up, const char *down)
 {
     int rc = 0;
+    // Will need to get duration update and set graph metadata when
+    // resource.acquire duration update is supported in the future.
     if (resources && (rc = grow_resource_db (ctx, resources)) < 0) {
         flux_log_error (ctx->h, "%s: grow_resource_db", __FUNCTION__);
         goto done;
--- a/resource/store/resource_graph_store.cpp
+++ b/resource/store/resource_graph_store.cpp
@@ -14,6 +14,21 @@
 using namespace Flux;
 using namespace Flux::resource_model;
 
+void resource_graph_metadata_t::set_graph_duration (
+                            graph_duration_t &g_duration)
+{
+    if ( (g_duration.graph_start == std::chrono::system_clock::from_time_t (0))
+                                 && (g_duration.graph_end
+                              == std::chrono::system_clock::from_time_t (0))) {
+        graph_duration.graph_start = std::chrono::system_clock::now ();
+        graph_duration.graph_end = graph_duration.graph_start +
+                            std::chrono::seconds (detail::SYSTEM_MAX_DURATION);
+    } else {
+        graph_duration.graph_start = g_duration.graph_start;
+        graph_duration.graph_end = g_duration.graph_end;
+    }
+}
+
 bool resource_graph_db_t::known_subsystem (const std::string &s)
 {
     return (metadata.roots.find (s) != metadata.roots.end ())? true : false;
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -279,6 +279,8 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
                           match_op_t op, int64_t jobid, int64_t *at)
 {
     const subsystem_t &dom = get_match_cb ()->dom_subsystem ();
+    graph_duration_t graph_duration =
+                                      get_graph_db ()->metadata.graph_duration;
     if (!get_graph () || !get_graph_db ()
         || (get_graph_db ()->metadata.roots.find (dom)
             == get_graph_db ()->metadata.roots.end ())
@@ -288,15 +290,17 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
     }
 
     int rc = -1;
+    int64_t graph_end = graph_duration.graph_end.time_since_epoch ().count ();
     detail::jobmeta_t meta;
     vtx_t root = get_graph_db ()->metadata.roots.at (dom);
     bool x = detail::dfu_impl_t::exclusivity (jobspec.resources, root);
     const std::set<std::string> exclusive_types =
                         detail::dfu_impl_t::get_exclusive_resource_types ();
     std::unordered_map<std::string, int64_t> dfv;
+
     detail::dfu_impl_t::prime_jobspec (jobspec.resources, dfv);
-    if (meta.build (jobspec,
-                    detail::jobmeta_t::alloc_type_t::AT_ALLOC, jobid, *at) < 0)
+    if (meta.build (jobspec, detail::jobmeta_t::alloc_type_t::AT_ALLOC, jobid,
+                    *at, graph_duration) < 0)
         return -1;
 
     if ( (op == match_op_t::MATCH_SATISFIABILITY)
@@ -304,6 +308,19 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
         detail::dfu_impl_t::update ();
     } else if ( (rc = schedule (jobspec, meta, x, op, root, dfv)) ==  0) {
         *at = meta.at;
+        if (*at < 0 or *at >= graph_end) {
+            detail::dfu_impl_t::reset_exclusive_resource_types
+                                                        (exclusive_types);
+            errno = EINVAL;
+            return -1;
+        }
+        // If job ends after the resource graph expires, reduce the duration
+        // so it coincides with the graph expiration. Note that we could
+        // arguably return ENOTSUP/EINVAL instead. Also note that we 
+        // know *at < graph_end from the previous check, and meta.duration 
+        // is < int64_t max from meta.build.
+        if ( (*at + static_cast<int64_t> (meta.duration)) > graph_end)
+            meta.duration = graph_end - *at;
         // returns 0 or -1
         rc = detail::dfu_impl_t::update (root, writers, meta);
     }
--- a/resource/config/system_defaults.hpp
+++ b/resource/config/system_defaults.hpp
@@ -15,8 +15,8 @@
 namespace Flux {
 namespace resource_model {
 namespace detail {
-    const uint64_t SYSTEM_DEFAULT_DURATION = 43200; // 12 hours
-    const uint64_t SYSTEM_MAX_DURATION = 604800;    //  7 days
+    const int64_t SYSTEM_DEFAULT_DURATION = 43200; // 12 hours
+    const int64_t SYSTEM_MAX_DURATION = 3153600000; // 100 years
 } // namespace detail
 } // namespace resource_model
 } // namespace Flux
--- a/resource/store/resource_graph_store.hpp
+++ b/resource/store/resource_graph_store.hpp
@@ -13,12 +13,21 @@
 
 #include <string>
 #include <memory>
+#include <chrono>
 #include "resource/schema/resource_graph.hpp"
+#include "resource/config/system_defaults.hpp"
 
 namespace Flux {
 namespace resource_model {
 
 class resource_reader_base_t;
+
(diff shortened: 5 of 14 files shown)


garlick:
Regarding starttime and duration types (part of _R_, specified [here](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#execution)) we are encoding them to/from doubles (jansson `F`) in flux-core, so fluxion should be doing the same.  My guess is it isn't critical though, since we're not likely to be assigning resources for fractions of a second in real life, and the spec indicates subsecond precision is optional.

milroy:
> Regarding starttime and duration types (part of R, specified [here](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#execution)) we are encoding them to/from doubles (jansson F) in flux-core, so fluxion should be doing the same.

I had read RFC 20, but the unpacking in `resource_match` [here](https://github.com/flux-framework/flux-sched/blob/69d4061cee71915f08093430dd823115d22b96e2/resource/modules/resource_match.cpp#L1500) confused me. That was why I used `json_number_value` to handle different `starttime` and `duration` types. I'll update the PR to unpack `starttime` and `duration` as optional jansson `F`.

garlick:
> will starttime and duration ever take negative values?

Shouldn't.  I'm not sure under what circumstances you could be handed invalid input but it does not hurt to be defensive, especially if the code's already there.

milroy:
As an update, adding the desired propagation of the the parent resource graph duration here: https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L18 results in numerous early testsuite failures. Even worse, the testsuite hangs at `t1017`.

Setting the resource graph `duration.second` in `resource_graph_store.hpp` to [`SYSTEM_MAX_DURATION`](https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L18) independently  results in numerous early testsuite failures and a hang at `t1017`.

In summary, we will need to figure out whether this functionality is correct, and if so what to do about the substantial requisite changes to testsuite.

milroy:
I pushed fixup commits so everyone can look at the complete set of changes and the testsuite failures.

resource/modules/resource_match.cpp
```diff
@@ -950,15 +962,17 @@ static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx,
 {
     int rc = 0;
     int saved_errno;
+    // Unused: necessary to avoid function overload and duplicated code
+    std::pair<uint64_t, uint64_t> duration {0, 0};
```
trws, line 966: Might I recommend a `std::chrono::duration` and perhaps a `std::chrono::time_point`?  Even if the contents are a double, that would at least encode, unambiguously, that it's seconds and prevent various classes of easy mistakes.

Also, this is partly personal preference, but could we avoid using pair? It's shockingly slow to compile, and has a tendency to obscure its contents.  A little struct, or two arguments, would be my preference.

trws:
My first thought on all those failures is to wonder what the duration they're receiving is.  If it's functionally "infinite" in a double, it might be overflowing the integral values in fluxion and coming out negative or doing something else nefarious that's causing a bunch of failures.

resource/modules/resource_match.cpp
milroy: I didn't know about `std::chrono::duration` or `time_point`, but `duration` looks like a good choice. A struct is also fine; I'm not attached to pair.

milroy:
> My first thought on all those failures is to wonder what the duration they're receiving is. If it's functionally "infinite" in a double, it might be overflowing the integral values in fluxion and coming out negative or doing something else nefarious that's causing a bunch of failures.

I did guard against overflow when unpacking the JSON doubles for `resource.acquire`, so it shouldn't be happening in `resource_match`. I've done checks (admitted not entirely thoroughly via `cout`s) of the values propagating to `dfu.cpp` and haven't seen any evidence of overflow. I should add an overflow check to this PR in `dfu.cpp` just in case.

What does appear to be happening in many of the cases is that the `starttime` and `expiration` of the resources from `resource.acquire` are both 0. That means the resource graph will have the default duration of (0, SYSTEM_MAX_DURATION) of 604800 seconds or 7 days. 

Now many of the jobspecs are scheduled `at` a time far in the future (1663489256 seconds) which is an invalid start time given the resource graph duration. I'm still looking into why the tests are scheduled so far in the future.

trws:
Chrono takes a bit of getting used to, but it works shockingly well to prevent common time manipulation errors, and we can set it to hold a double and it will maintain the expectations from the rest of flux.

Sounds good on the checks.


---

On September 18, 2022 at 12:49:48 AM PDT, Daniel Milroy ***@***.***> wrote:

@milroy commented on this pull request.

In resource/modules/resource_match.cpp<https://urldefense.us/v3/__https://github.com/flux-framework/flux-sched/pull/969*discussion_r973676916__;Iw!!G2kpM7uM-TzIFchu!j0U1bHf_vCJvFt8HPFh26N3p4mMN5PBAhBANZeuVGw5zPKUVlqg4uB6wL5dxm03VjA$>:

> @@ -950,15 +962,17 @@ static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx,
 {
     int rc = 0;
     int saved_errno;
+    // Unused: necessary to avoid function overload and duplicated code
+    std::pair<uint64_t, uint64_t> duration {0, 0};


I didn't know about std::chrono::duration or time_point, but duration looks like a good choice. A struct is also fine; I'm not attached to pair.

milroy:
> Even if the contents are a double, that would at least encode, unambiguously, that it's seconds and prevent various classes of easy mistakes.

The contents of `duration` are an important point to consider, too. Fluxion represents time as int types, and the fundamental operations in `planner` (especially, but there are other examples) require `int`s. I think it's best to translate the `float`s coming in from flux-core and handle int types within Fluxion.

I'm definitely open to other options, though.

grondo:
> I think it's best to translate the floats coming in from flux-core and handle int types within Fluxion.

This is fine IMO. The only reason we include fractional seconds in flux-core is because this allows more flexibility - e.g. when testing you can develop a test that runs less than a whole second.

trws:
As came up on slack, we probably should be moving toward int or struct of int representations for time everywhere eventually, but we need to be very careful about the conversions in the interim. One upside to Chrono is we can make a duration of type double representing ints, and safely assign it into a duration wrapping int types of an appropriately finer granularity and not worry about it, but we should probably decide what final representation we want to end up with and use that.


---

On September 18, 2022 at 5:09:25 PM PDT, Mark Grondona ***@***.***> wrote:

I think it's best to translate the floats coming in from flux-core and handle int types within Fluxion.

This is fine IMO. The only reason we include fractional seconds in flux-core is because this allows more flexibility - e.g. when testing you can develop a test that runs less than a whole second.

garlick:
Tentatively adding to an end of month release milestone.  If that's not feasible @milroy we can bump it to the next one.

milroy:
I've updated the PR with a cleaner and more naturally placed interface to `set_graph_duration` and taken advantage of  `std::chrono`. I've also cleaned up the checks and simplified the logic. The commit messages are now compliant with flux-framework RFC.

@jameshcorbett you can write testsuite checks against these commits.

resource/traversers/dfu.cpp
```diff
@@ -288,29 +290,47 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
...
+        if (meta.duration > std::numeric_limits<int64_t>::max ())
+            goto error;
+        if (*at < 0 or *at >= graph_end)
+            goto error;
+        // If job ends after the resource graph expires, reduce the duration
+        // so it coincides with the graph expiration. Note that we could
+        // arguably return ENOTSUP/EINVAL instead. Also note that we 
+        // know *at < graph_end from the previous check.
+        if ( (*at + static_cast<int64_t> (meta.duration)) > graph_end)
+            meta.duration = graph_end - *at;
```
jameshcorbett, line 322: @grondo @garlick Dan and I wanted to see what you think of this (see comment and above two lines of code). It would mean that more jobs would run, but not always for the duration that users requested. E.g. if the scheduler has 10 minutes left before resource expiration and a 30 minute job comes in, the job might be run (rather than rejected) but with a 10 minute duration.

jameshcorbett (approved):
No comment about the tests of course but the code LGTM.

resource/store/resource_graph_store.cpp
```diff
@@ -14,6 +14,21 @@
...
 
+void resource_graph_metadata_t::set_graph_duration (
+                            graph_duration_t &g_duration)
+{
+    if ( (g_duration.graph_start == std::chrono::system_clock::from_time_t (0))
+                                 && (g_duration.graph_end
+                              == std::chrono::system_clock::from_time_t (0))) {
+        graph_duration.graph_start = std::chrono::system_clock::now ();
+        graph_duration.graph_end = graph_duration.graph_start
+                          + std::chrono::seconds (detail::SYSTEM_MAX_DURATION);
```
milroy, line 25: @grondo and @garlick: another item that came up in the Fluxion biweekly hackathon today was the implementation of the `SYSTEM_MAX_DURATION` here. The duration is defined here: https://github.com/flux-framework/flux-sched/blob/6483346a3ecb9b7bccb103d760a0e87f92df6063/resource/config/system_defaults.hpp#L19
This short max duration of 7 days will cause increasingly unexpected and undesirable behavior for the system instance as the resource graph lifetime approaches 7 days. I propose changing this value to `std::numeric_limits<uint64_t>::max ()`. What do you think?

resource/traversers/dfu.cpp
milroy: One more thing we should address along these lines. The current check for jobspec duration here: https://github.com/flux-framework/flux-sched/blob/6483346a3ecb9b7bccb103d760a0e87f92df6063/resource/traversers/dfu_impl.hpp#L73
does effectively the same thing in that it shortens the job duration to `SYSTEM_MAX_DURATION` (Fluxion master) or the graph duration (this PR). Perhaps it should return an error, too?

grondo: > if the scheduler has 10 minutes left before resource expiration and a 30 minute job comes in, the job might be run (rather than rejected) but with a 10 minute duration.

This seems like a fine compromise. I should note that we do have the `duration-validator` jobtap plugin in flux-core, which will reject jobs with a duration longer than the remaining instance lifetime, so you won't have a job _submitted_ that could exceed the instance lifetime. However, you might have a job that fits within the instance expiration at the time of submission, but at the time resources are allocated it no longer does. In that case it makes total sense to reduce the expiration of the job to the expiration of the instance, but still let the job run.

resource/store/resource_graph_store.cpp
grondo: That seems fine to me, though I don't quite understand how SYSTEM_MAX_DURATION is being used. (We have a system instance on corona that has an uptime of 34 days, so I'm guessing this value doesn't apply to the actual system instance. Maybe we just can't create jobs with a duration longer than 7 days?)

grondo: Oh, I think I see. It wasn't clear to me at first that this code isn't in use yet.

resource/traversers/dfu.cpp
```diff
@@ -288,29 +290,47 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
...
         rc = detail::dfu_impl_t::update (root, writers, meta);
     }
     // returns 0 or -1
     rc += detail::dfu_impl_t::reset_exclusive_resource_types (exclusive_types);
 
     return rc;
+error:
+    detail::dfu_impl_t::reset_exclusive_resource_types (exclusive_types);
+    errno = EINVAL;
+    return -1;
```
grondo, line 333: Codecov is calling out that the error return here is not covered. I'd be interested to see that case tested (if possible). If setting up a test case is difficult then not a big deal. (You may have to disable the `.duration-validator` jobtap plugin in order to submit a job that exceeds the current instance expiration)

grondo (approved):
Nicely done! It was very easy to follow along with the commit history even though I'm not all too familiar with C++. Nice job catching many of the potential corner cases as well (I would have missed some of those).

It was also gratifying to see the evident teamwork here.

I just had one comment about potentially handling some of the other corner cases in the added tests. E.g. disable duration validator in flux-core, submit a job with a requested duration that exceeds the instance lifetime (and any other corner case you can think of.)

resource/traversers/dfu.cpp
milroy: Yes, I agree we should hit one of the error conditions that "goes to" the `error` condition. @jameshcorbett and I talked about this and it isn't easy to do without adding new functionality in `resource-query` and testing there.

resource/store/resource_graph_store.cpp
milroy: Yes, the issue will crop up pretty quickly if the commit as it is right now propagates to production.

resource/traversers/dfu.cpp
milroy: There are other use cases for Fluxion (like Kubernetes integrations) where the inputs probably won't be validated by the submitter.

> However, you might have a job that fits within the instance expiration at the time of submission, but at the time resources are allocated it no longer does. 

According to the scheduler the resources are allocated immediately after the code block in this comment thread, so I don't think that will happen in practice. 

I'm ambivalent about this approach because it seems like the user will expect that if they submit a job with a specified duration, the duration will either be honored or rejected. Getting the duration truncated might be confusing.

milroy: @trws, what are your thoughts?

resource/traversers/dfu.cpp
milroy: I'm happy to do that, but it will require a few more commits.

jameshcorbett: Yeah, me too, especially if it is faster than changes to resource-query. Any pointers on how to disable it?

milroy: You can also just use `resource-query` to do a `match allocate` with a duration longer than allowed. All you need to do is create a new jobspec.

milroy: I just realized that the check at line 313 above: 
```C++
        if (meta.duration > std::numeric_limits<int64_t>::max ())
            goto error;
```
is redundant because the duration validity checks are done in `meta.build` called at line 302. I'll add a fixup commit.

resource/traversers/dfu.cpp
trws: I'm missing a bit of context, if we're talking about submission time then I would probably prefer that the job be rejected.  If we're talking about processing the job later, maybe after a re-schedule where a higher priority job bumped the current one later, then giving it a chance to at least start sounds sensible.

resource/store/resource_graph_store.cpp
trws: If we're using chrono elsewhere, it might be better to use `std::chrono<>::max()` with the Rep and Period used elsewhere, that way we wont accidentally be setting the max above the max supported by our time representation, which is probably the signed 64 bit int rather than unsigned. If we use the uint max, we might get rollover issues or other problems in unexpected spots.

resource/traversers/dfu.cpp
trws: I started to say we could add a unit-test rather than an integration test, but it looks like sched doesn't have a unit test harness... That's going on my list, for now if we can test it through sharness, awesome, otherwise make a note or an issue to come back once we have gtest or catch2 set up in here.

resource/store/resource_graph_store.cpp
jameshcorbett: Tests failed when increasing SYSTEM_MAX_DURATION to numbers that would take 40+ bits to represent. I believe the failure comes in because according [to the standard](https://en.cppreference.com/w/cpp/chrono/duration), `std::chrono::seconds  has type  duration</*signed integer type of at least 35 bits*/> , and we pass SYSTEM_MAX_DURATION to std::chrono::seconds. I pushed a fixup commit that uses `std::chrono::time_point<std::chrono::system_clock>::max();` instead.

jameshcorbett:
Force-pushed to remove the `std::chrono<>::max()` commit I pushed before, and to rebase with master. Also added a few more tests.

If resource-query behaved the way I thought it did we could add the following tests:

```
+test_expect_success 'Misconfigured RV1 duration is handled correctly 1' '
+       cat > cmds022 <<-EOF &&
+       quit
+       EOF
+       flux R encode -r 0-1 -c 0-1 -H fluke[0-1] | \
+               jq ".execution.starttime |= -1" > out22.json &&
+       test_must_fail ${query} -L out22.json -f rv1exec -F rv1_nosched \
+               -t R22.out -P first < cmds022
+'
+
+test_expect_success 'Misconfigured RV1 duration is handled correctly 2' '
+       cat > cmds023 <<-EOF &&
+       quit
+       EOF
+       flux R encode -r 0-1 -c 0-1 -H fluke[0-1] | \
+               jq ".execution.starttime |= 1 | .execution.expiration |= 1" > out23.json &&
+       test_must_fail ${query} -L out23.json -f rv1exec -F rv1_nosched \
+               -t R23.out -P first < cmds023
+'
+
+test_expect_success 'Misconfigured RV1 duration is handled correctly 3' '
+       cat > cmds024 <<-EOF &&
+       quit
+       EOF
+       flux R encode -r 0-1 -c 0-1 -H fluke[0-1] | \
+               jq ".execution.expiration |= -1" > out24.json &&
+       test_must_fail ${query} -L out24.json -f rv1exec -F rv1_nosched \
+               -t R24.out -P first < cmds024
+'
+
+test_expect_success 'Misconfigured RV1 duration is handled correctly 4' '
+       cat > cmds025 <<-EOF &&
+       quit
+       EOF
+       flux R encode -r 0-1 -c 0-1 -H fluke[0-1] | \
+               jq ".execution.expiration |= 1.0E30" > out25.json &&
+       test_must_fail ${query} -L out25.json -f rv1exec -F rv1_nosched \
+               -t R25.out -P first < cmds025
+'
+
```

resource/traversers/dfu.cpp
jameshcorbett: > >     However, you might have a job that fits within the instance expiration at the time of submission, but at the time resources are allocated it no longer does.

> 
> According to the scheduler the resources are allocated immediately after the code block in this comment thread, so I don't think that will happen in practice.

Isn't that the problem though? A job is submitted to flux with duration T and is accepted by flux-core because the instance will not expire for another T + 5 seconds, but then it takes 10 seconds for the job to make it to the point of being scheduled, at which point the duration cannot be honored.

In any case I think it might not be a bad idea to ask Ryan and go with whatever he suggests. It's not really a system instance issue, but still.

ofaaland: I agree, better to let the job run and get some work done.  And I think that would work well with what libyogrt would expect.

I was going to say we would want to error out if the duration truncation were large, as a feedback mechanism.  But it seems like that means the scheduling delay is large, so either something is broken or the node is overloaded, and it would be better for the user/admin to learn about that in a way that does not prevent jobs from running.

That said, Ryan likely has a much better idea of what users expect than I do.

milroy:
I've cleaned up the commits and added a commit to `config`. The PR is ready for a final review so I removed the WIP tag and updated the title and description. @grondo and @trws: could you take a last look?

It would be good to merge this for the 0.25 milestone. I will see if I can get `resource-query` updated with @jameshcorbett ideas for additional checks, but that may not be possible before the deadline. We can make a follow-up `testsuite` PR for that.

resource/traversers/dfu.cpp
garlick: There seems to be a consensus and I'm not questioning that but for my own edification:

> A job is submitted to flux with duration T and is accepted by flux-core because the instance will not expire for another T + 5 seconds, but then it takes 10 seconds for the job to make it to the point of being scheduled, at which point the duration cannot be honored because the instance expires in T - 5 seconds.

When we say "scheduled" I assume that means "future start time planned" not "allocated"?  (Since obviously we expect jobs to be pending for long periods of time in a batch environment)

> But it seems like that means the scheduling delay is large, so either something is broken or the node is overloaded, and it would be better for the user/admin to learn about that in a way that does not prevent jobs from running.

Does fluxion's configured "queue-depth" still potentially mean that some pending jobs may remain "off the plan" for a long period after submission?  (The default queue depth appears to be [quite low](https://github.com/flux-framework/flux-sched/blob/409de4e9f32593e59603bdbfcd1809f4d88084b6/qmanager/config/queue_system_defaults.hpp#L17) so hopefully I'm misinterpreting what that means)

jameshcorbett: > When we say "scheduled" I assume that means "future start time planned" not "allocated"? (Since obviously we expect jobs to be pending for long periods of time in a batch environment)

Well technically by "scheduled" I meant "the job hits this check of ours that shortens the duration", rather ignorant of where exactly in the job lifecycle this check occurs. But yes I think that implies "future start time planned" rather than allocated. I have no idea how many jobs remain "off the plan" for long periods (maybe Dan knows), I was just thinking that it's not zero.

grondo (approved):
LGTM. I ran some testing locally and it seems to work as designed :+1:

resource/traversers/dfu.cpp
milroy: > When we say "scheduled" I assume that means "future start time planned" not "allocated"? (Since obviously we expect jobs to be pending for long periods of time in a batch environment)

I think either possibility is valid given what I know about the job lifecycle. 

> Does fluxion's configured "queue-depth" still potentially mean that some pending jobs may remain "off the plan" for a long period after submission?

I think this is possible given the queue depth, but I confess `qmanager` isn't a component I know much about yet.

Given your comments @garlick and @ofaaland I think shortening the duration is the right approach.

milroy:
I just force pushed a fixup to correct a typo. @trws if you get a chance to look at the `std::chrono` implementation that would be great, but we've got one outside review approval already so it probably isn't necessary.

resource/modules/resource_match.cpp
```diff
@@ -624,14 +628,48 @@ static int unpack_resources (json_t *resobj,
...
+                                       (std::numeric_limits<int64_t>::max ()))
+            goto inval;
+        // Ensure start and end are representable in system clock
+        if (start > static_cast<double>
+            (std::chrono::duration_cast<std::chrono::seconds>
+                (std::chrono::time_point<std::chrono::system_clock>::max ().time_since_epoch ()).count ())
+              || end > static_cast<double>
+                (std::chrono::duration_cast<std::chrono::seconds>
+                    (std::chrono::time_point<std::chrono::system_clock>::max ().time_since_epoch ()).count ()))
+            goto inval;
```
trws, line 661: Why the static_cast<double> after duration_cast rather than duration_cast<std::chrono::duration<double>>?  

Looking at this, I think we want some helpers to make things less verbose and a bit more readable.  I can work up something more comprehensive, but just defining a `flux::duration` and `flux::time_point` so we can use them in place of the full std::chrono::time_point<std::chrono::system_clock> mouthfull would be good.  It would also mean if we determined we want to be more specific about our duration type or change the base clock or something we could do it in one place.

resource/modules/resource_match.cpp
```diff
@@ -624,14 +628,48 @@ static int unpack_resources (json_t *resobj,
...
+        // Expects int type argument
+        duration.graph_start = std::chrono::system_clock::from_time_t (static_cast<int64_t> (start));
+        duration.graph_end = std::chrono::system_clock::from_time_t (static_cast<int64_t> (end));
```
trws, line 664: These should work directly as arguments to the constructor of a time_point or duration defined using a double as its storage, then let you do the conversion safely.  Something like `std::chrono::time_point<system_clock, duration<double>>(duration<double>(start))` should create the appropriate time_point.  Sadly a lot of the convenience stuff is in a newer header version, we should probably pull in the portable implementation from [date](https://github.com/HowardHinnant/date) so it's just `round<seconds>(sys_time<duration<double>>(duration<double>(start)))`, little more readable.

trws:
I'm good with this, it looks like it covers the case well, but I'm going to put up an issue to add some chrono helpers so we can pull down the verbosity a bit and keep things readable.

PR #971, opened 2022-09-17
qmanager: drop default-queue configuration

garlick:
This drops the `default-queue` key from the qmanager configuration since it is now the ingest module's job to assign a default queue if one is configured per RFC 33.  

In addition, update the sharness tests that use queues to include RFC 33 style config in addition to the qmanager config, which should help ease the transition to the new style.

### DIFF
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -147,10 +147,6 @@ qmanager_opts_t::qmanager_opts_t ()
               static_cast<int> (qmanager_opts_key_t::QUEUES)));
     inserted &= ret.second;
     ret = m_tab.insert (std::pair<std::string, int> (
-              "default-queue",
-              static_cast<int> (qmanager_opts_key_t::DEFAULT_QUEUE)));
-    inserted &= ret.second;
-    ret = m_tab.insert (std::pair<std::string, int> (
               "queue-policy",
               static_cast<int> (qmanager_opts_key_t::QUEUE_POLICY)));
     inserted &= ret.second;
@@ -179,11 +175,6 @@ qmanager_opts_t::qmanager_opts_t ()
         throw std::bad_alloc ();
 }
 
-void qmanager_opts_t::set_default_queue (const std::string &o)
-{
-    m_default_queue = o;
-}
-
 bool qmanager_opts_t::set_queue_policy (const std::string &o)
 {
     return m_queue_prop.set_queue_policy (o);
@@ -209,11 +200,6 @@ const queue_prop_t &qmanager_opts_t::get_queue_prop () const
     return m_queue_prop;
 }
 
-const std::string &qmanager_opts_t::get_default_queue () const
-{
-    return m_default_queue;
-}
-
 const std::string &qmanager_opts_t::get_queue_policy () const
 {
     return m_queue_prop.get_queue_policy ();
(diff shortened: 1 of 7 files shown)


garlick:
Tacked on a commit to stop annotating the job with the assigned queue, since queues aren't assigned by fluxion anymore.

grondo (approved):
This LGTM on a cursory inspection. Perhaps we should get @milroy's approval as well, but it seems important to get this moving forward for queues work in flux-core, so I'm fine if you want to set MWP now.

garlick:
I just pushed a few more commits to fix up a minor RFC 27 violation in the annotation, which was also complicating the annotation test, which was also not really working right.  This could be split off to a follow-on PR if that is preferred.

garlick:
Eh, yeah I'll split that out to another PR.

garlick:
(Sorry for the noise - the extra commits are gone now)

qmanager/modules/qmanager_opts.cpp
```diff
@@ -258,17 +239,14 @@ bool qmanager_opts_t::is_policy_params_set () const
 qmanager_opts_t &qmanager_opts_t::canonicalize ()
 {
     if (m_per_queue_prop.empty ()) {
-        std::string qn = is_default_queue_set ()? m_default_queue
-                                                : m_default_queue_name;
+        std::string qn = m_default_queue_name;
```
jameshcorbett, line 242: So what is this `m_default_queue_name` now, if the default queue can't be configured?

jameshcorbett (approved):
This looks good to me, but I have two questions. The first is in the comment below. The second is, if Fluxion is no longer assigning queues, do the qmanager queues have any role? Or are Fluxion and flux-core queues supposed to align?

qmanager/modules/qmanager_opts.cpp
garlick: `m_default_queue_name` is always set to the string "default".

When no queues are configured, flux-core will not set the jobspec `system.queue` attribute.  My assumption is that fluxion still needs its one queue to have a name, and and uses `m_default_queue_name` in this case.

I'm open to further cleaning this up if you think this leaves a lingering problem. I didn't dig too deeply here because I'm really not too familiar with the code and might need a little guidance.

garlick:
> if Fluxion is no longer assigning queues, do the qmanager queues have any role? Or are Fluxion and flux-core queues supposed to align?

Yes.  flux-core still has only one queue in the job manager.   fluxion receives all jobs that have entered SCHED state from the one job manager queue as alloc requests.  It then places those requests in the configured queues based on the jobspec `system.queue` value, in priority order.

This is a stopgap until we can expend  time properly building multiple queues into flux core, and it has the following limitations:
- the scheduler must use the "unlimited" mode of the RFC 27 resource allocation protocol, otherwise jobs that could run in one queue could be blocked by jobs that can't run in another queue
- `flux queue` can only control the one queue, not individual queues
- queues all use the same priority scheme (e.g. flux-accounting fair share applies to all or none)

qmanager/modules/qmanager_opts.cpp
jameshcorbett: No I don't think I have any special insight, just wanted to make sure I understood the code.

garlick:
I'll go ahead and set MWP, thanks!

PR #974, opened 2022-09-21
Fix REAPI C++ bindings 

tpatki:
C++ bindings have an issue where match_allocate doesn't set allocation, reservation or job info correctly. This was done in the C bindings, which is not the right place for it.

### DIFF
--- a/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/hlapi/bindings/c++/reapi_cli_impl.hpp
@@ -60,11 +60,13 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
     int rc = -1;
     at = 0;
     ov = 0.0f;
+    job_lifecycle_t st;
+    std::shared_ptr<job_info_t> job_info = nullptr;
     struct timeval start_time, end_time;
+    std::stringstream o;
 
     try {
         Flux::Jobspec::Jobspec job {jobspec};
-        std::stringstream o;
 
         if ( (rc = gettimeofday (&start_time, NULL)) < 0) {
             m_err_msg += __FUNCTION__;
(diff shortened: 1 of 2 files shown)


resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -97,7 +99,29 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
...
+            m_err_msg += ": ERROR: can't allocate memory: " 
+                         + std::string (strerror (errno))+ "\n"; 
+            rc = -1; 
+            goto out; 
+        } 
+ 
+        rq->set_job (jobid, job_info); 
+        rq->incr_job_counter ();
+                         
+        if ((rc = gettimeofday (&end_time, NULL)) < 0) {
```
milroy, line 124: Nit: Fluxion uses a space between double open parentheses for conditionals.

```suggestion
        if ( (rc = gettimeofday (&end_time, NULL)) < 0) {
```

resource/hlapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -97,7 +99,29 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
...
+        reserved = (at != 0)? true : false; 
+        st = (reserved)? 
+                    job_lifecycle_t::RESERVED : job_lifecycle_t::ALLOCATED; 
+        if (reserved) 
+            rq->set_reservation (jobid); 
+        else
+            rq->set_allocation (jobid);
+ 
+        job_info = std::make_shared<job_info_t> (jobid, st, at, "", "", ov); 
+        if (job_info == NULL) {
```
milroy, line 112: The original location wrapped this in a `try...catch`. The `make_shared` [documentation](https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared) mentions that it: "may throw [std::bad_alloc] or any exception thrown by the constructor of T. If an exception is thrown, the functions have no effect." I like the check for `NULL` since that appears to handle more potential failure modes. However, I think it's better to check for `nullptr`:

```suggestion
        if (job_info == nullptr) {
````

File: resource/hlapi/bindings/c++/reapi_cli_impl.hpp [hunk 3]
```diff
@@ -97,7 +99,29 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
 
         R = o.str ();
 
-        if ( (rc = gettimeofday (&end_time, NULL)) < 0) {
+
```
milroy, line 102: Since this PR is already an API fixup, perhaps we can address a potentially small problem here. I think it's better to end the top-level `try...catch` block between the `if (rq->get_traverser_err_msg () != "") {` block and the `if ( (rc = rq->writers->emit (o)) < 0) {` block.

What do you think @tpatki?

milroy (changes requested):
Thanks a lot for finding this problem and fixing it @tpatki! There are a few minor changes that need to be made but this looks great overall.

tpatki:
@milroy I incorporated all the changes you had requested. I believe this is ready to go.

jameshcorbett (approved):
Looks reasonable to me.

milroy:
The changes look good. Thanks again @tpatki for the quick fix! Setting MWP.

PR #976, opened 2022-09-26
testsuite: cover queues with non-overlapping resource constraints

garlick:
This adds a sharness test with  high level testing for queues with non-overlapping (partitioned) resources.  Nothing too fancy, just ensuring that the scheduler remains responsive when both queues are processing active jobs, a node gets drained, etc..

There's also some test cleanup, enabled by recent flux-core changes.


grondo (approved):
LGTM!

garlick:
Thanks!  I'll set MWP.

PR #978, opened 2022-09-30
testsuite: fix coverage method for queue exception

garlick:
Problem: t1006-qmanager-multiqueue.t exploits the fact that jobs may be submitted with incorrect queue specifications when the frobnicator is disabled, but this method won't work once the job manager becomes queue-aware.

Submit test jobs with qmanager unloaded, then reconfigure queues and reload qmanager to trigger the job exceptions.

(This test was failing in CI for flux-framework/flux-core#4627)

### DIFF
--- a/t/t1006-qmanager-multiqueue.t
+++ b/t/t1006-qmanager-multiqueue.t
@@ -157,38 +157,40 @@ test_expect_success 'qmanager: incorrect queue policy can be caught' '
 	flux dmesg | grep "Unknown queuing policy"
 '
 
-test_expect_success 'reconfigure queues with frobnicator disabled' '
+test_expect_success 'unload qmanager and deconfigure queues' '
+	remove_qmanager &&
+	cp /dev/null config/queues.toml &&
+	flux config reload
+'
+test_expect_success 'submit job with no queue' '
+	flux mini submit /bin/true >noqueue.jobid
+'
+test_expect_success 'reconfigure with one queue and load qmanager' '
 	cat >config/queues.toml <<-EOT &&
-	[ingest]
-	frobnicator.disable = true
-
 	[queues.foo]
-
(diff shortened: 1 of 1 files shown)


t/t1006-qmanager-multiqueue.t
```diff
@@ -157,38 +157,40 @@ test_expect_success 'qmanager: incorrect queue policy can be caught' '
...
 	queues = "foo"
 	EOT
 	flux config reload &&
-	reload_qmanager
+	load_qmanager
 '
-
 test_expect_success 'job submitted with no queue gets fatal exception' '
-	test_must_fail flux mini run /bin/true 2>noqueue.err &&
+	test_must_fail flux job attach $(cat <noqueue.jobid) 2>noqueue.err &&
```
grondo, line 178: Just for my own curiosity, how is `cat <noqueue.jobid` different from `cat noqueue.jobid`?

grondo (approved):
LGTM!

jameshcorbett (approved):
Looks reasonable to me. Jobs submitted when qmanager is not loaded are held in SCHED?

t/t1006-qmanager-multiqueue.t
garlick: Derp.  My fingers typed that.  I guess I'm used to doing that in these tests for grep, since if you put the grep file on the command line, it appears in the output and you have to parse it out again.  Will fix.

garlick:
> Jobs submitted when qmanager is not loaded are held in SCHED?

Correct.

t/t1006-qmanager-multiqueue.t
grondo: NP! (Darn, I thought I was going to learn something)

garlick:
Setting MWP, thanks!

PR #981, opened 2022-10-04
NEWS: add release notes for 0.25.0

garlick:
Here are some tentative release notes for 0.25.0 that assume we merge #980.  Feel free to push updates directly to this branch.

I tried building the current master against flux-core 0.43.0 on fluke and  encountered test failures, hence the note about the test suite requiring 0.44.0 or newer.  I didn't investigate deeply, but did note that some tests now use `flux mini CMD --queue` which  was only recently added.  I suspect the scheduler will work fine on older releases but since that has not been tried, I figured the test suite failures were enough to justify that note.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,31 @@
+flux-sched version 0.25.0 - 2022-10-04
+--------------------------------------
+
+Note: the flux-sched test suite requires flux-core 0.44.0 or newer.
+
+### New Features
+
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Thanks!

grondo:
Setting MWP so we can get v0.25.0 tagged

grondo:
@mergifyio refresh

PR #985, opened 2022-10-22
testsuite: update flux-tree-helper.py for new OutputFormat constructor

grondo:
Problem: The flux.util.OutputFormat class in flux-core will get an improved constructor where `headings` becomes an optional keyword argument, but this will break the PerfOutputFormat class in t/scripts/flux-tree-helper.py

Try the new-style constructor arguments in PerfOutputFormat. If that fails, fall back to the old style so that either works for the flux-tree-helper.py test script.

This PR will need to be merged to avoid a future PR in flux-core from breaking the `flux-tree` tests in flux-sched.


garlick (approved):
LGTM!

PR #989, opened 2022-12-01
Enable Fluxion resource graph elasticity

milroy:
This PR enables elasticity in the Fluxion resource graph by creating requisite ctors, dtors, and assignment overloads for `planner`, and `planner_multi`.

As discussed in PR #775, attaching a new subgraph to the boost resource graph without preallocating sufficient vertices in the `vecS` associative Boost container triggers a copy. Since `planner` did not have a copy constructor to deep copy the red-black trees and related data strctures, the copy cleared the RBtrees and allocations and reservations.

By defining copy constructors for these classes and data structures, subgraph `attach` and future subgraph `shrink` will preserve allocations and reservations.

This PR refactors `planner` and `planner_multi` from structs to classes. Operations on both classes (e.g., setters and getters) have been pulled into the class as member functions, and access controlled via `const` member functions where appropriate.

~The PR is WIP, as there are still several tasks to complete which may require reviewer feedback:~

- [x] Review the current header (`.h` and `.hpp`) structure for compatibility with external C clients.
- [x] Consider the current header (`.h` and `.hpp`) structure for compatibility with external C++ clients.
- [x] Review the usage of wrapper context structs `planner_t` and `planner_multi_t` to avoid errors related to forward declarations.
- [x] Add more robust error checking where noted in code comments
- [x] Add testsuite tests for the `planner_multi` and `pool_infra_t` ctors and assignment overload for elastic resources.

I've changed the `planner` interface as well as the build system to accommodate C and C++ clients. I added unit tests for ctors and assignment overload operator for `planner_t` and `planner_multi_t`, and created new unit tests for ctors and assignment overload operator for schema (both `schedule_t` and `pool_infra_t`).


milroy:
@JaeseungYeom and @tpatki the PR is ready for review.

resource/planner/c/planner_c_interface.cpp
```diff
@@ -377,6 +318,36 @@ extern "C" planner_t *planner_new (int64_t base_time, uint64_t duration,
     return ctx;
 }
 
+extern "C" planner_t *planner_copy (planner_t *p)
+{
+    planner_t *ctx = nullptr;
+
+    try {
+        ctx = new planner_t ();
```
trws, line 326: This doesn't get deleted if the planner below throws bad_alloc.  Might be easier to avoid if the planner were part of the constructor for the concrete planner_t?

resource/planner/c/planner_c_interface.cpp
```diff
@@ -377,6 +318,36 @@ extern "C" planner_t *planner_new (int64_t base_time, uint64_t duration,
...
+        ctx->plan = nullptr;
```
trws, line 327:
```suggestion
```

resource/planner/c/planner_c_interface.cpp
```diff
@@ -377,6 +318,36 @@ extern "C" planner_t *planner_new (int64_t base_time, uint64_t duration,
...
+
+    return ctx;
+}
+
+extern "C" planner_t *planner_new_empty ()
+{
+    planner_t *ctx = nullptr;
+
+    try {
+        ctx = new planner_t ();
```
trws, line 341: Same as above.

resource/planner/c/planner_multi.h
```diff
@@ -12,13 +12,12 @@
 #define PLANNER_MULTI_H
 
 #include "planner.h"
+#include "resource/planner/c++/planner_multi.hpp"
```
trws, line 15: Is this ok here? If this is meant to be included from C I mean.

resource/planner/c/planner_multi_c_interface.cpp
```diff
@@ -51,38 +42,25 @@ extern "C" planner_multi_t *planner_multi_new (
...
-    planner_t *p;
 
     if (duration < 1 || !resource_totals || !resource_types) {
         errno = EINVAL;
         goto error;
     } else {
         for (i = 0; i < len; ++i) {
-            if (resource_totals[i] > std::numeric_limits<int64_t>::max ()) {
+            if (resource_totals[i] >
+              static_cast<uint64_t> (std::numeric_limits<int64_t>::max ())) {
```
trws, line 52: I think this is the third or fourth of these I've seen in this review.  Might be good to have a helper function for this check.

resource/schema/infra_data.cpp
```diff
@@ -56,62 +56,94 @@ pool_infra_t::pool_infra_t ()
...
+        return false;
+    for (auto const &this_it : colors) {
+        auto const other = o.colors.find (this_it.first);
+        if (other == o.colors.end ())
+            return false;
+        if (this_it.first != other->first)
+            return false;
+        if (this_it.second != other->second)
+            return false;
+    }
```
trws, line 129: I'm relatively sure this is equivalent, if so it's a bit better self-documenting and somewhat faster (avoiding the O(logN) search on each element):

```suggestion
    if(colors.size() != o.colors.size()
       || !std::equal(this.colors.begin(), this.colors.end(), o.colors.begin()))
       return false;
```

Makes me want ranges, very badly, and autogenerated `<==>`, _sigh_.  Either way, same below.

trws: Actually, why is this one being compared differently than tags? The same `==` should work here as well for complete equality right?

resource/schema/sched_data.cpp
```diff
@@ -20,41 +20,70 @@ schedule_t::schedule_t ()
...
+    return true;
+}
+
 schedule_t::~schedule_t ()
 {
     if (plans)
         planner_destroy (&plans);
+
+    allocations.clear ();
+    reservations.clear ();
```
trws, line 86: These will get called automatically.
```suggestion
```

resource/planner/c++/planner.cpp
```diff
@@ -0,0 +1,536 @@
...
+    // Empty planner with essential members
+    // set for comparsons.
+    m_total_resources = 0;
+    m_resource_type = "";
+    m_plan_start = 0;
+    m_plan_end = 0;
+    m_p0 = nullptr;
+    m_avail_time_iter_set = 0;
+    m_span_counter = 0;
+}
```
trws, line 34: This is equivalent to `planner::planner() = default;`.  If you'd prefer to be explicit about these values, prefer initializing them in the class declaration in the header.

resource/planner/c++/planner.cpp
```diff
@@ -0,0 +1,536 @@
...
+            if ( (rc = m_mt_resource_tree.insert (new_point)) != 0)
+                return rc;
+            point = o.m_sched_point_tree.next (point);
+        }
+    } else {
+        // Erase if incoming planner is empty.
+        rc = erase ();
+    }
+    return rc;
+}
```
trws, line 397: We clearly need this, but perhaps make this the implementation of the copy constructor for the scheduled_point_tree_t type?

trws:
Just finished a pass over this, generally looks good but there are a number of arrangement bits I have questions about, and concerns about some of the C interface headers including C++ headers.

milroy:
@trws thanks for the detailed feedback. I'm in the process of updating the PR to implement your suggestions.

I rebased against master to check the PR against the updated CI tests and have run into compilation errors on the bookworm tests:

```bash
    CXX      libplanner_cxx_la-mintime_resource_tree.lo
  In file included from ../../../../../src/common/yggdrasil/rbtree.hpp:338,
                   from ../../../../../resource/planner/c++/scheduled_point_tree.hpp:15,
                   from ../../../../../resource/planner/c++/planner_internal_tree.hpp:14,
                   from ../../../../../resource/planner/c++/mintime_resource_tree.cpp:20:
  ../../../../../src/common/yggdrasil/rbtree.cpp: In member function 'void ygg::RBTree<Node, NodeTraits, Options, Tag, Compare>::insert(Node&, iterator<false>)':
  ../../../../../src/common/yggdrasil/rbtree.cpp:516:31: error: expected 'template' keyword before dependent template name [-Werror=missing-template-keyword]
    516 |                         this->insert_leaf_base<false>(node, parent);
        |                               ^~~~~~~~~~~~~~~~
```
The error seems odd because this PR doesn't change anything in yggdrasil and the CI tests passed for the latest commits to master.

trws:
It's possible that the warning suppression was in a header surrounding that include before, or that something else was suppressing it.  I can fix this more correctly in the cmake PR, if you would please mark a TODO to fix it in the automake file, and add the `-Wno-error=missing-template-keyword' to the CXXFLAGS for the appropriate file or files, that will get this done so we can make yggdrasil a system include in the other one (system includes don't get warning checked by default, there are boost issues in this class too, that's why I ended up going down the cmake rabbit hole in the first place).

milroy:
Adding `-Wno-error=missing-template-keyword` and the related `-Wno-error=maybe-uninitialized` to the appropriate makefiles breaks the Clang tests since Clang doesn't recognize those diagnostic flags. Is there a preferred way to conditionally add the diagnostic `CXXFLAGS` if the compiler is GNU?

trws:
It's more work, but the best way is probably to use the suppression directives. It needs to be something like this (clang first because it claims to be gcc, as many do)

~~~
#ifdef __clang__
/*clang*/
#elif __GNUC__

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wuninitialized"
    // include stuff here 
#pragma GCC diagnostic pop

#endif
~~~

Alternately just guarding the diagnostic ignored would work, but that would disable it for the entire TU, so I'd like to get rid of it fast if we do that.

milroy:
Adding the conditional pragma with the suppression directive works for the Clang and GCC 12 tests but breaks the earlier GCC tests. That's because `#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"` and `-Wmissing-template-keyword` are unrecognized directives for earlier versions of GCC (or conflict with `-Wno-error=maybe-uninitialized` from `generate-matrix.py`), but seem to be required for later versions (i.e., `-Wno-error=maybe-uninitialized"` doesn't work for GCC 12).

I added additional checks for `(__GNUC__ > 11)` or `(__GNUC__ <= 11)` which allows the CI tests to pass. I'm not entirely sure if the behavior changed with GCC 11, though, and haven't figured out exactly which version caused the change.

trws:
That sounds plenty good enough. Sorry that was a bit of a siege, we’ll get everything straightened out so we don’t need to deal with this kind of thing in external deps again.
---
On August 20, 2023 at 3:58:17 PM PDT, Daniel Milroy ***@***.***> wrote:

Adding the conditional pragma with the suppression directive works for the Clang and GCC 12 tests but breaks the earlier GCC tests. That's because #pragma GCC diagnostic ignored "-Wmaybe-uninitialized" and -Wmissing-template-keyword are unrecognized directives for earlier versions of GCC (or conflict with -Wno-error=maybe-uninitialized from generate-matrix.py), but seem to be required for later versions (i.e., -Wno-error=maybe-uninitialized" doesn't work for GCC 12).

I added additional checks for (__GNUC__ > 11) or (__GNUC__ <= 11) which allows the CI tests to pass. I'm not entirely sure if the behavior changed with GCC 11, though, and haven't figured out exactly which version caused the change.

resource/schema/infra_data.cpp
milroy: Actually, yes, `if (colors != o.colors)` works, too.

Regarding comparing `subplans`, this strategy doesn't work, because it's a map of `planner_multi_t` which is a wrapper struct around the `planner_multi` class. Adding an `==` operator to `planner_multi_t` doesn't work with `std::equal`, unfortunately.

resource/planner/c/planner_multi_c_interface.cpp
milroy: A search through the Fluxion source shows that checks like this are present in five lines of code across three files in this PR, but just once in each of the two `*_c_interface.cpp`. The others are in `mintime_resource_tree.cpp` and don't involve comparisons between `int64_t` and `uint64_t`. Should I create a helper function in `resource/planner/c++/`?

trws: Eh, if there are only 5 I guess leave it alone. It's not worth setting up new infrastructure for.

milroy: There are 10 other similar comparisons across Fluxion (mostly in the traverser). It may be worth while to have a helper function somewhere in top-level resource, but it's not clear to me where to put it. I'll create an issue and we can create a separate PR.

resource/planner/c++/planner.cpp
milroy: As we discussed the other week during the hackathon the necessity of inserting the `new_point` into both trees will make it difficult to implement. Perhaps we could create a wrapper class that contains `m_sched_point_tree` and `m_mt_resource_tree` as members and then define a copy constructor for that wrapper class.

That would require an overhaul of the planners' C interfaces (but not the public API at least), so maybe we can consider that for a future PR.

resource/planner/c++/planner.cpp
milroy: Implemented for `planner` and `planner_multi` classes.

milroy:
Thanks for your detailed and helpful feedback @trws. I think I've addressed and implemented your comments and suggestions.

The PR is ready for another review.

trws (approved):
I'm good with this, thanks @milroy!  There's always refactoring I'd love to do, but I don't think it makes sense to do it before we bump the C++ version.  Delegated constructors and some other things could help us with code like this in the future.

milroy:
@trws should I set MWP, or do you think this needs a review from someone else as well? 

Maybe @tpatki @JaeseungYeom or @jameshcorbett could spot check a few of the commits.

trws:
Usually I prefer extra reviews, but we all looked this over together and the issues were all cleanup (mostly in code from before too). It might make sense to get it in so we can get on rebasing the next set.
---
On August 21, 2023 at 7:24:08 PM PDT, Daniel Milroy ***@***.***> wrote:

@trws<https://urldefense.us/v3/__https://github.com/trws__;!!G2kpM7uM-TzIFchu!2zQzC-ZIOaOT214sAgnLGAfDlZuU70jILWgwjz2nYnwArOKxj4nJZ197s8yU631Ujh_SqGeX-ODCshk9ZOjtw0WVoys$> should I set MWP, or do you think this needs a review from someone else as well?

Maybe @tpatki<https://urldefense.us/v3/__https://github.com/tpatki__;!!G2kpM7uM-TzIFchu!2zQzC-ZIOaOT214sAgnLGAfDlZuU70jILWgwjz2nYnwArOKxj4nJZ197s8yU631Ujh_SqGeX-ODCshk9ZOjtyzAiDW0$> @JaeseungYeom<https://urldefense.us/v3/__https://github.com/JaeseungYeom__;!!G2kpM7uM-TzIFchu!2zQzC-ZIOaOT214sAgnLGAfDlZuU70jILWgwjz2nYnwArOKxj4nJZ197s8yU631Ujh_SqGeX-ODCshk9ZOjtsZA-8ao$> or @jameshcorbett<https://urldefense.us/v3/__https://github.com/jameshcorbett__;!!G2kpM7uM-TzIFchu!2zQzC-ZIOaOT214sAgnLGAfDlZuU70jILWgwjz2nYnwArOKxj4nJZ197s8yU631Ujh_SqGeX-ODCshk9ZOjt26CaU90$> could spot check a few of the commits.

tpatki:
+1 with merging @trws  @milroy -- we've reviewed it together, and looked at it a few times (I looked at it in Feb as well). The changes make sense to me and seem correct overall.  I haven't had a chance to test it locally at my end, and I will do that soon -- but we don't need to hold on merging this for my testing. We can always open an issue if something comes up when I try it out.

milroy:
Thanks everyone! Setting MWP.

trws:
@milroy, I just pushed up a CI config fix. This should let everything merge.

PR #992, opened 2022-12-08
testsuite: start/stop all queues with --all option

chu11:
Problem: flux-core PR 4776 added the ability for queues to be individually started and stopped.  As an update, starting/stopping all queues with the `flux queue` command now requires the --all option to be specified.  This is not specified in several flux-sched tests, leading to test failures.

Update all callers to specify --all when multiple queues are started or stopped.

***This should not be merged until https://github.com/flux-framework/flux-core/pull/4776 is merged***

### DIFF
--- a/t/sharness.d/sched-sharness.sh
+++ b/t/sharness.d/sched-sharness.sh
@@ -91,7 +91,7 @@ load_test_resources () {
 
 # N.B. this assumes that a scheduler is loaded
 cleanup_active_jobs () {
-    flux queue stop &&
+    flux queue stop --all &&
         flux job cancelall -f &&
         flux queue idle
 }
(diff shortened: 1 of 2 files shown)


garlick:
I think I'm contradicting myself from a couple of days ago, but should we just allow `flux queue start` and `flux queue stop` to apply to all queues as before and avoid this breakage?   If queues are configured and `--all` is unspecified, we could print a message like "warning: applying this command to all queues"?

It would be nice not to break an external interface, at least not without a transition period.  What do you think?

chu11:
ok, lets move to try and make `flux queue start/stop` backwards compatible.  closing

chu11:
re-opening, although we're supporting backwards compatibility, we would prefer that tests do not output an unnecessary warning message.  So we'd still like to do this change.  It is just no longer "breaking" if we don't do it.  We can merge at our leisure after https://github.com/flux-framework/flux-core/pull/4776 is merged.

garlick (approved):
LGTM!

PR #996, opened 2023-01-10
testsuite: do not assume queues started by default

chu11:
Problem: In several tests, it is assumed that newly created named queues are "started" by default.  This is not a safe assumption going forward.

Solution: In tests that create named queues, start the queues immediately after they are configured.

This PR is needed to ensure `flux-sched` can pass once PR https://github.com/flux-framework/flux-core/pull/4857 is merged.  It is safe to merge before that PR (and necessary for that PR to pass `flux-sched` tests).


garlick (approved):
Looks good!

Issue #991, opened 2022-12-01
Fluxion can't restart running jobs with match-format `rv1_nosched`


grondo:
The Flux system instance needed to be restarted on tioga recently and there were two active jobs in CLEANUP state. This caused Fluxion to fail to restart with the following errors:
```
[  +0.001134] job-manager[0]: scheduler: hello
[  +0.002056] sched-fluxion-resource[0]: parse_R: no scheduling key in R
[  +0.002077] sched-fluxion-resource[0]: run_update: parsing R: No such file or directory
[  +0.002084] sched-fluxion-resource[0]: update_request_cb: update failed (id=138630396509161472): No such file or directory
[  +0.002316] sched-fluxion-qmanager[0]: jobmanager_hello_cb: reconstruct (id=138630396509161472 queue=default): No such file or directory
[  +0.002330] sched-fluxion-qmanager[0]: hello: error loading R for id=138630396509161472: No such file or directory
[  +0.002347] sched-fluxion-qmanager[0]: handshake_jobmanager: schedutil_hello: No such file or directory
[  +0.002352] sched-fluxion-qmanager[0]: handshake: handshake_jobmanager: No such file or directory
[  +0.002362] sched-fluxion-qmanager[0]: mod_start: handshake: No such file or directory
[  +0.002728] sched-fluxion-qmanager[0]: service_unregister
[  +0.002759] sched-fluxion-qmanager[0]: module exiting abnormally
```

It appears that `parse_R()` in `resource/modules/resource_match.cpp` _always_ requires a `scheduling` key be set with JGF, but this will not be the case for any job when `sched-fluxion-resource.match-format = "rv1_nosched"`.

Since Fluxion supports a rv1exec reader, `parse_R()` should fall back to this reader when there is no scheduling key in R for a job.

grondo:
I verified this case (reloading `sched-fluxion-resource` with `match-format=rv1_nosched`) is missed in the Fluxion testsuite. Strangely, many other cases are tested, so I wonder if this was purposeful?

Anyway, adding this test to the testsuite demonstrates the issue:
```diff
diff --git a/t/t1007-recovery-full.t b/t/t1007-recovery-full.t
index 5b2de636..0923eeaa 100755
--- a/t/t1007-recovery-full.t
+++ b/t/t1007-recovery-full.t
@@ -150,6 +150,18 @@ test_expect_success 'recovery: qmanager restarts (rv1_nosched->rv1_nosched)' '
[...]

PR #1000, opened 2023-01-28
testsuite: adjust expectations of recovery in rv1_nosched mode

garlick:
This alters testing a bit in preparation for a workaround and ultimately a proper fix for #991. 

The `t1008-recovery-none.t` sharness test is dropped since it really just contains one recovery test that expects things to fail as they have been.  Replace that with a new expected-failure test in `t1007-recovery-full` that is a slightly modified version of the one proposed by @grondo in #991.  This test should fail as expected before and after the workaround proposed in flux-framework/flux-core#4894, and when we have a proper fix for #991, it can be flipped to expect success.

After this is merged, the failing sched CI test in flux-framework/flux-core#4894 should start working.


grondo (approved):
This looks reasonable to me.

PR #1005, opened 2023-02-07
resource: fix 'Internal match error' when hostlist constraint is provided

### LINKED ISSUE #1004: error messages from resource module need more improvement
As noted by @garlick in #1002, there are some other error paths in the Fluxion resource module that should return better or correct errors to the user:
```
# should this be unsatisfiable?
$ flux mini run --requires "host:picl1 rank:2" hostname
flux-mini: ERROR: [Errno 2] Internal match error: 

# should this be unsatisfiable?
$ flux mini run --requires host:xyz hostname
flux-mini: ERROR: [Errno 2] Internal match error: 
```

I believe these errors are coming from the feasibility jobspec plugin, so are the result of just a satisfiability match in resource module.

grondo:
This PR fixes cases of the resource module returning "Internal match error" when job requests should be "unsatisfiable" and a hostlist constraint is provided. The problem is that the hostlist constraint implementation overwrites `errno`, confusing the resource module which is looking specifically for `ENODATA`.

Fixes #1004

### DIFF
--- a/resource/libjobspec/hostlist_constraint.cpp
+++ b/resource/libjobspec/hostlist_constraint.cpp
@@ -34,9 +34,10 @@ HostlistConstraint::HostlistConstraint (const YAML::Node &values)
 
 bool HostlistConstraint::match (const Flux::resource_model::resource_t &r) const
 {
-    if (hostlist_find (hl, r.name.c_str()) < 0)
-        return false;
-    return true;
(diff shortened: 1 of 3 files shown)


garlick (approved):
Gave this a quick test and looks good!

grondo:
Thanks setting MWP!

Issue #1001, opened 2023-02-04
Job submission slows down on Hetchy


jameshcorbett:
```
[corbett8@hetchy12:Flux]$ for (( i=0; i<50; i++ )); do /usr/bin/time --format="%e" flux mini submit -N1 hostname; done
ƒMbBRgJXJ7Z
0.67
ƒMbBRrMARs5
0.39
ƒMbBS2HscVD
0.38
ƒMbBSGKnpCX
0.55
ƒMbBSfMZf7D
0.89
ƒMbBTD9rFgB
1.24
ƒMbBTusYXbV
1.59
ƒMbBUmcaRzX
1.94
ƒMbBVnFY2TZ
2.29
ƒMbBWwjTLRu
2.64
ƒMbBYGLeE1M
2.99
ƒMbBZjjprXR
3.34
... [snip]
ƒMbCkEwA7P5
11.29
ƒMbCqRWQW9u
11.75
... [snip]
ƒMbDajaVaLX
24.42
```

grondo:
A good test may be to try reloading `sched-fluxion-qmanager` and `sched-fluxion-resource` to see if the problem goes away. However, we may want to collect as much information from the affected system before doing that to better understand the problem. cc: @trws and @milroy to see if they have any other ideas.

grondo:
I started a test instance with the same R as configured on hetchy and could not reproduce the issue, so the cause here isn't the specific configuration of resources. Not sure how to debug the live system.

garlick:
In case it proves useful in recreating this, the current resource state and properties (queue) config is
```
$ flux resource list
     STATE PROPERTIES NNODES   NCORES    NGPUS NODELIST
      free windom         13     1664        0 hetchy[13,16-27]
      free bardpeak        2      128       16 hetchy[28-29]
 allocated                 0        0        0 
      down windom          2      256        0 hetchy[14-15]
$ flux resource status
    STATUS NNODES NODELIST
     avail     15 hetchy[13,16-29]
   offline      2 hetchy[14-15]
   exclude      2 hetchy[7,12]
   drained      2 hetchy[14-15]
```

grondo:
Getting a similar result from `perf`:
```
-   99.93%     0.00%  flux-broker-0  [.] ev_run                                                                                                         ▒
     ev_run                                                                                                                                             ▒
   - ev_run                                                                                                                                             ▒
      - 99.93% ev_invoke_pending                                                                                                                        ▒
[...]

PR #1007, opened 2023-02-08
Improve Fluxion match performance

milroy:
As reported in issue #1001 `flux mini submit` experiences pronounced slowdown for most of the Fluxion match policies except `first`. With the debugging help of @grondo and @trws we traced one source to an un-inlined accumulate operation. 
Adding the `inline` hint to the function allows the compiler to properly inline it and performance improves by a factor of >100 for all policies except `lonodex` and `hinodex` (or, e.g., `lonode` with `--exclusive -N1`.
Another hotspot we encountered in debugging is the `accumulate` operation in `calc_multilevel_scores`: https://github.com/flux-framework/flux-sched/blob/f7bdcb508d4e93ac880845a40d670c03bebf0306/resource/policies/dfu_match_multilevel_id_impl.hpp#L71
This O(_n_) `accumulate` operation is called for each `finish_vtx`, and even worse, currently scales linearly with the number of matches. This is because the accumulated vector isn't cleared after each traversal. We can replace the `accumulate` operation by a single `int64_t` which stores the prefix value.

While these changes improve performance for all policies, `lonodex` and `hinodex` are still slow. Future PRs will further improve performance by optimizing graph filtration and lookups on edge iteration.

### DIFF
--- a/resource/policies/dfu_match_multilevel_id.hpp
+++ b/resource/policies/dfu_match_multilevel_id.hpp
@@ -107,7 +107,8 @@ private:
                         unsigned int add_by,
                         unsigned int multiply_by);
         score_factor_t (const score_factor_t &o) = default;
-        int64_t calc_factor (int64_t base_factor, int64_t break_tie) const;
+        int64_t calc_factor (int64_t base_factor, int64_t break_tie);
+        int64_t m_factor = 0;
 
     private:
         std::string m_type;
@@ -116,11 +117,10 @@ private:
     };
 
     void set_base_factor (const std::string &type, unsigned int id);
-    int64_t calc_multilevel_scores () const;
 
     FOLD m_comp;
     unsigned m_stop_on_k_matches = 0;
-    std::vector<int64_t> m_multilevel_scores;
+    int64_t m_multilevel_scores = 0;
     std::unordered_map<std::string, score_factor_t> m_multilevel_factors;
 };
 
--- a/resource/evaluators/fold.hpp
+++ b/resource/evaluators/fold.hpp
@@ -20,21 +20,21 @@ namespace resource_model {
 
 namespace fold {
 struct greater {
-    bool operator() (const eval_egroup_t &a, const eval_egroup_t &b) const
+    inline bool operator() (const eval_egroup_t &a, const eval_egroup_t &b) const
     {
         return a.score > b.score;
     }
 };
 
 struct less {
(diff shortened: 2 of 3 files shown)


jameshcorbett (approved):
LGTM

Edit: sorry I didn't see all the slack conversation still ongoing, so review is a bit premature.

milroy:
@grondo and @garlick : @trws, @tpatki, @JaeseungYeom, and Zeke discussed `perf` traces and further performance improvements today.

We decided that those improvements will take longer to implement and will create further PRs. The contributions here will speed up most policies and as such can be reviewed.

resource/policies/dfu_match_multilevel_id.hpp
```diff
@@ -107,7 +107,8 @@ class multilevel_id_t : public dfu_match_cb_t
                         unsigned int add_by,
                         unsigned int multiply_by);
         score_factor_t (const score_factor_t &o) = default;
-        int64_t calc_factor (int64_t base_factor, int64_t break_tie) const;
+        int64_t calc_factor (int64_t base_factor, int64_t break_tie);
```
grondo, line 110: minor nit: It looks like your editor line wrapped the commit message at about 54 columns instead of 72, making it a little difficult to read.

grondo (approved):
This LGTM and already tested without issue on hetchy.

FYI - I pulled these 2 patches into our latest RPM build for TOSS 4 so we'll get these into the next release.

Just noticed a formatting error in the 2nd commit message, but otherwise this should get merged, IMO.

Perhaps we should also reword the PR title before merging since the title will be included in release notes (since this doesn't fully fix the perf degradation)

resource/policies/dfu_match_multilevel_id.hpp
milroy: Thanks for catching that.

milroy:
> Perhaps we should also reword the PR title before merging since the title will be included in release notes (since this doesn't fully fix the perf degradation)

I'll rename to "Improve Fluxion match performance" since the changes improve performance for every policy I tried. Of course, the `nodex` policies are still slow and need to be improved further.

resource/policies/dfu_match_multilevel_id.hpp
milroy: Fixed and force pushed.

grondo:
Thanks! Feel free to set MWP when you are ready.

PR #1010, opened 2023-02-28
avoid use of deprecated flux mini commands

grondo:
This PR replaces the use of `flux mini <cmd>` with `flux <cmd>` in flux-sched.

This will cause the testsuite to fail when run using versions of flux-core <= 0.47.0, but thought I'd go ahead and post the PR in case that is acceptable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,8 +134,8 @@ flux module list
 
 Submit jobs:
 ```
-flux mini submit -N3 -n3 hostname
-flux mini submit -N3 -n3 sleep 30
+flux submit -N3 -n3 hostname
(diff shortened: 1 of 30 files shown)


garlick (approved):
LGTM!

milroy:
Can we merge this so it goes in the April Fluxion release or is there more work to be done?

grondo:
We can merge now, I was waiting until the release of flux-core with `flux-mini` deprecated started rolling out to minimize the amount of time that `make check` would break here on our clusters. I'll set MWP.

PR #1012, opened 2023-03-02
add workflow dispatch

vsoch:
Problem: we should be able to build the container on request without needing to commit.
Solution: add a workflow_dispatch event to the main workflow.


grondo (approved):
LGTM!

Issue #987, opened 2022-11-04
Deferred allocation


JaeseungYeom:
Problem: support a scheduling request for an allocation to occur at a specific time in the future.

Currently, a reservation of resources occurs as early as possible. However, for supporting workflows that benefit from running tasks across heterogeneous platforms, it is desired to synchronize multiple allocations across different child instances. Such that task 1-10 run on corona while task 11-20 "simultaneously" run on another cluster managed by Flux.
To support such use cases, two things are needed.
One is the deferred allocation capability, and the other is a means to query the allocation delay.
A parent instance can query its remote child instances to find out when is the earliest by which all the children can allocate requested resources. Then, it should be possible to allocate synchronously across instances.

Pushing the reservation time back should also consider back-filing.
To be clear, this is not the same as to try allocating at the earliest after a specific point in time.
I am not entirely sure if the existing issue #963 is the latter case or the same as this.

milroy:
> A parent instance can query its remote child instances to find out when is the earliest by which all the children can allocate requested resources. Then, it should be possible to allocate synchronously across instances.

To make sure I understand the basics (without getting into too much complexity yet) of the deferred allocation capability, this is a three-part process: 
1) submit the jobspecs to all child instances with a new `match_reserve (jobspec)` request that reserves the requested resources on each child instance at the earliest time possible and return those times. 
2) Find the latest time returned (T), and for all the child instances that returned earlier times, issue a new `match_reserve_at (jobspec, T)` which moves the reservation back to time T. 
3) Handle the case where one or more children can't satisfy `match_reserve_at (jobspec, T)`. 

Is that basically correct?

milroy:
I've confirmed that by manipulating the `at` time in `dfu_traverser_t::run`: https://github.com/flux-framework/flux-sched/blob/35d3c96bb7378ec4f6803a7e340043fd4040e8c1/resource/traversers/dfu.cpp#L277 we can achieve the desired behavior. Here I've simulated this by hardcoding `at = 3600` in `dfu_traverser_t::run` and performing a `match allocate`:

```bash
resource-query> match allocate t/data/resource/jobspecs/basics/test001.yaml
      ---------------core35[1:x]
      ------------socket1[1:x]
      ---------node1[1:s]
      ------rack0[1:s]
      ---tiny0[1:s]
INFO: =============================
INFO: JOBID=1
INFO: RESOURCES=RESERVED
INFO: SCHEDULED AT=3600
INFO: =============================
``` 

[...]

Issue #963, opened 2022-08-23
Advanced reservation support


jameshcorbett:
Need to be able to make a reservation for resources at a fixed time in the future.  Useful for many things, one example of which is a planned downtime of a specific Rabbit in order to replace a failed drive.

Also useful in a more general context than El Cap for things like DATs. 

See also https://github.com/flux-framework/flux-core/discussions/3271#discussioncomment-101280

milroy:
We've discussed this issue several times in the last few weeks, and have come up with more features needed to support the desired behavior. There are some outstanding questions related to implementing these features.

1. Reserving the Advanced Reservation will require truncating running allocations
    - Requires new functionality in `planner` to consider active spans as available and then reducing the span end time scheduled point to coincide with the Advanced Reservation start time
    - Requires notifying core of the affected jobids' new end times
    - Requires handling the case where a job is running beyond its scheduled end point, but a `free` has not yet been received from core

2. Is an Advanced Reservation an allocation (an entry in the `allocations` map) or a reservation (`reservations` map)?
    - Are there disadvantages to modeling an Advanced Reservation on an allocation?
    - If it's a reservation, it will need to be rescheduled during each loop, which means it must have the highest priority. Also, in this case, can we guarantee that the set of truncated jobs does not change? Un-truncating jobs would be complex and ugly.

milroy:
We discussed this issue in the Fluxion group meeting today. I mentioned one more disadvantage of modeling a user (as opposed to an admin DAT) advanced reservation as a reservation rather than an allocation: re-reserving the reservation during each scheduling loop can cause the start time to change, necessitating feedback to the user submitting the request.

Modeling the user reservation off of an allocation has the added advantage that the start and end times won't change because it won't get re-reserved on each scheduling loop. The fixed, future allocation means that incoming jobs with a higher priority can't reserve those resources. As @grondo mentioned in a previous meeting, users will take advantage of this feature to game the scheduler. 

We can compensate for the fixed, future allocation preventing higher-priority jobs from being reserved by charging for the future allocation before it starts. This is fair, because the user is actually requesting occupying the scheduler state space via the future allocation.

A traditional charge model is just a rectangular step function: the charge for the full allocation starts at the start time and ends at the allocation end time. I propose a charge model for this type of allocation that convolves a step function with another function (say, a Gaussian), see: https://ietresearch.onlinelibrary.wiley.com/doi/10.1049/pel2.12348, Figure 4.

[...]

PR #1013, opened 2023-03-06
Support deferred job start time

milroy:
Issues [#3271 (flux-core)](https://github.com/flux-framework/flux-core/discussions/3271#discussioncomment-101280), #987, and #963 identify useful new capability for Fluxion: to allow a user to specify a time in the future for their job to start. If the job cannot start at that time, then Fluxion should reserve the job at the earliest time.

Functionally this is very similar to "allocate_orelse_reserve" with an "at" time in the future. This PR adds the desired capability.

TODO:
- [ ] is specifying `deferred_at` in the optional jobspec system attributes the right approach?
- [ ] how should we handle the Fluxion API? Currently a `bool` determines whether `match_allocate` also reserves.


milroy:
I realized the PR will need updating. As it stands the job will never run, because each time qmanager iterates through the jobs and reaches the deferred job it will set the at time to be _**n**_ (`deferred_start`) seconds in the future. We will need to specify `deferred_start` seconds in the future from the time at which the jobspec is first submitted, and then reduce the at time each time qmanager executes a loop.

PR #1022, opened 2023-04-12
testsuite: make module list awk less fragile

garlick:
Problem: if the number if fields output by 'flux module list' changes, some checks will fail.

Use $NF when grabbing the last field of output with awk instead of hard coding the column number of the last field.

flux-framework/flux-core#5085 proposes to change the output and would break this test.


grondo (approved):
Looks good!

PR #1023, opened 2023-04-12
testsuite: drop meaningless check

garlick:
Problem: t1010-sync-modules.t calls 'flux module info' to "check if the module is still intact", but this command is for static queries of the module DSO so will always succeed.

flux-framework/flux-core#5085 proposes to remove the info subcommand.

This test seems like it wasn't needed anyway, so just drop it.


garlick:
Apologies, this should have gone with the pr that was just merged.

PR #1027, opened 2023-04-17
Fix another typo

chu11:
Problem: Some typos were missed in a prior fix.

Fix more typos!

PR #1021, opened 2023-04-06
add spell check for news and readme

vsoch:
Problem: the NEWS.md is not spell checked, and goes out to releases.
Solution: add a strong spell checker to be run on CI so we catch issues.

Issue #1026, opened 2023-04-17
ci: add false positive list for typos


chu11:
just like https://github.com/flux-framework/flux-core/issues/5075

note that vendored typos in
- sharness
- json.hpp

fixed upstream and could be pulled down at a later time, yggdrasil typo fixes to be sent

edit: https://github.com/tinloaf/ygg/pull/10

edit: some typos in gnu autoconf stuffs: https://github.com/autoconf-archive/autoconf-archive/pull/277

PR #1028, opened 2023-04-18
Corrects some more typos.

dupgit:
Some more commits to correct typos. `supposedly` is corrected also here (#1027) and typos corrected by (#1021) were removed.

### DIFF
--- a/config/ax_boost_base.m4
+++ b/config/ax_boost_base.m4
@@ -10,7 +10,7 @@
 #
 #   Test for the Boost C++ libraries of a particular version (or newer)
 #
-#   If no path to the installed boost library is given the macro searchs
+#   If no path to the installed boost library is given the macro searches
 #   under /usr, /usr/local, /opt and /opt/local and evaluates the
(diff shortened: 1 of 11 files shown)


chu11:
hi @dupgit thanks for the fixes.  However, I think most of these were intentionally skipped in prior PRs.  The typos come from code we copied in from other projects, so we are working to get the typos fixed in those projects, and then we can bring them in via an update to that code.  See #1026.

I think the remaining one you found is handled in #1027 and the `strstr` typo is a false positive.

dupgit:
Hi @chu11,

> hi @dupgit thanks for the fixes. However, I think most of these were intentionally skipped in prior PRs. The typos come from code we copied in from other projects, so we are working to get the typos fixed in those projects, and then we can bring them in via an update to that code. See #1026.

Thanks for your answer and having taken the time to look at this PR.

> I think the remaining one you found is handled in #1027 and the `strstr` typo is a false positive.

Ok. Sorry for that I didn't realized that there are some vendor products in the project and strstr escaped my screening. I close this PR.

Regards.

PR #1030, opened 2023-04-21
 Add false positives typos config

chu11:
Problem: The current typos workflow check only checks a select number of files because there are so many false positives in flux-sched.

Solution: Add a .typos.toml configuration file to eliminate false positives so we can run the typo check on all of flux-sched.  The config file either skips files with lots of false positives or lists common false positives we wish to ignore.


chu11:
re-pushed, adding spellcheck to the mergify requirements too

PR #1031, opened 2023-04-24
ensure licenses appear in distribution tarball

grondo:
@jan-janssen noticed that the Fluxion `LICENSE` file is currently not included in the distribution tarball created by `make dist`.
Add this and other licenses to `EXTRA_DIST` so that they are included in the distribution.


garlick (approved):
LGTM!

PR #1008, opened 2023-02-22
Add Golang bindings

cmisale:
With this PR we want to integrate the Golang bindings we implemented to incorporate Flux APIs into Fluence, the plugin scheduler for Kubernetes.
We squashed all commits over the years into the ones in this PR to show only a single coherent change.

Commits breakdown
1. reapi: add Golang CLI bindings
    - resource/reapi/bindings/go/src/fluxcli/go.mod
    - resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
    - resource/reapi/bindings/go/src/fluxcli/reapi_cli_test.go
2. reapi: add Golang module bindings
    - resource/reapi/bindings/go/src/fluxmodule/reapi_module_test.go
    - resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
3. reapi: open Flux handle to complete reapi_module_ctx_t
    - resource/reapi/bindings/c/reapi_module.cpp
4. writers: add Kubernetes match writer
    - resource/writers/match_writers.cpp
    - resource/writers/match_writers.hpp
5. testsuite: add Golang client test outputs
    - resource/reapi/bindings/go/main/main.go
6. testsuite: add Go paths to Dockerfile
    - src/test/docker/bionic/Dockerfile

PR #1032, opened 2023-05-11
Update sphinx version

milroy:
Readthedocs CI tests are fail due to an old version of Sphinx. This quick PR Updates the version in `requirements.txt`

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==3.4.3
+sphinx<6.0.0
 sphinx-rtd-theme>=0.5.2
 docutils>=0.14,<0.18
 jinja2<3.1.0


grondo (approved):
Ah, in the other flux-framework projects we've instead been pinning `urllib3<2`, but if the build works it is probably better to update Sphinx as you've done here. Thanks!

jameshcorbett (approved):
LGTM! You verified that it wasn't just #989 right?

milroy:
> You verified that it wasn't just https://github.com/flux-framework/flux-sched/pull/989 right?

Yes, it affects PR #1008 as well.

milroy:
Thanks for the quick reviews @grondo and @jameshcorbett ! I'm setting MWP.

Issue #1035, opened 2023-06-08
fluxion can't restart with queues enabled


grondo:
Problem: If the Fluxion modules are reloaded, and queues are enabled, running jobs will be terminated with the following error (from #991):
```
[  +9.794686] job-manager[0]: scheduler: hello
[  +9.797168] sched-fluxion-qmanager[0]: jobmanager_hello_cb: ENOENT: map::at: No such file or directory
[  +9.797231] sched-fluxion-qmanager[0]: raising fatal exception on running job id=63504667937603584
```

This is using `match-format = "rv1"`

grondo:
I'll see if I can make a standalone reproducer a bit later.

trws:
@grondo, do you happen to have a testcase or reproducer that will produce this, or the original, issue? I'm looking over the code, and I think we must be having an issue with order of init or config or something, because the queue map is definitely populated in the normal path and load order, so something must change when we're in restart that sched doesn't expect.

grondo:
Yes, let me see if I can reproduce this in a test instance.

PR #1038, opened 2023-06-20
planner: ensure result in planner_avail_resources_at

trws:
During restart, we can end up in a situation where there is no prior schedule state, so there's no earlier time to return from get_state. It feels like get_state should always return *some kind of state* but everywhere else it's used implicitly checks for null, so fixing this one addresses the issue and keeps all planner tests passing.  Making the bigger change requires ensuring that the state that gets returned is not newer than the `at` passed in, which may require some new logic in initialization.

Fixes https://github.com/flux-framework/flux-sched/issues/1035#issuecomment-1590052478


trws:
Hold off on this for a minute.  It works, but I think it works for the wrong reason.  I'm pretty sure the planner is being asked for available resources at a negative time point.

trws:
Ok, this version is better.  I'm pretty sure there's still more to do to ensure planner always handles this right, but at least this case is addressed for now.

grondo (approved):
Testing shows this works _and_ the reproducer that showed Fluxion couldn't restart with running jobs + queues now passes.

I went ahead added the script as an issues reproducer along with a new test `t5100-issues-test-driver.t` (a copy of `t4000-issues-test-driver.t` from flux-core).

Mind if we cherry pick that commit and add it to this PR?
https://github.com/flux-framework/flux-sched/commit/f6d0367d06041b3b27f957004bdf1c34fa20bdfc

grondo:
I went ahead and pushed the testsuite commit on top here @trws. Thanks!

Issue #782, opened 2020-12-03
resource: traverser->run () segfaults with bidirectional edges


milroy:
As I mentioned in PR #775, passing a JGF with bidirectional edges (e.g. `contains` and `in` for the containment subsystem) to `update allocate` in `resource-query` results in a segfault.  After some debugging, I traced the problem to [`dfu_impl_t::upd_dfv`](https://github.com/flux-framework/flux-sched/blob/3d568512c838e7653185a897eb910e6dc192ac31/resource/traversers/dfu_impl_update.cpp#L245).  When the DFV reaches a leaf, it traverses back to its ancestor, then back to the leaf.  This cycle repeats until producing a segfault.

However, the fix appears simple: adding a gray coloration before the loop, i.e.:
```C++
(*m_graph)[u].idata.colors[dom] = m_color.gray ();
for (auto &subsystem : m_match->subsystems ()) {
```

breaks the cycle and eliminates the segfault.  All CI tests pass with this addition.

PR #1041, opened 2023-06-26
fix subsystems

trws:
This is a first cut at fixing multi-subsystem traversals.  When I tried one earlier today, resource-query segfaulted trying to access an uninitialized `shared_ptr`.  I'm not sure how the aux traversals got to this point, but it looks like they're pretty thoroughly broken right now.  Fixing just the segfault puts it into infinite recursion and hits a stack overflow, so added vertex coloring also.

Any feedback on this, or thoughts on what we can do here would be appreciated.  Especially the coloring I'm not sure is actually appropriate since a node colored exploring from one point might need to be explored from another point in the containment tree to find a proper match, but without it the rest of the design fails pretty spectacularly.

Has anyone done recent work with non-containment subsystems?

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -530,8 +530,9 @@ int dfu_impl_t::aux_upv (const jobmeta_t &meta, vtx_t u, const subsystem_t &aux,
     if ((prune (meta, x_in, aux, u, resources) == -1)
         || (m_match->aux_discover_vtx (u, aux, resources, *m_graph)) != 0)
         goto done;
+    (*m_graph)[u].idata.colors[aux] = m_color.gray ();
 
-    if (u != (*m_roots)[aux])
+    if (u != m_graph_db->metadata.roots[aux])
         explore (meta, u, aux, resources, pristine, excl, visit_t::UPV, upv);
 
     p = (*m_graph)[u].schedule.plans;
@@ -543,6 +544,7 @@ int dfu_impl_t::aux_upv (const jobmeta_t &meta, vtx_t u, const subsystem_t &aux,
         m_err_msg += ".\n";
         goto done;
     }
+    (*m_graph)[u].idata.colors[aux] = m_color.black ();
     if (m_match->aux_finish_vtx (u, aux, resources, *m_graph, upv) != 0)
         goto done;
     if ((rc = resolve (upv, to_parent)) != 0)


milroy:
> Has anyone done recent work with non-containment subsystems?

No recent work, and I'm not sure when/whether the non-containment subsystems worked.

Does the commit fix the observed behavior and you're mainly concerned about side-effects of coloring non-containment vertices? I'll dig into the graph coloring question later this week.

trws:
It makes it so a non-containment completes, though the post order of the aux subsystems pops some errors, they aren't fatal. It's mainly that I'm not sure the coloring gives us the behavior we want for non-containment subsystems.

tpatki:
Need to think a bit more about this from the perspective of coloring or tracking for multi-subsystems. Will take a look in the coming week. 

The non-containment subsystems of power and IO work correctly when tested independently ([power test case](https://github.com/flux-framework/flux-sched/blob/master/t/t3010-resource-power.t). I am fairly sure the multi-subsystem case didn't work and wasn't tested thoroughly. I recall testing with `-S CA+P` at some point during the var-aware development work and it failing, and us discussing that we need to add more support for it (I don't think we ever added that). 
 
 How are you testing the multi-subsystem case @trws? I don't see a test-case for it for (unless I missed it), I can help debug and test this too.

trws:
I was using the `mini-5subsystems-fine.graphml` file under the utilities as an input, and ensuring that query could at least run to completion against it with multiple subsystems active.  It's still not right, the coloring on auxiliary subsystems means the same vertex can't be matched from multiple sources, which is clearly wrong, but without something to serve as a termination condition the current version of the traversal hangs.

My current thought is that we should try to nail down what the actual traversal behavior is for the "upwalk" since I can't seem to find anything that documents how that's supposed to work.  As it was implemented (pre-coloring) the traverser was depth-first, walking all containment edges first, and walking all selected subsystems in subsystem order thereafter.  Using coloring for the depth-first portion is correct, though not actually necessary, but what should the behavior be for the other subsystems? I think we need an actual policy, and I'm not sure what that should be.  It could be limiting the number of steps away from the source vertex where the subsystem was entered, it could be based on what's in the request, honestly I'm not sure yet.

The main thing I really want to do to help reason about some of this, and make things more efficient, is actually break up the edges in the multigraph by subsystem.  We can either do this with the boost::graph "subgraph" feature, so we can treat each subsystem as an actual graph object, or I can throw together a container we can use as the storage for the graph that lets us efficiently access only the edges for one subsystem (probably just a vector with a side-table for offsets based on the subsystem ID).  If we have that division, then the DFS can operate directly on the containment tree, and wont need coloring, filtering, or dynamic checks because it's guaranteed to be a tree.  Then we can use custom logic only for the others where we need it.  What do you think?

milroy:
> I recall testing with -S CA+P at some point during the var-aware development work and it failing, and us discussing that we need to add more support for it (I don't think we ever added that).

I just tested `-S CA+P` as part of the investigation of PR #1046 and can confirm it does not work. I did it by testing one of the power-aware jobspecs against the power JGF with `-S CA+P` and `-S PA`:

```bash
$ ./resource/utilities/resource-query -L t/data/resource/jgfs/power.json -f jgf -S PA -P high
INFO: Loading a matcher: PA
resource-query> m allocate t/data/resource/jobspecs/power/test002.yaml
      ------------power0[10:x]
      ------------power1[10:x]
      ---------node53[1:x]
[...]
$ ./resource/utilities/resource-query -L t/data/resource/jgfs/power.json -f jgf -S C+PA -P high
INFO: Loading a matcher: C+PA
resource-query> m allocate t/data/resource/jobspecs/power/test002.yaml
INFO: No matching resources found
INFO: JOBID=1
```
I think this is a byproduct of the graph coloring.

>  Using coloring for the depth-first portion is correct, though not actually necessary

I think it's necessary for the Fluxion graph, since neighbor vertices form a cycle: https://github.com/flux-framework/flux-sched/issues/782. The edge multiplicity varies and is used for matching: https://github.com/flux-framework/flux-sched/blob/e0b499936b19fe7a8386d0c5e99d72ad8aae5594/resource/traversers/dfu_impl.cpp#L592

> but what should the behavior be for the other subsystems? I think we need an actual policy, and I'm not sure what that should be. It could be limiting the number of steps away from the source vertex where the subsystem was entered, it could be based on what's in the request, honestly I'm not sure yet.

I agree but don't immediately know how to approach this, either. I'll give it some thought.

trws:
Right, the other thing I didn’t mention was I think either way we should be using a bidirectional graph so the in edge is an in edge rather than an out edge. I  hadn’t seen the multi-edge code for pool resources though, that will need some thought for sure. The simplest option would be splitting the vertices, but that sounds awful. We do need a way to handle oversubscription too though, and that’s basically the same problem, so if we come up with a way to allocate part of a vertex that should cover pool resources as well.


---

On July 12, 2023 at 6:19:17 PM PDT, Daniel Milroy ***@***.***> wrote:

I recall testing with -S CA+P at some point during the var-aware development work and it failing, and us discussing that we need to add more support for it (I don't think we ever added that).

I just tested -S CA+P as part of the investigation of PR #1046<https://urldefense.us/v3/__https://github.com/flux-framework/flux-sched/pull/1046__;!!G2kpM7uM-TzIFchu!0lCBMBTXhYk87RXnzBlIY5tXDBlHybEoKXrzp0hji_K2ftN0XdmOjjnWby70IFgh98OV-x7I7N1XM60ITHtT2ubTM6g$> and can confirm it does not work. I did it by testing one of the power-aware jobspecs against the power JGF with -S CA+P and -S PA:

$ ./resource/utilities/resource-query -L t/data/resource/jgfs/power.json -f jgf -S PA -P high
INFO: Loading a matcher: PA
resource-query> m allocate t/data/resource/jobspecs/power/test002.yaml
      ------------power0[10:x]
      ------------power1[10:x]
      ---------node53[1:x]
[...]
$ ./resource/utilities/resource-query -L t/data/resource/jgfs/power.json -f jgf -S C+PA -P high
INFO: Loading a matcher: C+PA
resource-query> m allocate t/data/resource/jobspecs/power/test002.yaml
INFO: No matching resources found
INFO: JOBID=1

I think this is a byproduct of the graph coloring.

Using coloring for the depth-first portion is correct, though not actually necessary

I think it's necessary for the Fluxion graph, since neighbor vertices form a cycle: #782<https://urldefense.us/v3/__https://github.com/flux-framework/flux-sched/issues/782__;!!G2kpM7uM-TzIFchu!0lCBMBTXhYk87RXnzBlIY5tXDBlHybEoKXrzp0hji_K2ftN0XdmOjjnWby70IFgh98OV-x7I7N1XM60ITHtTPdEuSUg$>. The edge multiplicity varies and is used for matching: https://github.com/flux-framework/flux-sched/blob/e0b499936b19fe7a8386d0c5e99d72ad8aae5594/resource/traversers/dfu_impl.cpp#L592<https://urldefense.us/v3/__https://github.com/flux-framework/flux-sched/blob/e0b499936b19fe7a8386d0c5e99d72ad8aae5594/resource/traversers/dfu_impl.cpp*L592__;Iw!!G2kpM7uM-TzIFchu!0lCBMBTXhYk87RXnzBlIY5tXDBlHybEoKXrzp0hji_K2ftN0XdmOjjnWby70IFgh98OV-x7I7N1XM60ITHtTDtYYFb4$>


but what should the behavior be for the other subsystems? I think we need an actual policy, and I'm not sure what that should be. It could be limiting the number of steps away from the source vertex where the subsystem was entered, it could be based on what's in the request, honestly I'm not sure yet.

I agree but don't immediately know how to approach this, either. I'll give it some thought.

milroy:
Updating my comment above with a clearer explanation of one item.

>  Using coloring for the depth-first portion is correct, though not actually necessary

While coloring itself isn't necessary, there needs to be some capability for breaking cycles in the multigraph given the edge multiplicity between neighbors varies and is used for matching: https://github.com/flux-framework/flux-sched/blob/e0b499936b19fe7a8386d0c5e99d72ad8aae5594/resource/traversers/dfu_impl.cpp#L592.

PR #1042, opened 2023-06-27
mark all ranks known to graph when "all" is specified

### LINKED ISSUE #1040: fluxion should not use instance size to determine all ranks in resource set
As discussed in #1009, Fluxion currently uses the instance size attribute when marking "all" resources up or down.

This is fragile, since the instance size may not match the number of execution targets in the resource set acquired from the resource module, and also in the future the instance size may not be static.

Instead, the Fluxion resource module should somehow iterate over all resources when marking them UP or DOWN.

trws:
This changes the behavior a bit, in that it marks *all* ranks known to the database.  The original tried to mark all ranks active in flux, but marked the first N.  The new behavior seems more intuitive, but we may want to change it if we grow or shrink in the future.

fixes #1040

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1114,14 +1114,10 @@ done:
 static int decode_all (std::shared_ptr<resource_ctx_t> &ctx,
                        std::set<int64_t> &ranks)
 {
-    unsigned int size = 0;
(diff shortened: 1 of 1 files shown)


grondo (approved):
Tested this and it is working as expected, thanks!

Issue #1043, opened 2023-06-30
when a job exceeds its time limit, fluxion may reallocate resources before they have been freed


garlick:
Problem: when scheduling back to back jobs using Fluxion in `lonodex` mode with prolog and epilog scripts each containing a 10s sleep, I observed prolog and epilog scripts executing simultaneously, which seems like it might cause problems?

Here's a snapshot of `flux jobs` showing the effect:
```
 garlick@picl0:~$ flux jobs
       JOBID QUEUE    USER     NAME       ST NTASKS NNODES     TIME INFO
 ƒTF5edtS4D5 debug    garlick  stress      R      1      1   2.167s picl1
 ƒTF54wFgwgT debug    garlick  stress      C      1      1   1.177m picl1
```

This is on my `sdexec` working branch so it's possible (though I'm not seeing how) that I introduced something bad:
```
commands:    		0.51.0-160-g34e483a82
libflux-core:		0.51.0-160-g34e483a82
libflux-security:	0.9.0
build-options:		+systemd+hwloc==2.4.0+zmq==4.3.4

sched-fluxion-qmanager.info[0]: version 0.25.0-12-gecf20b37
```

grondo:
tagging @trws, @milroy or maybe @jameshcorbett for an assist on this one. This could be considered a critical issue if it is occurring in production (likely, since all jobs have time limits by default)

garlick:
A guess is that `fluxion-resource` is making resources available again once their allocation's time limit has been reached, rather than waiting for them to be explicitly freed.  Since the epilog in this case runs after the job has exceeded its time limit but before resources are freed, we have sad trombones. :sob: :trumpet:  :bone:

garlick:
For a reliable reproducer, see flux-framework/flux-core#5304.

milroy:
I can produce the reported behavior via the reproducer in flux-core #5304.

I can also reproduce the behavior with various `qmanager` options, e.g., running with `queue-policy=fcfs queue-params=queue-depth=1`. If I increase the job time limit to 3-4s the behavior stops, suggesting that this is a race condition.

[...]

PR #1044, opened 2023-07-05
testsuite: add test for double-booking

garlick:
Problem: there is no test script for ensuring fluxion never allocates the same resources to multiple jobs.

Add a sharness script that utilizes the alloc-check plugin to account for allocated resources and catch errors.  At this point, it just includes an "expected failure" test for #1043.

Note that this PR requires a version of flux-core with flux-framework/flux-core#5304 (just merged). I wanted to get this posted as I saw that @milroy was looking into #1043.

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -45,6 +45,7 @@ TESTS = \
     t1021-qmanager-nodex.t \
     t1022-property-constraints.t \
     t1023-multiqueue-constraints.t \
+    t1024-alloc-check.t \
     t2000-tree-basic.t \
     t2001-tree-real.t \
     t3000-jobspec.t \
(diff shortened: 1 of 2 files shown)


garlick:
The fedora34 builder apparently did not reproduce the issue - I'm guessing maybe the timing is a little tight:
>   XPASS: t1024-alloc-check.t 7 - no jobs received alloc-check exception # TODO known breakage vanished

garlick:
Well the 2s epilog I thought I had included was actually not enabled - fixed that.
Also fixed some issues with fluxion module loading/unloading and tidied a few things up.

garlick:
Test commits squashed, and a "skip" added for older flux-core as discused in :coffee: time today.

grondo (approved):
Tested this and LGTM. The `exit: Invalid number: ` error seems to be a sharness bug (`exit` is being called with an empty arg, i.e. `exit ""`, though why it only manifests here is a mystery to me. Doesn't seem too important for now.

garlick:
Thanks!  I'll set MWP.

PR #1046, opened 2023-07-08
[WIP] traverser: prevent allocation of currently allocated resources

milroy:
Problem: issue https://github.com/flux-framework/flux-sched/issues/1043 identified a scenario where Fluxion will grant a new allocation to a job while the resources are still occupied by the previous allocation. The double booking occurs due to the assumption Fluxion makes that a job will not run beyond its walltime. However, as the issue describes, an epilog script may cause a job to run beyond its walltime. Since Fluxion doesn't receive a `free` message until the epilog completes, the allocation remains in the resource graph but the scheduled point at allocation completion time is exceeded, allowing the resources to be allocated to another job. There are other common scenarios that can lead to multiple concurrent allocations, such as a job getting stuck in `CLEANUP`.

This PR adds a check for an existing allocation on each exclusive resource vertex during traversal pruning. The check only occurs for `AT_ALLOC` traversals to allow reservations and satisfiability traversals to succeed. The check for existing allocations prevents another job from receiving the resources and allows reservations and satisfiability checks to complete. It also ensures that the check will block `AT_ALLOC` traversals from matching the resources until the jobid is removed from the per-vertex `allocations` map via a `free` message from `job-manager`.

The check has the (I think) unavoidable side effect that reservations on the allocated subgraph will be allowed immediately after the `at` time of the `AT_ALLOC` traversal because the [`planner_multi_avail_time_first`](https://github.com/flux-framework/flux-sched/blob/65da8f26cd53737a1a29948ffbf7618fead6a13d/resource/traversers/dfu.cpp#L109) call will regard extra resources as available via incorrect aggregate counts. The side effect is minor, though, as the reservation will be blocked from transitioning to allocated until a `free` message cancels the existing allocation. During the subsequent scheduling loops, if the `AT_ALLOC` traversal fails the reservation will be pushed back during the `AT_ALLOC_ORELSE_RESERVE` traversal. That will continue until matching resources become available.

The PR fixes the reproducer from https://github.com/flux-framework/flux-core/pull/5304 and passes the test in PR #1044, but it's WIP because the behavior for non-exclusive resource requests is still suspect.

I chose to place the check in the `by_excl` part of `prune` because there can only be a single allocation in `(*m_graph)[u].schedule.allocations` for an exclusive vertex. While checking at non-exclusively allocated vertices is possible, the logic for determining which jobid is delayed via comparing availability counts returned by `planner` to the jobs in the `(*m_graph)[u].schedule.allocations` map and then determining the true number of available resources is much more complex.

Only checking at exclusive vertices can involve a deeper traversal than necessary (i.e. by checking at higher, non-exclusive vertices), but I think it will correctly block allocations provided that leaf-level vertices must be allocated exclusively. That appears to be ensured here: https://github.com/flux-framework/flux-sched/blob/65da8f26cd53737a1a29948ffbf7618fead6a13d/resource/traversers/dfu_impl.cpp#L116.

There are two items to verify before the WIP can be dropped:

- [ ] Make sure leaf-level vertices must be allocated exclusively
- [ ] Check the behavior for upwalks into an auxiliary subsystem

This PR should only be merged after PR #1044.


garlick:
Thanks @milroy for taking the time to write up the explanation.

#1044 has been merged so at some point, this PR can be rebased on master, and a commit can be added that changes the two `test_expect_failure` tests in `t1024-alloc-check.t` to `test_expect_success`.

grondo:
Just checking in here since deadline for new packages for the next TOSS release is tomorrow. I think the CI failures are just do to the fact that the expect failure alloc-check tests need to be updated to expect success:

> ```# 2 known breakage(s) vanished; please update test(s)```

Happy to push an update here if that would help @milroy.

milroy:
@grondo: just got out of meetings and rebased and pushed the update to the testsuite. I'm going to check the two remaining items in the description and drop the WIP tag if possible.

milroy:
I've confirmed that the exclusivity check here: https://github.com/flux-framework/flux-sched/blob/65da8f26cd53737a1a29948ffbf7618fead6a13d/resource/traversers/dfu_impl.cpp#L116
ensures that everything under `slot` must be allocated exclusively. For example, the following simple jobspec:
```YAML
version: 9999
resources:
  - type: cluster
    count: 1
    with:
      - type: rack
        count: 1
        with:
          - type: node
            count: 1
            with:
              - type: socket
                count: 1
                with:
                  - type: slot
                    label: corelevel
                    count: 1
                    with:
                     - type: core
                       exclusive: false
                       count: 1

# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: [ "app" ]
    slot: corelevel
    count:
      per_slot: 1
````
Produces the follow (overly verbose- we'll need to fix that) match error: 
```bash
resource-query> m allocate t/data/resource/jobspecs/exclusive/test001.yaml
ERROR: by_excl: exclusivity conflicts at jobspec= : vertex=core0by_excl: exclusivity conflicts at jobspec= : vertex=core1
[...]
```

From that standpoint this PR should work as intended. However, is Fluxion conforming to RFC 14 with the jobspec above? In particular, here: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html#reserved-resource-types, which states:

> A task slot SHALL have at least one edge specified using with:, and the resources associated with a slot SHALL be exclusively allocated to the program described in the jobspec, unless otherwise specified in the exclusive field of the associated resource.

The last sentence implies `exclusive: false` at a leaf vertex should be supported by Fluxion. @grondo and @garlick is my assessment correct here?

garlick:
Honestly I do not know.  We don't have command line tooling that can express that AFAIK.  Maybe we could open an issue on that one for later?

milroy:
> Maybe we could open an issue on that one for later?

My concern is that if it's decided that Fluxion needs to support requests like the one above then the fix in this PR won't work.

garlick:
I could be mistaken but I _think_ jobspec v1 would forbid that type of slot definition.  See https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html#v1-specific-resource-graph-restrictions.  flux-core ensures that only v1 compliant jobspec can be submitted to the scheduler, so...is that sufficient justification for deferring this concern?

milroy:
It looks like the intention was to make an `exclusive: false` request under `slot` invalid in RFC 25: https://github.com/flux-framework/rfc/pull/189 and https://github.com/flux-framework/flux-sched/pull/502/commits/3552b2cb107277456435b2dd696dfd398094e186#r309027939 but I don't see that it ever happened.

milroy:
> so...is that sufficient justification for deferring this concern?

Yes, it is. I was a bit confused by the Resource Graph Restrictions section, but I think that together with:

> a node type resource vertex MAY also contain the following optional keys:
exclusive
 (implying no other vertices can contain that key) is a strong enough indication.

garlick (approved):
LGTM!

Just gave this a quick sanity check on my test cluster with the artificially extended prolog/epilog and it seems to work well.   (Not too different from the sharness test so maybe that doesn't add much).

grondo (approved):
LGTM too! Perhaps @jameshcorbett should add his review in case there is something here that might affect rabbit allocations.

trws:
Yeah, this looks good.  We may end up deciding to tweak it as part of the traversal refactor I want to do, but it should be easier to express after that too.  Really good idea @milroy!

jameshcorbett (approved):
Looks reasonable to me!

milroy:
Thanks all for the discussion and reviews! Setting MWP.

PR #1047, opened 2023-07-09
testsuite: t1024-alloc-check.t: do not fail if rank 0 not drained

grondo:
Problem: Once flux-framework/flux-core#5319 is fixed, broker ranks will not be drained if the epilog is running when a job is canceled. However, this causes t1024-alloc-check.t to fail since `flux resource undrain 0` will fail.

Don't fail if the undrain of rank 0 fails, so that this test will pass even after fixing the flux-core bug.


garlick (approved):
Thanks!

grondo:
Thanks, setting MWP.

PR #1049, opened 2023-07-12
Add cmake build system

trws:
This is an experiment to see if we can use a fake configure script to get the benefits of both cmake, and all the CI tooling flux projects rely on.


trws:
Hey @grondo, I'm playing with this a bit as a compromise alternative.  There are still some rough edges on it, but what would you think of an approach like this that can still be used as though it had an autoconf-generated configure script? Non-trivial upside is it's already almost fully compatible with the CI scripts without modifying them.

grondo:
note that if we go this route another TODO will be to update the RPM spec file in the TOSS build farm.

Also, not sure we can drop the bionic CI builds since we haven't dropped support for that distro yet. Was there something that doesn't work there?

trws:
That error is because your cmake is too old, I bumped it down to 3.10 in hopes it would work with the default cmake on focal and bionic but they are just too old, it needs 3.12. The env var actually worked, this just means we can’t name the target afterwards.


---

On July 13, 2023 at 8:05:40 AM PDT, Mark Grondona ***@***.***> wrote:

note that if we go this route another TODO will be to update the RPM spec file in the TOSS build farm.

Also, not sure we can drop the bionic CI builds since we haven't dropped support for that distro yet. Was there something that doesn't work there?

trws:
> I was just going to try this out and I can't figure out how to get it to configure against a side installed flux-core. The autotools build configures flux-sched using the flux-core corresponding to the first `flux` found in `PATH`. Therefore you can do
> 
> ```
> $ /path/to/flux-core/bin/flux ./configure ...
> ```

Is this supposed to work? I just tried it, and sched's autoconf doesn't configure under those conditions, it seems to look for flux-core only in PKG_CONFIG_PATH and the selected PREFIX.  I just pushed an update that matches the ax_flux_core macro in the cmake search, it also now checks for the correct cmake version.  Given anything 3.12 or higher basic cases like that or with compiler or prefix flags should work.  Coverage, sanitizers, and "--with" options would still be TODO.

grondo:
Yes, it works for me (I tend to use it most of the time to configure against a side-installed flux-core). E.g. for flux-core installed with `prefix=/tmp/grondo`:

```
$ /tmp/grondo/bin/flux ./configure
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
[snip]
checking for FLUX_CORE... yes
checking for flux... /tmp/grondo/bin/flux
checking for FLUX_IDSET... yes
checking for FLUX_SCHEDUTIL... yes
checking for FLUX_OPTPARSE... yes
checking for FLUX_HOSTLIST... yes
[snip]
  flux-sched version 0.28.0
  Prefix...........: /tmp/grondo
  Debug Build......: 
  C Compiler.......: gcc
  C++ Compiler.....: g++
  CFLAGS...........: -g -O2
  CPPFLAGS.......... 
  CXXFLAGS.......... -g -O2
  FLUX.............: /tmp/grondo/bin/flux
  FLUX_VERSION.....: 0.52.0-1-ga35e212a0
  FLUX_CORE_CFLAGS.: -I/tmp/grondo/include
  FLUX_CORE_LIBS...: -L/tmp/grondo/lib -lflux-core
  LIBFLUX_VERSION..: 0.52.0-1-ga35e212a0
  FLUX_PREFIX......: /tmp/grondo
  LDFLAGS..........: 
  LIBS.............: 
  Linker...........: /usr/bin/ld -m elf_x86_64

```

grondo:
On TOSS4 systems, `PYTHON=/usr/bin/python3 ./configure` is required to successfully configure flux-core and flux-sched (the TCE python doesn't have the correct python deps). How do we do this with cmake? It seems to ignore `PYTHON` when run as `cmake .` or `./configure`.

trws:
I'll add support for the env var, the cmake way is to do one of these:

```
-D Python_EXECUTABLE=<path to exe>
-D Python_ROOT_DIR=<path to install prefix>
```

grondo:
Thanks, that worked!

trws:
Clearly working, but when I try exactly the same it doesn't work... 
somehow I'm missing something.  Looking through ax_flux_core I only see 
it using pkg-config to look it up, for the "checking for FLUX_CORE" then 
using the result to look for the rest, but that can't be all if that 
works for you.  Do you happen to remember where the autoconf code for 
that is? It would help me debug my environment.

grondo:
Ok, I had forgotten how this works. The main support was introduced in 

b4bd5f09a2743600bd96d44f763d01e4d8ced501

which sets the default `--prefix` to the first `flux(1)` found in `PATH` :

```
commit b4bd5f09a2743600bd96d44f763d01e4d8ced501
Author: Mark A. Grondona <mark.grondona@gmail.com>
Date:   Fri Feb 28 10:07:40 2020 -0800

    build: set --prefix to same prefix as flux(1) by default
    
    Use AC_PREFIX_PROGRAM([flux]) to have the default prefix of flux-sched
    set to the same prefix as the flux executable in PATH. For example,
    for a system with flux installed as /usr/bin/flux, prefix will default
    to /usr instead of /usr/local for flux-sched, unless
    
     ./configure --prefix=/usr/local
    
    is used.
    
    This also allows flux-sched to be configured against side-installed
    flux with simply:
    
     /path/to/side/install/bin/flux env ./configure
```

Then the stuff in `ax_flux_core.m4` which prepends to `PKG_CONFIG_PATH` based on `prefix` takes over.

So, if you are using `--prefix=/somewhere/else` then that overrides the default `prefix` and the convenience hack to configure against side-installed flux-core is overridden.

trws:
Thanks @grondo, that helps a lot!  I think I was using a wrapper script that set prefix and expecting it to still work because of the flux in path.  I can replicate that in cmake pretty easily.  Also tempted to add something to the cmake to find flux in path, then add `$(dirname $(dirname $FLUX))/lib/pkgconfig` to the `PKG_CONFIG_PATH`, that way even if they're in a separate prefix we get the "use the flux-core in path" behavior.

grondo:
Sounds great to me!

trws:
Ok, finding flux-core by finding flux and matching prefix if it isn't explicitly set implemented.  This allows flux-core to be found by path *even when a different prefix is selected* if the prefix for flux-core isn't explicitly specified.  The order of priority is: FLUX_CORE_PREFIX > PKG_CONFIG_PATH > flux in path > default locations

trws:
Ok, this is now all there.  I want to add a better C++ test system (Catch2 probably unless there are objections), include-what-you-use for CI, and a few other things, but this is a coherent build system and set of fixes. Given where it is, and the issues with boost I keep running into on some platforms, I would like to remove the autotools setup at some point, but *not now*.  If we have a successful release using the cmake wrappers and everything, that's when I'd like to do it if that works for everyone.

milroy:
I'm in the process of reviewing the PR, but I ran into a small issue in my `bookworm` Docker build:
```bash
milroy1@docker-desktop:/usr/src$ git config --global --add safe.directory /usr/src
milroy1@docker-desktop:/usr/src$ ./autogen.sh 
milroy1@docker-desktop:/usr/src$ ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var
-- VER 37a272b7
CMake Error at CMakeLists.txt:11 (project):
  VERSION "37a272b7" format invalid.
```
It looks like this is caused by the fork not containing the upstream tags. To get the project to build and test successfully, I needed to set the version to something like `0.28.0` in a new file (`flux-sched.ver`).

You might want to allow a user-specified environment variable that overrides the version returned by `git describe --always  ...`.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 cmake_minimum_required(VERSION 3.14)
```
milroy, line 1: Nit: wrap the commit body at 72 or 80 characters (can't remember which is preferred for git).

.configure-custom.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
milroy, line 1: It's probably better for commit intelligibility to turn the refactoring commits into fixups.

CMakeLists.txt
```diff
@@ -68,7 +68,6 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   message(STATUS "flux core found and no specific prefix specified, using ${FLUX_CORE_PREFIX}")
```
milroy, line 63: Nit: wrap the commit body at 72 or 80 characters (can't remember which is preferred for git).

codecov.yml
```diff
@@ -7,7 +7,7 @@ coverage:
  - "**/tests/*.c"
  - "src/common/libtap/.*"
  - "src/common/libutil/json.hpp"
- - "src/common/yggdrasil/.*""
+ - "src/common/yggdrasil/.*"
```
milroy, line 10: Nice catch!

resource/planner/c/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+add_library(planner STATIC
+  ./planner_c_interface.cpp
+  ./planner_multi_c_interface.cpp
+	./planner.h
+	./planner_multi.h
```
milroy, line 5: Nit: spacing.
```suggestion
  ./planner.h
  ./planner_multi.h
``

File: src/test/docker/bookworm/Dockerfile [hunk 6]
```diff
@@ -10,9 +10,10 @@ RUN sudo apt-get -qq install -y --no-install-recommends \
 	libboost-system-dev \
```
milroy, line 10: High-level comment on this commit: it's not clear why the `bionic`,  `el7`, and `fedora33` dockerfiles are deleted in this commit.

milroy (changes requested):
This is a huge improvement to the build system in terms of speed, and I like the maintainability improvements, too. Thanks again for taking the time to walk me through the contributions last week!

I identified what I think is a minor problem with builds on Fluxion forks without tags, as well as a few small issues and nits related to commit messages. In general, the commit messages need to be wrapped, and in the standard Flux Framework Problem: ... format.

From a high-level, it's somewhat difficult to follow the PR because files are updated and lines changes several times along the course of the commits. This PR should be restructured so that the refactor commits are fixups that can be squashed in a rebase. I think that will condense the changes and help make them more understandable.

CMakeLists.txt
```diff
@@ -22,8 +22,15 @@ file(APPEND ${CONFIG_H} "/* Define _GNU_SOURCE so that we get all necessary prot
 file(APPEND ${CONFIG_H} "#define _GNU_SOURCE 1 \n\n ")
```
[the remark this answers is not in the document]
trws: These comments are ending up on lines of code rather than on commits, not sure if there's anything to do about that but it would be good to find a way to make the link to the commit more obvious.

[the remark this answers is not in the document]
grondo: It has always been annoying that there's no way to comment on a commit message itself. However, going through the commits in the "Commits" tab will only show the comments pertaining to that commit (though it still appears attached to a line of code)

grondo:
One random other suggestion. We typically autogenerate release notes from PR titles (since they are used by default as the merge commit message), so before a merge the PR should possibly be titled something other than  "Cmake revert checks"  (which admittedly I'm not sure I understand what that means). Maybe just "switch build to cmake"

trws:
I’ll fix that. The current title is because this started as a fork of the previous cmake PR where I reverted all the changes to the *checks.sh scripts to try the configure script approach.

---
On September 6, 2023 at 2:47:42 PM PDT, Mark Grondona ***@***.***> wrote:

One random other suggestion. We typically autogenerate release notes from PR titles (since they are used by default as the merge commit message), so before a merge the PR should possibly be titled something other than "Cmake revert checks" (which admittedly I'm not sure I understand what that means). Maybe just "switch build to cmake"

src/test/docker/bookworm/Dockerfile
trws: That's an artifact of this not having been rebased on top of the build-system update PR yet.  This should evaporate after the rebase I just did.  Will post after some cleanup.

CMakeLists.txt
trws: Done, this was an artifact of trying out lazygit, which apparently doesn't do this automatically, rather than using my vim/fugitive setup which does.  Lesson learned.

CMakeLists.txt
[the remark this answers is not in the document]
trws: I keep hoping we can do something better for that.  Tools like stacked git, jujutsu, and even facebook's sapling with their experimental example stacked PR UI for github look promising, but nothing first party. 😢 

Short of going as far as reviewing with a third-party UI or switching to gerritt it feels like we're a bit stuck.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
...
-project(flux-sched VERSION 0.40.0 LANGUAGES CXX C)
+set(VER_FILE ${CMAKE_SOURCE_DIR}/flux-sched.ver)
+if(EXISTS ${VER_FILE})
+  file(READ ${VER_FILE} FLUX_SCHED_VER)
+  string(STRIP ${FLUX_SCHED_VER} FLUX_SCHED_VER)
+else()
+  execute_process(COMMAND sh -c [=[git describe --always | awk '/.*/ {sub(/^v/, ""); printf "%s",$1; exit}']=] OUTPUT_VARIABLE FLUX_SCHED_VER)
+endif()
+string(REGEX REPLACE "-.*$" "" FLUX_SCHED_VER_NOGIT ${FLUX_SCHED_VER})
+project(flux-sched VERSION ${FLUX_SCHED_VER_NOGIT} LANGUAGES CXX C)
```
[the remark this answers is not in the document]
trws: That was a mistake on my part, it should support the same version override env var as core does, so I'll add that.  It should still work on forks though as long as one runs `git fetch --tags upstream` or similar, same as core does.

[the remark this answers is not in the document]
trws: I'm also making it possible by specifying a cmake define.

trws:
Ok, that was one heck of a rebase... This may be the first time I've actually saved a backup of the origin branch just in case something got munged up in all the rewriting, but there are a *lot* less commits now and should be less files actually changed in this PR.

CMakeLists.txt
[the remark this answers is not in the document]
trws: Here's the UI I was referring to from facebook: https://reviewstack.dev/ it also works with ghstack, if only

milroy:
I started taking a look at the updates. The monster rebase definitely cleaned up the commits!

I ran into a problem with the update to `CMakeLists.txt`:
```bash
$ ./configure
-- VER dc8f6c88
CMake Error at CMakeLists.txt:14 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.
CMake Warning at CMakeLists.txt:15 (project):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.
-- Building flux-sched version dc8f6c88
-- Boost version: 1.74.0
-- Configuring incomplete, errors occurred!
See also "/usr/src/CMakeFiles/CMakeOutput.log".
```
Which you don't hit unless all three are true: undefined `FLUX_SCHED_VER` environment variable, no `VER_FILE`, and no tags present in the fork. Quoting the final `REGEX` argument fixes the issue:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 07b32932..6c299b01 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ else()
   execute_process(COMMAND sh -c [=[git describe --always | awk '/.*/ {sub(/^v/, ""); printf "%s",$1; exit}']=] OUTPUT_VARIABLE FLUX_SCHED_VER)
 endif()
 message(STATUS "VER ${FLUX_SCHED_VER}")
-string(REGEX REPLACE "-.*$" "" FLUX_SCHED_VER_NOGIT ${FLUX_SCHED_VER})
+string(REGEX REPLACE "-.*$" "" FLUX_SCHED_VER_NOGIT "${FLUX_SCHED_VER}")
 project(flux-sched VERSION ${FLUX_SCHED_VER_NOGIT} LANGUAGES CXX C)
 message(STATUS "Building flux-sched version ${FLUX_SCHED_VER}")
```

trws:
Thanks for that @milroy, patched up and push incoming.  I'll set MWP after that.  If you get through a review and you're happy, approving should send it through.

milroy (approved):
The first commit "ensure superceded PR workflows are cancelled" needs to wrap at 72 characters. Beyond that this PR is great and ready to merge. I dropped the MWP label and am approving this PR. 

@trws, you can re-add MWP when you fix first commit message.

trws:
wrapped and pushed, resetting MWP

PR #1051, opened 2023-07-18
testsuite: drop resource exclusion test

garlick:
Problem: flux-core is deprecating the capability to change resource.exclude on the fly, but t1013-exclusion.t relies on this capability.

Since the scheduler is not informed of excluded resources, this test duplicates flux-core coverage and does not add anything scheduler-specific.
    
Remove test.


garlick:
Another option would be to remove this test.  Since the scheduler is simply not told about excluded resources, the test is just duplicating coverage that is in flux-core.  There's no exclusion specific code to test in the scheduler that I know of.  (There shouldn't be)

grondo:
That sounds fine to me.

garlick:
OK, now this PR removes the test instead of tweaking it.

grondo (approved):
Seems good to me!

PR #1052, opened 2023-07-26
Add iwyu

trws:
Depends on #1049, uses the cmake built-in support for iwyu to add it, added to fedora38-clang because that's the fastest target on a distro that packages it at a reasonable version.  It also shares deps with clang, so that helps.


grondo:
Cool. I tried to set this up with flux-core but way too many false positives 😞

trws:
I think we’ll need a “mapping file” to fix that, hopefully it won’t be too bad to mark flux/core.h as the main header for flux and a few other things like that.
---
On July 25, 2023 at 5:28:10 PM PDT, Mark Grondona ***@***.***> wrote:

Cool. I tried to set this up with flux-core but way too many false positives 😞

PR #1053, opened 2023-07-27
add delete subgraph functionality

zekemorton:
This PR adds delete subgraph functionality to the jgf reader and resource query. This is a part of the elasticity effort and allows removal of a piece of the resource graph. This function works by passing in a path of a node in the graph (ex /medium0/rack3) and will remove the tree with that node as root (ex /medium0/rack3/*). 

This implementation requires two main steps. The first step is to identify all boost graph nodes in the subgraph. This is done via a DFS on the out edges of each node. Note that one of the out edges of a node is its parent, so in this case a string comparison of the path was used to determine that an out edge is not the parent. The second step is to remove the nodes from the graph. Because we are using listS instead of vecS, actually deleting the nodes would invalidate the vertex descriptors, which would invalidate the resource graph metadata used for vertex descriptor look ups. Instead, this PR clears all edges from "deleted nodes" and removes them from the resource graph metadata. This makes them unaccessible from traversals from the main graph and unaccessible from resource graph metadata look ups. 

This PR also adds this functionality to resource query and adds new tests for these features. 

WIP to do items:

- [x] Add automated tests for remove subgraph


resource/readers/resource_reader_jgf.cpp
```diff
@@ -1053,6 +1089,34 @@ int resource_reader_jgf_t::update_edges (resource_graph_t &g,
...
+{
+    vtx_t next_node;
+    boost::graph_traits<resource_graph_t>::out_edge_iterator ei, ei_end;
+    boost::tie (ei, ei_end) = boost::out_edges (node, g);
+
+    for (; ei != ei_end; ++ei) {
+        next_node =  boost::target (*ei, g);
+        
+        for (auto const &paths_it : g[next_node].paths) {
+            if (paths_it.second.find(g[node].name) !=std::string::npos &&
```
zekemorton, line 1104: Add a comment to explain this check

resource/readers/resource_reader_jgf.cpp
```diff
@@ -1053,6 +1089,35 @@ int resource_reader_jgf_t::update_edges (resource_graph_t &g,
...
+    for (; ei != ei_end; ++ei) {
+        next_node =  boost::target (*ei, g);
+        
+        for (auto const &paths_it : g[next_node].paths) {
+            // check that we don't recurse on parent edges 
+            if (paths_it.second.find(g[node].name) != std::string::npos &&
+                paths_it.second.find(g[node].name) < paths_it.second.find(g[next_node].name)) {
+                
+                node_list.push_back(next_node);
+                get_subgraph_nodes(g, next_node, node_list);
```
milroy, line 1109:
```suggestion
            if (paths_it.second.find (g[node].name) != std::string::npos &&
                paths_it.second.find (g[node].name) < paths_it.second.find (g[next_node].name)) {

                node_list.push_back (next_node);
                get_subgraph_nodes (g, next_node, node_list);
```

milroy (changes requested):
This is looking very good. After you fix a few nits and typos and address one question this should be good to merge. Thanks a lot!

zekemorton:
After finding a segfault centered around not deleting elements from the by_outedge member of the resource graph metadata, I added in a few new changes to solve this issue. 
 - remove the by_outedge for each vertex that we remove from the resource graph
 - remove any edges from the vertex who is parent to the root of the removed subgraph to the root vertex (note that there can be more than one)
 - add in tests that try to fully allocate the resource graph after removing a resource

jameshcorbett:
> After finding a segfault

Pincer maneuver complete

resource/utilities/command.cpp
```diff
@@ -430,14 +432,40 @@ static int attach (std::shared_ptr<resource_context_t> &ctx,
...
+    }
+
+    if ( (rd->remove_subgraph (ctx->db->resource_graph, ctx->db->metadata,
+                               node_path)) != 0) {
+        std::cerr << "ERROR: can't remove subgraph " << std::endl;
+        std::cerr << "ERROR: " << rd->err_message ();
+        return -1;
+    }
+    if (ctx->traverser->initialize (ctx->fgraph, ctx->db, ctx->matcher) != 0) {
+        std::cerr << "ERROR: can't reinitialize traverser after remove"
```
milroy, line 468: Let's remove this `initialize` call now because it corrupts `idata` planners. See issue #1075 .

zekemorton:
@milroy I add the most recent changes requested including the two TODO messages and spacing fixes. 
I am not sure what the issue is with the python version tests though. Any thoughts there?

milroy:
Looks like you figured out the runner issue with PR #1080. I'm approving this PR, but let's wait to merge it until #1080 is merged.

milroy (approved):
Thanks for the changes! LGTM.

milroy:
@zekemorton since PR #1080 is merged you can rebase this PR and set MWP.

zekemorton:
> Approving reviews have been dismissed because this pull request was updated.

@milroy looks like it might need another approval since doing the rebase

milroy (approved):
Approving again in the hopes that this will trigger MWP action.

PR #1055, opened 2023-08-02
Add first nodex policy

trws:
First cut to add a node exclusive first policy.  This seems to work, refactored one of the nodex tests to not be quite so duplicative and test this too.  In cases with no overlapping jobs, firstnodex is the same as hinodex, but it will only look for one match rather than scoring matches.

Need to test the performance, but this should be significantly faster than the other nodex policies.

### DIFF
--- a/resource/policies/dfu_match_policy_factory.cpp
+++ b/resource/policies/dfu_match_policy_factory.cpp
@@ -38,11 +38,13 @@ std::shared_ptr<dfu_match_cb_t> create_match_cb (const std::string &policy)
     std::shared_ptr<dfu_match_cb_t> matcher = nullptr;
 
     try {
-        if (policy == FIRST_MATCH) {
+        if (policy == FIRST_MATCH || policy == FIRST_NODEX_MATCH) {
             std::shared_ptr<high_first_t> ptr
                 = std::make_shared<high_first_t> ();
             ptr->add_score_factor (std::string ("node"), 1, 10000);
             ptr->set_stop_on_k_matches (1);
+            if (policy == FIRST_NODEX_MATCH)
+               ptr->add_exclusive_resource_type ("node");
             matcher = ptr;
         } else if (policy == HIGH_ID_FIRST) {
             matcher = std::make_shared<high_first_t> ();
--- a/resource/utilities/resource-query.cpp
+++ b/resource/utilities/resource-query.cpp
@@ -54,132 +54,131 @@ static const struct option longopts[] = {
 
 static void usage (int code)
 {
-    std::cerr <<
-"usage: resource-query [OPTIONS...]\n"
-"\n"
-"Command-line utility that takes in an HPC resource request written in\n"
-"Flux's Canonical Job Specification (or simply a jobspec) (RFC 14) and\n"
-"selects the best-matching compute and other resources in accordance\n"
-"with a selection policy.\n"
-"\n"
-"Read in a resource-graph generation recipe written in the GRUG format\n"
-"and populate the resource-graph data store representing the compute and\n"
-"other HPC resources and their relationships (RFC 4).\n"
-"\n"
-"Provide a simple command-line interface (cli) to allow users to allocate\n"
-"or reserve the resource set in this resource-graph data store \n"
-"using a jobspec as an input.\n"
-"Traverse the resource graph in a predefined order for resource selection.\n"
-"Currently only support one traversal type: depth-first traversal on the\n"
-"dominant subsystem and up-walk traversal on one or more auxiliary \n"
-"subsystems.\n"
-"\n"
-"OPTIONS allow for using a predefined matcher that is configured\n"
-"to use a different set of subsystems as its dominant and/or auxiliary\n"
-"ones to perform the matches on.\n"
-"\n"
-"OPTIONS also allow for instantiating a different resource-matching\n"
-"selection policy--e.g., select resources with high or low IDs first.\n"
-"\n"
-"OPTIONS allow for exporting the filtered graph of the used matcher\n"
-"in a selected graph format at the end of the cli session.\n"
-"\n"
-"To see cli commands, type in \"help\" in the cli: i.e.,\n"
-"  % resource-query> help\n"
-"\n"
-"\n"
-"\n"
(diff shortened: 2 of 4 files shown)


trws:
Looking at the code for this, I'm thinking it would be really useful to add a new "policy" that lets us build one of these up more dynamically. It would let us factor out duplication here and let us play with the number of matches to look for and things like that.  I kinda want to get this in so we _have_ it, then build that up.

trws:
Rebased.

trws:
Quick resource-bench.sh numbers:

```
firstnodex
Min. Match Time: 1.90735e-06
Max. Match Time: 0.000230789
Avg. Match Time: 0.000187913
first
Min. Match Time: 6.60419e-05
Max. Match Time: 0.000261068
Avg. Match Time: 9.12583e-05
hinodex
Min. Match Time: 1.90735e-06
Max. Match Time: 0.0184488
Avg. Match Time: 0.00863134
```

So, firstnodex is about 450 times faster than hinodex, but still about 10x slower than first, at least for this little microbenchmark.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -38,11 +38,13 @@ std::shared_ptr<dfu_match_cb_t> create_match_cb (const std::string &policy)
     std::shared_ptr<dfu_match_cb_t> matcher = nullptr;
```
milroy, line 38: Please add a description to the body of the commit following the Flux "Problem:, Add:" convention.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -38,11 +38,13 @@ std::shared_ptr<dfu_match_cb_t> create_match_cb (const std::string &policy)
...
 
     try {
-        if (policy == FIRST_MATCH) {
+        if (policy == FIRST_MATCH || policy == FIRST_NODEX_MATCH) {
             std::shared_ptr<high_first_t> ptr
                 = std::make_shared<high_first_t> ();
             ptr->add_score_factor (std::string ("node"), 1, 10000);
             ptr->set_stop_on_k_matches (1);
+            if (policy == FIRST_NODEX_MATCH)
+               ptr->add_exclusive_resource_type ("node");
```
milroy, line 47: I'm really impressed that this is all it takes to generate the correct behavior!

t/t3033-resource-nodex.t
```diff
@@ -9,7 +9,36 @@ exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/nodex"
...
+
+    cmds003="${cmd_dir}/cmds03.in"
+    test003_desc="match allocate 7 jobs -- last fails (pol=$pol)"
+    test_expect_success "${test003_desc}" '
+        sed "s~@TEST_SRCDIR@~${SHARNESS_TEST_SRCDIR}~g" ${cmds003} > cmds003 &&
+        ${query} -L ${grugs} -S CA -P $pol -t ${prefix}3.R.out < cmds003 &&
+        test_cmp ${prefix}3.R.out ${exp_dir}/${prefix}3.R.out
+    '
+}
+
```
milroy, line 41: I like that this function reduces duplicate code.

t/t3033-resource-nodex.t
```diff
@@ -65,28 +72,8 @@ test_expect_success "${test003_desc}" '
...
+run_tests_with_policy lonodex 01
 
-cmds013="${cmd_dir}/cmds03.in"
-test013_desc="match allocate 7 jobs -- last fails (pol=lonodex)"
-test_expect_success "${test013_desc}" '
-    sed "s~@TEST_SRCDIR@~${SHARNESS_TEST_SRCDIR}~g" ${cmds013} > cmds013 &&
-    ${query} -L ${grugs} -S CA -P lonodex -t 013.R.out < cmds013 &&
-    test_cmp 013.R.out ${exp_dir}/013.R.out
-'
+run_tests_with_policy firstnodex 00
```
milroy, line 77: I tested the outputs in the `jammy` container and can reproduce the results comparing the output to `hinodex`. I don't think there's a guarantee `firstnodex` will always select the same resources as `hinodex`.

milroy (changes requested):
This looks good overall and is an elegant implementation. I did find a few things that need to be changed, which I indicated in the line-specific comments. 

Two other higher-level changes:
1. Please add `firstnodex` and description here: https://github.com/flux-framework/flux-sched/blob/76f2e3bf0e5698bd0a9821f6e2392a1d91706e59/resource/utilities/resource-query.cpp#L126
2. Please break up the commit such that there is one for "policies" (changes to `dfu_match_policy_factory.*pp`, "testsuite" (changes to `t3033-resoure-nodex.t`), and "resource-query" (changes to `resource-query.cpp`).

t/t3033-resource-nodex.t
milroy: That said, it's no different than how the current `first` policy check works in `t3001-resource-basic.t`.

resource/policies/dfu_match_policy_factory.cpp
trws: Much as the current traverser is really complex, this is why.  Dong designed it to be amazingly flexible and generic.  Hopefully we can take a bit better advantage of that if we expose it a bit more.

trws:
It ended up being four commits, because I decided to convert the usage message string into a C++11 raw string and made that a separate commit.  This should be good to go unless I botched something in rebasing it.

resource/utilities/resource-query.cpp
```diff
@@ -713,7 +712,7 @@ static std::shared_ptr<resource_context_t> init_resource_query (int c,
 
     try {
         ctx = std::make_shared<resource_context_t> ();
-	ctx->db = std::make_shared<resource_graph_db_t> ();
+    ctx->db = std::make_shared<resource_graph_db_t> ();
```
milroy, line 717: Looks like VScode messed up the spacing here.

milroy (approved):
Thanks for the reorganization!

There's a spacing issue that your editor introduced on one line. Please fix the alignment because it makes the block slightly confusing.

PR #1056, opened 2023-08-03
Remove tree

trws:
What it says on the tin, remove the tree scripts and tests for same.  In trying to test that removal, the run_sanity_check test was broken, removing the unnecessary templating fixed it.


resource/utilities/test/Makefile.am
```diff
@@ -3,12 +3,6 @@ TESTS = run_sanity_check.sh
 check_SCRIPTS = $(TESTS)
 
 CLEANFILES = $(check_SCRIPTS)
-EXTRA_DIST = resource-bench.sh benchmark.graphml.in benchmark.yaml run_sanity_check.sh.in
+EXTRA_DIST = resource-bench.sh benchmark.graphml.in benchmark.yaml
```
grondo, line 6: Maybe `run_sanity_check.sh` should be added here (`make distcheck` is failing)

resource/utilities/test/Makefile.am
```diff
@@ -3,12 +3,6 @@ TESTS = run_sanity_check.sh
 check_SCRIPTS = $(TESTS)
 
 CLEANFILES = $(check_SCRIPTS)
```
grondo, line 5: I think this line needs to be removed now that `run_sanity_check.sh` isn't generated..

grondo (commented):
A couple issues noted seem to be causing the failures in CI...

resource/utilities/test/Makefile.am
trws: Yes, this one too, ouch, not sure how I missed this one.

trws:
Looks like you got it in one @grondo, everything's passing.

grondo:
Last step would be to add something to the commit message body -- even when the commit title is self explanatory we try to have a descriptive body in the main flux-frramework projects. Lately we've adopted a `Problem:`/solution style format like:

> Problem: flux-tree is a research tool that is no longer installed and thus is unused
> and its presence is a maintenance burden.
> 
> Remove flux-tree and associated tests.

Though that style is probably not mandatory here.

trws:
Fair enough, done.  I used that format for the sanity check, since it really weirds me out that it happened as part of this PR.  Why it didn't show up before I have no idea, so having some explanation of the issue seems particularly appropriate.

grondo (approved):
Thanks!

PR #1057, opened 2023-08-04
transition bionic containers to jammy

trws:
This may end up having to merge before the tests on flux-core's version can pass, because of the flux-sched test run.  Hoping to get a jump on it here, somehow need to break the cycle.


grondo (commented):
LGTM, but honestly I'd consider just squashing all these commits into one big commit with a nice descriptive commit message saying something like `ci: replace bionic with jammy and bookworm` with a nice explanation in the message "bionic is out of LTS, etc" so update to jammy and add a bookworm image since flux-core has one, etc.

If you want to keep a couple separate commits that's fine too, but they'd need commit message to give slightly more detail.

And THANK YOU for doing this!

trws:
Thanks @grondo.  Most of these are squashed out, all the container oriented commits are.  There's a larger commit message, and updated to only build one jammy output along with using bookworm for most tests.

grondo (approved):
LGTM! Thanks @trws

PR #1060, opened 2023-08-22
mergify: remove rule listing, use branch protection

trws:
Rather than listing everything by name in two places, rely on the branch protections in the repo.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,17 +2,6 @@ queue_rules:
   - name: default
     conditions:
       - base=master
-      - status-success="validate commits"
-      - status-success="spelling"
-      - status-success="python format"
-      - status-success="python lint"
-      - status-success="bookworm - gcc-12,distcheck"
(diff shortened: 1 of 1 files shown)


trws:
Since this one is simple cleanup, I'm going to set MWP.  @grondo, let me know if you see any issues here but otherwise an approval should move it through.

jameshcorbett (commented):
This looks reasonable to me but I noticed we don't actually have branch protections on for these checks--only for "spelling".

trws:
Thanks for that @jameshcorbett, apparently I forgot to hit "save changes" in that tab. 🤦

jameshcorbett (approved):
LGTM then!

grondo:
Looks like `spelling` and  `jammy - test-install` are still missing from branch protections.

grondo:
I'll add those and press the button here

grondo (approved):
LGTM, I'll press the button!

PR #1062, opened 2023-08-23
Updating branch of golang release

vsoch:
This is an update to #1008 a̶n̶d̶ ̶f̶o̶r̶ ̶t̶h̶e̶ ̶t̶i̶m̶e̶ ̶b̶e̶i̶n̶g̶ ̶j̶u̶s̶t̶ ̶a̶ ̶t̶e̶s̶t̶  to see if a rebase with master here fixes our current testing issues. I had not made any changes to code aside from the Go, so my spidey senses are saying because the PR is old, the testing / CI is failing for niche reasons (possibly using an upstream container with incompatible changes) that might be fixed with a rebase. I'm flagging this for now since 1. it could entirely fail, and 2. I am absolutely terrible (as we have established before) with this git history thing, so I'm sure that is wrong. Safe to ignore for now :)

PR #1067, opened 2023-09-08
fix: devcontainer to use jammy and install cmake

vsoch:
Problem: the current development environment (VSCode Development Container) is not using the latest base (jammy) and is missing dependencies python3-yaml and an updated cmake and ninja for building.
Solution: Update Dockerfile to use flux-core jammy and add missing dependencies!

This will be needed for me to test or work on #1062

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:focal
+FROM fluxrm/flux-core:bookworm
 
 LABEL maintainer="Vanessasaurus <@vsoch>"
 
@@ -10,7 +10,6 @@ ENV USERNAME=${USERNAME}
 ENV USER_UID=${USER_UID}
 ENV USER_GID=${USER_GID}
 USER root
-RUN apt-get update
 
 # Install extra buildrequires for flux-sched:
 RUN sudo apt-get update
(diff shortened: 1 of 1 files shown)


trws:
I'd somewhat prefer we use the bookworm base for this TBH, ubuntu keeps breaking us, and the jammy image is less tested.  If you have a strong preference for jammy we can probably keep it, but it was pretty much the straw for me.

Also, the base package for cmake in either jammy or bookworm is easily new enough, so the apt package would be ok.  What are your thoughts?

vsoch:
@trws I chose jammy thinking we'd eventually need to upgrade to it, might as well sooner than later. Are you thinking that won't be the case an we'd remove / switch to another base?

trws:
We've already upgraded in that bionic (the old LTS release) is completely gone, as are old Fedora and RedHat, and Jammy is one of our base tested containers now.  In doing that though, we ran into even more problems dealing with Ubuntu's container images and found out that the even more recent release of Debian bookworm took a whole lot less fixing, so it's currently the "default" container on flux-core.

Anyway, I suppose the question is should we upgrade to Ubuntu Jammy or Debian Bookworm more than it is should we upgrade.

trws:
Actually, I'm ok with this basically whatever you decide.  Would like to have a PR to add the go toolchain into the sched containers and maybe this one too if we're hosting code for that though.

trws (approved):

vsoch:
@trws I am I indifferent so I will do whatever you think is best! Let me know your preference.

trws:
I'd prefer the bookworm container personally.  The ubuntu containers forcing "minimal" on us is especially painful in a devcontainer.  This is a new thing they're doing where there's no regular container anymore, the base jammy container is now more like bionic-minimal than bionic, and it's kinda painful interactively as a result.

vsoch:
Ok will update soon! Interviewing now.

vsoch:
@trws all set! I kept the direct install of cmake because we use that exact version in other places, and might be safer to ensure they are the same.

trws:
Seems reasonable.  At some point I'm going to switch the el8 container to use the system one, since that's what the TOSS4 buildbot will do, but this looks great.  Thanks @vsoch!

PR #1070, opened 2023-09-11
rc: call set -e after shebang in 01-sched-fluxion

### LINKED ISSUE #1069: avoid `#!/bin/bash -e` in rc1 script
As noted in https://github.com/flux-framework/flux-core/issues/5447, use of `#!/bin/bash -e` in Flux rc scrips can break for users that set `BASH_ENV=/etc/bashrc` since `bashrc` is not "errexit clean"

The proposed solution is to drop `-e` for `set -e` after the she-bang.

garlick:
Problem: If BASH_ENV is set to a script that is not errexit clean, the 01-sched-fluxion script may exit immediately.

Drop -e from the shebang and add a "set -e" afterwards.

Fixes #1069


grondo (approved):
LGTM! Thanks!

garlick:
Reworked as in flux-core to use `#!/bin/sh` in the rc scripts.

garlick:
I'll go ahead and set mwp here.

PR #1076, opened 2023-09-15
docker: avoid sudo in Dockerfile for arm runner

trws:
Problem: Github actions build for arm is dying on a nosuid mounted filesystem under the container trying to use sudo

Solution: Set user to root, avoid sudo.  We may want to do this on more Dockerfiles, but it has only showed up in the arm build so far.

This specifically came up here: https://github.com/flux-framework/flux-sched/actions/runs/6189435988/job/16803524042

Apparently overlayfs honors nosetuid in mounts, why this only shows up in the virtualized runner I have no idea but it's easier to fix it this way than try and debug the runners themselves.

### DIFF
--- a/src/test/docker/bookworm/Dockerfile
+++ b/src/test/docker/bookworm/Dockerfile
@@ -2,10 +2,11 @@ FROM fluxrm/flux-core:bookworm
 
 ARG USER=flux
 ARG UID=1000
+USER root
 
 # Install extra buildrequires for flux-sched:
-RUN sudo apt-get update
-RUN sudo apt-get -qq install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


grondo (approved):
This seems good to me. I actually wonder why we didn't do it this way from the beginning since it seems simpler (as long as the build and tests aren't run as root)

trws:
To be honest, I don't think it occurred to me until much later.  I always treated `USER` directives as being "inward" or "away from root" but since they can arbitrarily switch like `su` it does clean things up some.  If this ends up working, and especially if we run into this on other containers, it might be the way to go in future.

PR #1077, opened 2023-09-18
Change binfmt flags to ensure setuid is enabled in qemu-user-static supported CI runs

trws:
Second try to fix the aarch64 container builders.  AFAICT this is an issue with the way the binfmt_flags are set up.  I actually like this solution if it works, it's a lot easier to maintain than the opaque github action reference, and gives us more control over the behavior.

NOTE: I changed the behavior to build the arm container on push for debugging, ensure that's rebased out before this merges!


trws:
Here's a link to the successful test of this change: https://github.com/flux-framework/flux-sched/actions/runs/6220263727/job/16879950413?pr=1077 

It's not directly accessible from here anymore due to a rebase.  Hopefully this will keep similar problems from coming back, and we might want to apply a similar change to the actions file on core.

trws:
@grondo, apparently ReadTheDocs is switching what they require in their YAML at the same time.  Fix included here as well since it was required to get the tests green.  We'll want that on core for sure.  https://blog.readthedocs.com/use-build-os-config/

grondo:
We've already incorporated th build.os requirement in flux-core (actually, we only added the config file itself recently when it became a requirement), but thanks!

.github/workflows/main.yml
```diff
@@ -123,9 +123,9 @@ jobs:
       uses: docker/setup-buildx-action@v2
       if: matrix.needs_buildx
 
-    - uses: dbhi/qus/action@main
-      with:
-        targets: aarch64
+    - name: setup qemu-user-static
+      run: |
+        docker run --rm --privileged aptman/qus -s -- -p --credential aarch64
```
grondo, line 128: Question: The commit message references "This adds the `--capability yes` argument to the script", but I don't see anything in the actual commit that does something like that (this seems to just replace the `dbhi/qus/action` action with a `docker run` comand. Just checking if the commit message is outdated.

grondo (approved):
LGTM! One question about a commit message inline.

.github/workflows/main.yml
trws: Ah, I'll fix that, I meant "--credential" but mistyped it.

grondo:
Also: update PR title for release notes inclusion :wink:

PR #1080, opened 2023-09-26
ci: style checks change python version from 3.6 to 3.8

zekemorton:
Problem: python 3.6 is end of life and is now cauing failures
for the ci pipeline linting and formatting stages.

Change python version to 3.8. Change black version to 22.3.0
to fix formatting error from new click release. This only
affects the ci linting and formatting checks.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,13 +29,13 @@ jobs:
     steps:
     - uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
     - uses: actions/checkout@v2
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
     - name: install black
-      run: python3 -m pip install 'black==20.08.b1' --force-reinstall
+      run: python3 -m pip install 'black==22.3.0' --force-reinstall
     - name: check format
       run: ./scripts/check-format
 
@@ -45,7 +45,7 @@ jobs:
     steps:
     - uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
     - uses: actions/checkout@v2
       with:
         ref: ${{ github.event.pull_request.head.sha }}


chu11:
Perhaps this discussion needs to be brought up again, it's been almost 2 years

https://github.com/flux-framework/flux-core/issues/3929

grondo:
Python 3.6 is still the `/usr/bin/python3` in TOSS 4, so I don't think we can remove support for Python 3.6 any time soon since this is production software for those machines.

grondo:
Eh, though if this is just for style checks maybe its fine? The only worry would be if the new style or lint checks caused a developer to write python that didn't work on 3.6.

zekemorton:
That makes sense, unfortunately it looks like the change was made in the runner image, see here: https://github.com/actions/runner-images/issues/8238 on September 18th. It looks like the other ubuntu image does not support python 3.6 either, so I am not sure if there is a better alternative?

grondo:
Yes good point. I don't know of a better alternative at the moment. It does seem fairly low risk to update the runner python versions :shrug: I was just confused at first.

milroy:
The [runner issue](https://github.com/actions/runner-images/issues/8238) intimates that it may be possible to download and install a non-supported, archived version of Python on the runner: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python. I can't find any examples of how to do that, though. That approach might extend the action runtime considerably, though.

vsoch:
> Python 3.6 is still the /usr/bin/python3 in TOSS 4, so I don't think we can remove support for Python 3.6 any time soon since this is production software for those machines.

You could make the same argument backwards - if we only use 3.6 here basically any system that isn't a crustacean could run into unforseen issues. A lot of new features have been added since then.

If it's just an issue of getting a working (old python from _seven years ago_ lol then why not just install with conda (or miniconda is smaller)? They have versions that go back until the dawn of time. (cue sunrise with dinosaurs).

https://repo.anaconda.com/miniconda/ and then

grondo:
> You could make the same argument backwards - if we only use 3.6 here basically any system that isn't a crustacean could run into unforseen issues

We do test on python versions 3.6 - 3.10 in CI, so this is unlikely

grondo (approved):
This seems fine to me! Thanks!

vsoch:
yay! I love it when the answer to the problem is... what problem? :D

vsoch:
@grondo can we set merge-when-passing? I ask selfishly because I need this commit :)

grondo:
Yes! This is approved so that should work

vsoch:
boorah!! Thanks everyone!!

PR #1082, opened 2023-09-28
build: resurrect `make deb`

grondo:
This PR brings back the `make deb` target, which we've been using to build and install packages on test systems.

A `make dist` target, required by `scripts/debbuild.sh`, is added (and is now used by `make distcheck`), and `make dist` is fixed to work with slightly older versions of `git-archive` that do not support the `--add-virtual-file` option. Finally, the build system is forced to `cmake` in `debian/rules` since the fake `configure` script doesn't support all the requisite options.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ find_package(Boost 1.50 REQUIRED COMPONENTS
 message(STATUS "Boost version: ${Boost_VERSION}")
 
 # Install paths
-set(FLUX_CMD_DIR "${CMAKE_INSTALL_PREFIX}/libexec/cmd")
+set(FLUX_CMD_DIR "${CMAKE_INSTALL_LIBEXECDIR}/flux/cmd")
 set(FLUX_LIB_DIR "${CMAKE_INSTALL_LIBDIR}/flux")
 set(FLUX_MOD_DIR "${FLUX_LIB_DIR}/modules")
 set(FLUX_SHELL_PLUGIN_DIR "${FLUX_LIB_DIR}/shell/plugins")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,15 +197,25 @@ add_custom_target(check
   )
 add_custom_target(installcheck
   COMMAND env FLUX_SCHED_TEST_INSTALLED=1 ${CMAKE_CTEST_COMMAND} ${CTEST_COMMON_FLAGS})
+add_custom_target(dist
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  COMMAND echo ${FLUX_SCHED_VER} > flux-sched.ver
+  COMMAND git archive --format=tar.gz
+  --prefix=flux-sched-${FLUX_SCHED_VER}/
+  --add-file=flux-sched.ver
+  --output=${CMAKE_BINARY_DIR}/flux-sched-${FLUX_SCHED_VER}.tar.gz
+  HEAD .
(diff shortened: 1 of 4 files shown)


grondo:
Hold off on merging this until #1083 is solved

trws (approved):
Oh, I missed that you had already put in a fix for #1083 here @grondo, I'm good with this as a fix as well if it takes care of the issue.  Thanks for the fix!

grondo:
Thanks @trws! The one issue I'm still fighting with is #1083.

So far, from lots of online searching I've come up with this as the best way to install a link:
```diff
diff --git a/src/python/CMakeLists.txt b/src/python/CMakeLists.txt
index b4228f69..3efd2dd4 100644
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -12,3 +12,22 @@ install(DIRECTORY fluxion
   DESTINATION "${PYTHON_INSTALL_SITELIB}"
   FILES_MATCHING PATTERN "*.py")
 
+set(PYNAME "python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}")
+set(PYDIR "${CMAKE_INSTALL_FULL_LIBDIR}/flux/${PYNAME}")
+set(LINKTARGET "${PYTHON_INSTALL_SITELIB}/fluxion")
+
+INSTALL(CODE "\
+  execute_process( \
+  COMMAND \
+    mkdir -p \$ENV{DESTDIR}/${PYDIR} \
+  )
+  execute_process( \
+  COMMAND \
+    rm -f \$ENV{DESTDIR}/${PYDIR}/fluxion \
+  )
+  MESSAGE (STATUS \
+    \"linking ${PYDIR}/fluxion -> ${PYTHON_INSTALL_SITELIB}/fluxion\")
+  execute_process( \
+  COMMAND \
+    ln -s ${LINKTARGET} \$ENV{DESTDIR}/${PYDIR}/fluxion \
+  )")
```


Actually I was going to report that this works for `make install DESTDIR=` but not for `make deb`, but this morning it started working in both cases. However, this seems to be magically working in both cases now, so I'll probably incorporate these changes here. However, does the above look correct to you?

Also, is there some way to prevent cmake from rebuilding the whole project when you touch one file? That makes it really painful to work on the install target when you end up building from scratch with each test (especially with Fluxion)...

grondo:
Ok, I've pushed the latest hopefully working set of changes. @garlick, want to give this branch a try?

trws:
That looks reasonable.  Usually it's best practice to use the `cmake -E ...` commands for that kind of thing for portability, but we require a unix-like environment in so many places it really doesn't matter.

I'm looking into the rebuild issue.  To be honest I'm not sure, I have a feeling it's being defensive due to the use of some of the custom commands, but I'm not completely sure _why_ yet.

trws:
Blarg, yes, there are some dirty edges because I missed specifying explicit inputs for a couple of targets.  Hopefully that will take care of it once I clean those up.

garlick:
Looking good!  I did notice a couple formerly "noinst" programs have leaked into the install
```
-rwxr-xr-x root/root     35392 2023-09-29 16:04 ./usr/bin/grug2dot
-rwxr-xr-x root/root    495480 2023-09-29 16:04 ./usr/bin/resource-query
```
but that's not an issue with the deb per se so I'll approve and set MWP.

Thank you!  It will be good to be able to install test releases of fluxion on my test cluster again.

garlick (approved):
:+1:

grondo:
Awesome. Thanks @trws and @garlick!

PR #1086, opened 2023-09-28
build: don't look for things we don't use

garlick:
This drops support from the build system for finding flux-core's libpmi.so and libczmq.so, neither of which are used by the scheduler.

Also, drop the deprecated workflow-examples readthedocs repo from the sphinx config.


grondo (approved):
LGTM!

trws (approved):
This looks great @garlick, thanks!

Note to self: pull out version of FindFluxCore from just before this as something we can provide from flux-core, like a pkg-config file.

Issue #1093, opened 2023-10-01
libresource.so should not be installed or should be renamed and versioned


grondo:
I could be incorrect, but I don't think the `libresource.so` convenience library used be installed by the autotools based build system.
This library now appears to be installed directly into `$libdir`, and it does appear that the `sched-fluxion-resource.so` is linked to it:
```
$ ldd /usr/lib/aarch64-linux-gnu/flux/modules/sched-fluxion-resource.so   | grep resource
	libresource.so => /lib/aarch64-linux-gnu/libresource.so (0x0000007f8b826000)
```
If this library is to now be installed, I'd suggest naming it less generically to `libfluxion-resource.so`. If the library is meant for general linking then it should probably also get a version (not sure if this is true at this point though).

grondo:
It also seems to be leaking a lot of symbols, though maybe this is normal for a c++ library?
```
$ nm -D  /usr/lib/aarch64-linux-gnu/libresource.so  | wc -l
2165
```

Issue #1092, opened 2023-10-01
fluxion modules installed to different path than expected in jammy container


grondo:
As reported by @vsoch on slack, the Fluxion modules are not being found in the default module search path in (at least) the jammy container.

This appears to be because the flux-core install is using a non-multiarch path for `$libdir` (`/usr/lib`), while the new cmake-based build system in flux-sched is detecting the multi-arch path (`/usr/lib/x86_64-linux-gnu`). In this case, the default in the flux-sched build is correct, but that is immaterial since it isn't installing to the default module path supported by the broker.

The flux-core pkg-config file does export a series of paths which could be used here rather than implied defaults. E.g. in the jammy image:

```console
$ cat /usr/lib/pkgconfig/flux-core.pc 
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
datarootdir=${prefix}/share
mandir=${datarootdir}/man
fluxrcdir=/etc/flux
fluxrc1dir=/etc/flux/rc1.d
fluxrc3dir=/etc/flux/rc3.d
fluxcmddir=${exec_prefix}/libexec/flux/cmd
fluxlibdir=${exec_prefix}/lib/flux
fluxmoddir=${exec_prefix}/lib/flux/modules
fluxcmdhelpdir=${datarootdir}/flux/help.d
fluxshellrcdir=/etc/flux/shell
fluxshellpluginpath=${exec_prefix}/lib/flux/shell/plugins

Name: flux-core
Description: Flux Resource Manager Framework Core
Version: 0.54.0-155-g544f916dd
Libs: -L${libdir} -lflux-core
Cflags: -I${includedir}
```

Using these pkg-config vars will probably be the fool proof way to set several of these paths instead of the `GNUInstallDirs` defaults.

vsoch:
Thanks @grondo !

vsoch:
Is there a quick fix or do we need to rebuild? My focal workaround will not work without the higher python version provided by jammy:

```
 > [5/5] RUN poetry lock &&     poetry install &&     chown -R fluxuser /home/fluxuser:                                                                                                                                                                                                
0.431 Creating virtualenv snakemake-executor-plugin-flux-6hbiHWzC-py3.8 in /root/.cache/pypoetry/virtualenvs                                                                                                                                                                           
0.698 Updating dependencies
0.698 Resolving dependencies...
6.843 
[...]

Issue #1091, opened 2023-10-01
docs are not built by default with `make`?


grondo:
I may be doing something wrong, but it appears `make` doesn't build the docs automatically like it used to, and you have to run `make docs` to get them built:

```console
$ make -j 6
[snip]
[ 98%] Building CXX object resource/utilities/CMakeFiles/grug2dot.dir/grug2dot.cpp.o
[ 98%] Built target schema_test02
[ 98%] Built target matcher_util_api_test
[100%] Linking CXX executable grug2dot
[100%] Built target sched-fluxion-resource
[100%] Built target grug2dot
[100%] Built target resource-query
$ ls doc/man5/*.5
ls: cannot access 'doc/man5/*.5': No such file or directory
$ make docs
[ 50%] Generating man5/flux-config-sched-fluxion-qmanager.5, man5/flux-config-sched-fluxion-resource.5, ALL
Running Sphinx v3.4.3
[snip]
[100%] Built target html
[100%] Built target docs
$ ls doc/man5/*.5
doc/man5/flux-config-sched-fluxion-qmanager.5
doc/man5/flux-config-sched-fluxion-resource.5
```

PR #1094, opened 2023-10-02
build: fixes for minor build issues

grondo:
This PR fixes a series of minor build issues discovered recently and reported in #1091, #1092, and #1093.

I'm not actually sure if there was some motivation for installing libresource.so, a difference from what we've done previously. For now, since this matches previous behavior, it seems low risk to fall back to making libresource a convenience library.


trws:
All of these work for me. The docs issue was a mistake in breaking the dependencies, and Libre source I was trying to match what autotools did but must have misinterpreted. Thanks for these @grondo, good stuff!

grondo:
Thanks @trws! Also hitting an issue in the fedora builders that I am looking into.

grondo:
To get the fedora builds to work I had to `pip3 install sphinx_rtd_theme`, though I don't know why this wasn't an issue before. I also tried to 

The `iterator` deprecation message may be a red herring, looks like the reapi build is failing here:

Also notice a link to `czmq` which should no tbe necessary.

```
cd /usr/src/resource/reapi/bindings/go/src/test && GOPATH=/usr/src/go:/usr/src/resource/reapi/bindings/go GOOS=linux G0111MODULE=off CGO_CFLAGS="-I/usr/src/resource/reapi/bindings/c" CGO_LDFLAGS='-Wl,-R /usr/src/resource -L/usr/src/resource/reapi/bindings -L/usr/src/resource/libjobspec -ljobspec_conv -lreapi_cli -L/usr/src/resource -lresource -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp' go build -ldflags '-w' -o /usr/src/resource/reapi/bindings/go/src/test/main main.go
  # command-line-arguments
  /usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
  /usr/bin/ld: /usr/src/resource/libresource.a(infra_data.cpp.o): in function `Flux::resource_model::pool_infra_t::~pool_infra_t()':
  /usr/src/resource/schema/infra_data.cpp:143: undefined reference to `planner_multi_destroy'
```

vsoch:
> I approve, but it looks like we're now hitting some new reapi issue that some how slipped in? @vsoch, @milroy have you seen the boost iterator deprecation issue elsewhere?

I'm not sure that was from the golang PR, at least the portion I'm familiar with we didn't touch c++ stuff, just Go! But I think it looks like this?

https://github.com/boostorg/graph/issues/228

Can the version of boost be bumped? I'm guessing it's just that fedora 34 is kind of old now - I don't even see it in their list! https://src.fedoraproject.org/rpms/boost. Can we use a newer set of versions? I think it's end of life? https://docs.fedoraproject.org/en-US/releases/eol/

vsoch:
Wow they do a new release every 6 months and only maintain for 13! That is some fast moving right there: https://docs.fedoraproject.org/en-US/releases/lifecycle/. And then the schedule: https://fedorapeople.org/groups/schedule/

It's kind of nuts when you consider ubuntu LTS and _five years_ ! https://ubuntu.com/about/release-cycle

trws:
For context, we have fedora 34 because that's the version RHEL _9_ forked from, which is one rhel version above what we use in TOSS4. We could swap that for alma or whatever redhat won't sue us for that's actually usable for CI, but haven't done the legwork for that yet.

trws:
Wait, it looks like that link line doesn't have planner in it. Maybe that was being exposed by libresource as a shared lib but isn't as static for some reason? We just got back from the louvre and I need to put Kim to bed but that's where I would look.

vsoch:
> We just got back from the louvre 

!!! :raised_hands: :art:

grondo:
> Wait, it looks like that link line doesn't have planner in it. Maybe that was being exposed by libresource as a shared lib but isn't as static for some reason?

Well I don't know CMake very well (at all), but `resource/CMakeLists.txt` does have this:

https://github.com/flux-framework/flux-sched/blob/b29177b184e3093b99339ebfe4ec630c64fcad7a/resource/CMakeLists.txt#L75C1-L77C6
```cmake
target_link_libraries(resource PRIVATE
    planner
    )
```

trws:
Yup, that's almost certainly it.  Either that needs to be `PUBLIC` or `reapi` needs an extra dep on `planner`.  The former is simpler to deal with it because of having to extract everything from targets for golang, but either should work.

grondo:
I did try this on a whim:
```diff
diff --git a/resource/CMakeLists.txt b/resource/CMakeLists.txt
index 4c1cfc97..36759a71 100644
--- a/resource/CMakeLists.txt
+++ b/resource/CMakeLists.txt
@@ -72,7 +72,7 @@ add_library(resource STATIC
     ${RESOURCE_HEADERS}
     )
 target_include_directories(resource PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(resource PRIVATE
+target_link_libraries(resource PUBLIC
     planner
     )
 target_link_libraries(resource PUBLIC

```

But it didn't seem to do anything, do I need to clear some kind of CMake cache?

trws:
I wouldn't have thought so, but maybe.  Generally it's best to just wipe out the build directory as a full clean, but removing CMakeCache.txt should make that regenerate as well, it just wont necessarily rebuild everything afterwards.

vsoch:
@grondo I might be wrong, but I've had a lot of trouble with cmake because typically you do something like:

```bash
mkdir -p build
cd build
cmake ../
```

And then the build (cmake files) are generated separately from the source. But for the setup here, they get mixed, and although I tried a lot of clean and similar commands, I could never get it to a "clean" state (removing whatever cache is there) unless I nuked and started from scratch. I bet there is some expert / advanced way to do it that I'm not experienced enough to know. That might be good to know - but maybe @trws we could next try to prioritize fixing this setup?

grondo:
I used `git clean -xdf`.

trws:
You're right @vsoch, it's a really bad idea to build in-source with `./configure`, using `cmake -B build` or going into a new directory and using `cmake ..` works with our current setup already.  The configure script also works if you go into a directory and use `../configure` just like autotools, but to make it work like autotools it has to allow `./configure` in the source tree because our CI and so-forth do that all over the place.  Normally I have the cmake setup throw an error if someone tries it, but I can't without refactoring everything to always build in a separate directory.

Anyway, the only reliable way I know to get a repo clean after that, short of blowing it away, is to use `git clean` to wipe out everything that isn't checked in.

grondo:
Well, I'm still getting link errors, even after moving the libplanner link to the existing PUBLIC stanza: (and adding libplanner_cxx just in case)
```diff
diff --git a/resource/CMakeLists.txt b/resource/CMakeLists.txt
index 4c1cfc97..850175cd 100644
--- a/resource/CMakeLists.txt
+++ b/resource/CMakeLists.txt
@@ -72,10 +72,9 @@ add_library(resource STATIC
     ${RESOURCE_HEADERS}
     )
 target_include_directories(resource PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(resource PRIVATE
-    planner
-    )
 target_link_libraries(resource PUBLIC
+    planner
+    planner_cxx
     flux::hostlist
     flux::idset
     PkgConfig::JANSSON
```

```
/usr/bin/ld: /usr/src/resource/libresource.a(resource_reader_jgf.cpp.o): in function `Flux::resource_model::resource_reader_jgf_t::update_vtx_plan(unsigned long, boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, Flux::resource_model::resource_pool_t, Flux::resource_model::resource_relation_t, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, boost::listS>&, fetch_helper_t const&, unsigned long, long, unsigned long, bool)':
resource_reader_jgf.cpp:(.text+0x541e): undefined reference to `planner_avail_resources_during'
/usr/bin/ld: resource_reader_jgf.cpp:(.text+0x5511): undefined reference to `planner_add_span'
/usr/bin/ld: /usr/src/resource/libresource.a(resource_reader_jgf.cpp.o): in function `Flux::resource_model::resource_reader_jgf_t::undo_vertices(boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, Flux::resource_model::resource_pool_t, Flux::resource_model::resource_relation_t, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, boost::listS>&, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, vmap_val_t, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, vmap_val_t> > >&, unsigned long, bool)':
resource_reader_jgf.cpp:(.text+0x5c01): undefined reference to `planner_rem_span'
/usr/bin/ld: /usr/src/resource/libresource.a(resource_reader_rv1exec.cpp.o): in function `Flux::resource_model::resource_reader_rv1exec_t::add_vertex(boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, Flux::resource_model::resource_pool_t, Flux::resource_model::resource_relation_t, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, boost::listS>&, Flux::resource_model::resource_graph_metadata_t&, unsigned long, long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > const&, int, int)':
resource_reader_rv1exec.cpp:(.text+0x488): undefined reference to `planner_new'
/usr/bin/ld: resource_reader_rv1exec.cpp:(.text+0x4cb): undefined reference to `planner_new'
```

trws:
Ok, that means it isn't getting propagated into reapi the way it would be normally by cmake.  @vsoch it looks like there is an explicit list of `-l` and `-L` flags in there rather than pulling them from the targets? If so, we'll need to add explicit flags for planner, rework the build for libresource to include the objects of planner, or extract the flags from the dep targets.  Also it looks like there's a list of targets being passed at the end of BUILD_GO_PROGRAM that the function doesn't read? Not sure what's going on there.

grondo:
@vsoch, @trws: Any ideas for a quick workaround? Trying to prep flux-sched for an RPM build tomorrow and this PR would be helpful on that front. Unfortunately, I may not know enough cmake to be helpful here - though maybe with a hint I could put together a workaround for now...

trws:
Best I have for quick is to add `-l planner` and the appropriate path if necessary to the CGO_LIBRARY_FLAGS to just do it explicitly for now.  I'm looking at it, and the longer-term solution is probably going to be to link a library of all the stuff that reapi will use, then link that to reapi, so we can use all the logic to pull in deps properly, but that may take more work.  Or I guess really it might be easier to just do that, make a shared lib that has no code but links resource and just `-l` that instead of resource?

vsoch:
> Ok, that means it isn't getting propagated into reapi the way it would be normally by cmake. @vsoch it looks like there is an explicit list of -l and -L flags in there rather than pulling them from the targets? 

Can you tell me what you mean by "in there" ?  This is the one that I have knowledge about:

https://github.com/flux-framework/flux-sched/blob/b29177b184e3093b99339ebfe4ec630c64fcad7a/resource/reapi/bindings/go/src/test/CMakeLists.txt#L6

And I don't know how to do it any other way, so your suggestions are appreciated.

> If so, we'll need to add explicit flags for planner, rework the build for libresource to include the objects of planner, or extract the flags from the dep targets. 

If you mean here: https://github.com/flux-framework/flux-sched/tree/master/resource/planner

I don't think we touched that for the Go work.

> Also it looks like there's a list of targets being passed at the end of BUILD_GO_PROGRAM that the function doesn't read? Not sure what's going on there.

Those are supposed to be added here as ARGN deps to `add_dependencies`: https://github.com/flux-framework/flux-sched/blob/b29177b184e3093b99339ebfe4ec630c64fcad7a/cmake/GolangSimple.cmake#L24-L26. So I think they are used, but I couldn't tell you if they do anything meaningful.

grondo:
> Best I have for quick is to add -l planner and the appropriate path if necessary to the CGO_LIBRARY_FLAGS to just do it explicitly for now. 


That worked for me. I seemed to have to add both `-lplanner` and `-lplanner_cxx`

```diff
--- a/resource/reapi/bindings/go/src/test/CMakeLists.txt
+++ b/resource/reapi/bindings/go/src/test/CMakeLists.txt
@@ -3,7 +3,7 @@ set(SRCS main.go)
 set(CGO_CFLAGS "-I${CMAKE_BINARY_DIR}/resource/reapi/bindings/c")
 
 # This is currently passed but not used because when passed into add_custom_com
mand the spaces are escaped
-set(CGO_LIBRARY_FLAGS "-Wl,-R ${CMAKE_BINARY_DIR}/resource -L${CMAKE_BINARY
_DIR}/resource/reapi/bindings -L${CMAKE_BINARY_DIR}/resource/libjobspec -ljobspe
c_conv -lreapi_cli -L${CMAKE_BINARY_DIR}/resource -lresource -lflux-idset -lstdc
++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-
cpp")
+set(CGO_LIBRARY_FLAGS "-Wl,-R ${CMAKE_BINARY_DIR}/resource -L${CMAKE_
BINARY_DIR}/resource/reapi/bindings -L${CMAKE_BINARY_DIR}/resource/libjobspec -l
jobspec_conv -lreapi_cli -L${CMAKE_BINARY_DIR}/resource -lresource -L${CMAKE_BIN
ARY_DIR}/resource/planner/c -lplanner -L${CMAKE_BINARY_DIR}/resource/planner/c++
 -lplanner_cxx -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lf
lux-hostlist -lboost_graph -lyaml-cpp")
 
 # This ensures the main binary is cleaned up
 set_directory_properties(
```

trws:
That would be perfectly fine for this then. I'll take a look at making dependency flag propagation work for reapi but that's going to take a bit.

vsoch:
> That would be perfectly fine for this then. I'll take a look at making dependency flag propagation work for reapi but that's going to take a bit.

Thank you! And apologies for the issues - I'm not sure why these didn't bubble up in previous CI.

grondo:
> I'm not sure why these didn't bubble up in previous CI.

That must be my fault due to the change here to make libresource a non-installed, static library.
Unfortunately, I don't know enough CMake to fix it properly.

The last fix needed here is to rectify the sphinx build in the Fedora builders. I'm not sure why it fails to find `sphinx_rtd_theme` here but works in flux-core CI (maybe we're not using that theme there?) I'll check. Anyway, worst case I'll disable docs build on these builders until we figure out the correct fix.

vsoch:
If it fails again (sphinx) I can take a look. I use quite a bit of that ecosystem and I suspect it would come down to version differences. Especially in the last few releases of sphinx there have been several hard bugs with pydantic and other underlying deps. Anyhoo, I can take a look and see if anything makes sense.

grondo:
Yes, if you don't mind taking a look. The thing making me scratch my head is that sphinx-build works in the flux-core build in the Fedora  images, but not here. Since the CI here is based on the docker images from flux-core, the sphinx version and everything _should_ be the same..

grondo:
Ah, the issue might be that in flux-core, only man pages are built by default but here I've got the html building as well (unnecessary) -- I'll try to fix that.

vsoch:
fedora builders say...

image

grondo:
Ok, let's try this -- `make` and `make docs` now. just build manpages by default and `make html` (in the docs dir) builds html docs. I think this was how things worked previously and why the build used to work on fedora (which is missing the sphinx_rtd_theme package)

grondo:
Yay :green_circle: all systems green.

grondo:
Since @trws approved I'm going to set MWP. Any missteps here can be fixed up later.

trws:
I'm guessing the docs issue is that we had `ALL` on the manpage target rather than the docs target before, it fits what you describe as symptoms.  We'll get this straightened out, I need to chat with @vsoch and @milroy a bit to figure out how the go build is intended to work and be maintained so we can make it all fit, but after a bit of digging it looks like we can make it pretty nice.

grondo:
> I'm guessing the docs issue is that we had ALL on the manpage target rather than the docs target before, it fits what you describe as symptoms.

Yes that was the issue. Here I added ALL to the `docs` target and removed `html` from it to get the previous behavior (only manpages are built by default, but you can do `make html` in the `docs` directory to generate html docs. Hopefully that was correct!

trws:
Works for me.

-Tom (mobile)

vsoch:
@trws what was the ultimate fix for boost? I think we need an update to the spack spec - the new build is failing. https://github.com/flux-framework/spack/actions/runs/6400783639/job/17375398411 I can do this update if you can tell me what kind of pin or similar we should do.

trws:
I don’t see the error message in the output. If it’s running with auto tools then there is a persistent issue with some distros and the autoconf macros for boost, I think that one was ensuring that rather than just —with-boost we pass each boost package’s —with- option separately. If it’s the warning turning error, then we need to add a -Wno-error= for the warning code. If it’s with the cmake build, I haven’t seen that cause issues with boost yet so I need the error to debug.

vsoch:
Oh gosh I didn’t even consider that! We need to totally redo the spack recipe for cmake, and a non-traditional approach to building… that’s most likely the issue.

trws:
What do you mean non-traditional? The configure script is 100% optional, everything is implemented in the cmake, so it will build as a bog-standard cmake package. It will need some of the new build system directives that slack added to handle transitions from one to another though so we don’t break previous versions building with the package.

Sent from my iPhone

On Oct 4, 2023, at 9:54 AM, Vanessasaurus ***@***.***> wrote:

﻿

Oh gosh I didn’t even consider that! We need to totally redo the spack recipe for cmake, and a non-traditional approach to building… that’s most likely the issue.

vsoch:
Oh sorry I meant to say we don’t make a separate build root first, which I’d assume is what spack does by default. Probably we can override that step.

trws:
We support it though. That’s what I was getting at earlier. Spack does it by default for cmake, and that’s what we want. For auto tools it doesn’t because many auto tools projects break with a separate build root. In our case all flux projects build either in tree or out of tree, and I always build all of them out of tree, so I see no reason to override it unless we want to override the unnecessary in-tree auto tools default.

Sent from my iPhone

On Oct 4, 2023, at 9:59 AM, Vanessasaurus ***@***.***> wrote:

﻿

Oh sorry I meant to say we don’t make a separate build root first, which I’d assume is what spack does by default. Probably we can override that step.

vsoch:
hey @grondo ! It looks like another issue slipped in from this PR - specifically we need to build *.so for each of the reapi_cli bindings, jobspec_conv.so and resource (e.g., libresource.so and libreapi_cli.so). Specifically this is how we build the fluence sidecar server:

```
GOOS=linux CGO_CFLAGS="-I/home/flux-sched/resource/reapi/bindings/c" CGO_LDFLAGS="-L/usr/lib -L/home/flux-sched/resource -lresource -L/home/flux-sched/resource/libjobspec -ljobspec_conv -L//home/flux-sched/resource/reapi/bindings -lreapi_cli -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp" go build -ldflags '-w' -o bin/server cmd/main.go
```
And these are the changes that I make:

```
sed -i 's/add_library(resource STATIC/add_library(resource SHARED/g' resource/CMakeLists.txt
sed -i 's/add_library ( reapi_module STATIC/add_library ( reapi_module SHARED/g' resource/reapi/bindings/CMakeLists.txt
sed -i 's/add_library ( reapi_cli STATIC/add_library ( reapi_cli SHARED/g' resource/reapi/bindings/CMakeLists.txt
sed -i 's/add_library ( jobspec_conv STATIC/add_library ( jobspec_conv SHARED/g' resource/libjobspec/CMakeLists.txt
```

Enter Dr. Jankmaster! :woman_health_worker:  In the above, I'm using sed to change STATIC to SHARED, but ideally I could set an envar or similar. Is this something we could consider?

grondo:
This would be something to ask @trws. I don't know enough cmake to answer, sorry!

vsoch:
@grondo I think I could give it a shot (it could just be an envar or similar I think?) but mostly wanted to get permission if there is some licensing issue or other thing I don't know about. But I'm going back to sleep for a bit anyway so... happy to wait for the wisdom of @trws ! :raised_hands:

grondo:
There is probably a "cmake" way to do this, with autotools it would be something like `--enable-shared-all` or something. With Cmake I gather it is something you specify with `-D` on the cmake command line. Doing it that way would be preferred IMO.

I don't know enough about Fluxion to make an informed opinion on whether all the convenience libraries can be made into DSOs and installed, but the other thing to consider is that typically libraries build statically in a project are purposefully not made into shared libs and installed because they do not have good public interfaces, or they expose structures and classes that should be internal-only. Consider well if you want to expose an interface publicly before doing so, because then ofttimes you are locked into maintaining that API (and struct if you are exposes structures). One good strategy is to expose a small, well designed portion of the interface at first, and only install that. That is just my :coin: :coin:

trws:
I put something up on slack about this, but generally with cmake you either specify it for the library in the source, or you can make it a define.  If you need both the best way is usually to make an object library then use its targets in one static and one shared library, that way it only builds the objects once.

PR #1097, opened 2023-10-04
ci: remove deprecated syntax and update release action

grondo:
This PR updates to the simpler `softprops/gh-release-action` action to create a release after a tag (as used by flux-core and other framework projects) and also removes the use of `set-output` which has been deprecated for awhile.


garlick (approved):
LGTM

PR #1100, opened 2023-10-09
ci: do not force configure --prefix to /usr

### LINKED ISSUE #1099: Quick start container broken (module not found)
The first quick start step is broken: https://flux-framework.readthedocs.io/en/latest/tutorials/containers/flux-sched-container.html

 Running on Ubuntu 22.04:

```
$ docker run -ti fluxrm/flux-sched:latest
Oct 06 02:11:37.348654 broker.err[0]: rc1.0: flux-module: load sched-fluxion-resource: module not found in search path
Oct 06 02:11:37.349482 broker.err[0]: rc1.0: /etc/flux/rc1 Exited (rc=1) 4.8s
```

```
$ docker image ls | grep fluxrm
fluxrm/flux-sched                                         latest    0dd6172688fa   33 hours ago    2.13GB
```

Guessing a path issue: 
```
$ docker run -ti --entrypoint bash fluxrm/flux-sched:latest
 fluxuser@9112d7d31b1e:~$ flux module list
flux-module: flux_open: No such file or directory
```

grondo:
Problem: When flux-sched is built in ci using the configure wrapper for cmake, a prefix of /usr is forced. This bypasses the automatic setting of libdir to match flux-core and results in Fluxion modules installed in the wrong path.

Drop the use of --prefix=/usr in generate-matrix.py. This isn't necessary since the prefix will be set to that of the first flux(1) found in PATH, which is /usr/bin/flux.

Fixes #1099


grondo:
Verified the distcheck build is now using `/usr/lib` from "configure" output:
```
  -- flux core found and no specific prefix specified, using /usr
  -- Using flux-core libdir of /usr/lib
```

trws:
Hm, I'm good with this, but we probably also want to make the libdir *always* take the flux-core libdir don't you think? If someone uses the explicit prefix otherwise we'll still get the wrong behavior won't we?

trws (approved):

grondo:
>  I'm good with this, but we probably also want to make the libdir always take the flux-core libdir don't you think? I

I was thinking the same thing, but didn't go that far at this point (was mainly trying to fix the docker images generated by CI).

Happy to make that change here -- I'll give it a go unless you'd like to propose something.

Alternately, the flux-core pkgconfig file explicitly lists the flux-core module dir as `fluxmoddir`:
```
fluxmoddir=${prefix}/lib/aarch64-linux-gnu/flux/modules
```
which could allow the fluxion modules to go the right spot without overriding the default `libdir` set by prefix. However, I wasn't sure how to get extra pkgconfig variables from cmake. Anyway, a change like that could always be done later if, for some reason, it was deemed necessary.

grondo:
Ok, I've added a commit that always forces the flux-sched `libdir` to that of the detected flux-core.

grondo:
I'll go ahead and set MWP here so that our docker images get fixed. If the forcing of `libdir` was a misstep we could easily remove that in the future.

Issue #1098, opened 2023-10-04
reapi cli: match allocate job info and return bug 


zekemorton:
Found a few bugs in  `reapi_cli_t::match_allocate`
 - The return code does not indicate whether or not the job spec was successful in allocation. It seems like the variable `at` is being used within the function to determine if the job spec was allocated or reserved, with no check for failure. Note that it seems like `at` does not get changed in case of failure, so if used in something like resource query, it will likely default to 0 to indicate allocated. A failed allocation will still create a `job_info_t` even though nothing has been scheduled and label it as allocated.  
 - the parameter `jobspec` is meant to be a string representation of the job spec itself rather than the file name for the job spec. This makes it so that `jobspec_fn` cannot be set in the `job_info_t` for the job. This can possibly be addressed by changing the parameter or adding in an `update_info_jobspec_fn` type of function to `reapi_cli_t`. This is important for any tests that use resource query's info command to match
 - the elapsed time is set in the variable `ov` after the `job_info_t` is set, making the value always 0. This should be an easy fix with some reordering

zekemorton:
Addressed in merged PR noted above. 

Note this from a comment in the PR:
"we found that changing the interface for reapi_clit_t::match_allocate has bigger implications and the return behavior should be left as is. A non 0 return code represents an error when running the code rather than a failure to find a match.
rq2 now determines if a match was found by checking errno for ENODEV, EBUSY, EINVAL, all of which indicate a failure to match."

PR #1101, opened 2023-10-11
RQ2 minimal functionality and bug fix to reapi match allocate 

zekemorton:
This MR includes changes to the reapi to support baseline rq2 functionality and some basic tests to test out rq2. 

rq2 in this MR can handle all of the cli arguments, though the write_to_graph function is not yet implemented so the `-g` and `-o` flags do not yet do anything. rq2 only implements 3 commands, `help`, `quit`, and `match`. For `match` only the sub commands `allocate` and `allocate_orelse_reserve` are currently supported. This MR introduces a lot of infrastructure around rq2 and should set the stage for future development in this effort. 

The test expected output was generated with resource-query and should be a good measure of whether rq2 is behaving properly.


zekemorton:
@vsoch do you mind taking a look at #1098 and [this commit ](https://github.com/flux-framework/flux-sched/pull/1101/commits/aae650605dc344576bd8b3429fff83eae2c23608). It changes the behavior of the return value for `reapi_cli_t::match_allocate` and is now causing one of the go bindings test to fail with error. The second match allocate (line 53 in main.go for testing) is now resulting in error because the match allocate fails. 

What are your thoughts on this? Should `reapi_cli_t::match_allocate` return non 0 if `traverser_run` returns non 0? I want to be careful about making changes to the reapi that would affect existing code thats using it

vsoch:
Sure! So that function is used in the go bindings here: https://github.com/flux-framework/flux-sched/blob/aae650605dc344576bd8b3429fff83eae2c23608/resource/reapi/bindings/go/src/fluxcli/reapi_cli.go#L102-L109. The general logic is that we expect match_allocate (and more generally, calls to flux) to return an integer return code. The function here https://github.com/flux-framework/flux-sched/blob/aae650605dc344576bd8b3429fff83eae2c23608/resource/reapi/bindings/go/src/fluxcli/reapi_cli.go#L40-L44 then turns that into a proper go error - nil if the return code is 0 and an error otherwise.

>  The second match allocate (line 53 in main.go for testing) is now resulting in error because the match allocate fails.

Can you link me to the failing test? Why would it be correct that match allocate fails (returns non zero?)

> What are your thoughts on this? Should reapi_cli_t::match_allocate return non 0 if traverser_run returns non 0? I want to be careful about making changes to the reapi that would affect existing code thats using it

I don't know the details of traverser_run - if it fails does that mean the match allocate has failed too? In that case why would we want to hide the error from the Go bindings? Looking closer - it looks like you are adding an rc2 to the function, and returning the sum of rc1 and rc2. Can we make a small table of the possible return values given this addition, and the action (if it's considered a failure or not) for each? I'm guessing if rc is -1 and rc2 is 0 we'd get a value of -1, and maybe that doesn't indicate an actual error has occurred (and we just need to account for that case in the Go bindings)?

zekemorton:
> Can you link me to the failing test?
https://github.com/flux-framework/flux-sched/actions/runs/6512317565/job/17689731284?pr=1101#step:7:4178
> Why would it be correct that match allocate fails (returns non zero?)
It should fail if you try a match allocate and no resources are available for that allocate. In this case, nothing is actually scheduled or reserved. In any case, whether we determine this from return codes or from some other means, it would be important to know if a match was successful or not. Note that a `MATCH_ALLOCATE_ORELSE_RESERVE` in this case would succeed since the job can be scheduled for the future. 

> don't know the details of traverser_run - if it fails does that mean the match allocate has failed too?
traverser_run is a pass through function for `dfu_traverser_t::run` and this function calls a schedule function. If that schedule function fails, it will return -1 I think. This function failing means that the match allocate has failed.  https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu.cpp#L309C7-L309C7

> Can we make a small table of the possible return values? 
rc = 0 on success or -1 on failure and has to do more infrastructure types of things, memory error, not able to get time of day, etc. Failure indicates that the process is partially completed and sets `m_err_msg` with an appropriate message.
rc2 = 0 on success or -1 on failure and strictly has to do with weather or not the traverser run function was successful. 
having rc =-1 is a weird failure that shouldn't ever really happen, whereas rc2 will fail if you give it an unsatisfiable request if that make sense? 
Without these changes, what is now rc2 never actually gets checked and get over written for the other checks with rc. 

> why would we want to hide the error from the Go bindings?
I guess it depends on what you consider an error, is it an error for the function to try the match allocate and not do anything if the allocate is not possible? Or is that expected behavior and the results should be checked by some other means?

vsoch:
I don't think I know enough about flux-sched to address your questions, going to ping @milroy here. I can help with any changes to the Go that are warranted when it's decided how to handle it.

zekemorton:
@milroy this should be ready for a review! 

As discussed, I changed the go tests to set orelse_reserve to true and updated the expected output accordingly.

resource/utilities/rq2.cpp
```diff
@@ -0,0 +1,599 @@
...
+                }
+                break;
+            case 'g': /* --graph-format */
+                rc = string_to_graph_format (optarg, format);
+                if ( rc != 0) {
+                    std::cerr << "[ERROR] unknown format for --graph-format: ";
+                    std::cerr << optarg << std::endl;
+                    usage (1);
+                }
+                graph_format_to_ext (format, token);
```
jameshcorbett, line 496: Could this be simplified a bit? It looks like you take the string, turn it into a format, and then turn the format back into the same string. I don't see `graph_format_to_ext()` and `string_to_graph_format()` being used elsewhere.

jameshcorbett (dismissed):
LGTM generally, some small comments. And I'm sure you'll still want Dan to review.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -140,8 +127,24 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
...
+            m_err_msg += ": ERROR: can't allocate memory: " 
+                        + std::string (strerror (errno))+ "\n"; 
+            rc = -1; 
+            goto out; 
+        }
+
+    rq->set_job (jobid, job_info); 
+    }
+
+    rq->incr_job_counter ();
```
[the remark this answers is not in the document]
zekemorton: Yes, that is at least the behavior from resource-query, which this is trying to match. 
https://github.com/flux-framework/flux-sched/blob/0b70113bdf7eef5d501815d50155bdf124b14bae/resource/utilities/command.cpp#L153C5-L153C5

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -109,28 +110,14 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
```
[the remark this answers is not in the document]
zekemorton: There are still somethings that need to be run, but now that you mention it, I don't remember 100% if I already looked at it all. It's been a little while since I worked on this, so I will play around with it and see if it acts as expected.

resource/utilities/rq2.cpp
zekemorton: I believe this was part of the functionality for writing out graphs of different formats from resource query. I had started to fill in the functionality entirely, but decided to add that in later. Must not have removed it all yet. I can take this out and save it for later. Thanks for catching it!

resource/reapi/bindings/c++/reapi_cli_impl.hpp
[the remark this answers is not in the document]
zekemorton: Actually, looks like you're right. I think we just need `rq->incr_job_counter ();` which should probably be moved to the out section anyways. @milroy do you see anything that would cause any problems if we make this change?

resource/utilities/rq2.cpp
zekemorton: Ahh I remember now, this does convert it from string to format object and back to string. This was beneficial in original rq because it stored the format object along the way and this only required a single string comparison.  Since we don't have the functionality to use the format object, it doesn't really make a lot of sense to do it this way right now, but this does in its current form accomplish *something*. `boost::iequals` is case insensitive, so this does convert the string to lowercase. 

I am happy to change it to something that's a bit more clear or keep it as is, though we will want to store that format object at some point

resource/reapi/bindings/c++/reapi_cli_impl.hpp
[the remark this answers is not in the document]
jameshcorbett: Oh, thanks. I think it makes sense now. It looks like the job counter just keeps track of how many jobs have come through.

zekemorton:
After discussing the changes with @milroy, we found that changing the interface for reapi_clit_t::match_allocate has bigger implications and the return behavior should be left as is. A non 0 return code represents an error when running the code rather than a failure to find a match. 
rq2 now determines if a match was found by checking errno for ENODEV, EBUSY, EINVAL, all of which indicate a failure to match.

resource/reapi/bindings/c++/reapi_cli.hpp
```diff
@@ -141,6 +141,8 @@ class reapi_cli_t : public reapi_t {
     static int find (void *h, std::string criteria, json_t *&o );
```
milroy, line 143: Nit on commit message: "sets a shared pointed for" (pointer). More importantly, it's not really a setter. It's mutating a function out parameter.

resource/utilities/rq2.cpp
```diff
@@ -0,0 +1,631 @@
...
+            std::cout << "INFO:" << " ELAPSE=" << std::to_string (elapse)
+                      << std::endl;
+            std::cout << "INFO:" << " PREORDER VISIT COUNT=" << pre
+                      << std::endl;
+            std::cout << "INFO:" << " POSTORDER VISIT COUNT=" << post
+                      << std::endl;
+        }
+        out << "INFO:" << " =============================" << std::endl;
+        st = (at == 0)? job_lifecycle_t::ALLOCATED : job_lifecycle_t::RESERVED;
+        if (reapi_cli_t::info(&ctx, jobid, job) == 0)
```
milroy, line 273: Nit: spacing.

```suggestion
        if (reapi_cli_t::info (&ctx, jobid, job) == 0)
```

milroy (changes requested):
The PR is looking really good and just requires a few minor changes.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -99,6 +99,21 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
...
+        rc = -1;
+        goto out;
+    }
+
+    // Check for an unsuccessful match 
+    if (rc != 0) {
+        rq->incr_job_counter ();
+        rc = 0;
+        goto out;
+    }
```
milroy, line 116: An unsuccessful match should increment the counter, but I don't see why this code is needed. I also think it will break future tests. The flow is expected to continue execution until the final `return` line unless an error is encountered. See this case line in the current `resource-query`:  https://github.com/flux-framework/flux-sched/blob/0b70113bdf7eef5d501815d50155bdf124b14bae/resource/utilities/command.cpp#L309

If the execution reaches line 155 below then there were no errors so the counter will be incremented. I think that's the desired behavior.

milroy (changes requested):
I think the added check for the return code from `rq->traverser_run` will break the future implementation of `match_multi`.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
milroy: That's also a good point, but I think we can handle incorrectly setting the allocation and job with a straightforward check (tested). Let's schedule a quick chat.

zekemorton:
@milroy I added in the most recent changes requested, should be ready now!

milroy (approved):

milroy:
Thanks for the great work! The PR looks good and is ready to merge. Go ahead and set MWP.

Issue #994, opened 2022-12-09
remove data-staging shell plugin


garlick:
Problem: fluxion installs the data-staging plugin, but it requires post-v1 jobspec and at this point, is not part of the plan  to schedule coral2 rabbits wtih Fluxion.

At a minimum, the plugin should not be installed.

We should also consider if it and its sharness test ought to be removed from the source tree.  I am not sure if it is providing the only test coverage for fluxion's storage scheduling  capability or if that is covered elsewhere.

See also flux-framework/flux-core#3232

PR #1102, opened 2023-10-19
shell: remove data-staging shell plugin

grondo:
The data-staging plugin does not appear to have a real-world use-case and duplicates somewhat the existing `stage-in` shell plugin from flux-core. Additionally, the tests use the `--standalone` job shell option which will be removed. Therefore, this PR proposes just removing it.


garlick (approved):
Since there was no discussion in #994 I think that's an indication that there are no plans for this.  Anyway it can be easily brought back if needed.

grondo:
Thanks! I'll go ahead and set MWP then.

jameshcorbett (commented):
Yeah LGTM. Sorry I'm late!

Issue #1103, opened 2023-10-30
fluixion assigns jobs requesting unlimited duration an expiration time >  instance expiration


garlick:
Problem: when a job is submitted with unlimited duration, fluxion assigns it an expiration time that is (start time + instance duration), which does not account for the time elapsed between instance start and job start.  As a result, the job's expiration time is after the instance is no longer running.
```
$ flux alloc -t 30m -N 1
ƒ(s=1,d=1) $ flux kvs get resource.R | jq '.execution | .expiration - .starttime'
1800
ƒ(s=1,d=1) $ flux submit sleep inf
ƒZ5R5knb
ƒ(s=1,d=1) $ flux job info ƒZ5R5knb R | jq '.execution | .expiration - .starttime'
1800
# Wait a few minutes
ƒ(s=1,d=1) $ flux submit sleep inf
ƒ2oCAKVu9
ƒ(s=1,d=1) $ flux job info ƒ2oCAKVu9 R  | jq '.execution | .expiration - .starttime'
1800
```

grondo:
I think the problem is likely here:

https://github.com/flux-framework/flux-sched/blob/310cc2de216d8cee16e98bae8c816f203302a6d9/resource/traversers/dfu_impl.hpp#L73-L74

https://github.com/flux-framework/flux-sched/blob/310cc2de216d8cee16e98bae8c816f203302a6d9/resource/traversers/dfu_impl.hpp#L89-L92

If a jobspec does not have a duration set, then the _duration_ is set to the whole graph duration instead of the _remaining_ time.
A possible fix, though I don't know what I'm doing:

```diff
diff --git a/resource/traversers/dfu_impl.hpp b/resource/traversers/dfu_impl.hpp
index 23048c01..7ef64948 100644
--- a/resource/traversers/dfu_impl.hpp
+++ b/resource/traversers/dfu_impl.hpp
@@ -70,8 +70,9 @@ struct jobmeta_t {
         now = t;
         jobid = id;
         alloc_type = alloc;
+        const auto now = std::chrono::system_clock::now();
         int64_t g_duration = std::chrono::duration_cast<std::chrono::seconds>
-            (graph_duration.graph_end - graph_duration.graph_start).count ();
+            (graph_duration.graph_end - now).count ();
 
         if (g_duration <= 0) {
             errno = EINVAL;

```

PR #1104, opened 2023-11-01
traverser: fix default job duration when no duration specified

grondo:
When a resource graph has a fixed duration (i.e. a non-zero expiration), Fluxion currently assigns the graph _duration_ to job requests that do not specify a duration. This results in job expirations that exceed the instance lifetime.

This PR fixes that by adjusting unspecified jobspec durations to the _remaining_ graph duration, instead of the entire graph duration. This results in the expiration of these jobs matching that of the instance.

The existing test that checked this behavior was also updated and fixed.

### DIFF
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -290,7 +290,9 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
     }
 
     int rc = -1;
-    int64_t graph_end = graph_duration.graph_end.time_since_epoch ().count ();
+    int64_t graph_end = std::chrono::duration_cast<std::chrono::seconds>
+                            (graph_duration.graph_end
+                                .time_since_epoch ()).count ();
     detail::jobmeta_t meta;
     vtx_t root = get_graph_db ()->metadata.roots.at (dom);
     bool x = detail::dfu_impl_t::exclusivity (jobspec.resources, root);
--- a/t/t4011-match-duration.t
+++ b/t/t4011-match-duration.t
@@ -9,17 +9,58 @@ test_description='Test that parent duration is inherited according to RFC14'
 #
 test_under_flux 1
 
-test_expect_success HAVE_JQ 'parent duration is inherited when duration=0' '
+export FLUX_URI_RESOLVE_LOCAL=t
+
+# Ensure fluxion modules are loaded under flux-alloc(1)
+test_expect_success 'set FLUX_RC_EXTRA so Fluxion modules are loaded under flux-alloc' '
+	mkdir rc1.d &&
+	cat <<-EOF >rc1.d/rc1-fluxion &&
+	flux module unload -f sched-simple
+	flux module load sched-fluxion-resource
+	flux module load sched-fluxion-qmanager
+	EOF
+	mkdir rc3.d &&
+	cat <<-EOF >rc3.d/rc3-fluxion &&
+	flux module remove -f sched-fluxion-qmanager
+	flux module remove -f sched-fluxion-resource
+	flux module load sched-simple
+	EOF
+	chmod +x rc1.d/rc1-fluxion rc3.d/rc3-fluxion &&
+	export FLUX_RC_EXTRA=$(pwd)
+'
+test_expect_success 'load fluxion modules in parent instance' '
+	flux module remove sched-simple &&
+	load_resource &&
+	load_qmanager &&
+	test_debug "flux dmesg -H | grep version"
+'
+test_expect_success HAVE_JQ 'parent expiration is inherited when duration=0' '
 	cat >get_R.sh <<-EOT &&
 	#!/bin/sh
-
 	flux job info \$FLUX_JOB_ID R
 	EOT
 	chmod +x get_R.sh &&
-	out=$(flux run -t20s -n1 flux start flux run -n1 ./get_R.sh) &&
-	echo "$out" | jq -e ".execution.expiration - .execution.starttime <= 20" &&
-	out=$(flux run -t30s -n1 flux start flux run -n1 ./get_R.sh) &&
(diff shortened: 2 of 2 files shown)


resource/traversers/dfu_impl.hpp
```diff
@@ -70,8 +70,9 @@ struct jobmeta_t {
         now = t;
```
milroy, line 70: Nit on the commit message: this is more specifically the `traverser` (which is part of `resource`, but we've typically indicated this code scope in the commit messages. (So it would be `traverser: set default...`)

milroy (dismissed):
This is a good catch and the fix is elegant.

I'm requesting a few minor changes: a slight change to the `now` computation, a nit on the second commit message, and also that the commit order be swapped so the testsuite changes come after what they're testing.

grondo:
Thanks @milroy! I've made the suggested changes and pushed the result.

resource/traversers/dfu_impl.hpp
```diff
@@ -71,7 +71,8 @@ struct jobmeta_t {
         jobid = id;
         alloc_type = alloc;
         int64_t g_duration = std::chrono::duration_cast<std::chrono::seconds>
-            (graph_duration.graph_end - graph_duration.graph_start).count ();
+            (graph_duration.graph_end -
+             std::chrono::system_clock::now()).count ();
```
milroy, line 75: Sorry @grondo, but after more thought my previous suggestion results in a small drift, too. Can you do this instead:
```suggestion
            (graph_duration.graph_end - now);
 ```

Comment (grondo) [reply resource/traversers/dfu_impl.hpp:75]: Will do! However, for the uninitiated, can you describe the problem being fixed by this change? I only understand enough of `std::chrono` to make this one change, and with this subtle of a change it seems this might be fairly brittle. At least I could add a comment here for future code spelunkers.

Comment (milroy) [reply resource/traversers/dfu_impl.hpp:75]: The motivation for my request was to reduce the number of independent measurements of `now` throughout Fluxion which I hope makes drift debugging easier.

Your question is good and caused me to reconsider the `g_duration` computation and the downstream use. I think there's a bigger problem with the logic of aligning the job duration with the `graph_end` time. Issue #1103 reveals that the check here (which was designed to catch the problem) isn't working correctly: https://github.com/flux-framework/flux-sched/blob/bf62d8a74591353cf2aff43e9aeb41be81da8bef/resource/traversers/dfu.cpp#L323

Comment (grondo) [reply resource/traversers/dfu_impl.hpp:75]: I only just realized what you said initially: that I was redeclaring `now` in the original patch. Somehow I missed that there was already a `now` declared in the function. :facepalm: Sorry about that!

However, using the existing `int64_t` now results in a compilation error. I'm guessing some kind of cast needs to be used, but being pretty unfamilar with `std::chrono` I'm not sure what to use.

Since `g_duration` and `now` are already `int64_t` type, would it make sense to just cast `graph_duration.graph_end` to `int64_t` as well before taking the difference, e.g.

```diff
diff --git a/resource/traversers/dfu_impl.hpp b/resource/traversers/dfu_impl.hpp
index 098fcd41..f9fa8f9b 100644
--- a/resource/traversers/dfu_impl.hpp
+++ b/resource/traversers/dfu_impl.hpp
@@ -70,9 +70,9 @@ struct jobmeta_t {
         now = t;
         jobid = id;
         alloc_type = alloc;
-        int64_t g_duration = std::chrono::duration_cast<std::chrono::seconds>
-            (graph_duration.graph_end -
-             std::chrono::system_clock::now()).count ();
+        int64_t g_end = std::chrono::system_clock::to_time_t
+                          (graph_duration.graph_end);
+        int64_t g_duration = g_end - now;
 
         if (g_duration <= 0) {
             errno = EINVAL;

```

I really know nothing about `std::chrono` and was honestly having trouble casting `now` to the appropriate type otherwise. Please feel free to suggest anything you'd like (or even go ahead and  push the correct fix to this branch).

grondo: > Your question is good and caused me to reconsider the g_duration computation and the downstream use. I think there's a bigger problem with the logic of aligning the job duration with the graph_end time. Issue https://github.com/flux-framework/flux-sched/issues/1103 reveals that the check here (which was designed to catch the problem) isn't working correctly:

Ah, yes, I see now that the "fix" attempted in this PR won't really fix anything. If the start of the job is delayed, then its expiration may still be greater than the instance expiration because real time has passed between when `build()` is called and the job's actual starttime.

I think the real fix is in the code you've called out above, or to have a way to just set the job _expiration_ to the instance expiration, regardless of the job's duration (though I don't know the code well enough to determine how to do that).

Also, unless I'm mistaken, it seems like setting the job's duration to the instance duration in `build()`  isn't doing anything, and that code should be removed to avoid confusion.

Ideally, we'd like to get this and the follow on PR #1105 fixed before the tag tomorrow. Maybe I'll look into why the duration adjustment in `run()` isn't working (as best I can), but could use help if you're available.

grondo: Well, when I print some debug from that line, `graph_end` seems to have the wrong scale:
```
at=1699291701 meta.duration=120 graph_end=1699291819000000000
```
Does `graph_end` need one of those crazy `std::chrono::duration_cast<std::chrono::seconds>`?

grondo:
I also noticed that the added test here does not work because `flux alloc` and `flux batch` in the flux-sched testsuite do not load Fluxion module by default. (Other tests that use `flux batch` load fluxion modules manually).

So, I'll be force-pushing this branch with a completely different fix and the testsuite improved.

garlick (approved):
I'll add my approval FWIW since we're on a tight timeline for a tag today, we need this fixed, and there's a regression test.  Perhaps if @milroy thinks follow up is needed can do that in another PR?

grondo:
Thanks! I'll set MWP.

PR #1105, opened 2023-11-01
resource: support resource expiration updates

grondo:
This PR adds support for the `expiration` key in the [RFC 28](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_28.html#additional-acquire-responses) "Additional acquire responses".

Presence of this key indicates that the expiration of the instance resource set has been updated.

This is built on top of #1104 since this PR adds to the sharness tests which is modified there.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1236,13 +1236,15 @@ static void update_resource (flux_future_t *f, void *arg)
     int rc = -1;
     const char *up = NULL;
     const char *down = NULL;
+    double expiration = -1.;
     json_t *resources = NULL;
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
-    if ( (rc = flux_rpc_get_unpack (f, "{s?:o s?:s s?:s}",
+    if ( (rc = flux_rpc_get_unpack (f, "{s?:o s?:s s?:s s?:F}",
                                            "resources", &resources,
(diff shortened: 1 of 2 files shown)


grondo:
Oh, and obviously the new tests here won't pass CI until flux-framework/flux-core#5522 is merged. Actually, perhaps I should make this test conditional on support for running job updates of duration.

grondo:
Ok, that test now checks if running job updates are allowed and the test is skipped if not.

grondo:
Rebased on updated testing in #1103

resource/modules/resource_match.cpp
```diff
@@ -1236,13 +1236,15 @@ static void update_resource (flux_future_t *f, void *arg)
     int rc = -1;
```
garlick, line 1236: Typo in commit message s/acquirej/acquire/

garlick (approved):
Not spotting any problems here except a trivial typo in a commit message.  LGTM

grondo:
Thanks! I pushed a fix for the commit message typo and will set MWP.

grondo:
Hm, the new test is hanging at shutdown occasionally in CI (i.e. it appears all tests have completed, but the test never exits). I'll see if I can reproduce this locally.

grondo:
Ok, the hang was due to no scheduler loaded at the end of the test. I've fixed that and forced a push.

PR #1106, opened 2023-11-02
ci: fix coverage build

grondo:
I noticed that the `coverage` build is always reporting `Warning: make check failed, trying recheck in ./t`. Apparently, this isn't because `make check-code-coverage` is failing, but because one of the subsequent `coverage blah` steps is.

Possibly some change in the cmake `check-code-coverage` target is causing one of these steps to fail, or maybe it was always like this for flux-sched. Anyway, this PR just removes some of the post-check coverage steps and makes a couple others optional to avoid the extra unnecessary work of `make recheck`.

### DIFF
--- a/src/test/checks_run.sh
+++ b/src/test/checks_run.sh
@@ -118,14 +118,14 @@ if test "$COVERAGE" = "t"; then
 
 	ARGS="$ARGS --enable-code-coverage"
 	CHECKCMDS="\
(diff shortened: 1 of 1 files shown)


trws (approved):
Looks good to me, good catch @grondo!

grondo:
Thanks! I'll set MWP.

PR #1109, opened 2023-11-14
testsuite: fix a couple race conditions in `t4011-match-duration.t`

grondo:
This PR fixes a couple race conditions in `t/t4011-match-duration.t` which cause the test to fail occasionally.


jameshcorbett (approved):
LGTM, thanks @grondo !

grondo:
Thanks @jameshcorbett! I had to restart the spelling and pr-validator checks because they failed for unknown reason. I'll set MWP now.

PR #1110, opened 2023-12-02
Add a set-status RPC for marking vertices up or down

jameshcorbett:
The motivation for this PR comes from [this flux-coral2 issue](https://github.com/flux-framework/flux-coral2/issues/99). Rabbits can (and very often do) go down, and fluxion needs to avoid scheduling rabbit jobs on dead rabbits. Rabbit status is monitored by HPE software and reported in kubernetes objects, which flux-coral2 software checks. The flux-coral2 software needs a way to report to Fluxion that a rabbit has died (or resurrected).

This PR provides a way to do that. Flux-coral2 can now send a RPC like `sched-fluxion-resource.set_status /path/to/rabbit down` to mark a rabbit as down, and similarly to mark it up.

However, I did notice that marking all the nodes or racks in the resource graph didn't prevent jobs from running, hence the test failures. @milroy do you know whether this is expected? Is marking a resource as 'down' purely informational--does it have no effect on actual resource choices?

As an aside, one tricky aspect is that with this PR, flux-coral2 will need to know the resource graph layout. It will know the hostname of the rabbit that has gone down, but it won't necessarily know that the rabbit 'rzvernal201' is '/rzvernal/rack3/rzvernal201` in the resource graph. I think I can work around this though--the resource graph is already written out to a file (because it needs to contain information about rabbit locality) so the flux-coral2 software can read in that graph and figure things out.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -255,6 +255,9 @@ static void satisfiability_request_cb (flux_t *h, flux_msg_handler_t *w,
 static void params_request_cb (flux_t *h, flux_msg_handler_t *w,
                                const flux_msg_t *msg, void *arg);
 
+static void set_status_request_cb (flux_t *h, flux_msg_handler_t *w,
+                                   const flux_msg_t *msg, void *arg);
+
 static const struct flux_msg_handler_spec htab[] = {
     { FLUX_MSGTYPE_REQUEST,
       "sched-fluxion-resource.match", match_request_cb, 0 },
@@ -288,6 +291,8 @@ static const struct flux_msg_handler_spec htab[] = {
       "sched-fluxion-resource.satisfiability", satisfiability_request_cb, 0 },
     { FLUX_MSGTYPE_REQUEST,
       "sched-fluxion-resource.params", params_request_cb, 0 },
+    { FLUX_MSGTYPE_REQUEST,
+      "sched-fluxion-resource.set_status", set_status_request_cb, 0 },
     FLUX_MSGHANDLER_TABLE_END
 };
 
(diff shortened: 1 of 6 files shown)


jameshcorbett:
@grondo can you describe what the core-sched interactions are when a node goes down?

jameshcorbett:
Thanks @grondo , that actually gave me the information I needed to fix the tests! I think this PR is a pretty harmless addition now.

zekemorton (approved):
I looked through all of the commits, and the code looks clean and all of the changes make sense! I couldn't find anything to complain about. 

I think it's still a good idea for @milroy to give it a look over

milroy:
I'll add my review today.

resource/modules/resource_match.cpp
```diff
@@ -2633,6 +2638,58 @@ static void params_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+            flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
+        return;
+    }
+    // mark the vertex
+    if (ctx->traverser->mark (resource_path, status_it->second) < 0) {
+        flux_log_error (h, "%s: traverser::mark: %s", __FUNCTION__,
+                                ctx->traverser->err_message ().c_str ());
+        ctx->traverser->clear_err_message ();
+        errmsg = "Failed to set status of resource vertex";
+        if (flux_respond_error (h, msg, EINVAL, errmsg.c_str()) < 0)
```
milroy, line 2684: Nit: spacing.
```suggestion
        if (flux_respond_error (h, msg, EINVAL, errmsg.c_str ()) < 0)
```

resource/modules/resource_match.cpp
```diff
@@ -2633,6 +2638,58 @@ static void params_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+        errmsg = "could not find path '" + resource_path + "' in resource graph";
+        if (flux_respond_error (h, msg, EINVAL, errmsg.c_str()) < 0)
+            flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
+        return;
+    }
+    // check that the status given is valid ('up' or 'down')
+    auto status_it = sts.find (status);
+    if (status_it == sts.end ()) {
+        errmsg = "unrecognized status '" + status + "'";
+        if (flux_respond_error (h, msg, EINVAL, errmsg.c_str()) < 0)
```
milroy, line 2674: Nit: spacing.
```suggestion
        if (flux_respond_error (h, msg, EINVAL, errmsg.c_str ()) < 0)
```

resource/modules/resource_match.cpp
```diff
@@ -2633,6 +2638,58 @@ static void params_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+    std::string resource_path = "", status = "", errmsg = "";
+    std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
+    resource_pool_t::string_to_status sts = resource_pool_t::str_to_status;
+
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s}",
+                                        "resource_path", &rp,
+                                        "status", &st) < 0){
+        if (flux_respond_error (h, msg, EINVAL, "malformed RPC") < 0)
+            flux_log_error (h, "%s: flux_respond_error", __FUNCTION__);
+        return;
```
milroy, line 2657: You could save the repetition of the `flux_respond_error` and `flux_log_error` by setting `errmsg` and using `goto error` like this block: https://github.com/flux-framework/flux-sched/blob/0b70113bdf7eef5d501815d50155bdf124b14bae/resource/modules/resource_match.cpp#L1863

```suggestion
        errmsg = "malformed RPC";
        goto error;
```
That would apply to the following blocks, too.

milroy (changes requested):
This PR is almost ready to merge. I'm requesting a quick change to remove duplicate code and it will be ready to merge.

resource/modules/resource_match.cpp
jameshcorbett: I originally wrote this function that way, but the compiler complained that I was jumping over the initialization of some variables, and so it prevented me from using the `auto` type. Do you have any recommendations to avoid that?

milroy: I've run into that a lot, too. You'll need to declare the variable before the `if` blocks. Sometimes that isn't feasible; if that's true in this case then we can merge this PR.

jameshcorbett: Done! And your spacing comments addressed.

milroy (approved):

milroy:
LGTM! Approved.

Once you rebase against the current `master` you can set MWP.

jameshcorbett:
Done, setting MWP. Thanks @milroy !

PR #1120, opened 2023-12-16
Build reapi_cli as a shared library

milroy:
PR #1094 Updated Fluxion to be build as a static library. Some projects like Fluence are clients of Fluxion and its REAPI and need to link Fluxion libraries as shared libraries to preserve license boundaries. This PR adds the capability to build the necessary shared libraries.

### DIFF
--- a/resource/reapi/bindings/go/src/test/CMakeLists.txt
+++ b/resource/reapi/bindings/go/src/test/CMakeLists.txt
@@ -3,15 +3,24 @@ set(SRCS main.go)
 set(CGO_CFLAGS "-I${CMAKE_BINARY_DIR}/resource/reapi/bindings/c")
 
 # This is currently passed but not used because when passed into add_custom_command the spaces are escaped
-set(CGO_LIBRARY_FLAGS "-Wl,-R ${CMAKE_BINARY_DIR}/resource -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -L${CMAKE_BINARY_DIR}/resource/libjobspec -ljobspec_conv -lreapi_cli -L${CMAKE_BINARY_DIR}/resource -lresource -L${CMAKE_BINARY_DIR}/resource/planner/c -lplanner -L${CMAKE_BINARY_DIR}/resource/planner/c++ -lplanner_cxx -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp")
+set(CGO_LIBRARY_FLAGS -Wl,-R
+    -lflux-idset
+    -lstdc++
+    -lczmq
+    -ljansson
+    -lhwloc
+    -lboost_system
+    -lflux-hostlist
+    -lboost_graph
+    -lyaml-cpp
+    -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -lreapi_cli
+    )
 
 # This ensures the main binary is cleaned up
 set_directory_properties(
     PROPERTIES
-      ADDITIONAL_CLEAN_FILES "${CMAKE_CURRENT_BINARY_DIR}/main"
+    ADDITIONAL_CLEAN_FILES "${CMAKE_CURRENT_BINARY_DIR}/main"
   )
 
-# GO_GET(go_redis github.com/hoisie/redis)
-#                          main      main.go  
 # We add the dependencies at the end so the build can run in parallel and we don't try to build this before they are ready!
-BUILD_GO_PROGRAM(${TARGET} ${SRCS} "${CGO_CFLAGS}" "${CGO_LIBRARY_FLAGS}" jobspec_conv reapi_module reapi_cli resource flux::core)
+BUILD_GO_PROGRAM(${TARGET} ${SRCS} "${CGO_CFLAGS}" "${CGO_LIBRARY_FLAGS}" reapi_cli)
--- a/cmake/GolangSimple.cmake
+++ b/cmake/GolangSimple.cmake
@@ -11,11 +11,8 @@ function(BUILD_GO_PROGRAM NAME MAIN_SRC CGO_CFLAGS CGO_LIBRARY_FLAGS)
   get_filename_component(MAIN_SRC_ABS ${MAIN_SRC} ABSOLUTE)
   add_custom_target(${NAME})
 
-  # IMPORTANT: the trick to getting *spaces* to render in COMMAND is to convert them to ";"
-  # string(REPLACE <match-string> <replace-string> <out-var> <input>...)
-  STRING(REPLACE " " ";" CGO_LIBRARY_FLAGS ${CGO_LIBRARY_FLAGS})  
   add_custom_command(TARGET ${NAME}
-                    COMMAND GOPATH=${GOPATH}:${CUSTOM_GO_PATH} GOOS=linux G0111MODULE=off CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS='${CGO_LIBRARY_FLAGS}' go build -ldflags '-w'
+                    COMMAND GOPATH=${GOPATH}:${CUSTOM_GO_PATH} GOOS=linux G0111MODULE=off CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LIBRARY_FLAGS}" go build -ldflags '-w'
                     -o "${CMAKE_CURRENT_SOURCE_DIR}/${NAME}"
                     ${CMAKE_GO_FLAGS} ${MAIN_SRC}
                     WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
--- a/resource/reapi/bindings/CMakeLists.txt
+++ b/resource/reapi/bindings/CMakeLists.txt
@@ -1,21 +1,25 @@
-add_library ( reapi_cli STATIC
+add_library ( reapi_cli SHARED
     c++/reapi.hpp
     c++/reapi_cli.hpp
     c++/reapi_cli_impl.hpp
     c/reapi_cli.cpp
     c/reapi_cli.h
     )
-target_link_libraries(reapi_cli PRIVATE
+target_link_libraries( reapi_cli PRIVATE
+    resource
+    jobspec_conv
     flux::core
     )
-add_library ( reapi_module STATIC
+install(TARGETS reapi_cli LIBRARY DESTINATION "${FLUX_LIB_DIR}")
+
+add_library (reapi_module STATIC
     c++/reapi.hpp
     c++/reapi_module.hpp
     c++/reapi_module_impl.hpp
     c/reapi_module.cpp
     c/reapi_module.h
     )
-target_link_libraries(reapi_module PRIVATE
+target_link_libraries( reapi_module PRIVATE
     flux::core
     )
 
--- a/t/t9001-golang-basic.t
+++ b/t/t9001-golang-basic.t
@@ -14,6 +14,8 @@ if ! which go >/dev/null; then
     test_done
 fi
 
+export LD_LIBRARY_PATH=../../resource/reapi/bindings
+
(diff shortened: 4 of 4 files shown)


vsoch:
@milroy this is fantastic! :raised_hands: For a quick example of what this enables, I was (very easily) able to build a go module out of tree:

https://github.com/converged-computing/flex-ice-cream

And specifically, the compile command from the root:

```bash
GOOS=linux CGO_CFLAGS="-I/opt/flux-sched/resource/reapi/bindings/c" \
CGO_LDFLAGS="-L/usr/lib -L/opt/flux-sched/resource -lfluxion-resource \
    -L/opt/flux-sched/resource/libjobspec -ljobspec_conv -L//opt/flux-sched/resource/reapi/bindings \
    -lreapi_cli -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist \
    -lboost_graph -lyaml-cpp" go build -ldflags '-w' -o bin/server src/fluence/cmd/main.go
```
For go libraries, it is much nicer (and expected by developers) to have go.mod and go.sum in the root of the repository. I tested the above and found I could use the (new?) `libfluxion-resource.so` and I still needed to link to JobSpec and reapi cli binding libraries for the compile to work.

This dummy example builds fluence, and I'll quickly be changing that to something that is not fluence (and likely deleting the git history in case there is issue having it in two spots). I just fluence because it was the quickest way to prototype/test the bindings and give feedback here, and could be compared against the "fudged version" from https://github.com/flux-framework/flux-k8s/pull/47. When work is merged here, we will want to use these bindings there (and good to sanity check it will build). 

For now I'm using a fork of the work here, and only because I need to change the module name to match the repository in go.mod (go does not allow you to use, for example, a go.mod with `flux-framework/flux-sched` that is cloned from `milroy/flux-sched`) 

@trws and @grondo I hope you are excited about this, because it means we (and other developers) can easily build cool stuff with fluxion! I am going to make an attempt soon to prototype what I wanted to do with ice cream, and from that we will likely identify some of the missing pieces (go or c bindings here) that are needed. We've already anticipated some of these needs (and I opened issues earlier this week). Thank you in advance for the reviews!

vsoch:
Update: library is now officially ice-cream-ized: 
image

https://github.com/converged-computing/flex-ice-cream

It's not perfect, but it's a start! I made sure to nuke the fluence history so there wouldn't be worry about having it there. Going back to sleep, thanks again for the work here and TBA review!

trws:
Glad it works, it feels a little odd to need to make the cli a shared library? Is there something in there we should move to something meant to be an external interface?

vsoch:
Here is what we needed in fluence: https://github.com/flux-framework/flux-k8s/blob/105562e4662e86a1b8ce1e19762678a6c0dd6309/src/fluence/fluxion/fluxion.go

Primarily to make a reapi client and then interact with it!

```go
f.cli = fluxcli.NewReapiClient()
```

milroy:
Dropping WIP.

I drastically simplified this PR and reduced the API surface by linking the static Fluxion `resource` in `reapi_cli` and building `reapi_cli` as a a shared library. @vsoch let me know if installing `libreapi_cli.so` works.

vsoch:
The build works - this is a development container that installs flux-sched from your branch (with the go path tweaked) and then just runs make:

image

The compile command:

```console
GOOS=linux CGO_CFLAGS="-I/opt/flux-sched/resource/reapi/bindings/c" CGO_LDFLAGS="-L/usr/lib -L/opt/flux-sched/resource/libjobspec -ljobspec_conv -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp -L//opt/flux-sched/resource/reapi/bindings -lreapi_cli" go build -ldflags '-w' -o bin/icecream src/cmd/main.go
```

And can confirm the .so is installed.

```console
find /usr -name libreapi_cli.so
/usr/lib/flux/libreapi_cli.so
```

The flux .so is installed to a non-traditional place, so not an issue to add that to `LD_LIBRARY_PATH`:

```console
./bin/icecream 
./bin/icecream: error while loading shared libraries: libreapi_cli.so: cannot open shared object file: No such file or directory
export LD_LIBRARY_PATH=/usr/lib/flux
```

But I think we are still missing one library:
```console
./bin/icecream 
./bin/icecream: error while loading shared libraries: libjobspec_conv.so: cannot open shared object file: No such file or directory
```
Which seems to still be in the install root:

```console
find /opt/flux-sched/ -name libjobspec_conv.so
/opt/flux-sched/resource/libjobspec/libjobspec_conv.so
find /usr -name libjobspec_conv.so
# no output
```

milroy:
> Which seems to still be in the install root:

Ok, I updated the PR to install `libjobspec_conv.so` to `FLUX_LIB_DIR`. I think we need to install the Flux and Fluxion libraries to `FLUX_LIB_DIR`, but @trws might have an opinion on whether we could provide an override flag.

milroy:
@vsoch I did some additional simplification and cleanup that further reduces the complexity of linking Fluxion. Can you test again and make sure this works?

resource/reapi/bindings/go/src/test/CMakeLists.txt
```diff
@@ -3,15 +3,23 @@ set(SRCS main.go)
...
+    -lflux-idset \
+    -lstdc++ \
+    -lczmq \
+    -ljansson \
+    -lhwloc \
+    -lboost_system \
+    -lflux-hostlist \
+    -lboost_graph \
+    -lyaml-cpp \
+    -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -lreapi_cli")
```
trws, line 16: This is stylistic, but do you strongly prefer having this as a quoted string rather than unquoted here?  The extra noise of having to have a `\` on the end of each line is rather unfortunate.

trws:
Pending the style question and the commit validation getting fixed up I'm good here, and will approve.

resource/reapi/bindings/go/src/test/CMakeLists.txt
vsoch: I think it looks nice! It's much easier to read than the previous lump.

grondo:
> I think we need to install the Flux and Fluxion libraries to FLUX_LIB_DIR, but @trws might have an opinion on whether we could provide an override flag.

Just an opinion, but yes, installing Flux libraries with generic names like `libjobspec_conv.so` and `libreapi_cli.so` is probably better done under a subdirectory. If you do want to install these in a standard path, you might consider adding a prefix like we do for `libflux-*` libraries (`libflux-core.so`, `libflux-idset.so`, etc.), e.g. `libfluxion-jobspec_conv.so` and `libfluxion-reapi_cli.so`.

vsoch:
> Just an opinion, but yes, installing Flux libraries with generic names like libjobspec_conv.so and libreapi_cli.so is probably better done under a subdirectory. If you do want to install these in a standard path, you might consider adding a prefix like we do for libflux-* libraries (libflux-core.so, libflux-idset.so, etc.), e.g. libfluxion-jobspec_conv.so and libfluxion-reapi_cli.so.

Oh just to be clear, I'm okay with it being under flux, and think that's the better design.

resource/reapi/bindings/go/src/test/CMakeLists.txt
trws: I agree, but I think we're hitting a mismatch of expectations on cmake syntax.  This following `set` command should build the same string as the one currently in the PR, and is slightly more idiomatic:

```suggestion
set(CGO_LIBRARY_FLAGS -Wl,-R
    -lflux-idset
    -lstdc++
    -lczmq
    -ljansson
    -lhwloc
    -lboost_system
    -lflux-hostlist
    -lboost_graph
    -lyaml-cpp
    -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -lreapi_cli)
```

trws:
> > I think we need to install the Flux and Fluxion libraries to FLUX_LIB_DIR, but @trws might have an opinion on whether we could provide an override flag.
> 
> Just an opinion, but yes, installing Flux libraries with generic names like `libjobspec_conv.so` and `libreapi_cli.so` is probably better done under a subdirectory. If you do want to install these in a standard path, you might consider adding a prefix like we do for `libflux-*` libraries (`libflux-core.so`, `libflux-idset.so`, etc.), e.g. `libfluxion-jobspec_conv.so` and `libfluxion-reapi_cli.so`.

Agreed.

milroy:
> Just an opinion, but yes, installing Flux libraries with generic names like libjobspec_conv.so and libreapi_cli.so is probably better done under a subdirectory. If you do want to install these in a standard path, you might consider adding a prefix like we do for libflux-* libraries (libflux-core.so, libflux-idset.so, etc.), e.g. libfluxion-jobspec_conv.so and libfluxion-reapi_cli.so.

> Oh just to be clear, I'm okay with it being under flux, and think that's the better design.

@vsoch do you mean that you're okay with installation of `libreapi_cli.so` being under `FLUX_LIB_DIR`? 

@grondo regardless of the installation path do you recommend that I add a prefix (`libfluxion-reapi_cli.so`)?

resource/reapi/bindings/go/src/test/CMakeLists.txt
milroy: @trws for some reason neither your comment nor @vsoch displayed when I was responding. I tried the more cmake-idiomatic `set` command you suggested as well as a few others I devised, and couldn't get them to configure or link.

vsoch:
> @vsoch do you mean that you're okay with installation of libreapi_cli.so being under FLUX_LIB_DIR?

Yes.

resource/reapi/bindings/go/src/test/CMakeLists.txt
trws: Ah, I made a mistake in the above example.  That actually gets stored as `"-Wl,-R;-flux-idset;-lstdc++"` etc., as a list, so it needs to be _unquoted_ on the usage to separate the list into args in the usage.

Alternately, this is a cmake-style multiline-string, it will include newlines in the string, but usually that's not really an issue as long as it isn't spliced into a string or similar.
```suggestion
set(CGO_LIBRARY_FLAGS [[ -Wl,-R
    -lflux-idset
    -lstdc++
    -lczmq
    -ljansson
    -lhwloc
    -lboost_system
    -lflux-hostlist
    -lboost_graph
    -lyaml-cpp
    -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -lreapi_cli
    ]])
```

milroy: Unfortunately, the bracket argument method results in 
```bash
resource/reapi/bindings/go/src/test/CMakeFiles/main.dir/build.make:72: *** missing separator.  Stop.
```
The other issue is that bracket arguments don't support variable references (needed for `CMAKE_BINARY_DIR`): https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#bracket-argument

trws: Fair enough. If you don't want to do the list expansion on the usage then leave as is. 👍

milroy: Ah, I got it! The list expansion style does work, but GolangSimple.cmake also needs to be changed to get rid of the replacement of spaces with `;`.

vsoch: oops, that is definitely my bad :grimacing: :laughing:

milroy:
Thanks for the feedback and helpful cmake tips, @trws @grondo and @vsoch!

This PR is yet simpler and I thikn addresses all your comments. I've re-requested a final review.

vsoch (approved):

trws (approved):
Looks great to me, good stuff. 👍

vsoch:
image

milroy:
Thanks for the reviews!

Setting MWP.

PR #1123, opened 2023-12-20
Resource pool status

jameshcorbett:
Problem: the FluxionResourcePoolV1 does not support setting
vertex status. However, some resources, such as rabbits, should
be initialized as down, and are only known to Fluxion through
JGF produced by the FluxResourcePool class.

Add a 'status' parameter to the FluxionResourcePoolV1 constructor,
with a default of 'up' to maintain backwards compatibility.

It sounds like there's a chance we'll change the JGF reader's default to 'down' in which case we'd want to change the behavior of the ResourceGraph class as well, I think.

### DIFF
--- a/src/python/fluxion/resourcegraph/V1.py
+++ b/src/python/fluxion/resourcegraph/V1.py
@@ -35,6 +35,7 @@ class FluxionResourcePoolV1(Node):
         size,
         properties,
         path,
+        status=0,
     ):
         """Constructor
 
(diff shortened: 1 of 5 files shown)


milroy (changes requested):
This PR is looking good overall, but the test isn't currently being run.

You'll need to add the test to CMakeLists here: https://github.com/flux-framework/flux-sched/blob/d51c81f398240ae4765975ab88365c7a86e8ccae/t/CMakeLists.txt#L88

and then make a simple sharness test (i.e. `t10001-resourcegraph.t`) that runs `t10001-resourcegraph.py` like this: https://github.com/flux-framework/flux-sched/blob/d51c81f398240ae4765975ab88365c7a86e8ccae/t/t3301-system-latestart.t#L31

milroy:
> It sounds like there's a chance we'll change the JGF reader's default to 'down' in which case we'd want to change the behavior of the ResourceGraph class as well, I think.

I think we'll end up leaving the JGF reader's default to `up` to avoid having to change a lot of the current sharness tests.

jameshcorbett:
> You'll need to add the test to CMakeLists here:

Done and pushed!

> and then make a simple sharness test (i.e. `t10001-resourcegraph.t`) that runs `t10001-resourcegraph.py` like this:

How come? The python test outputs TAP already, since I copied the pycotap package over from flux-core.

milroy:
> How come? The python test outputs TAP already, since I copied the pycotap package over from flux-core.

You're right. I messed up the `PATH` in my testing which confused me about how sharness was executing the test.

milroy (approved):
LGTM! This is ready to be merged.

jameshcorbett:
Thanks! Setting MWP

PR #1127, opened 2023-12-22
flux-ion-resource: move to src/cmd and install

### LINKED ISSUE #1122: install and package `flux ion-resource`
While debugging and working around #1121 on a production cluster, the `flux ion-resource find` and `set-status` commands were very useful, but this command is not installed with Fluxion, so it had to be used from a build directory.

We should consider installing and packaging this command so it is available for easy future use.

trws:
problem: recent production debugging and fixes have benefitted from having flux ion-resource available, but it is not installed.

solution: install it, and since it's no longer a test-only script, move it to src/cmd/

fixes #1122


trws:
Ok, this should be good to go now I hope.  It's notable that the first commit, the one that's a move and one line, is all it takes to install the script.  The second commit is required for `flux-ion-resource` to pass pylint, it was in pretty tough shape and this was the minimum to make it pass (almost, bit of extra cleanup too).

grondo (approved):
LGTM! I verified that the `make deb` target works and now includes `flux ion-resource` (and that the script is functional)

Thanks!

Issue #1129, opened 2023-12-27
match failed: Value too large for defined data type


grondo:
In preparation for some possible scaling tests, I was running a test Flux instance with 48 brokers per node across 64 nodes on corona. When trying to run a job with one task per rank, Fluxion failed with the following error:

```
$ flux run -N3072 -n3072 hostname
Dec 27 08:59:43.540844 sched-fluxion-resource.err[0]: match_multi_request_cb: match failed due to match error (id=4541860806656): Value too large for defined data type
0.112s: job.exception type=alloc severity=0 alloc denied due to type="match error"
```

FYI - 32 nodes with 48 brokers per node does work.

grondo:
Here's a standalone reproducer. The problem seems to require a larger number of cores per node. That is, I was able to run this test with 48 cores per node up to 10000 nodes. However, with 112 cores per node, the max `TEST_SIZE` that can be used here is 813:
```sh
#!/bin/bash

TEST_SIZE=${TEST_SIZE:-2000}

flux module remove -f sched-fluxion-qmanager
flux module remove -f sched-fluxion-resource

flux module remove resource
flux config load <<EOF
[sched-fluxion-qmanager]
queue-policy = "easy"
#[sched-fluxion-resource]
#match-format = "$MATCH_FORMAT"

[resource]
noverify = true
norestrict = true

[[resource.config]]
hosts = "test[1-${TEST_SIZE}]"
cores = "0-112"
gpus = "0-8"
EOF

flux module load resource monitor-force-up
flux module load sched-fluxion-resource
flux module load sched-fluxion-qmanager
flux queue start --all --quiet
flux resource list
flux resource status

flux run -vvv -N${TEST_SIZE} -n${TEST_SIZE} \
	--setattr=exec.test.run_duration=1ms \
	true

flux jobs -a
```

```console
$ TEST_SIZE=813 flux start ./test.sh 
     STATE NNODES   NCORES    NGPUS NODELIST
      free    813    91869     7317 test[1-813]
 allocated      0        0        0 
      down      0        0        0 
     STATE UP NNODES NODELIST
     avail  ✔    813 test[1-813]
jobid: f3m7fSF1
0.000s: job.submit {"userid":6885,"urgency":16,"flags":0,"version":1}
0.013s: job.validate
0.024s: job.depend
0.024s: job.priority {"priority":16}
0.095s: job.alloc
0.095s: job.prolog-start {"description":"cray-pals-port-distributor"}
[...]

PR #1132, opened 2024-01-08
Fix overflow in `calc_factor`

milroy:
In issue #1129 @grondo reported `EOVERFLOW` errors for match requests for thousands of CPUs and GPUs. 

Currently, `calc_factor` computes the tie breaking factor with modular arithmetic. The computation returns -1 when `break_tie` is divisible by `m_multiply_by` (or in other circumstances in modular arithmetic).

The negative value of `tie` causes `std::numeric_limits<int64_t>::max () - tie` to overflow in the following integer check, generating a spurious `EOVERFLOW` errno: https://github.com/flux-framework/flux-sched/blob/3d953e34c8da698276c1ad8e17c14173c36580c5/resource/policies/dfu_match_multilevel_id_impl.hpp#L60

This PR fixes the computation to ensure `tie` is strictly positive by using `abs()`.

### DIFF
--- a/resource/policies/dfu_match_multilevel_id_impl.hpp
+++ b/resource/policies/dfu_match_multilevel_id_impl.hpp
@@ -55,7 +55,7 @@ int64_t multilevel_id_t<FOLD>::score_factor_t::calc_factor (
         return -1;
     }
     mul = add * m_multiply_by;
-    tie =  break_tie % m_multiply_by - 1;
+    tie = abs (break_tie % static_cast<int64_t> (m_multiply_by) - 1);
 
     if (mul > (std::numeric_limits<int64_t>::max () - tie)) {
         errno = EOVERFLOW;
--- a/t/CMakeLists.txt
+++ b/t/CMakeLists.txt
@@ -80,6 +80,7 @@ set(ALL_TESTS
   t4011-match-duration.t
   t4012-set-status.t
   t5000-valgrind.t
+  t5100-issues-test-driver.t
   t6000-graph-size.t
   t6001-match-formats.t
   t6002-graph-hwloc.t
--- a/t/issues/t1035-fluxion-reload.sh
+++ b/t/issues/t1035-fluxion-reload.sh
@@ -16,8 +16,7 @@ fi
 test $(flux resource list -no {nnodes}) -eq 4 || die "test requires 4 nodes"
 
 log "Unloading modules..."
-flux module remove sched-fluxion-qmanager
-flux module remove sched-fluxion-resource
+flux module remove sched-simple
 flux module remove resource
 
 log "Amending instance resource set with properties: batch, debug..."
--- a/t/issues/t1129-match-overflow.sh
+++ b/t/issues/t1129-match-overflow.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+#  Ensure fluxion calc_factor executes without overflow
+#
+
+log() { printf "issue#1129: $@\n" >&2; }
+
+TEST_SIZE=${TEST_SIZE:-900}
+
+log "Unloading modules..."
+flux module remove sched-simple
+flux module remove resource
+
+flux config load <<EOF
+[sched-fluxion-qmanager]
+queue-policy = "easy"
+
+[resource]
+noverify = true
+norestrict = true
+
+[[resource.config]]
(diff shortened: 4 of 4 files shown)


grondo (commented):
This LGTM! Nice debugging @milroy!

I'd perhaps suggest addition of a reproducer to the testsuite. If, based on the fix, you have an easy reproducer that can be integrated with another test, that might be best. If not, then the reproducer script could be adapted and placed into e.g. `t/issues/t1129-match-overflow.sh`, e.g. something like:
```sh
#!/bin/bash -e

if test -z "$T1129_RELAUNCHED"; then
	export T1129_RELAUNCHED=t
	exec flux start $0
fi

TEST_SIZE=${TEST_SIZE:-900}

flux module remove -f sched-fluxion-qmanager
flux module remove -f sched-fluxion-resource

flux module remove resource
flux config load <<EOF
[sched-fluxion-qmanager]
queue-policy = "easy"
#[sched-fluxion-resource]
#match-format = "$MATCH_FORMAT"

[resource]
noverify = true
norestrict = true

[[resource.config]]
hosts = "test[1-${TEST_SIZE}]"
cores = "0-112"
gpus = "0-8"
EOF

flux module load resource monitor-force-up
flux queue start --all --quiet
flux resource list
flux resource status

flux run -vvv -N${TEST_SIZE} -n${TEST_SIZE} \
	--setattr=exec.test.run_duration=1ms \
	true
```

milroy:
Yes, we should add a check to the testsuite. I'll figure out if I can make an easy change to existing checks to test for this issue and if not I'll adapt your reproducer.

milroy:
@grondo I think your reproducer is the best way to test for this problem in the testsuite. While integrating it into the testsuite, I noticed that `t/issues/t1035-fluxion-reload.sh` isn't being run becuse it got left out of `CMakeLists`: https://github.com/flux-framework/flux-sched/blob/master/t/CMakeLists.txt

Should I re-enable that test in this PR?

grondo:
The tests under `t/issues` should be run by the issues test driver `t5100-issues-test-driver.t`. Is that the test that is actually missing?

milroy:
I added `t1129-match-overflow` based on your reproducer, @grondo. It works as expected in my development environment, but fails the GitHub checks. The output looks identical between local execution and on GitHub. Do you know what might be causing the failures?

grondo:
Yes, this was my fault, sorry. As you noticed, the CI environment has sched-simple loaded, so at the end of the test we have to unload Fluxion modules and reload sched-simple to avoid an error.

I also noticed that my test invokes an unnecessary `flux start`, so try the following patch and see if this works:

```diff
diff --git a/t/issues/t1129-match-overflow.sh b/t/issues/t1129-match-overflow.sh
index bc40a960..11314035 100755
--- a/t/issues/t1129-match-overflow.sh
+++ b/t/issues/t1129-match-overflow.sh
@@ -5,11 +5,6 @@
 
 log() { printf "issue#1129: $@\n" >&2; }
 
-if test -z "$T1129_RELAUNCHED"; then
-       export T1129_RELAUNCHED=t
-       exec flux start $0
-fi
-
 TEST_SIZE=${TEST_SIZE:-900}
 
 log "Unloading modules..."
@@ -41,3 +36,8 @@ log "Running test job."
 flux run -vvv -N${TEST_SIZE} -n${TEST_SIZE} \
        --setattr=exec.test.run_duration=1ms \
        true
+
+log "reloading sched-simple..."
+flux module remove sched-fluxion-qmanager
+flux module remove sched-fluxion-resource
+flux module load sched-simple
```

milroy:
Thanks for the fix! I wonder why the previous version worked on my local `bookworm` container.

Anyway, if this looks good to you now I'll set MWP.

grondo:
>  I wonder why the previous version worked on my local bookworm container.

If you were using a flux-sched container perhaps the fluxion rc3 script is present and removes the Fluxion modules for us. Perhaps the reload of sched-simple isn't actually required. (I'm not sure)

grondo:
Yes, looks great, thanks! Feel free to set MWP.

milroy:
Maybe MWP won't merge unless there's an explicit reviewer approval. @grondo can you submit an approval for the PR?

grondo (approved):
LGTM!

grondo:
Apologies @milroy, I had meant to approve!

PR #1133, opened 2024-01-09
Add match options and match satisfy to reapi

zekemorton:
This PR adds changes to the different bindings for the reapi. These changes center around how match allocate was originally implanted for both the cli and the module. In both bases, the only options are to use either "allocate_orelse_reserve" or "allocate".

This PR introduces 2 new functions to the bindings for go and C and overloads the match allocate function for c++. 

C/Go:
Added functions Match and MatchSatisfy. 
Match is the general purpose match function that interfaces with the C++ bindings
MatchSatisfy is a new function that wraps Match using the "satisfiability" option. 
MatchAllocate is now a wrapper to Match that converts the orelse_reserve bool into the appropriate match option.

C++: 
Added an overloaded match allocate function that acts much like Match as mentioned above.
Edited match allocate to like mentioned above.


resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
```diff
@@ -115,6 +113,62 @@ func (cli *ReapiClient) MatchAllocate(
 
 }
 
+// Match matches a jobspec to the "best" resources and either allocate
```
vsoch, line 116:
```suggestion
// MatchAllocate matches a jobspec to the "best" resources and either allocate
```

vsoch: And looks like the parameters / returns need to be updated.

resource/reapi/bindings/c/reapi_module.h
```diff
@@ -32,6 +32,39 @@ reapi_module_ctx_t *reapi_module_new ();
  */
 void reapi_module_destroy (reapi_module_ctx_t *ctx);
 
+/*! Match a jobspec to the "best" resources and either allocate
+ *  orelse reserve them. The best resources are determined by
+ *  the selected match policy.
+ *
+ *  \param ctx       reapi_module_ctx_t context object
+ *  \param cmd       const char *: set to specify the specific match option
```
vsoch, line 40: "cmd" for a job / workload manager might be confusing - perhaps just explicitly call this `match_option` or similar.

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
```diff
@@ -115,6 +113,62 @@ func (cli *ReapiClient) MatchAllocate(
...
+//	                 in terms of elapse time needed to complete
+//	                 the match operation.
+//	\return          0 on success; -1 on error.
+func (cli *ReapiClient) MatchAllocate(
+	orelse_reserve bool,
+	jobspec string,
+) (reserved bool, allocated string, at int64, overhead float64, jobid uint64, err error) {
+	var match_op string
+
+	if orelse_reserve {
```
vsoch, line 142: These should be const defined by the module at the top.

resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
```diff
@@ -80,6 +79,36 @@ func (m *ReapiModule) MatchAllocate(
...
+// int reapi_module_match_allocate (reapi_module_ctx_t *ctx, bool orelse_reserve,
+// at: is the scheduled time "at"
+func (m *ReapiModule) MatchAllocate(
+	orelse_reserve bool,
+	jobspec string,
+	jobid int,
+) (reserved bool, allocated string, at int64, overhead float64, err error) {
+	var cmd string
+
+	if orelse_reserve {
```
vsoch, line 92: Ditto.

resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
```diff
@@ -80,6 +79,36 @@ func (m *ReapiModule) MatchAllocate(
...
+}
+
+// MatchSatisfy runs satisfiability check on jobspec
+// int reapi_module_match_allocate (reapi_module_ctx_t *ctx, bool orelse_reserve,
+// at: is the scheduled time "at"
+func (m *ReapiModule) MatchSatisfy(
+	jobspec string,
+	jobid int,
+) (reserved bool, allocated string, at int64, overhead float64, err error) {
+	cmd := "satisfiability"
```
vsoch, line 108: Constant here as well.

vsoch (commented):
Did a quick review!

zekemorton:
@vsoch @milroy I think this one is finally ready for another review! Let me know what you think or if you have any questions

resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
```diff
@@ -14,7 +14,13 @@ package fluxcli
 import "C"
 import (
 	"fmt"
-	"unsafe"
+)
+
+const (
```
vsoch, line 19: Ditto here.

resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
```diff
@@ -80,6 +86,32 @@ func (m *ReapiModule) MatchAllocate(
...
+		match_op = MATCH_ALLOCATE_ORELSE_RESERVE
+	} else {
+		match_op = MATCH_ALLOCATE
+	}
+
+	return m.Match(match_op, jobspec, jobid)
+}
+
+// MatchSatisfy runs satisfiability check on jobspec
+func (m *ReapiModule) MatchSatisfy(
```
vsoch, line 107: I have a dumb question - I've never used the module (only the client) - what's the difference, and how does that lead to the different designs for the (equivalently named) functions here? What I'm wanting to understand is if this is something we should be aiming to maintain similar interfaces for (or not).

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
```diff
@@ -115,6 +121,69 @@ func (cli *ReapiClient) MatchAllocate(
...
+		match_op = MATCH_ALLOCATE_ORELSE_RESERVE
+	} else {
+		match_op = MATCH_ALLOCATE
+	}
+
+	return cli.Match(match_op, jobspec)
```
vsoch, line 156: I like this!

resource/reapi/bindings/go/src/test/main.go
```diff
@@ -61,7 +61,25 @@ func main() {
 	}
 	printOutput(reserved, allocated, at, jobid, err)
 
-	reserved, allocated, at, overhead, jobid, err = cli.MatchAllocate(reserve, string(jobspec))
+	reserved, allocated, at, overhead, jobid, err = cli.Match("allocate", string(jobspec))
+	fmt.Println("Errors so far: \n", cli.GetErrMsg())
+
+	if err != nil {
+		fmt.Printf("Error in ReapiClient MatchAllocate: %v\n", err)
```
vsoch, line 68: Just to sanity check out tests here (I haven't looked in a while) does a return ensure there is a test failure? Or the design is just based on the output (and we don't need to check for failure because the output would be wrong, that would make sense too).

t/data/resource/expected/rq2/003.R.out
```diff
@@ -0,0 +1,60 @@
+      ---------------core35[1:x]
+      ------------socket1[1:x]
+      ---------node1[1:s]
```
vsoch, line 3: Yes, operator, I'd like my new wallpaper in "flux R output" design. Yes, extra squiggles please.

resource/reapi/bindings/c/reapi_cli.h
```diff
@@ -67,6 +100,22 @@ int reapi_cli_match_allocate (reapi_cli_ctx_t *ctx, bool orelse_reserve,
                               bool *reserved,
                               char **R, int64_t *at, double *ov);
 
+/*! Match a jobspec to the "best" resources and either allocate
+ *  orelse reserve them. The best resources are determined by
+ *  the selected match policy.
```
milroy, line 105: Typo in description:
```suggestion
/*! Run Satisfiability check for jobspec. 
```

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
milroy: Agreed!

resource/reapi/bindings/go/src/fluxmodule/reapi_module.go
milroy: The module is designed to send RPCs to and receive responses from Fluxion (running in a Flux instance) rather than start/stop/maintain a context. It would be useful to maintain similar interfaces because it may be fun to try wiring up multiple module clients to the same Flux + Fluxion instance. Like if we wanted to maintain coherent scheduling state between Flux resources (e.g., on bare metal) and Fluence resources.

milroy: (The module is what Flux uses to call procedures in Fluxion.)

milroy (changes requested):
Adding a change request to resolve the item with CamelCase.

Beyond a few nits and typos this looks really good to me. In particular I like your API design with overloads to maintain backwards compatibility. It may be possible to use an interface instead, though.

@trws, do you have a preference or another design suggestion here?

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -110,7 +104,7 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
     }
 
     // Check for an unsuccessful match 
-    if (rc == 0) {
+    if ((rc == 0) && (match_op != match_op_t::MATCH_SATISFIABILITY)) {
```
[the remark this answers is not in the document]
trws: If it needs a pre-space, doesn’t it need a post-space also? Is this what gofmt would output?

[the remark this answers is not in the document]
trws: Bah, not go, sorry.  Still the symmetric spacing point stands.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -154,12 +148,26 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
...
     
-    rq->incr_job_counter ();
+    if (match_op != match_op_t::MATCH_SATISFIABILITY)
+        rq->incr_job_counter ();
 
 out:
     return rc;
 }
 
+int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
```
trws, line 158: This whole function shows up as untested.

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -90,26 +90,39 @@ extern "C" int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph,
     return rc;    
 }
 
-extern "C" int reapi_cli_match_allocate (reapi_cli_ctx_t *ctx,
-                   bool orelse_reserve, const char *jobspec,
+extern "C" int reapi_cli_match (reapi_cli_ctx_t *ctx,
+                   const char *match_op, const char *jobspec,
```
trws, line 94: Could we pass the match_op_t enum rather than a string around?  Using all the strcmps to turn it back into an enum seems a bit painful.  Maybe a helper function to convert the string to enum if we need to do that elsewhere then pass it through?

trws:
I think I’m generally ok here, would strongly prefer that we avoid stringly-typed interfaces where we can, but the general design seems like a notable improvement in how we can grow the interface so that’s great.

resource/reapi/bindings/c/reapi_cli.cpp
vsoch: I also really like the idea of enums. I haven't used them in cgo yet, but they seem fairly straight forward. @zekemorton here is an example: https://github.com/giorgisio/cgo/blob/master/enums.go. If we have enums for the match options instead of strings, of course no need to worry about camel case.

resource/reapi/bindings/go/src/test/main.go
zekemorton: The design us just based on the output, so it compares what running it prints out compared to a file of expected output.

vsoch: yeah - I remembered this shortly after I commented (and updated my comment). Thanks for the reminder!

resource/reapi/bindings/c/reapi_cli.cpp
zekemorton: I think that makes a lot of sense! I can change the C bindings cli to use to use enum as a parameter for match op. Do you think I should also change the go bindings to use enum? I wasn't sure if we wanted to use any C types for the go function parameters? Is there a go specific version of an enum that might work better here?

resource/reapi/bindings/c++/reapi_cli_impl.hpp
zekemorton: Yes, I am not sure what the best way to add tests for this is. This function was added for backwards compatibility reasons, but the other changes in this PR seem to move at least what is tested away from it. I am not sure if there is anything using this function directly with the c++ bindings or if it would be safe to remove or not.

resource/reapi/bindings/c/reapi_cli.cpp
vsoch: I'm not sure, because there isn't technically an enum type in Go, but we do have iota and constants! Let me mess around a bit tonight and put together an idea or proposal.

vsoch:
@zekemorton @trws @milroy I put together a quick prototype for what I think might be a good pattern for adding the faux enum, and also testing it: https://github.com/zekemorton/flux-sched/pull/1

The problem and summary are in the description there, and I'll provide some bullets here.

## Organization

- I think we will eventually want to move shared logic between reapi_cli and reapi_module into a common place. The convention for that in go is a "pkg" directory so I added that to src.

## Enum Design

- Reading about enum types, the most widely accepted (but not loved, overall, people are not happy there is no native type) is to use iota with const. A slight detail I picked up on is having an indicator for "we forgot to set this" - so instead of MatchAllocate being 0, we set MatchUnknown to 0.
- The other pattern I saw was for human readability, and that is having a String() function so if someone passes this new `MatchType` to a fmt as a string, it will not throw an error.

## Testing

This was a bit hard. I started with a strategy that would (akin to build) generate a binary, but (to be frank) I don't like the complexity of this approach. We should not need to have CMakeLists.txt and Makefile.am in _every_ single go package directory with a `something_test.go` forever on out. Instead, I want to start on the premise that when a pkg is imported into a main function (e.g., into the main test directory) it would be used (and thus tested, as a binary) there. And then to supplement that, I want to do the vanilla, expected setup that you see with go - a run of `go test -v`. So that's the [test that I added](https://github.com/researchapps/flux-sched/blob/1c0570eea5a0ce0548928641efdd7a8b9d6d8c75/t/t9001-golang-basic.t#L38-L43). The start of that file checks for go on the path, so we already know we have it. The test assumes that the test itself is run in a subshell (so any changes of directory are essentially thrown away).

If we like this design, I can detail the next steps. The main step is to add the enum to C, and then to integrate the types module into each of the reapi module and cli.  Instead of adding it (I am bad/slow at C) I include an example (and example usage) [here](https://github.com/researchapps/flux-sched/blob/1c0570eea5a0ce0548928641efdd7a8b9d6d8c75/resource/reapi/bindings/go/src/pkg/types/types.go#L13-L26). In addition to the [tests that I wrote](https://github.com/researchapps/flux-sched/blob/go-enum-prototype-with-test/resource/reapi/bindings/go/src/pkg/types/types_test.go) we likely would want to have tests that sanity check the C enums in some way against Go to make sure we don't mess them up.

Finally, we could likely add some nice helper functions to that, to assist with the conversion of checking if something is a type (e.g., it would be easier to call `mt.IsMatchAllocate()` and get a boolean vs. something else.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
milroy: This function will get tested once we move to the new `resource-query` (`rq2`) that Zeke is developing. `rq2` is a client of REAPI CLI, so we'll get coverage testing via the testsuite.

zekemorton: rq2 is already represented in the test suite, at least in regards to match_allocate. The match allocate referenced above uses the orelse_reserve flag rather that the match_op parameter. It's possible that it can be safely removed if rq2 and reapi c bindings are the only clients of reapi c++ bindings, though I am not sure if that is the case.

milroy: Got it. I don't know of any other reapi C++ clients.

resource/reapi/bindings/c/reapi_cli.cpp
trws: AFAIK the best you can do is a newtype-like thing with say `type MatchType int` or similar, then make a set of constants based on that, or re-use the C ones maybe by re-exporting their names from the module.  Either way, I'd say ensure that the values in go are taken from C automatically if possible, just to keep maintenance low.

vsoch: lol, like this?

https://github.com/researchapps/flux-sched/blob/1c0570eea5a0ce0548928641efdd7a8b9d6d8c75/resource/reapi/bindings/go/src/pkg/types/types.go#L31-L40

image

vsoch: Side note - did that two days ago - [full PR here](https://github.com/researchapps/flux-sched/commit/f16741eee8b5ca427fbc75579bea4334ded70848). I think everyone missed it here. I want to again emphasize that we need a setup that 1. introduces more modular (shared) types / components, and 2. allows for standard go style testing and 3. makes a distinction between the variable being undefined and the first type.

vsoch: I'm happy to follow up with a PR for whichever of those features are out of scope here.

zekemorton: Yeah, I think would prefer a follow on PR with the conversion to enum if possible. I have a couple of things that I think take higher priority in the near term and would like to get this merged since it's a blocker for some of the things @vsoch has been working on

resource/reapi/bindings/c++/reapi_cli_impl.hpp
zekemorton: Then should I remove it? This would remove the need for overloading the function in the first place which might be better in the long run

resource/reapi/bindings/c/reapi_cli.cpp
zekemorton: After a hack session with @milroy I was able to convert the C and C++ bindings to use enum and adding in the changes for golang should hopefully be pretty smooth. I'll hack on it a bit and see if it can fall under the scope of this PR

resource/reapi/bindings/c++/reapi_cli_impl.hpp
zekemorton: I removed this function from the cli c++ bindings but kept it for the module c++ bindings since it seems its was being used by the queue manager.

resource/reapi/bindings/c/reapi_cli.cpp
zekemorton: Just added a new commit with some work in progress on using the interface with enums. I am running into what I think is a linking error during the build process with the go bindings. I'd love a second set of eyes on it if anyone gets a chance. The error I am getting is: 

```
/usr/lib/golang/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/tce/packages/gcc/gcc-10.3.1/bin/ld: /usr/tce/packages/gcc/gcc-10.3.1/bin/ld: DWARF error: could not find variable specification at offset 55654
/usr/tce/packages/gcc/gcc-10.3.1/bin/ld: DWARF error: could not find variable specification at offset 4a4d
...
/usr/tce/packages/gcc/gcc-10.3.1/bin/ld: DWARF error: could not find variable specification at offset 4ffd0
/g/g0/morton30/flux/flux-sched/resource/reapi/bindings/libreapi_cli.a(reapi_cli.cpp.o): in function `Flux::resource_model::detail::resource_query_t::traverser_run(Flux::Jobspec::Jobspec&, match_op_t, long, long&)':
/g/g0/morton30/flux/flux-sched/resource/reapi/bindings/c++/reapi_cli_impl.hpp:725: undefined reference to `Flux::resource_model::dfu_traverser_t::run(Flux::Jobspec::Jobspec&, std::shared_ptr<Flux::resource_model::match_writers_t>&, match_op_t, long, long*)'

File: resource/reapi/bindings/go/src/test/CMakeLists.txt [hunk 15]
```diff
@@ -1,9 +1,11 @@
...
+# set(CGO_CFLAGS "-I${CMAKE_BINARY_DIR}/resource/reapi/bindings/c") # -I${CMAKE_BINARY_DIR}")
+set(CGO_CFLAGS "-I${CMAKE_BINARY_DIR}")
+
 
 # This is currently passed but not used because when passed into add_custom_command the spaces are escaped
-set(CGO_LIBRARY_FLAGS "-Wl,-R ${CMAKE_BINARY_DIR}/resource -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -L${CMAKE_BINARY_DIR}/resource/libjobspec -ljobspec_conv -lreapi_cli -L${CMAKE_BINARY_DIR}/resource -lresource -L${CMAKE_BINARY_DIR}/resource/planner/c -lplanner -L${CMAKE_BINARY_DIR}/resource/planner/c++ -lplanner_cxx -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp")
+set(CGO_LIBRARY_FLAGS "-Wl,-R ${CMAKE_BINARY_DIR}/resource -L${CMAKE_BINARY_DIR}/resource/reapi/bindings -L${CMAKE_BINARY_DIR}/resource/libjobspec -ljobspec_conv -lreapi_cli -L${CMAKE_BINARY_DIR}/resource -lresource -L${CMAKE_BINARY_DIR}/resource/planner/c -lplanner -L${CMAKE_BINARY_DIR}/resource/planner/c++ -lplanner_cxx -L${CMAKE_BINARY_DIR}/resource/traversers -L${CMAKE_BINARY_DIR}/resource/policy/base -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp")
```
vsoch, line 9: okay here is what has worked for me:

```
# GOOS=linux CGO_CFLAGS="-I/home/flux-sched/resource/reapi/bindings/c" CGO_LDFLAGS="-L/usr/lib -L/home/flux-sched/resource -lresource -L/home/flux-sched/resource/libjobspec -ljobspec_conv -L/home/flux-sched/resource/reapi/bindings -lreapi_cli -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp" go build -ldflags '-w' -o bin/server cmd/main.go

BUILDENVVAR=CGO_CFLAGS="-I${FLUX_SCHED_ROOT}/resource/reapi/bindings/c" CGO_LDFLAGS="-L${INSTALL_PREFIX}/lib -L${FLUX_SCHED_ROOT}/resource -lfluxion-resource -L${FLUX_SCHED_ROOT}/resource/libjobspec -ljobspec_conv -L/${FLUX_SCHED_ROOT}/resource/reapi/bindings -lreapi_cli -lflux-idset -lstdc++ -lczmq -ljansson -lhwloc -lboost_system -lflux-hostlist -lboost_graph -lyaml-cpp"
```
The order matters too I think. And what about changing this line?

```
set(CGO_CFLAGS "-I${CMAKE_BINARY_DIR}/resource/reapi/bindings/c")
```
This step makes me run into trouble too, and what I usually do is figure out which symbol it is asking for, where it is being asked for from, and then 1. ensure that I'm adding that to the include or linked library path, and 2. ensure the order is correct so the thing needed is in the command string before where it's needed, if that makes sense.

vsoch: And then it's like, 20-30 minutes of trying random things until it works! :laughing:

vsoch: From a quick glance I think you'd want `-lreapi_cli` later, to come after the dependencies that it needs.

trws: > the thing needed is in the command string before where it's needed

Quick warning here, if this is a normal gcc-style command line, things earlier on the line can link to symbols provided by libraries and objects later on the line.  Writing that down it sounds backwards, but it is true.

vsoch: Note that @milroy just [pushed an update](https://github.com/flux-framework/flux-sched/blob/6b23d0d3502ae715f7aef165253024a9ba3d430d/resource/reapi/bindings/go/src/test/CMakeLists.txt) to his PR for the bindings that might help here too (I'm testing it out locally with flex-ice-cream, just waiting for the container environment to boot up)!

zekemorton:
Alright, I just pushed some changes and now it seems to build correctly and passes the test! 
I am not so familiar with CMake so I might have done something silly, but seems to be working. Let me know what you think!

resource/reapi/bindings/go/src/test/CMakeLists.txt
zekemorton: Thanks so much for the tips here, this helped a lot!

zekemorton:
I just rebased and pushed changes that take advantage of @milroy 's recent PR. The cmake build commit should be a bit cleaner. This should now be ready for another review

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
```diff
@@ -72,6 +74,53 @@ func (cli *ReapiClient) InitContext(jgf string, options string) (err error) {
...
+//	 				 in terms of elapse time needed to complete
+//					 the match operation.
+//
+// \return          0 on success; -1 on error.
+func (cli *ReapiClient) Match(
+	match_op types.MatchType,
+	jobspec string,
+) (reserved bool, allocated string, at int64, overhead float64, jobid uint64, err error) {
+	var r = C.CString("")
+	spec := C.CString(jobspec)
```
vsoch, line 106: I think for `C.CString` it expects us (the developer) to free the memory, but [I'm not sure](https://groups.google.com/g/golang-nuts/c/bBylU8kE5uk/m/DFC3FvZTDH0J)

image

Ping @trws I always get this wrong.

vsoch:
I forgot the most important thing! :bird: :exploding_head: 

image

vsoch:
And to discuss some planning, I'm hoping to start work on moving the go bindings out of tree after we are merged here, and after that continue work on fluence. I'm going to start from the same git history and be very careful to rename / move so that the original authors / commits are preserved. It would be really great if we could finish this one up today (and then I could work over the weekend) but if that's not possible it's ok too.

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
milroy: I think it needs to be freed here, but am curious what Tom thinks. 

This raises an interesting issue: I think we should try running Valgrind checks against the CGo code. There's evidence that it works: https://github.com/golang/go/issues/60600. That would be good to have in CI testing, but I think is beyond the scope of this PR.

resource/policies/base/match_op.h
```diff
@@ -0,0 +1,20 @@
+#ifndef MATCH_OP_H
+#define MATCH_OP_H
+
+enum class match_op_t { MATCH_UNKNOWN,
```
milroy: I left a comment on an earlier commit, but I'm not sure we need `MATCH_UNKNOWN`. If we leave it in, we need a check for this before incrementing the job counter. See my earlier comment.

resource/reapi/bindings/c++/reapi_module.hpp
```diff
@@ -21,20 +21,28 @@ extern "C" {
...
+                               const uint64_t jobid, bool &reserved,
+                               std::string &R, int64_t &at, double &ov);
     static int match_allocate (void *h, bool orelse_reserve,
                                const std::string &jobspec,
                                const uint64_t jobid, bool &reserved,
                                std::string &R, int64_t &at, double &ov);
     static int match_allocate_multi (void *h, bool orelse_reserve,
                                      const char *jobs,
                                      queue_adapter_base_t *adapter);
+    static int match_allocate_multi (void *h, match_op_t match_op,
```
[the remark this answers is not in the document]
milroy: It's really helpful for reducing the number of RPCs to Fluxion.

resource/reapi/bindings/c++/reapi_module_impl.hpp
```diff
@@ -70,6 +69,18 @@ int reapi_module_t::match_allocate (void *h, bool orelse_reserve,
     return rc;
 }
 
+int reapi_module_t::match_allocate (void *h, bool orelse_reserve,
+                                    const std::string &jobspec,
+                                    const uint64_t jobid, bool &reserved,
+                                    std::string &R, int64_t &at, double &ov)
+{
+    match_op_t match_op = (orelse_reserve)? match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE
```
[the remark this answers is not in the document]
milroy: I believe this is to support `rq2`.

resource/reapi/bindings/c/reapi_cli.h
```diff
@@ -46,15 +47,15 @@ int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph,
...
- *  \param match_op  const char *: set to specify the specific match option
+ *  \param match_op  enum match_op_t: set to specify the specific match option
  *                   from 1 of 4 choices:
- *                   allocate: try to allocate now and fail if resources 
+ *                   MATCH_ALLOCATE: try to allocate now and fail if resources 
  *                   aren't available. 
- *                   allocate_orelse_reserve : Try to allocate and resever if 
- *                   resources aren't available now.
- *                   satisfiability: Do a satisfiablity check and do not 
+ *                   MATCH_ALLOCATE_ORELSE_RESERVE : Try to allocate and reseve
```
milroy, line 54: Nit: typo "reseve".

resource/reapi/bindings/c/reapi_module.h
```diff
@@ -37,15 +38,15 @@ void reapi_module_destroy (reapi_module_ctx_t *ctx);
...
- *  \param match_op  const char *: set to specify the specific match option
+ *  \param match_op  enum match_op_t: set to specify the specific match option
  *                   from 1 of 4 choices:
- *                   allocate: try to allocate now and fail if resources 
+ *                   MATCH_ALLOCATE: try to allocate now and fail if resources 
  *                   aren't available. 
- *                   allocate_orelse_reserve: Try to allocate and resever if 
- *                   resources aren't available now.
- *                   satisfiability: Do a satisfiablity check and do not 
+ *                   MATCH_ALLOCATE_ORELSE_RESERVE : Try to allocate and reseve
```
milroy, line 45: Nit: typo.

resource/policies/base/match_op.h
```diff
@@ -1,20 +1,22 @@
 #ifndef MATCH_OP_H
```
milroy, line 1: It would be cleaner to rebase and squash this commit into 04c7a3510fac6b9ad4ae790a6a7f9ad6e67950e3 as a fixup (to drop this commit message). I don't think that will cause conflicts, but if it does you can leave this commit as is.

milroy (changes requested):
This is getting close and just requires a few formatting fixes and an additional guard check for a `match` function. Please also try squashing one commit as a fixup to reduce the complexity a bit.

This is really good work!

resource/policies/base/match_op.h
zekemorton: The go bindings need to have a default value, though we weren't entirely sure what that value should be. I think MATCH_ALLOCATE would end up being the default option if we just removed MATCH_UNKNOWN

trws: Why do we need a default?

zekemorton: I think it's because the way we define the `MatchType` in go using `iota` means that an initialized but unset MatchType variable will use the 0 value or what we are calling a default. We can set that value to be one of the actual match options or some other value, in this case `MATCH_UNKNOW`. Is that about right @vsoch ?

vsoch: Yes that's correct! We want to be able to distinguish between "this variable was not set" (0) vs. "this variable was set as match allocate" (1). If we had MatchAllocate as the first, then you couldn't tell between those two cases.

resource/reapi/bindings/go/src/fluxcli/reapi_cli.go
trws: Yup, spec needs to be freed.  Should be the usual `defer C.free(unsafe.Pointer(str))` or similar, at least unless things have changed.  The exception to this is cases where an API takes ownership of it and promises to free it later. The `var r` bit seems a bit off to me though.  Since that string is never used, a `var r *C.char = nil` might make more sense, no reason to pay to allocate it and free it if it's unused.  Whether the string placed there by reapi_cli_match needs to be freed I'd have to dig in to check but probably.

resource/policies/base/match_op.h
trws: Hm, I suppose I can see where you're coming from. It's a downside to using an int for this rather than a true algebraic data type sadly. I guess this is ok, just wish we could make an unknown state unrepresentable.

vsoch: If you always instantiate like:

```go
popcornState := Buttered
```
You are probably OK. The issue comes with if/for loops or similar where you need to define the variable at the top of the statement (unset) or similar, and then it may not always be clear if it was set to an explicit value (MatchAllocate == 0) or there is a bug and you forgot to set it.

trws: Right, it makes sense, and FWIW  I have pushed for similar things in APIs where there needs to be an initial but still invalid value.  Suppose I'm just wishing for the rust or OCaml or haskell-style algebraic types where the language doesn't allow for a default-initialized instance, so you can't write `let mut a: PopcornFlavor;` and get away with it.  You can write it, but the compiler will inform you of your misdeeds rather rapidly:

```rust
 #[derive(Debug)]
 enum PopcornFlavor {
    Buttered,
    Salted,
    Kettle
 }
 pub fn main() {
    let mut a: PopcornFlavor;
    println!("{:?}", a)
 }
```

Result:
```shell
$ rustc /home/scogland1/test.rs
error[E0381]: used binding `a` isn't initialized
 --> /home/scogland1/test.rs:9:22
  |
8 |     let mut a: PopcornFlavor;
  |         ----- binding declared here but left uninitialized
9 |     println!("{:?}", a)
  |                      ^ `a` used here but it isn't initialized
  |
  = note: this error originates in the macro `$crate::format_args_nl` which comes from the expansion of the macro `println` (in Nightly builds, run w
ith -Z macro-backtrace for more info)
help: consider assigning a value
  |
8 |     let mut a: PopcornFlavor = todo!();
  |                              +++++++++

error: aborting due to previous error

For more information about this error, try `rustc --explain E0381`.
```

Ah well, better to have it so we can catch it when we can't prevent it otherwise.

vsoch: Ah if only the compiler would tell me about my misdeeds!! That's neat with rust, I wonder why Go didn't do similar.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -154,7 +148,8 @@ int reapi_cli_t::match_allocate (void *h, bool orelse_reserve,
 
     }
     
-    rq->incr_job_counter ();
+    if (match_op != match_op_t::MATCH_SATISFIABILITY)
```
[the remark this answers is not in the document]
zekemorton: Instead of doing this check before job increment, I added a check for a valid match option near the top of the function. Otherwise, it looks like a an invalid match option like `MATCH_UNKNOWN` would end up triggering logic for `MATCH_ALLOCATE`.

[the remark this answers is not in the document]
milroy: I can't find that check in `reapi_cli_t::match_allocate `, but maybe it's in a different commit. Can you point it out to me?

milroy (approved):
Great work iterating on this and integrating everyone's suggestions @zekemorton! This has been an epic PR conversation.

zekemorton:
Looks like I accidentally squashed the commit it was a part of, so I just un-squashed it! Should now be in https://github.com/flux-framework/flux-sched/pull/1133/commits/e7e78f82986feb4e5e12d67f340a83e540d27258

resource/policies/base/match_op.h
```diff
@@ -0,0 +1,34 @@
...
+#define MATCH_OP_H
+
+
+enum match_op_t { MATCH_UNKNOWN,
+                  MATCH_ALLOCATE,
+                  MATCH_ALLOCATE_W_SATISFIABILITY,
+                  MATCH_ALLOCATE_ORELSE_RESERVE,
+                  MATCH_SATISFIABILITY };
+
+static const char *match_op_to_string (enum match_op_t match_op) {
```
milroy, line 11: No need to specify `enum` type here:
```suggestion
static const char *match_op_to_string (match_op_t match_op) {
```

resource/policies/base/match_op.h
```diff
@@ -0,0 +1,34 @@
...
+    switch (match_op) {
+        case MATCH_ALLOCATE:                  return "allocate";
+        case MATCH_ALLOCATE_ORELSE_RESERVE:   return "allocate_orelse_reserve";
+        case MATCH_ALLOCATE_W_SATISFIABILITY: return "allocate_with_satisfiability";
+        case MATCH_SATISFIABILITY:            return "satisfiability";
+        default: return "error";
+    }
+}
+
+static bool match_op_valid (enum match_op_t match_op) {
```
milroy, line 21: No need to specify `enum` type here:
```suggestion
static bool match_op_valid (match_op_t match_op) {
```

resource/reapi/bindings/c++/reapi.hpp
```diff
@@ -99,9 +99,16 @@ class reapi_t {
      *                   detail. However, when it is used within a Flux's
      *                   service module, it is expected to be a pointer
      *                   to a flux_t object.
-     *  \param orelse_reserve
-     *                   Boolean: if false, only allocate; otherwise, first try
-     *                   to allocate and if that fails, reserve.
+     *  \param match_op  enum match_op_t: set to specify the specific match option
```
milroy, line 102:
```suggestion
     *  \param match_op match_op_t: set to specify the specific match option
```

resource/reapi/bindings/c++/reapi.hpp
```diff
@@ -131,9 +138,15 @@ class reapi_t {
...
```
milroy, line 141:
```suggestion
     *  \param match_op match_op_t: set to specify the specific match option

File: resource/reapi/bindings/c/reapi_cli.cpp [hunk 28]
```diff
@@ -91,33 +91,21 @@ extern "C" int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph,
 }
 
 extern "C" int reapi_cli_match (reapi_cli_ctx_t *ctx,
-                                const char *match_op, const char *jobspec,
+                                enum match_op_t match_op, const char *jobspec,
```
milroy, line 94: No need for `enum` type in function signature.
```suggestion
                               match_op_t match_op, const char *jobspec,
```

resource/reapi/bindings/c/reapi_cli.h
```diff
@@ -69,7 +70,7 @@ int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph,
  *                   the match operation.
  *  \return          0 on success; -1 on error.
  */
-int reapi_cli_match (reapi_cli_ctx_t *ctx, const char *match_op,
+int reapi_cli_match (reapi_cli_ctx_t *ctx, enum match_op_t match_op,
```
milroy, line 73: No need for `enum` type in function signature.
```suggestion
int reapi_cli_match (reapi_cli_ctx_t *ctx, match_op_t match_op,
```

resource/reapi/bindings/c/reapi_module.cpp
```diff
@@ -48,7 +48,7 @@ extern "C" void reapi_module_destroy (reapi_module_ctx_t *ctx)
 }
 
 extern "C" int reapi_module_match (reapi_module_ctx_t *ctx,
-                                   const char *match_op, const char *jobspec,
+                                   enum match_op_t match_op, const char *jobspec,
```
milroy, line 51: No need for `enum` type in function signature.
```suggestion
                                 match_op_t match_op, const char *jobspec,
```

resource/reapi/bindings/c/reapi_module.h
```diff
@@ -60,7 +61,7 @@ void reapi_module_destroy (reapi_module_ctx_t *ctx);
  *                   the match operation.
  *  \return          0 on success; -1 on error.
  */
-int reapi_module_match (reapi_module_ctx_t *ctx, const char *match_op,
+int reapi_module_match (reapi_module_ctx_t *ctx, enum match_op_t match_op,
```
milroy, line 64: No need for `enum` type in function signature.
```suggestion
int reapi_module_match (reapi_module_ctx_t *ctx, match_op_t match_op,
```

resource/reapi/bindings/c/reapi_module.h
```diff
@@ -37,15 +38,15 @@ void reapi_module_destroy (reapi_module_ctx_t *ctx);
  *  the selected match policy.
  *
  *  \param ctx       reapi_module_ctx_t context object
- *  \param match_op  const char *: set to specify the specific match option
+ *  \param match_op  enum match_op_t: set to specify the specific match option
```
milroy, line 41:
```suggestion
 *  \param match_op match_op_t: set to specify the specific match option
```

resource/reapi/bindings/c/reapi_cli.h
```diff
@@ -46,15 +47,15 @@ int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph,
  *  the selected match policy.
  *
  *  \param ctx       reapi_cli_ctx_t context object
- *  \param match_op  const char *: set to specify the specific match option
+ *  \param match_op  enum match_op_t: set to specify the specific match option
```
milroy, line 50:
```suggestion
 *  \param match_op match_op_t: set to specify the specific match option
```

milroy (commented):
Some `enum` type specifiers crept into function signatures and should be removed. It is correct to include the type specifier as well (see https://en.cppreference.com/w/cpp/language/elaborated_type_specifier), but it is inconsistent with the majority of Fluxion code. 

Once that's fixed this should be ready to merge.

resource/policies/base/match_op.h
trws: Is this a C++-only header? I thought this was C, where it's required unless there's a `typedef` to alias the enum to the unqualified name.

trws: We do usually do that though... Oh I see, it used to be in a C++ file and was moved here.  The equivalent would be to make it:

```c
typedef enum match_op_t { MATCH_UNKNOWN,
                  MATCH_ALLOCATE,
                  MATCH_ALLOCATE_W_SATISFIABILITY,
                  MATCH_ALLOCATE_ORELSE_RESERVE,
                  MATCH_SATISFIABILITY } match_op_t;
```

That way it can be used bare, which is what we do most places.

zekemorton: This is super helpful, thanks so much for the suggestion! This is definitely cleaner and more constant with the rest of the code, I just wasn't sure that I could/how to do it before. Went ahead and made those changes, so should be ready now!

milroy: Yeah, that's a clean solution!

milroy (approved):
LGTM. Thanks for all the hard work pushing this through!

vsoch:
yes!! :partying_face:

PR #1139, opened 2024-02-06
debbuild: fix Python purelib paths on Debian systems

grondo:
Problem: Debian based systems may use the 'posix_local' scheme for Python installations, which causes `make deb` to break because the 'purelib' path is returned under /usr/local even when the flux-core prefix is /usr.

Use the workaround described in

 https://gitlab.kitware.com/cmake/cmake/-/issues/25113

and set DEB_PYTHON_INSTALL_LAYOUT=deb to fix `make deb`.

There is probably some standard way to actually fix this issue, but since the deb builds are purely for testing at this time, this solution works for now.

### DIFF
--- a/scripts/debbuild.sh
+++ b/scripts/debbuild.sh
@@ -5,6 +5,19 @@ DEBFULLNAME=$USER
 EMAIL=$(git config --get user.email)
 DEBEMAIL=$EMAIL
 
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! I've set MWP.

PR #1141, opened 2024-02-07
configure: drop src/shell requirements

wihobbs:
Problem: autotools borks because it goes looking for a Makefile for the fluxion shell, but support for that was dropped in #1102.

Drop it from the configure script as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,7 +193,6 @@ AC_CONFIG_FILES([Makefile
   src/cmd/Makefile
   src/python/Makefile
   src/python/fluxion/Makefile
-  src/shell/Makefile
   resource/Makefile
   resource/planner/Makefile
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -193,7 +193,6 @@ AC_CONFIG_FILES([Makefile
   src/cmd/Makefile
   src/python/Makefile
   src/python/fluxion/Makefile
-  src/shell/Makefile
```
grondo, line 196: Commit message correction: There was no "fluxion shell", what was in `src/shell` was the data-staging plugin for the Flux shell.

grondo (approved):
LGTM! Thanks!

wihobbs:
Thanks @grondo!

Issue #1137, opened 2024-02-01
possible performance issue in `sched.resource-status` RPC


grondo:
As mentioned in a meeting yesterday, during preparation for scale testing it was noted that `flux resource list` response times were excessively slow when Fluxion was the loaded scheduler module. To probe the performance of this RPC, the following script was developed which can test the response times of the RPCs used by `flux resource list` at different node sizes. Results for `sched-simple` are included as well for comparison:

```sh
#!/bin/bash
PROG=$(basename $0)
VERBOSE=0

log() { 
    test $VERBOSE -eq 0 && return
    local fmt=$1;
    shift
    printf >&2 "$PROG: $fmt\n" $@
}
rpc() { flux python -c "import flux; flux.Flux().rpc(\"$1\").get()"; }

runtest() {
    SCHEDULER=$1
    NNODES=$2
    log "Starting test of ${SCHEDULER} with ${NNODES} nodes"
    log "Removing modules..."
    flux module remove -f sched-fluxion-qmanager
    flux module remove -f sched-fluxion-resource
    flux module remove -f sched-simple
    flux module remove resource

    log "Loading fake resources via config..."
    flux config load <<EOF
[resource]
noverify = true
norestrict = true

[[resource.config]]
hosts = "test[1-${NNODES}]"
cores = "0-63"
gpus = "0-8"
EOF

    log "Reloading resource module..."
    flux module load resource noverify

    log "Loading ${SCHEDULER} modules..."
    if test "$SCHEDULER" = "sched-simple"; then
        flux module load sched-simple
    else
        flux module load sched-fluxion-resource
        flux module load sched-fluxion-qmanager
    fi

    log "Timing sched.resource-status RPC"
    t0=$(date +%s.%N)
    rpc "sched.resource-status"
    t1=$(date +%s.%N)
    dt1=$(echo "$t1 - $t0" | bc -l)

    log "Timing resource.status RPC"
    t0=$(date +%s.%N)
    rpc "resource.status"
    t1=$(date +%s.%N)
    dt2=$(echo "$t1 - $t0" | bc -l)

    printf "%-13s %8s %28.3f %18.3f\n" $SCHEDULER $NNODES $dt1 $dt2
}

printf "%-13s %8s %28s %18s\n" \
       SCHEDULER NNODES "T(sched.resource-status)" "T(resource.status)"
for scheduler in sched-simple fluxion; do
    for nnodes in 128 256 512 1024 2048 4096 8192 16384; do
[...]

PR #1144, opened 2024-02-20
Improve sched.resource-status RPC and search performance

milroy:
As reported in issue #1137, the `sched.resource-status` RPC is so slow that Fluxion becomes unusable at large scale (16K simulated nodes).

This PR improves performance by implementing caching of `R all`, `R down`, and `R alloc` payloads. It also improves traversal performance for single-subsystem resource graphs by short-circuiting subsystem selector and edge comparisons.

Note that the search performance improvement applies to `find` searches as well as other traversals like `match allocate`.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -105,6 +105,15 @@ struct resource_ctx_t : public resource_interface_t {
     std::map<uint64_t, uint64_t> allocations;  /* Allocation table */
     std::map<uint64_t, uint64_t> reservations; /* Reservation table */
     std::map<std::string, std::shared_ptr<msg_wrap_t>> notify_msgs;
+    bool m_resources_updated = true; /* resources have been updated */
+    bool m_resources_down_updated = true; /* down resources have been updated */
+    /* last time allocated resources search updated */
+    std::chrono::time_point<
+                    std::chrono::system_clock> m_resources_alloc_updated;
+    /* R caches */
+    json_t *m_r_all;
+    json_t *m_r_down;
+    json_t *m_r_alloc;
 };
 
 msg_wrap_t::msg_wrap_t (const msg_wrap_t &o)
@@ -317,6 +326,7 @@ static void set_default_args (std::shared_ptr<resource_ctx_t> &ctx)
     ct_opts.set_match_policy ("first");
     ct_opts.set_prune_filters ("ALL:core");
     ct_opts.set_match_format ("rv1_nosched");
+    ct_opts.set_update_interval (0);
     ctx->opts += ct_opts;
 }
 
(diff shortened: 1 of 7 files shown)


milroy:
I timed 100 invocations of `find` (by timing the `find` call in `resource-query`) with current `master` against this PR. The average and standard deviation of the `master` branch are 4.7805882 and 0.113447098 seconds, respectively. The average and standard deviation for this PR are 4.0092688 and 0.103034633 seconds. 

That's a performance improvement of about 16%.

milroy:
The subsystem selector short circuit in the current PR does this:
```C++
        if (m_subsystem_size == 1)
           return true;
        typedef typename boost::property_traits<inframap>::value_type infra_type;
        const infra_type &inf = get (m_imap, ent);
        const multi_subsystems_t &subsystems = inf.member_of;
```

Which requires creating new member data (`m_subsystem_size`). I tried the same check without the member data, e.g.,
```C++
        typedef typename boost::property_traits<inframap>::value_type infra_type;
        const infra_type &inf = get (m_imap, ent);
        const multi_subsystems_t &subsystems = inf.member_of;
        if (subsystems.size () == 1)
           return true;
```
which resulted in lower performance. I assumed this was due to eliminating the declarations. However, codecov indicates these lines are covered anyway. Opinions on whether codecov is wrong, or  whether my "optimization" isn't working as I expect?

jameshcorbett:
How much lower performance?

milroy:
> How much lower performance?

I'll have to rerun the tests, but about 10% if I remember correctly.

trws:
That's a surprisingly large difference for that change.  Implies a lot more cost in get than I would have guessed.

jameshcorbett (approved):
Looks reasonable!

milroy:
I didn't remember correctly.

The current PR short circuit (check and `return` before the declarations) yields an average `find` time (average of 50 searches) across 16,384 nodes with the graph defined in issue #1137 of 3.9163794 seconds. Moving the check and `return` after the declarations yields a 50-`find` average time of 4.0694326 seconds. That's about 3.8% different.

However, the standard deviations are very different. The current PR version exhibits a standard deviation of 0.065674029 versus 0.098490942 for the version with `return` after the declarations. That's a 33.3% difference. I say let's stick with the current version unless you think the overhead of storing the additional `int` isn't worth it.

milroy:
I should add that I did the tests on my laptop, which isn't an ideal test environment.

tpatki:
Just a note: working on adding PFA annotations to Fluxion, will update on when that's up and running. Does anyone have a recommendation of where to build/test? I need clang, and was trying on Quartz/Tioga. @milroy What's your local test setup?

milroy:
I switched the member data type from `int` to `bool` in `subsystem_selector_t` and plan to set MWP tomorrow if there are no objections.

milroy:
> Does anyone have a recommendation of where to build/test? I need clang, and was trying on Quartz/Tioga. @milroy What's your local test setup?

@tpatki I use the CI docker images on my laptop. The `bookworm - clang-15` test would be a good base to use. You should be able to pull the image and start a container in podman on quartz or tioga and develop and test there.

trws:
Yeah, I'm mostly using the bookworm environment myself just because it's a nice common baseline.  Quartz or any toss4 system, other than tioga apparently, should work well too.

The lack of clang compilers on there confuses me greatly.  I can add one to my devtools env which I think you both have access to, but they should really be in TCE.  Thanks for reporting that one @tpatki.

trws (approved):

milroy:
Thanks for the feedback and reviews! Setting MWP.

PR #1145, opened 2024-02-21
testsuite: remove trailing ampersands in tests

wihobbs:
Problem: there are two tests that have trailing `&&` in the test_debug calls, and when these tests are run with the --debug flag in GitLab CI, they error out.

Remove those trailing ampersands.


grondo (approved):
Great find Thanks!

PR #1147, opened 2024-02-23
testsuite: prepare for flux_job_wait(3) change

garlick:
Problem: some tests presume that only jobs submitted with the 'waitable' flag will contribute to the return value of flux job wait --all, but flux-framework/flux-core#8357 proposes to make all jobs waitable, regardless of the flag.

Ensure that t1023-multiqueue-constraints.t consumes residual wait state before starting the array of waitable jobs.


grondo (approved):
LGTM, but I hope there aren't other assumptions like this one in use in the wild! :grimacing:

garlick:
Yeah it's a bit of a concern....  Thanks, I'll set MWP.

PR #1149, opened 2024-03-04
reader: fix jgf properties

jameshcorbett:
Given this vertex in JGF, resource-query reports it has no properties.

```json
      {
        "id": "35",
        "metadata": {
          "type": "node",
          "basename": "node",
          "name": "compute-01",
          "id": 1,
          "uniq_id": 35,
          "rank": 0,
          "exclusive": true,
          "unit": "",
          "size": 1,
          "properties": ["pdebug"],
          "paths": {
            "containment": "/ElCapitan0/rack0/compute-01"
          }
        }
      }
```

```
resource-query> g /ElCapitan0/rack0/compute-01
No properties were found for /ElCapitan0/rack0/compute-01 (vtx's uniq_id=35).
```

Looking at the reader, I found that it expects properties to be given as a mapping with keys the name of the property. When I changed the vertex in JGF to the following, the property was recognized. It looks like the reader had been silently failing on the type.

```json

      {
        "id": "35",
        "metadata": {
          "type": "node",
          "basename": "node",
          "name": "compute-01",
          "id": 1,
          "uniq_id": 35,
          "rank": 0,
          "exclusive": true,
          "unit": "",
          "size": 1,
          "properties": {
            "pdebug": ""
          },
          "paths": {
            "containment": "/ElCapitan0/rack0/compute-01"
          }
        }
      },
```

```
resource-query> g /ElCapitan0/rack0/compute-01
pdebug=
```

### DIFF
--- a/t/python/t10001-resourcegraph.py
+++ b/t/python/t10001-resourcegraph.py
@@ -45,6 +45,17 @@ RV1_2 = {
     },
 }
 
+RV1_3 = {
+    "version": 1,
+    "execution": {
+        "R_lite": [{"rank": "0-19", "children": {"gpu": "0-1", "core": "0-7"}}],
+        "starttime": 0.0,
+        "expiration": 0.0,
+        "nodelist": ["compute[0-19]"],
+        "properties": {"pdebug": "0-9", "pbatch": "10-19"},
+    },
+}
+
 
 class TestResourceGraph(unittest.TestCase):
     """Test for the ResourceGraph class."""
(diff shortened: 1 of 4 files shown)


jameshcorbett:
@milroy this is ready for review whenever.

t/python/t10001-resourcegraph.py
```diff
@@ -45,6 +45,17 @@
     },
```
milroy, line 45: Typo in the commit message. I think you mean "there are no tests where the JGF writer handles properties correctly."

milroy (approved):
Apart from a typo in one commit message this PR LGTM!

t/python/t10001-resourcegraph.py
jameshcorbett: Good catch, thanks! Force-pushed with the correction.

jameshcorbett:
Thanks! Setting MWP.

PR #1161, opened 2024-04-06
policies: add firstnodex to the known policies

trws:
problem: the firstnodex policy was added and tested, but couldn't be selected by users because it wouldn't be seen as a known policy.

solution: accept the firstnodex value as a known policy in `known_match_policy`


milroy (approved):
Thanks for the quick PR. LGTM!

PR #1162, opened 2024-04-06
Improve match performance with node constrains when feasibility is enabled

milroy:
Matching jobspecs with node constraints is currently implemented in such a way that the recursive `dom_*` call occurs before constraint checking.

This PR moves the node constraint check within the prune function to prevent unnecessary recursive dom_* calls, speeding up feasibility checks.

The following are comparative performance tests run on a laptop. The following test configuration is used to test without feasibility checking:
```bash
flux config load config.toml
flux module load resource noverify monitor-force-up
```

and the following is is contents of `config.toml`:
```toml
[resource]
noverify = true
norestrict = true

[[resource.config]]
hosts = "test[1-16384]"
cores = "0-63"
gpus = "0-8"

[sched-fluxion-qmanager]
# easy backfill
queue-policy = "easy"

[sched-fluxion-resource]
match-policy = "firstnodex"
match-format = "rv1_nosched"
```
### Test results without the change in this PR:

- Feasibility enabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 --requires="hosts:test[16001-16016]" hostname; done
real	0m27.414s
```
- Feasibility disabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 --requires="hosts:test[16001-16016]" hostname; done
real	0m7.360s
```
- Feasibility enabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
real	0m8.184s
```
- Feasibility disabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
real	0m7.136s
```
### Test results WITH the change in this PR:

- Feasibility enabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 --requires="hosts:test[16001-16016]" hostname; done
real	0m10.710s
```
- Feasibility disabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 --requires="hosts:test[16001-16016]" hostname; done
real	0m7.344s
```
- Feasibility enabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
real	0m8.155s
```
- Feasibility disabled:
```bash
time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
real	0m7.231s
```

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -209,6 +209,14 @@ int dfu_impl_t::prune (const jobmeta_t &meta, bool exclusive,
         rc = -1;
         goto done;
     }
+    //  RFC 31 constraints only match against type == "node"
+    //  unspecified constraint matches everything
+    if (meta.constraint != nullptr
+        && (*m_graph)[u].type == "node"
+        && !meta.constraint->match ((*m_graph)[u])) {
+        rc = -1;
+        goto done;
+    }
     // if rack has been allocated exclusively, no reason to descend further.
     if ( (rc = by_avail (meta, s, u, resources)) == -1)
         goto done;
@@ -704,13 +712,6 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta, vtx_t u,
         dom_exp (meta, u, next, check_pres, &x_inout, dfu);
     *excl = x_in;
     (*m_graph)[u].idata.colors[dom] = m_color.black ();
-
-    //  RFC 31 constraints only match against type == "node"
-    //  unspecified constraint matches everything
-    if (meta.constraint != nullptr
-        && (*m_graph)[u].type == "node"
-        && !meta.constraint->match ((*m_graph)[u]))
-        goto done;
 
     p = (*m_graph)[u].schedule.plans;
     if ( (avail = planner_avail_resources_during (p, at, duration)) == 0) {


grondo:
> ```
> time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
> real	0m8.184s
> ```
Nice!

Just a few quick comments about testing job submission and scheduling:

Submitting jobs with `flux submit` back-to-back like this may not be fully exercising the system. This is because the time to submit a single job here is bounded by the time to run `flux submit`, not necessarily the actual submission RPC. A better test of the impact of these changes might be to run a throughput test that tests how long it takes the scheduler to schedule all the submitted jobs. To do that, use the `--cc` option. For example on a size=1 instance on quartz:

```console
$ time for i in {1..100}; do flux submit -N 16 -n 64 hostname; done
real	0m34.698s
$ time flux submit --cc=1-100 -N16 -n 64 hostname
real	0m0.847s
```

You can see that submitting multiple jobs at once with `--cc` is a few orders of magnitude faster than `flux submit` in a loop. This is not only because we don't have to pay the half second penalty for starting `python` for each job, but also because the submission RPCs are sent asynchronously.

If you add `--wait-event=alloc` then the command will wait until all jobs have been scheduled, you can get the time for the scheduler to schedule all jobs. You could also add `--setattr=exec.test.run_duration=10s` to bypass the execution system so that jobs actually stay in RUN state instead of getting an exception since all the fake resources don't really exist.

Using these together might give us a better idea of the full impact of these changes, i.e. I imagine things will look much better...

grondo (approved):
Without really fully understanding the change :grin: LGTM!

trws (approved):
I like it! This makes me realize that the pruning plays an interesting conflicting role with the current coloring scheme though, resources that are pruned are never colored, so they could be explored by later passes. It's not an issue here because the nodes would just be skipped by the constraint anyway, but something to keep in mind as we work through subsystem match and reworking coloring. 

I think this is GTG if you're happy @milroy

milroy:
> Just a few quick comments about testing job submission and scheduling:

@grondo I now realize you've explained that to me before but I didn't think of it when I was running the initial test. Thanks for the additional details so I understand better!

I reran the tests above with the same setup using your suggestions, and the improvements from the change in this PR are uniform and better than I reported above. Note that I ran 1000 jobs with feasibility disabled because the tests ran so fast:

### Test results WITHOUT the change in this PR:

- Feasibility enabled (100 jobs):
```bash
time flux submit --cc=1-100 -N16 -n 64 --requires="hosts:test[16001-16016]" hostname
real	0m33.823s
```
- Feasibility disabled (1000 jobs):
```bash
time flux submit --cc=1-1000 -N16 -n 64 --requires="hosts:test[16001-16016]" hostname
real	0m1.382s
```
- Feasibility enabled (100 jobs):
```bash
time flux submit --cc=1-100 -N16 -n 64 hostname
real	0m1.896s
```
- Feasibility disabled (1000 jobs):
```bash
time flux submit --cc=1-1000 -N16 -n 64 hostname
real	0m0.395s
```
### Test results WITH the change in this PR:

- Feasibility enabled (100 jobs):
```bash
time flux submit --cc=1-100 -N16 -n 64 --requires="hosts:test[16001-16016]" hostname
real	0m8.807s
```
- Feasibility disabled (1000 jobs):
```bash
time flux submit --cc=1-1000 -N16 -n 64 --requires="hosts:test[16001-16016]" hostname
real	0m1.372s
```
- Feasibility enabled (100 jobs):
```bash
time flux submit --cc=1-100 -N16 -n 64 hostname
real	0m1.873s
```
- Feasibility disabled (1000 jobs):
```bash
time flux submit --cc=1-1000 -N16 -n 64 hostname
real	0m0.379s
```

That's a throughput improvement of almost 4x for 100 jobs with feasibility checking.

milroy:
Thanks for the reviews! Setting MWP.

PR #1169, opened 2024-04-12
WIP fix: convert strings to boost:flyweight

vsoch:
Problem: string comparison is immensely inefficient, taking 6-10% of resources (reported by Tom) for a trace in the scoring module..
Solution: start with the root of the issue in the scoring module and work backwards to convert string types to boost:: flyweight. I tried to have a minimal footprint here but it spread really quickly

Background: [boost::flyweight](https://www.boost.org/doc/libs/1_79_0/libs/flyweight/doc/tutorial/index.html) is a cool type that essentially maps a string to an internal hashmap. @trws chose it because it still works with a lot of native string stuff. The large change was converting the subsystem_t to use it, which bubbles into many places. Design decisions we will need to make:

- Changes to top level interfaces vs. parsing the strings as flyweights (as I do now)
- Creating classes for groups of flyweights (e.g., subsystem, relation) unlike the flattened versions I have
- When to do a conversion vs. a `.get()`

There is a lot of room for improvement, but I'm really new to C++ so I'm trying to keep this as simple as possible to start, and first getting something working, and then make it better. The build currently works, but tests fail.

Opening this as a draft because I need help with debugging tests - we have segfaults that need gdb-fu and I've only used -disas - ping @trws when you have time (or others that might be interested - I mostly need to learn how to do it).

Also [this is how I feel](https://youtu.be/ZC8jwLHyJgA) about C++ strings... :laughing:  (sorry still laughing about Monk / this entire find)! Gotta make the painful things fun, I think.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,8 @@
         "terminal.integrated.defaultProfile.linux": "bash"
   },
       "extensions": [
-         "ms-vscode.cmake-tools"
+        "ms-vscode.cmake-tools",
+        "ms-vscode.cpptools-extension-pack"
       ]
     }
   },
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,11 @@ pkg_check_modules(HWLOC REQUIRED IMPORTED_TARGET hwloc>=1.11.1)
 pkg_check_modules(JANSSON REQUIRED IMPORTED_TARGET jansson>=2.10)
 pkg_check_modules(UUID REQUIRED IMPORTED_TARGET uuid)
 
+# threads needed for boost flyweight
+set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads REQUIRED)
+
 set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
@@ -209,6 +214,10 @@ add_custom_target(dist
   COMMAND rm flux-sched.ver
   COMMENT "Generated flux-sched-${FLUX_SCHED_VER}.tar.gz"
   )
+
+add_compile_options(-fsanitize=address)
+add_link_options(-fsanitize=address)
+
 # run installcheck, if it passes then write out version information and pack up
 # a tarball with the result
 add_custom_target(distcheck
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -24,9 +24,12 @@ extern "C" {
 #include <string>
 #include <memory>
 #include <cstdint>
+#include <boost/flyweight.hpp>
 
 #include "resource/reapi/bindings/c++/reapi.hpp"
 #include "qmanager/config/queue_system_defaults.hpp"
+
+static boost::flyweight<std::string>::initializer  fwinit;
 
 namespace Flux {
 namespace queue_manager {
--- a/resource/CMakeLists.txt
+++ b/resource/CMakeLists.txt
@@ -46,6 +46,7 @@ add_library(resource STATIC
     policies/dfu_match_policy_factory.cpp
     jobinfo/jobinfo.cpp
     schema/resource_data.cpp
+    schema/data_std.cpp
     schema/infra_data.cpp
     schema/sched_data.cpp
     schema/color.cpp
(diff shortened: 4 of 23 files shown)


vsoch:
Some tiny progress! Thanks to @milroy for seeing this. Here is the first failure to build (this is for the focal build)

```console
 /usr/include/boost/flyweight/detail/recursive_lw_mutex.hpp:57: undefined reference to `pthread_mutexattr_init'
```

We are missing Threads! :thread: No, not the Facebook chat app - nobody wants that kind of threads. :P 

image

So I very simply added it to the CMakeLists.txt, and specifically to check that it was supported in the top level one, and then as a library that is needed for linking for the resource public library. That seemed to get through that first failure (this is the same focal build), which now gets through that, albeit the tests still fail:

image

So probably I need to figure out how to use valgrind to debug test failures next.

trws:
Ah, it must default to using the mutex protection for flyweights. If we only use them in the resource module, we can use the `nolock` one to not need that, but if we use them in more places it will be useful.

vsoch:
I wonder if the equality check is just failing because we are still using strings in the tests? https://github.com/flux-framework/flux-sched/blob/cb951738bbba0e296678b2e5120484da9e1950e5/resource/schema/test/schema_test02.cpp#L101-L116 Will try some things after dinner!

trws:
That’s one of the tests that doesn’t run mod main. So it’s comparing uninitialized flyweights.

---
On April 12, 2024 at 7:59:07 PM PDT, Vanessasaurus ***@***.***> wrote:

I wonder if the equality check is just failing because we are still using strings in the tests? https://github.com/flux-framework/flux-sched/blob/cb951738bbba0e296678b2e5120484da9e1950e5/resource/schema/test/schema_test02.cpp#L101-L116<https://urldefense.us/v3/__https://github.com/flux-framework/flux-sched/blob/cb951738bbba0e296678b2e5120484da9e1950e5/resource/schema/test/schema_test02.cpp*L101-L116__;Iw!!G2kpM7uM-TzIFchu!0hDtEWYzaSi9HWDuBdOhvHR3PZBfivFdlRjQ7FmKxm_BUN1IvU8tsNg4v3BW2Hx8F4ph_5we_nzfaqUfAup_CZbrBLk$> Will try some things after dinner!

vsoch:
Yeah I think I'm reading the test wrong too, I think "ok" is from this tap thing: https://github.com/flux-framework/flux-sched/blob/cb951738bbba0e296678b2e5120484da9e1950e5/src/common/libtap/tap.h#L48

vsoch:
> That’s one of the tests that doesn’t run mod main. So it’s comparing uninitialized flyweights.
It's strange that doesn't throw an error. 

What about if we try an init() somewhere? I'm reading here: https://live.boost.org/doc/libs/1_74_0/libs/flyweight/doc/tutorial/technical.html

vsoch:
Albeit the design needs some work, I wanted to test that doing an init for flywheel in the "one off" places (resource query utiliity and tests) would resolve those tests (and I think it did).
image

vsoch:
I don't know how to debug this more because my local run doesn't reproduce what happens here:

image

I also have jammy, so not sure why it would be different. But I can't compare apples to oranges.

vsoch:
There is also something called a key value flyweight, I wonder if we need to use that for some of the subsystem maps? https://www.boost.org/doc/libs/1_79_0/libs/flyweight/example/key_value.cpp. I also don't know the difference between when they show:

```
boost::flyweights::flyweight
```
vs. what we are doing, which is missing the middle level:


```
boost::flyweight
```
It could be the first is just an old design - I don't see it here: https://www.boost.org/doc/libs/1_79_0/libs/flyweight/doc/reference/index.html. They also have a note about the threads library there:

> In order to use the serialization capabilities of Boost.Flyweight, the appropriate Boost.Serialization library module must be linked. Other than that, Boost.Flyweight is a header-only library, requiring no additional object modules. The default [simple_locking](https://www.boost.org/doc/libs/1_79_0/libs/flyweight/doc/reference/locking.html#simple_locking) locking policy introduces a dependency on the Pthreads library on those POSIX-compliant environments where the [Boost.Config](https://www.boost.org/doc/libs/1_79_0/libs/config/doc/html/index.html) macro BOOST_HAS_PTHREADS is set.

Maybe we want a holder too?

> intermodule_holder_class<C> maintains a C instance which is unique even across different dynamically linked modules of the program using this same type. In general, this guarantee is not provided by [static_holder_class](https://www.boost.org/doc/libs/1_79_0/libs/flyweight/doc/reference/holders.html#static_holder_class), as most C++ implementations are not able to merge duplicates of static variables stored in different dynamic modules of a program.

PR #1174, opened 2024-04-12
up the required C++ standard to 20

trws:
I realize this is odd timing for this, but it's been sitting for a while, and I keep tripping over myself trying to use interfaces that I know exist but don't in C++14.  IIRC @grondo did a test that we can build with gcc12+ everywhere we care about through the devtoolset to build our RPMs, it would help me a lot if we could finish this through as long as it doesn't cause deployment concerns.

problem: Many interfaces are made more cumbersome or slower by forcing the use of only up to c++14.

solution: As previously discussed, up the required C++ standard to 20 since we have devtoolset gcc everywhere that can compile to that standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ if(POLICY CMP0140)
   cmake_policy(SET CMP0140 NEW)
 endif()
 
-set(CMAKE_CXX_STANDARD 14 CACHE STRING "The C++ standard to use")
+set(CMAKE_CXX_STANDARD 20 CACHE STRING "The C++ standard to use")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
--- a/resource/planner/c++/planner_multi.hpp
+++ b/resource/planner/c++/planner_multi.hpp
@@ -35,6 +35,19 @@ struct planner_multi_meta {
 struct idx {};
 struct res_type {};
 
+template<typename T>
+struct polyfill_allocator : std::allocator<T> {
+        using std::allocator<T>::allocator;
+        template<typename U>
+        struct rebind {
+            using other = polyfill_allocator<U>;
+        };
+        using pointer = T*;
+        using const_pointer = T const*;
+        using reference = T&;
+        using const_reference = T const&;
+};
+
 using boost::multi_index_container;
 using namespace boost::multi_index;
 typedef multi_index_container<
@@ -47,7 +60,8 @@ typedef multi_index_container<
             tag<res_type>, // index nametag
             member<planner_multi_meta, std::string, &planner_multi_meta::resource_type> // index's key
         >
-    >
+    >,
+    polyfill_allocator<planner_multi_meta>
 > multi_container;
 
 class planner_multi {


grondo (approved):
LGTM. @jameshcorbett may want to double check that RPMs can be built with this change before merging the result, especially if you want to release another patch version before Monday.  (but like @trws mentioned, I _think_ I remember checking this before.)

trws:
Excellent point, @jameshcorbett would you be willing to give it another run through to make sure this doesn't cause problems spinning a new RPM? The last thing I want to do is block us being able to push out an update right now.

jameshcorbett:
Working through some issues in the buildfarm, it didn't work out of the box:

```
/builddir/build/BUILD/flux-sched-0.33.1/resource/planner/c++/planner_multi.hpp:96:21:   required from here
/usr/include/boost/detail/allocator_utilities.hpp:125:31: error: no class template named 'rebind' in 'class std::allocator<planner_multi_meta>'
  125 |           rebind<Type>::other other;
      |                               ^~~~~
/usr/include/boost/detail/allocator_utilities.hpp: In instantiation of 'struct boost::detail::allocator::rebinder<std::allocator<planner_multi_meta> >::result<planner_multi_meta>':
/usr/include/boost/detail/allocator_utilities.hpp:133:49:   required from 'struct boost::detail::allocator::compliant_allocator_rebind_to<std::allocator<planner_multi_meta>, planner_multi_meta>'
/usr/include/boost/mpl/eval_if.hpp:60:31:   required from 'struct boost::mpl::eval_if_c<false, 
```

My only option here is to up the `boost` version to whatever is standard on LC right? Which on Ruby is 1.66.

trws:
Try it, I’ll see if I can factor around it. What boost are we currently building with?

---
On April 15, 2024 at 3:50:24 PM PDT, James Corbett ***@***.***> wrote:

Working through some issues in the buildfarm, it didn't work out of the box:

/builddir/build/BUILD/flux-sched-0.33.1/resource/planner/c++/planner_multi.hpp:96:21:   required from here
/usr/include/boost/detail/allocator_utilities.hpp:125:31: error: no class template named 'rebind' in 'class std::allocator<planner_multi_meta>'
  125 |           rebind<Type>::other other;
      |                               ^~~~~
/usr/include/boost/detail/allocator_utilities.hpp: In instantiation of 'struct boost::detail::allocator::rebinder<std::allocator<planner_multi_meta> >::result<planner_multi_meta>':
/usr/include/boost/detail/allocator_utilities.hpp:133:49:   required from 'struct boost::detail::allocator::compliant_allocator_rebind_to<std::allocator<planner_multi_meta>, planner_multi_meta>'
/usr/include/boost/mpl/eval_if.hpp:60:31:   required from 'struct boost::mpl::eval_if_c<false,


My only option here is to up the boost version to whatever is standard on LC right? Which on Ruby is 1.66.

jameshcorbett:
We had `buildrequires >= 1.53.0` but I wonder if we were just picking up the default version of 1.66.0 (assuming I understand how the buildfarm works)? Either way, upping it to `buildrequires >= 1.66.0` hit the same issue.

grondo:
> but I wonder if we were just picking up the default version of 1.66.0

The buildfarm installs buildrequires using yum/dnf, so will probably pick the latest available version already. Keep in mind that a distro usually has ABI guarantees, so it is unlikely the packaged version of libboost would ever get a big version bump if it would break that guarantee (so usually libs stick at a given version for an entire major release lifetime)

trws:
Ok, we have an unexpected wrinkle.  Boost multiindex apparently expects the rebind property to exist on allocators until 1.67, which was deprecated in C++17 and removed in C++20.

Thankfully this is an interface I know reasonably well, and the whole thing is user-customizable, so I made a little wrapper allocator that provides the missing fields to keep the old version of boost happy.  We can remove it if and when we can use a boost that's less behind, but it's more important to keep things working and fix issues right now than muck around with getting new boost versions.

jameshcorbett (approved):
Looks reasonable and it works in the buildfarm! I can also bump the `gcc-toolset` to 13 as we discussed on a call, that seems to work as well.

trws:
Excellent! Ability to format strings with the good version of fmt or std::format will make life much nicer!

---
On April 16, 2024 at 10:38:34 AM PDT, James Corbett ***@***.***> wrote:

@jameshcorbett approved this pull request.

Looks reasonable and it works in the buildfarm! I can also bump the gcc-toolset to 13 as we discussed on a call, that seems to work as well.

PR #1175, opened 2024-04-13
qmanager: split remove to separate pending

trws:
problem: The "remove" method was used for both handling the `free` and `cancel` RPCs. Under some code paths, remove de-allocates resources, so using it for a pending cancellation, even if it seems to be protected by checks, is unnecessarily dangerous.  Given the recent double-booking issue and a lack so far of other theories on how we get there, this is a possible culprit.

solution: Re-factor queue_policy_base_t remove into remove and remove_pending, the latter of which only operates on the pending states and has no code-paths that can result in a resource.cancel RPC being sent.  The remove_pending is used for both the cancel RPC and the part of the free RPC where the code was used before, should be a pure refactor.

### DIFF
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -316,7 +316,7 @@ void qmanager_cb_t::jobmanager_cancel_cb (flux_t *h, const flux_msg_t *msg,
     if ((job = queue->lookup (id)) == nullptr
         || !job->is_pending ())
         return;
-    if (queue->remove (id) < 0) {
+    if (queue->remove_pending (job.get ()) < 0) {
         flux_log_error (h, "%s: remove job (%jd)", __FUNCTION__,
                        static_cast<intmax_t> (id));
         return;
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -101,7 +101,8 @@ class queue_policy_base_impl_t
 {
 public:
     int insert (std::shared_ptr<job_t> job);
-    int remove (flux_jobid_t id);
+    int remove_pending(job_t *job);
+    int remove(flux_jobid_t id);
     const std::shared_ptr<job_t> lookup (flux_jobid_t id);
     bool is_schedulable ();
     void set_schedulability (bool scheduable);
@@ -116,7 +117,7 @@ protected:
     int process_provisional_cancel ();
     int process_provisional_reprio ();
     int insert_pending_job (std::shared_ptr<job_t> &job, bool into_provisional);
-    int erase_pending_job (std::shared_ptr<job_t> &job, bool &found_in_prov);
+    int erase_pending_job (job_t *job, bool &found_in_prov);
     std::shared_ptr<job_t> pending_pop ();
     std::shared_ptr<job_t> alloced_pop ();
     std::shared_ptr<job_t> rejected_pop ();
@@ -269,6 +270,17 @@ public:
      *                       EINVAL: invalid argument.
      */
     int insert (std::shared_ptr<job_t> pending_job);
+
+    /*! Remove a job whose jobid is id from the pending or maybe_pending queues.
+     * If succeeds, it changes the pending queue or resource state. This queue
+     * becomes "schedulable" if pending job queue is not empty: i.e.,
+     * is_schedulable() returns true;
+     *
+     *  \param id        jobid of flux_jobid_t type.
+     *  \return          0 on success; -1 on error.
+     *                       ENOENT: unknown id.
+     */
+    int remove_pending (job_t *job);
 
     /*! Remove a job whose jobid is id from any internal queues
      *  (e.g., pending queue, running queue, and alloced queue.)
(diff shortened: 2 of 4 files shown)


garlick (commented):
I'm not at all sure I understand the qmanager code being changed here.

Is there some way we can add a test case to `t1024-alloc-check.t`?  Maybe such a test could run a sleep job, then use a python one liner to send a `sched.cancel` RPC for the sleep job, then try to run another job and ensure it is not double booked?

trws:
Sure, I can add that in no problem.  I have a test that should do that, but I haven't managed to make that test fail even with the old code. This ensures there's no code path where it could get from `sched.cancel` to `sched-fluxion-resource.cancel`, but I haven't found the race or whatever it is that makes that actually occur. Either way I think this is a test we want, adding shortly.

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -316,7 +316,7 @@ void qmanager_cb_t::jobmanager_cancel_cb (flux_t *h, const flux_msg_t *msg,
     if ((job = queue->lookup (id)) == nullptr
         || !job->is_pending ())
         return;
-    if (queue->remove (id) < 0) {
+    if (queue->remove_pending (job.get()) < 0) {
```
milroy, line 319: Nit:
```suggestion
    if (queue->remove_pending (job.get ()) < 0) {
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -116,7 +117,7 @@ class queue_policy_base_impl_t
     int process_provisional_cancel ();
     int process_provisional_reprio ();
     int insert_pending_job (std::shared_ptr<job_t> &job, bool into_provisional);
-    int erase_pending_job (std::shared_ptr<job_t> &job, bool &found_in_prov);
+    int erase_pending_job(job_t *job, bool &found_in_prov);
```
milroy, line 120: Nit:
```suggestion
    int erase_pending_job (job_t *job, bool &found_in_prov);
```
Apart from the overhead of `shared_ptr`, is there an advantage to using a raw pointer here?

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -270,6 +271,17 @@ class queue_policy_base_t : public detail::queue_policy_base_impl_t,
...
+    /*! Remove a job whose jobid is id from the pending or maybe_pending queues.
+     * If succeeds, it changes the pending queue or resource state. This queue
+     * becomes "schedulable" if pending job queue is not empty: i.e.,
+     * is_schedulable() returns true;
+     *
+     *  \param id        jobid of flux_jobid_t type.
+     *  \return          0 on success; -1 on error.
+     *                       ENOENT: unknown id.
+     */
+    int remove_pending(job_t *job);
```
milroy, line 283: Nit:
```suggestion
    int remove_pending (job_t *job);
```

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -15,6 +15,7 @@
 #include <cerrno>
 #include <algorithm>
 #include "qmanager/policies/base/queue_policy_base.hpp"
+#include "queue_policy_base.hpp"
```
milroy, line 18: Is the double inclusion intentional?

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -593,7 +613,7 @@ int queue_policy_base_impl_t::pending_reprioritize (flux_jobid_t id,
         }
     } else {
         bool found_in_prov = false;
-        if (erase_pending_job (job, found_in_prov) < 0)
+        if (erase_pending_job (job.get(), found_in_prov) < 0)
```
milroy, line 616: Nit:
```suggestion
        if (erase_pending_job (job.get (), found_in_prov) < 0)
```

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -614,7 +634,7 @@ int queue_policy_base_impl_t::process_provisional_cancel ()
         auto job = m_jobs[id];
         if (job->state == job_state_kind_t::PENDING) {
             bool found_in_provisional = false;
-            if (erase_pending_job (job, found_in_provisional) < 0)
+            if (erase_pending_job (job.get(), found_in_provisional) < 0)
```
milroy, line 637: Nit:
```suggestion
            if (erase_pending_job (job.get (), found_in_provisional) < 0)
```

qmanager/policies/base/queue_policy_base_impl.hpp
```diff
@@ -644,7 +664,7 @@ int queue_policy_base_impl_t::process_provisional_reprio ()
...
```
milroy, line 667: Nit:
```suggestion
            if (erase_pending_job (job.get (), found_in_provisional) < 0)
```

milroy (approved):
I found a few nits and have one question, but pending the addition of a test @garlick recommended this looks good to go.

qmanager/policies/base/queue_policy_base_impl.hpp
trws: I have no idea how that happened, will remove.

trws:
Test pushed.

milroy:
@trws I have a nit on the new commit message, but apart from that this is good to go.

t/t1024-alloc-check.t
```diff
@@ -60,9 +60,33 @@ test_expect_success 'some jobs received timeout exception' '
 test_expect_success 'no jobs received alloc-check exception' '
```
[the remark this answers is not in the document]
trws: Done, though that does put it over 50 characters.

PR #1177, opened 2024-04-16
refactor: optmgr to simpler design, one file

vsoch:
Problem: the current optmanager is split into two files and it adds unneeded complexity. There might have been some initial intention for such a design, but it seems overkill for a small class that manages options.
Solution: refactor all functions to just be in one file.

I also added the C++ extensions to be provided by the default for the DevContainer - without it I miss the "customize your environment" option that pops up and then allows me to choose my compiler, etc. to get the nice testing buttons on the side. :red_circle: :green_circle: 

I chat with @trws about this in our last meeting.


trws (approved):
Love it, this looks great!  If you didn't mind doing this, there are a few others that could use the same treatment  `queue_policy_base[_impl]` comes to mind.

vsoch:
I'd love to! This was easy and fun. I really appreciate small tasks like this that will (slowly) get me more familiar with fluxion and the code base. I'm hoping with enough exposure things will start to click at some point.

vsoch:
Do I need to rebase and then do merge when passing?

trws:
I actually don't see any need for a rebase, it's all up to date, and the commit messages look good.  Feel free to set MWP.

vsoch:
If I remember, my researchapps org might not work with the merge train, but we can try again. If it doesn't I can switch over to manbat.

vsoch:
woot!

PR #1178, opened 2024-04-16
queue_base_manager: refactor to remove impl

vsoch:
Problem: the multiple inheritance adds complexity without obvious benefit.
Solution: remove it for the queue_base_manager.hpp (removing *_impl.hpp)

I tried to keep everything in "protected" that I could (that was originally there) until it needed to be used in qmanager, and previously this was done with detail, so I moved those up to public.


vsoch:
Woot!! Ping @trws :green_circle:

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -379,29 +455,44 @@ class queue_policy_base_t : public detail::queue_policy_base_impl_t,
...
+    {
+        m_scheduled = false;
+    }
 
     /*! Implement queue_adapter_base_t's pure virtual method
      *  so that this queue can be adapted for use within high-level
      *  resource API. Return true if the scheduling loop is active.
      */
-    virtual bool is_sched_loop_active ();
+    bool is_sched_loop_active ()
```
trws, line 492: The ones that were virtual need to stay virtual. Looks like there's at least one more, but we need to be careful about that, could make the behavior very strange otherwise.

```suggestion
    virtual bool is_sched_loop_active ()
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -414,29 +505,519 @@ class queue_policy_base_t : public detail::queue_policy_base_impl_t,
...
+    }
+
+
+    /*! Pop the first job from the pending job queue. The popped
+     *  job is completely graduated from the queue policy layer.
+     *
+     *  \return          a shared pointer pointing to a job_t object
+     *                   on success; nullptr when the queue is empty.
+     */
+    std::shared_ptr<job_t> pending_pop () {
```
trws, line 546: I think this and a few of the methods under it were protected in the impl class were they not?

vsoch: yes and I methodically started with them all there, but because they were used in qmanager (via the now removed class) I had to move them out of protected.

vsoch: That was this comment in the description:

> I tried to keep everything in "protected" that I could (that was originally there) until it needed to be used in qmanager, and previously this was done with detail, so I moved those up to public.

trws: Ah, ok.  Thanks for that.

qmanager/policies/base/queue_policy_base.hpp
vsoch: This should be fixed. There was one other function.

trws (approved):
Looks good to me, this one is big enough and moves enough around it would be nice to have another pair of eyes though.  @milroy could you take a glance?

milroy:
Looking now.

qmanager/modules/qmanager.cpp
```diff
@@ -20,7 +20,6 @@ extern "C" {
 #include <sstream>
```
milroy, line 20: Nit: the commit message should be wrapped at 72 characters to conform with RFC 1: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -737,7 +737,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
             }
```
milroy, line 737: Nit: commit message should indicate the component being changed, so something like: "qmanager: erase_pending_job now takes regular pointer"

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -737,7 +737,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
...
         } else {
             bool found_in_prov = false;
-            if (erase_pending_job (job, found_in_prov) < 0)
+            if (erase_pending_job (job.get(), found_in_prov) < 0)
```
milroy, line 740: RFC 7 states the C code "shall separate function names and the opening parenthesis": https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_7.html#c-coding-style-recommendations. Fluxion is almost all C++, but it follows the flux-core convention here.
```suggestion
            if (erase_pending_job (job.get (), found_in_prov) < 0)
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -806,7 +806,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
             auto job = m_jobs[id];
             if (job->state == job_state_kind_t::PENDING) {
                 bool found_in_provisional = false;
-                if (erase_pending_job (job, found_in_provisional) < 0)
+                if (erase_pending_job (job.get(), found_in_provisional) < 0)
```
milroy, line 809: Ditto above.
```suggestion
                if (erase_pending_job (job.get (), found_in_provisional) < 0)
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -836,7 +836,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
...
```
milroy, line 839: Ditto above.
```suggestion
                if (erase_pending_job (job.get (), found_in_provisional) < 0)
```

milroy (approved):
These changes are a step in the right direction for flattening out some of the Fluxion classes!

The changes look good functionally, but there are a few style changes needed before merging.

qmanager/policies/base/queue_policy_base.hpp
vsoch: Fixed.

qmanager/policies/base/queue_policy_base.hpp
vsoch: Fixed.

qmanager/policies/base/queue_policy_base.hpp
vsoch: Went back and fixed both commits (hopefully not borking anything) - I wasn't sure how to do an older one, but if you rebase with `reword` that seems to do the trick!

vsoch:
With two approvals I can set MWP?

vsoch:
:partying_face:

qmanager/modules/qmanager.cpp
trws: What do you use for producing commits @vsoch? Occurs to me I used to run into this _every time_ until I got configs for all my editors to automate it.

qmanager/policies/base/queue_policy_base.hpp
trws: That's exactly the right way to do it, good stuff!  Also lazygit is surprisingly good at stuff like that, convenient for moving bits around interactively until it's good.

trws:
Awesome, thanks again @vsoch!

qmanager/modules/qmanager.cpp
vsoch: For the initial commit I just make it in my terminal, and then if I `--amend` or ``reword` it is typically vim.

vsoch: I can see the line length at the bottom, so it's no issue to just be weary of that from now on.

trws: Eeeeexcelent.  That's the one I have the best stuff for 😈 

There's a decent gitcommit.vim syntax file in any recent vim or neovim, but I'm guessing you have something older maybe, the vim-polyglot plugin will make essentially everything nicer, including auto-setting textwidth to 72 and syntax highlighting the first 50 characters of the subject so you can tell when it gets too long. https://github.com/sheerun/vim-polyglot

If you just want the minimum, I still keep this around in my .vimrc because it does the minimum and at least makes text auto-wrap at 72 chars, and makes it so the `gq` command will wrap at the same:

```vim
au FileType gitcommit setlocal textwidth=72
```

PR #1181, opened 2024-04-18
test: run t10001 under flux

### LINKED ISSUE #1172: python/t10001-resourcegraph.py test fails outside of flux environment
Problem: when running `make check` on my Ubuntu desktop, this test fails for me, unless I run it in flux instance or via the `flux(1)` command front end.

```
      Start 92: python/t10001-resourcegraph.py
92/92 Test #92: python/t10001-resourcegraph.py ........***Failed    0.07 sec
Traceback (most recent call last):
  File "/home/garlick/proj/flux-sched/t/python/t10001-resourcegraph.py", line 22, in <module>
    from fluxion.resourcegraph.V1 import (
  File "/home/garlick/proj/flux-sched/src/python/fluxion/resourcegraph/V1.py", line 13, in <module>
    from flux.idset import IDset
ModuleNotFoundError: No module named 'flux'
```

jameshcorbett:
Problem: t10001-resourcegraph does not run under flux by default, leading to test failures when the flux package cannot be imported.

Add a script for running Python tests under flux and make t10001 use it.

Fixes #1172.

### DIFF
--- a/t/python/subflux.py
+++ b/t/python/subflux.py
@@ -1,0 +1,58 @@
+"""
+Utility for re-running python tests under flux.
+
+Copied from flux-core.
+"""
+
+###############################################################
+# Copyright 2014 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+
+import argparse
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM! You might want to note in the commit message the source of the subflux script. (Just in case some future person wonders)

jameshcorbett:
> LGTM! You might want to note in the commit message the source of the subflux script. (Just in case some future person wonders)

Good idea! Done. I also added a header docstring to the module saying so.

Setting MWP.

PR #1183, opened 2024-04-19
resource: ensure all resources start in DOWN state when some ranks are excluded by configuration

### LINKED ISSUE #1182: fluxion loses track of down nodes
We've seen a couple instances now where Fluxion loses track of one or more down nodes and allocates a down node to a job, which promptly fails, allocates the down node to the next job, which fails, etc.

This mismatch can be seen currently on the fluke cluster with the following script:
```python
import flux
from flux.resource.list import ResourceListRPC

h = flux.Flux()

rpc1 = ResourceListRPC(h, "resource.sched-status", nodeid=0)
rpc2 = ResourceListRPC(h, "sched.resource-status", nodeid=0)

rset = rpc1.get()
fluxion = rpc2.get()

def symmetric_diff(a, b):
    return (a|b) - (a&b)

diff = symmetric_diff(rset.down, fluxion.down)
if diff.ranks:
    print("difference detected between fluxion and core down ranks:")
    print(f"hosts: {diff.nodelist}")
    print(f"ranks: {diff.ranks}")
```

```console
$ flux python ./rcheck.py
difference detected between fluxion and core down ranks:
hosts: fluke103
ranks: 100
```

The host `fluke103` has been 'offline' since Nov  8 2023 (actually I need to double check that exact date, but the node has definitely not been online since the last Flux restart).

Since all resources should be marked down by fluxion until the resource.acquire protocols tells them they are up, this seems to be due to fluxion marking resources up that are not included in the resource.acquire protocol. Of course, there is no way to prove the core resource module did not send this rank in an up idset in a response, so perhaps more investigation needed.

I'm also unsure why the other down nodes did not have the same result.

We could perhaps do some debugging by reloading fluxion to see if the problem corrects itself on fluke. If not then we can enable more debugging, etc.

grondo:
This PR fixes #1182 and adds a test for the condition (verified that the test fails on current master).

This is a critical issue since currently Fluxion could assign down resources to jobs. The jobs then immediately fail and the next job is scheduled, eating through the entire queue.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1144,14 +1144,10 @@ done:
 static int decode_all (std::shared_ptr<resource_ctx_t> &ctx,
                        std::set<int64_t> &ranks)
 {
-    int64_t size = ctx->db->metadata.by_rank.size();
-    
-    for (int64_t rank = 0; rank < size; ++rank) {
-        auto ret = ranks.insert (rank);
-        if (!ret.second) {
-            errno = EEXIST;
-            return -1;
-        }
+    ranks.clear ();
+    for (auto const& kv: ctx->db->metadata.by_rank) {
+        if (kv.first >= 0)
+            ranks.insert (kv.first);
     }
     return 0;
 }
--- a/t/issues/t1182-exclude-with-down-ranks.sh
+++ b/t/issues/t1182-exclude-with-down-ranks.sh
@@ -1,0 +1,69 @@
+#!/bin/bash -e
+#
+#  Ensure Fluxion marks all ranks down even if some ranks are excluded
+#
+
+log() { printf "issue#1182: $@\n" >&2; }
+
+# Need a few ranks for this test, so start a new instance of size=4
+if test "$ISSUE_1182_ACTIVE" != "t"; then
+    export ISSUE_1182_ACTIVE=t
+    log "Re-launching test script under flux-start"
+    exec flux start -s 4 $0
+fi
+
+cat <<'EOF' >rcheck.py
+import sys
+import flux
+from flux.resource.list import ResourceListRPC
+
+h = flux.Flux()
+
+rpc1 = ResourceListRPC(h, "resource.sched-status", nodeid=0)
+rpc2 = ResourceListRPC(h, "sched.resource-status", nodeid=0)
+
+rset = rpc1.get()
+fluxion = rpc2.get()
+
+def symmetric_diff(a, b):
+    return (a|b) - (a&b)
+
+diff = symmetric_diff(rset.down, fluxion.down)
+if diff.ranks:
+    print("difference detected between fluxion and core down ranks:")
+    print(f"hosts: {diff.nodelist}")
+    print(f"ranks: {diff.ranks}")
+    sys.exit(1)
+sys.exit(0)
+EOF
+
+log "Unloading modules..."
+flux module remove sched-simple
+flux module remove resource
+
+# Exclude rank 0
+flux config load <<EOF
+[resource]
+exclude = "0,2"
+EOF
+
+flux module load resource monitor-force-up
+
+# Drain rank 3. Scheduler should only see rank 1 as up
+log "draining rank 3"
+flux resource drain 3
+
+flux resource status
+
+flux module load sched-fluxion-resource
+flux module load sched-fluxion-qmanager
+
+log "comparing fluxion down ranks with flux-core resource module:"
+flux resource list
+FLUX_RESOURCE_LIST_RPC=sched.resource-status flux resource list
+flux python ./rcheck.py
+
+log "reloading sched-simple..."
+flux module remove sched-fluxion-qmanager
+flux module remove sched-fluxion-resource
+flux module load sched-simple


resource/modules/resource_match.cpp
```diff
@@ -1144,13 +1144,13 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
...
 {
-    int64_t size = ctx->db->metadata.by_rank.size();
-    
-    for (int64_t rank = 0; rank < size; ++rank) {
-        auto ret = ranks.insert (rank);
-        if (!ret.second) {
-            errno = EEXIST;
-            return -1;
+    for (auto const& kv: ctx->db->metadata.by_rank) {
+        if (kv.first >= 0 && kv.first < IDSET_INVALID_ID) {
```
garlick, line 1148: So `ctx->db->metadata.by_rank` can include the value IDSET_INVALID_ID (as a terminator?)  That seems pretty strange!

garlick (approved):
This LGTM!

resource/modules/resource_match.cpp
grondo: Uh.. actually I'm not sure. If I didn't add this check then all resources stayed down. It seemed safe to check for a valid id before adding it to the idset. (maybe the invalid value was `-1`)

garlick: Seems like maybe that's what's going on here?

https://github.com/flux-framework/flux-sched/blob/master/resource/modules/resource_match.cpp#L568

grondo: Ok, I made that change and forced a push.

garlick: Sorry, I wasn't suggesting a change.  I was just trying to peel back the layers a bit to understand what is happening?  Well, I guess if the tests pass this is OK though?

Maybe @milroy or @trws ought to have a look?

grondo: Yeah, agreed. (_tap tap. Is this thing on?_)

trws: Looking now.

trws: This looks correct, but makes me wonder how we're ending up with a -1 in that map.  It shouldn't happen.  The other thing, I'd probably just do a `ranks.clear()` before the loop and delete the ret.second check.  Keys can never be duplicates in a map, so there's no way it could double-insert from there, and the only usage of this currently passes in an empty set anyway.

milroy:
Looking now as well.

resource/modules/resource_match.cpp
milroy: Assuming the reader is JGF, `-1` rank could get into `by_rank` via the reader since the default rank value is `-1`: https://github.com/flux-framework/flux-sched/blob/1d628f7f4ebc8674956ce779b8ef9d5f7ba3a5ad/resource/readers/resource_reader_jgf.cpp#L47-L66

If the reader isn't JGF I'm not sure how that happened.

milroy: > ... and delete the ret.second check

We do need some check for the insertion, however. If the insertion failed that means the key already existed, which means there was a duplicate in `by_rank`. That is an indicator of a corrupted graph. In the current implementation that can't happen since the loop is just iterating an integer, but if we're inserting from `by_rank` the check should exist.

milroy: > I'd probably just do a ranks.clear() before the loop and delete the ret.second check.

I agree with this. I was about to say that having the check would be a good guard against duplicates in the graph, but there can't be duplicates in the `by_rank` map, either.

That said, the commit looks good to me.

grondo: Logging the values inserted into the ranks set without the `-1` check I see:
```
$ FLUX_MODULE_PATH_PREPEND=$(pwd)/resource/modules flux start -s 4
Apr 19 18:37:27.413233 sched-fluxion-resource.err[0]: ranks.insert(-1)
Apr 19 18:37:27.413240 sched-fluxion-resource.err[0]: ranks.insert(0)
Apr 19 18:37:27.413241 sched-fluxion-resource.err[0]: ranks.insert(1)
Apr 19 18:37:27.413243 sched-fluxion-resource.err[0]: ranks.insert(2)
Apr 19 18:37:27.413245 sched-fluxion-resource.err[0]: ranks.insert(3)
```

If `-1` is in the ranks set, all resources are marked down in the scheduler without any further errors:
```
f(s=4,d=0) grondo@corona212:~/git/flux-sched$ flux resource list 
     STATE NNODES   NCORES    NGPUS NODELIST
      free      0        0        0 
      down      4      192        0 corona[212,212,212,212]
f(s=4,d=0) grondo@corona212:~/git/flux-sched$ flux resource status
     STATE UP NNODES NODELIST
     avail  ✔      4 corona[212,212,212,212]
```

I guess another thing to fix here is to make this condition a fatal error.

grondo: Well I spent a bit trying to figure out why a `-1` included in the "all" rank set then causes all resources to remain down, but I got nowhere and frankly I don't understand how it could happen.

Anyway, protecting against adding `-1` in "all" fixes this, but we might want to run this down at some point.

grondo: I force pushed a change with the suggestions for `decode_all()`

trws (approved):
Looks great, thanks for this @grondo, and for trying to track down the -1 issue as well. I'll take a look later but this is all good for now.

grondo:
Thanks, I'll set MWP

PR #1184, opened 2024-04-20
reapi c++: add satisfy endpoint

vsoch:
Problem: the c++ bindings do not have satisfy support.
Solution: add satisfy to them.

In practice I was adding this for exposure to the Go bindings, but I do not think it is necessary, because the Go bindings use the C bindings, which already have reapi_cli_match_satisfy. I saw that match_allocate seems to have support to provide the SATISFIABILITY match_op, which is provided to the traverser, so I tried to call that same function. I am opening this PR in case it is interesting or useful. If not,
please close and disregard.

If the PR is desired, here are questions I have:

- I notice the general pattern for functions is to return a return code (0 success, -1 failure) and then take references to variables provided by the user (that are updated). I assume since we don't know what those are coming in as, we should set them to what we want for a default (for example, -1 for satisfied "no").
- I'm not adding Go tests here because (at least I want) the main Go wrappers to be in fluxion-go. I am going to do some PRs there shortly to update the go version.
- That said, how are the bindings tested? I couldn't find the tests in here.
- The reapi_cli seems to support the overhead variable since it calls match_allocate under the hood, which (I think?) sets it. But the module (by way of the rpc) doesn't seem to have that, so the signatures are different. Is that an issue?
- As best as I could tell, the RPC only returns 0 when it is a satisfy "YES"
- It's not clear to me what the "at" variable would be given a satisfy match_op. Reading the docstring, it sounds like a "yes" is >= 0 so I used that, but I'd like to understand why. It's hard to read this code because the variable "at" is really tiny and a substring of other variables and my eyesight is bad in my elderly dinosaur years. :laughing: 

And again, if this is not of interest, please disregard! Going to do some work on fluxion-go next.


vsoch:
I am sorry codecov bot, please forgive my sins :pray:

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -163,6 +163,34 @@ int reapi_cli_t::match_allocate (void *h, match_op_t match_op,
...
+    // This return code says that something wrong happened with 
+    // this function call - it doesn't reflect satisfy status
+    return_code = match_allocate (h, match_op, jobspec, jobid, reserved,
+                                  R, at, ov);
+
+    // The "at" variable reflects satisfy status.
+    // A value of at >= 0 is supposed to mean match_allocate success
+    // (0 is allocated, >0 is timestamp of reservation). Assuming
+    // satisfy could be either of these cases, we check for that here
+    if ((return_code == 0) && (at >= 0)) {
```
vsoch, line 188: I think I might need to check that `errno` thing here instead of the response? I need to read about what it is.

PR #1187, opened 2024-04-26
Update stats output to include data for failed matches

milroy:
Performance testing with Fluxion has revealed the need to save and output data on failed matches. This PR updates the resource module to track and report these stats.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -42,17 +42,37 @@ using namespace Flux::opts_manager;
  ******************************************************************************/
 
 struct match_perf_t {
-    double load;                   /* Graph load time */
-    uint64_t njobs;                /* Total match count */
-    uint64_t njobs_reset;          /* Jobs since reset match count */
+    double load = 0.0;                     /* Graph load time */
     /* Graph uptime in seconds */
-    std::chrono::time_point<std::chrono::system_clock> graph_uptime;
+    std::chrono::time_point<std::chrono::system_clock> graph_uptime
+                                        = std::chrono::system_clock::now ();
     /* Time since stats were last cleared */
-    std::chrono::time_point<std::chrono::system_clock> time_since_reset;
-    double min;                    /* Min match time */
-    double max;                    /* Max match time */
-    double mean;                   /* Mean match time */
-    double M2;                     /* Welford's algorithm */
+    std::chrono::time_point<std::chrono::system_clock> time_since_reset
+                                        = std::chrono::system_clock::now ();
+    struct time_stats {
+        void update_stats (double elapsed) {
+            /* Update using Welford's algorithm */
+            double delta = 0.0;
+            double delta2 = 0.0;
+            njobs++;
+            njobs_reset++;
+            min = (min > elapsed)? elapsed : min;
+            max = (max < elapsed)? elapsed : max;
+            // Welford's online algorithm for variance
+            delta = elapsed - avg;
+            avg += delta / (double)njobs;
+            delta2 = elapsed - avg;
+            M2 += delta * delta2;
+        }
+        uint64_t njobs = 0;                /* Total match count */
+        uint64_t njobs_reset = 0;          /* Jobs since match count reset */
+        double min = std::numeric_limits<double>::max (); /* Min match time */
+        double max = 0.0;                  /* Max match time */
+        double avg = 0.0;                  /* Average match time */
+        double M2 = 0.0;                   /* Welford's algorithm */
+    };
+    time_stats succeeded;
+    time_stats failed;
 };
 
 class msg_wrap_t {
@@ -353,15 +373,6 @@ static std::shared_ptr<resource_ctx_t> getctx (flux_t *h)
         ctx->h = h;
         ctx->handlers = NULL;
         set_default_args (ctx);
-        ctx->perf.load = 0.0f;
-        ctx->perf.njobs = 0;
-        ctx->perf.njobs_reset = 0;
-        ctx->perf.min = std::numeric_limits<double>::max ();
-        ctx->perf.max = 0.0f;
-        ctx->perf.mean = 0.0f;
-        ctx->perf.M2 = 0.0f;
-        ctx->perf.graph_uptime = std::chrono::system_clock::now ();
-        ctx->perf.time_since_reset = std::chrono::system_clock::now ();
         ctx->matcher = nullptr; /* Cannot be allocated at this point */
         ctx->fgraph = nullptr;  /* Cannot be allocated at this point */
         ctx->writers = nullptr; /* Cannot be allocated at this point */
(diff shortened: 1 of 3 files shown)


resource/modules/resource_match.cpp
```diff
@@ -2203,19 +2246,26 @@ static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+                            "min-match", min,
+                            "max-match", ctx->perf.max,
+                            "avg-match", mean,
+                            "match-variance", variance,
+                            "njobs-failed", ctx->perf.njobs_failed,
+                            "njobs-reset-failed", ctx->perf.njobs_reset_failed,
+                            "min-match-failed", min_failed,
+                            "max-match-failed", ctx->perf.max_failed,
+                            "avg-match-failed", mean_failed,
+                            "match-variance-failed", variance_failed) < 0) {
```
garlick, line 2268: Quck thought  - it might make sense to break the flat list up into some objects like
```
match.success.min
match.success.avg
match.success.max
match.failed.min
match.failed.avg
match.failed.max
```
etc or wherever grouping makes sense.   That would make it a little easier to access specific data with queries e.g.
```
flux module stats sched-fluxion-resource | jq .match.failed
```

milroy: Good suggestion! Let me know if the implementation makes sense.

resource/modules/resource_match.cpp
```diff
@@ -2203,19 +2249,38 @@ static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+                            "njobs-reset-failed", ctx->perf.njobs_reset_failed,
+                            "min-match-failed", min_failed,
+                            "max-match-failed", ctx->perf.max_failed,
+                            "avg-match-failed", mean_failed,
+                            "match-variance-failed", variance_failed))) {
+        errno = ENOMEM;
+        goto error;
+    }
+
+    if (flux_respond_pack (h, msg, "{s:I s:I s:o s:f s:I s:I s:{s:o s:o}}",
```
garlick, line 2274: Up to you but it might be nice to have the min/max/avg values in sub-objects.

Also it looks like the sub-objects could be leaked on error.

Packing objects with little o (which "steals" the reference) complicates object reference count management on error because the original reference is not always restored when pack fails (at least that was found to be the case in some jansson versions).  So a suggestion is to either pack the response in one go or use big O to pack the sub objects and decrement their reference counts on both success and failure paths.

milroy: @garlick, I think I implemented your first suggestion as you intended. Let me know what you think and I'll discard or squash the fixup commits.

resource/modules/resource_match.cpp
```diff
@@ -53,6 +53,13 @@ struct match_perf_t {
     double max;                    /* Max match time */
     double mean;                   /* Mean match time */
     double M2;                     /* Welford's algorithm */
+    // Track failed matches
+    uint64_t njobs_failed;
+    uint64_t njobs_reset_failed;
+    double min_failed;
+    double max_failed;
+    double mean_failed;
+    double M2_failed;
```
trws, line 62: I like this, but reducing the duplication might be nice, maybe:

```suggestion
struct time_stats {
    void add_time(double t) {
        double delta = 0.0f;
        double delta2 = 0.2f;
        njobs++;
        njobs_reset++;
        min = (min > elapse)? elapse : min;
        max = (max < elapse)? elapse : max;
        // Welford's online algorithm for variance
        delta = elapse - mean;
        mean += delta / (double)njobs;
        delta2 = elapse - mean;
        M2 += delta * delta2;
    }
    uint64_t njobs = 0;                /* Total match count */
    uint64_t njobs_reset = 0;          /* Jobs since reset match count */
    double min = std::numeric_limits<double>::max ();                    /* Min match time */
    double max = 0;                    /* Max match time */
    double mean = 0;                   /* Mean match time */
    double M2 = 0;                     /* Welford's algorithm */
};
time_stats succeeded;
time_stats failed;
```

src/cmd/flux-ion-resource.py
```diff
@@ -224,12 +224,62 @@ def stat_action(_):
...
     print("Time Since Stats Reset: ", resp["time-since-reset"], "Secs")
-    print("Num. of Total Jobs Matched: ", resp["njobs"])
-    print("Num. of Jobs Matched Since Reset: ", resp["njobs-reset"])
-    print("Min. Match Time: ", resp["min-match"], "Secs")
-    print("Max. Match Time: ", resp["max-match"], "Secs")
-    print("Avg. Match Time: ", resp["avg-match"], "Secs")
-    print("Match Variance: ", resp["match-variance"], "Secs^2")
+    print(
+        "Num. of Total Jobs Successfully Matched: ",
+        json.dumps(resp["match"]["succeeded"]["njobs"]),
```
trws, line 229: Why the dumps usage here? Is it not a number at this point?

resource/modules/resource_match.cpp
```diff
@@ -2203,19 +2246,45 @@ static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
...
-                                   "graph-uptime", graph_uptime_s,
-                                   "time-since-reset", time_since_reset_s,
-                                   "njobs", ctx->perf.njobs,
-                                   "njobs-reset", ctx->perf.njobs_reset,
-                                   "min-match", min,
-                                   "max-match", ctx->perf.max,
-                                   "avg-match", mean,
-                                   "match-variance", variance) < 0) {
+    
+    if (flux_respond_pack (h, msg, "{s:I s:I s:o s:f s:I s:I"
```
trws, line 2250: I get why the indentation, and why this is like this, but my goodness this is complex.  I'm guessing you had it making two sub-objects earlier and backed away based on @garlick's comment. This is clearly right, so I'm ok with taking it, but breaking deep ones like this down would be nice.

milroy: Ok, I updated this with sub-objects. @trws  and @garlick let me know if this implementation looks better.

src/cmd/flux-ion-resource.py
milroy: Yes, it's a number. The `json.dumps` is a mistake.

resource/modules/resource_match.cpp
milroy: Good suggestion. Implemented.

milroy:
Thanks for the second round of feedback!

resource/modules/resource_match.cpp
```diff
@@ -2236,12 +2288,17 @@ static void stat_clear_cb (flux_t *h, flux_msg_handler_t *w,
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
     // Clear the jobs-related stats and reset time
-    ctx->perf.njobs_reset = 0;
-    ctx->perf.min = std::numeric_limits<double>::max ();
-    ctx->perf.max = 0.0f;
-    ctx->perf.mean = 0.0f;
-    ctx->perf.M2 = 0.0f;
-    ctx->perf.time_since_reset = std::chrono::system_clock::now ();
+    ctx->perf.succeeded.njobs_reset = 0;
```
trws, line 2291: All of these init lines are now redundant, they're set by the initializers in the struct. They don't hurt anything necessarily but they aren't necessary.

resource/modules/resource_match.cpp
```diff
@@ -2177,53 +2189,93 @@ static void stat_request_cb (flux_t *h, flux_msg_handler_t *w,
...
+                                "njobs-reset", ctx->perf.succeeded.njobs_reset,
+                                "stats",
+                                    "min", min,
+                                    "max", ctx->perf.succeeded.max,
+                                    "avg", mean,
+                                    "variance", variance))) {
+        errno = ENOMEM;
+        goto error_free;
+    }
+    if (!(match_failed = json_pack ("{s:I s:I s:{s:f s:f s:f s:f}}",
```
trws, line 2240: This is a lot easier for me to follow, good stuff.

trws (approved):
I'm happy. Thanks for the updates @milroy!

resource/modules/resource_match.cpp
milroy: Fixed!

trws:
@garlick, would you like one more pass here or are we ready for MWP?

garlick:
Go ahead.  Nice work!

trws:
Ready when you are then @milroy!

milroy:
Thanks for the reviews! 

I added a tiny commit to test populating the failed stats path. Once that succeeds I'll set MWP.

PR #1188, opened 2024-04-27
qmanager: defer jobs with unmatchable constraints

trws:
This is the in-progress PR for the constraint job blocking problem.  Full description below, but we're still lacking two things I really want to have:

1. [ ] A test that will surface this if it comes back, other than looking at timing, this is _very_ hard to make an integration test for.  A unit test would be easier, trying to come up with something.
2. [ ] A fix for the stats collection logic that includes failed matches, both the number of them and times of them since that being missing is part of what helped this hide so well.

problem: Jobs with constraints that can't be matched because nodes are
down or drained are currently considered every time we enter the
scheduling loop.  If they reach the head of the queue, which is likely
because we currently only configure one sched queue, they get
re-considered over and over despite the fact they can't run, which
greatly slows down scheduling, and can cause severe blocking observed up
to 20 seconds of delay for a single submission.  This change also
exposed a bug with the duration calculation for duration=0.0 jobs, which
used to be set to the full duration of the graph rather than the
remaining duration of the graph.

solution: Add a new "m_blocked" member to the qmanager base which holds
jobs which return EBUSY from an alloc_orelse_reserve. This state can
only happen when the job is blocked by a constraint requiring a node in
an unusable state.  The jobs in m_blocked are moved into m_pending
(ready to be considered) by the notify callback.  Currently they are
moved regardless of what status changes occurred (a node being drained
causes them to move) but it's a relatively small cost to move them back
afterward and simplifies the logic considerably. The duration for 0.0
duration jobs is now set to the remaining time rather than the total
time during meta build time.


trws:
I see the coverage failure, and will see what I can do about that, but that's going to be a few days.  Looks like the uploader we're using is deprecated, and the actual error is because the uploader is using too old a version of gcov compared to the version of gcc on bookworm, but it only matters on qmanager for some reason. 🤷

grondo:
We switched to `codecov-actionW@v4` in flux-core awihle ago. I can attempt to do an update here as well (along with the other deprecated actions) then we can rebase this one.

trws:
That would be awesome if you could @grondo!

t/python/t10001-resourcegraph.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S flux python
```
grondo, line 1: TIL! I didn't know about the `env -S` option. That's handy.

Possibly fixes #1172?

grondo: Hm, though I tested on alpine and the `-S` seems to be GNU coreutils only option (and BSD since the option was copied from that version)

grondo (approved):
This LGTM,  I'm not that familiar with the code in question, but these look like quite clever fixes!

Not sure if you came up with any way to test the deferral of unreservable jobs due to other jobs with no timelimit though. Would it be enough to submit a job without a timelimit and using all resources, submit another job, then cancel the first job and ensure the 2nd is started? (I'm sure I'm oversimplifying, apologies if so).

resource/traversers/dfu_impl.hpp
```diff
@@ -86,10 +86,13 @@ struct jobmeta_t {
...
-        if (jobspec.attributes.system.duration == 0.0f)
-            duration = g_duration;
-        else
+        if (jobspec.attributes.system.duration == 0.0f) {
+            // set to the amount of time left, full graph duration can cause EINVAL from match
+            duration = (graph_duration.graph_end - std::chrono::system_clock::from_time_t (t))
+                           .count();
+        } else {
             duration = (int64_t)jobspec.attributes.system.duration;
+        }
```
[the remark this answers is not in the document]
trws: It does, it actually wasn't a problem until the other changes in this same commit because it would be corrected automatically by a check after the at-time validity check to always fit in the remaining time.  The problem is that once the other changes were made, it became possible to actually have it fail inside the match unless this was fixed here.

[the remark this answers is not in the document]
grondo: Great, thanks for the explanation!

t/python/t10001-resourcegraph.py
trws: Ah, darn. I knew it worked on GNU and BSD and didn't think about busybox.  I could have cmake run the python ones with `flux python` pretty easily but was hoping to get it directly in the shebang.

trws:
I think I have a test to use, I just have to incorporate it into sharness.  Thankfully the unreservable aspect is an easy test, submit two jobs requiring the same resource with no time limit first one that runs at least until the second is considered, wait for both.  If they both succeed it's all good.

The performance regression case I'm not sure how to test in a way that's reliable.  Maybe with the updates @milroy is adding for stats, that way we could watch for number of failed matches and see if it's repeatedly trying to schedule when it can't?  We may have to revisit that one.

t/python/t10001-resourcegraph.py
grondo: Well tests passed, so maybe it is fine ? 🤷

trws: It should only come up when invoking directly or in specific situations I guess... I could grab sbang to let us do this as an alternative.

trws:
Ok, added the test I was using locally basically as-is.  I didn't see anywhere that it clearly fit, but if anyone knows a spot I can merge it into some other file.

resource/traversers/dfu.cpp
```diff
@@ -310,6 +310,11 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
         detail::dfu_impl_t::update ();
     } else if ( (rc = schedule (jobspec, meta, x, op, root, dfv)) ==  0) {
         *at = meta.at;
+        if (errno == ENOENT && *at == graph_end) {
+            // no schedulable point found even at the end of the time, return EBUSY
+            errno = EBUSY;
+            return -1;
+        }
```
milroy, line 317: > This seems to only happen when `schedule` returns -1, an errno of ENOENT and the `at` value equal to the end of the graph.

I'm a bit confused about what's happening here. If the job constraint is still unsatisfied, I'd expect `schedule` to return `-1` with `errno != ENODEV` (I think `ENOENT`, but it's hard to trace), signifying satisfiable.

For this check: ```if (errno == ENOENT && *at == graph_end) {```, I think `schedule` should return `0` and an `at` reservation time (e.g., `at` value equal to the end of the graph) for the jobs with a duration set to the remaining graph lifetime. Where is `errno` set to `ENOENT` in this case?

If `errno != ENOENT && *at >= graph_end` should that return `EINVAL`?

milroy (commented):
I found a few nits which should be quick to fix.

I do have some questions about the behavior of some of the checks in this PR. I don't think they warrant changes, but I'd like to understand the behavior a bit better before the PR gets merged.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -85,8 +85,15 @@ queue_policy_bf_base_t<reapi_type>::allocate_orelse_reserve (void *h,
...
-            // The semantics of our backfill policies is to skip this job
-            iter++;
+            // The semantics of our backfill policies is to skip this job,
+            // add it to the blocked list, and re-consider when resource
+            // status changes
+
+            // copy and advance iterator before extract invalidates it
+            auto next = iter;
+            ++next;
+            m_blocked.insert(m_pending.extract(iter));
```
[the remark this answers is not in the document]
trws: Yup, that was the original goal actually because it avoids reconsidering jobs that can't run due to DOWN resources until at least one resource state changes. Plus yeah extract is awesome.  Honestly `merge` is even better, so nice to have an efficient way to move nodes around.

resource/traversers/dfu_impl.hpp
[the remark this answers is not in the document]
trws: Ok, I misspoke here.  The failure happens [here](https://github.com/flux-framework/flux-sched/pull/1188/files#diff-d5108dd91cff331d5c2d234a5c449051c915fb664d96ca93bed689b05d21a03bR318-R320). There are two things going on at the same time that made me decide this was necessary, which is part of why it took so long to come to this as a solution.  When a job with an open-ended runtime gets reserved, it now reserves until exactly the end of the graph.  In principle that's what happened before with the adjusted duration, just below that, maybe with the other changes this is no longer necessary, will test that this morning.  The other thing is a job that _can_ run but doesn't fit in the graph. I just re-traced it and the result is not making me happy, will probably re-factor this.  The reason this specific condition happens, is that the `at` time will be set to the end of the graph because of the previous job taking all time until that point, and the ENOENT  happens because of an unplanned path for errno to get set, it's set by the `planner_multi_avail_time_next` in the increment portion of the for loop.  Since the select actually succeeded before that, rc is now zero and ENOENT is set, so the block after that leaves it alone and we have a success with at set to the end of the graph and ENOENT as the errno.

This is spaghetti errno handling, but as far as I can tell it's reliable.

I'm looking into whether the duration change is actually necessary, that might actually be an ordering issue and we can leave it.

trws:
Ok, I think this is all cleaned up.  Here are the highlights:

1. @milroy was completely right, there was no need for the extra `==ENOENT` check or the change to duration logic after the other changes, so I removed those.
2. rather than trying to fix up the python script shebang, I added an invocation of `flux env` around the test script so that we always get the flux-side python path added on
3. formatting nits applied, planning to come back to this and get clang-format and clang-tidy working on this repo so I stop missing things like that
4. commit adding the test now actually _runs_ the test (slightly important :facepalm:)

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -472,7 +472,10 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
         default:
             break;
         }
-
+        // blocked jobs must be reconsidered after a job completes
+        // this covers cases where jobs that couldn't run because of an
+        // existing job's reservation can when it completes early
+        reconsider_blocked_jobs();
```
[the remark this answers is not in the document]
milroy: Looks like you missed this one.

resource/traversers/dfu.cpp
```diff
@@ -310,6 +310,11 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
         detail::dfu_impl_t::update ();
     } else if ( (rc = schedule (jobspec, meta, x, op, root, dfv)) ==  0) {
         *at = meta.at;
+        if (*at == graph_end) {
+            // no schedulable point found even at the end of the time, return EBUSY
+            errno = EBUSY;
+            return -1;
+        }
         if (*at < 0 or *at >= graph_end) {
```
milroy, line 318: I don't think this really matters, but `at == graph_end` satisfies both the `if` condition above and this one.
```suggestion
        if (*at < 0 or *at > graph_end) {
```

milroy:
This looks cleaner and easier to understand; thanks! I also like that you renamed the `ov` variable. Fluxion has too many inscrutable variables.

I'm wondering if there is a way to include a test for jobs with unsatisfiable constraints being reconsidered many times. You could submit a few jobs that `require` a down node and then undrain the node. After the stats update PR #1187 gets merged you could then check for the number of failed matches.

milroy:
> In fact I think my existing test would work, it just needs some extra jobs added and a stats check.

The stats PR got merged. I agree that just adding some extra jobs should be sufficient to test this.

t/t4013-unreservable.sh
```diff
@@ -0,0 +1,72 @@
...
+
+test_expect_success 'ensure all three succeeded' '
+  flux job wait -av
+'
+test_expect_success 'unload fluxion' '
+	flux module remove sched-fluxion-qmanager &&
+	flux module remove sched-fluxion-resource &&
+	flux module load sched-simple
+'
+test_done
```
jameshcorbett, line 72: No newline at end of file? Also I think the `fixes #1172` in the commit message is no longer correct?

trws: It is still correct, solution just moved to maybe-installtest

jameshcorbett: Oh ok, obviously trivial but I believe the `fixes` message is in the wrong commit (should be on https://github.com/flux-framework/flux-sched/pull/1188/commits/d42210034a20356dd7052e1caed90dd5bac9b326?).

trws: Ah, good point, fixed.

trws:
Ok, the test has been extended.  We'll have to see how well it holds. As far as I understand the system, it should reliably hit 10, but it's possible it's partially timing dependent. If we find that it's unreliable we may want to adjust the test to <= something.

trws:
@milroy, any chance I could convince you to take a (hopefully) last pass over here? I think this is about where we need it.

milroy:
@trws: yes, looking now.

CMakePresets.json
```diff
@@ -6,8 +6,6 @@
     "patch": 0
```
milroy, line 6: The commit message should be wrapped at 72 characters.

CMakePresets.json
```diff
@@ -14,7 +14,9 @@
       "description": "Default build using Ninja generator",
```
milroy, line 14: The commit message should start with a problem: statement.

milroy (approved):
A few commit messages need to be cleaned up, but once that's done I think this PR is ready to merge.

PR #1190, opened 2024-04-29
docs/cmake:stop constantly rebuilding manpages

trws:
problem: manpages rebuild on every invocation of the build

solution: remove the accidental ALL from the add_custom_command for manpages, that was causing cmake to constantly rebuild them trying to create a file named "ALL"


grondo (approved):
LGTM!

PR #1191, opened 2024-04-29
ci: update deprecated github actions

grondo:
This PR attempts to update all deprecated actions in `workflow/main.yml`. 

As part of the update, the codecov uploader is updated from the bash version to the most recent action uploader. Be aware that this requires that `CODECOV_TOKEN` be set in the environment of the action (where I don't think the bash uploader did). The main repo gets this from the organization secrets, but if you have a personal fork, you'll have to set the `CODECOV_TOKEN` secret to continue getting codecov uploads to your user org in codecov.

We'll see how this works in a PR here...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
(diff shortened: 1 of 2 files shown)


grondo:
There's a big drop in coverage here in [qmanager/policies/base/queue_policy_base.hpp](https://app.codecov.io/gh/flux-framework/flux-sched/pull/1191/blob/qmanager/policies/base/queue_policy_base.hpp) and [src/common/liboptmgr/optmgr.hpp](https://app.codecov.io/gh/flux-framework/flux-sched/pull/1191/blob/src/common/liboptmgr/optmgr.hpp) and I'm not sure the cause, but it should be looked into.

trws:
That is probably expected, because those files now include all the former contents of the `<name>_impl.hpp` files, that no longer exist.  If the coverage is lower than _those_ files, then we have a serious problem, but if it's not then it should just be that refactor showing up.

trws (approved):
This looks great, thanks so much @grondo!

grondo:
Thanks, I'll set MWP and we can rebase the other pending PRs.

PR #1192, opened 2024-04-30
[WIP] policy: refactor of policy options and matching

wihobbs:
Problem: the current matching functions for config-provided policies don't take advantage of some C++20 features, are a bit confusing, and have no unit tests.

Refactor and add unit testing.


resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -20,53 +20,48 @@ extern "C" {
...
             ptr->add_score_factor (std::string ("node"), 1, 10000);
-            if (policy == HIGH_NODEX_FIRST)
+            if (policy == "HIGH_NODEX_FIRST")
                ptr->add_exclusive_resource_type ("node");
             matcher = ptr;
-        } else if (policy == LOCALITY_AWARE) {
+        } else if (policy == "LOCALITY_AWARE") {
             matcher = std::make_shared<greater_interval_first_t> ();
-        } else if (policy == VAR_AWARE) {
+        } else if (policy == "VAR_AWARE") {
```
jameshcorbett, line 64: I think it's usually discouraged to have lots of string literals lying around. The more times you type out the string, the greater the chance that you'll mistype it once and it can sometimes be pretty annoying to debug. If you refer to the string using a variable, at least the compiler will tell you. So in this switchyard I would probably go for something like `if policy == policies["firstnodex"]` rather than `if policy == "FIRST_NODEX_MATCH"`.

wihobbs:
Oh, oops, I had forgotten this was open as a PR. Ignore that push, that was just to sync across machines.

PR #1195, opened 2024-05-03
Repo cleanup

trws:
Now that we've had the cmake build stable for a while, it's time to remove the autotools build. With that gone, reorganization becomes feasible, so the resource and qmanager directories can now finally live under src.


garlick:
Just a suggestion but maybe split off the repo reorg to a separate PR since the automake removal is an obvious immediate yes and the reorg conflicts with all(?) pending prs and might require more discussion.

trws:
Sure, no problem.

PR #1196, opened 2024-05-03
autoconf removal

trws:
As suggested by @garlick over in #1195, splitting out the autotools removal since that's a much easier unit if it doesn't have all the file moves mixed in.


trws:
The extra commit is here because this is where the problem surfaced. I'm not sure _why_ it surfaced here, but the cgo build for the go test was looking for libczmq, which is apparently no longer found in the focal image for some reason.

grondo:
flux-core dropped the czmq dependency awhile ago.

trws:
I saw that, it's extremely odd to me that this just happened now... Maybe it's just when the containers got poked? Not sure.

chu11 (commented):
a little late to the party here, a few things

- remove autogen.sh
- remove INSTALL, since that's the GNU autoconf boilerplate one
- update README.md with new build instructions (it still mentions autogen.sh, ./configure)
- Update: remove configure? rename configure-cmake to configure?  

hopefully not something else I'm forgetting

trws:
Ok, I wasn't planning to rework docker-run-tests.sh for this, so I'll do whatever of that I can without breaking it, or only small tweaks.  The `./configure` workflow actually works, that's part of why that stuff is there, though yeah it's not necessary.

trws:
Turns out autogen being missing just pops a warning in `docker-run-checks.sh` AFAICT, so assuming this passes that should be cleaned up.

trws:
Anyone willing to sign off here?

grondo (approved):
This LGTM!

PR #1197, opened 2024-05-04
docker-run-checks: ensure we match user's home

trws:
problem: docker-run-checks.sh currently assumes that mounting $HOME to /home/$USER will match the outer system. Sometimes that is not the case. Example, my lima-managed vm uses /home/$USER.linux as the home directory for the vm user, which causes strange errors because of the mismatch.

solution: map $HOME to $HOME


trws:
Assuming everything passes here, would be nice to get this one matching flux-core.

grondo (approved):
:+1:

PR #1199, opened 2024-05-07
cmake: look for ver file in source directory

trws:
problem: cmake was looking for the version file in the build directory rather than the source directory, causing it not to be found in some common configurations.

solution: have it look in the source directory instead


grondo (approved):
LGTM! Just to double check though, `flux-sched.ver` is only written out by `make dist` so it won't "sometimes" be in the build dir, correct?

trws:
It will not, the `make dist` command is the one (hopefully only one) command that modifies the source directory.  Mainly because it uses archive, and grabs the flux-sched.ver from the root.  Thinking back on that now, I really am not fond of that... I'm going to add a command to not write into the source directory, that's just not kosher.

trws:
I almost merged this, but to my utter shock I think I just ran into a bug in `git` of all things... revising.

PR #1203, opened 2024-05-13
cmake: ensure we install libraries to our prefix

trws:
problem: in some cases we used flux-core's libdir as ours, which works great when both are in the same prefix but causes big problems when they do not

solution: always install to our own prefix

Related to #1202


grondo:
I think the original change was to fix #1092. I fear I've forgotten the context of that one though.

trws:
Thanks for that @grondo, I'm digging a bit here. It looks like that's where the lines got added, but at some point they ended up _outside the conditional_ which is what was really causing the problem.  I'm trying to figure out why, will come back here with more.

trws:
Found it, this PR to remove the forcing of `/usr` in the wrapper script also moved the forcing of libdir out of the conditional: #1100 

Now that the cmake system is using the GNUInstallDirs package to place things, it *should* match everything anyway, and the jammy installtest seems to work.  That said, if we want to make sure that a prefix-matching install always puts things in flux-core's mod directory I can put those lines back and move them back into the conditional.

grondo:
> Now that the cmake system is using the GNUInstallDirs package to place things, it _should_ match everything anyway,

We could test that, though it would be safer to choose the same module path as the targeted flux-core when the prefixes match, unless otherwise told so. I guess we'd find out quickly since I think at least one of the container broke with the other bug.

trws:
Any remaining concerns here?

grondo (approved):
This seems like a good approach to me!

PR #1205, opened 2024-05-15
Add jobid and schedule iteration count to match stats; make perf a global variable

milroy:
Match performance variability observed on a large cluster has pointed to the need for storing and reporting jobids corresponding to the max succeeded and failed math times. This WIP also adds the capability to report the number of `select` iterations performed for the jobid corresponding to the max match times.

This PR is WIP as the current implementation for passing the iteration count may not be desirable.

### DIFF
--- a/resource/CMakeLists.txt
+++ b/resource/CMakeLists.txt
@@ -16,6 +16,7 @@ set(RESOURCE_HEADERS
     schema/sched_data.hpp
     schema/resource_base.hpp
     schema/resource_data.hpp
+    schema/perf_data.hpp
     schema/color.hpp
     schema/ephemeral.hpp
     traversers/dfu.hpp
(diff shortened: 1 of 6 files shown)


milroy:
The current state of this PR puts the iteration count into jobspec to avoid altering the traverser `run` function signature or return type. This works but feels hackish. @trws do you have ideas for a cleaner implementation, or should I change `run`?

trws:
I'll have to look over the options. Maybe we should have the traverser interface populate a stats object? Not sure. It seems like something we might want anyway.

trws:
Honestly, I'm seriously considering saying we should just have a global (or class static maybe) instance that we use for stats.  It's global context.  Much as I still feel the pangs of the classic "global state is evil" this is a case where even if we go multi-threaded, I think I would still want it to be global and just add some atomics.

milroy:
@trws this is ready for a first-round review. Let me know if you want me to refactor the `match_perf_t` to use a factory to create stats variables.

trws (approved):
Works for me.  I think we'll want to have a self-enumerating object for additional stats at some point, meaning something where we can add a new counter or lime point or similar just by name and have it show up, but we don't need to have it right now and this gets us some things we really do _need_.

trws:
@milroy ready for mwp?

milroy:
Thanks for the review! Adding MWP now.

PR #1206, opened 2024-05-16
avoid copying of std::map in out_edges

trws:
This is stacked on top of #1203, and is a performance enhancement.  For large numbers of constraint jobs especially out_edges was becoming a bottleneck.  Using this test script I saw a drop from ~7 minutes 30 seconds to 90 seconds on my laptop and a very significant shift in the weights in the perf output.

<details>
<summary> Test script </summary>

```bash
#!/usr/bin/env bash
#

test_description='
'

. `dirname $0`/sharness.sh

export TEST_UNDER_FLUX_QUORUM=1
export TEST_UNDER_FLUX_START_MODE=leader
rpc() {
  flux python -c \
    "import flux, json; print(flux.Flux().rpc(\"$1\").get_str())"
}

test_under_flux 16384 system

test_expect_success 'unload sched-simple' '
	flux module remove -f sched-simple
'

test_expect_success 'update configuration' '
	flux config load <<-'EOF'
	[[resource.config]]
	hosts = "fake[0-16383]"
	cores = "0-63"
	gpus = "0-3"

	[[resource.config]]
	hosts = "fake[0-9999]"
	properties = ["compute"]

	[[resource.config]]
	hosts = "fake[10000-16000]"
	properties = ["test"]
	
	[[resource.config]]
	hosts = "fake[16001-16383]"
	properties = ["debug"]

	[sched-fluxion-qmanager]
	queue-policy = "easy"

	[sched-fluxion-resource]
	match-policy = "firstnodex"
	prune-filters = "ALL:core,ALL:gpu,cluster:node,rack:node"
	match-format = "rv1_nosched"
	EOF
'

test_expect_success 'reload resource with monitor-force-up' '
	flux module reload -f resource noverify monitor-force-up
'
test_expect_success 'drain a few nodes' '
	flux resource drain 1-1000 test with drained nodes
'
test_expect_success 'load fluxion modules' '
	flux module load sched-fluxion-resource &&
	flux module load sched-fluxion-qmanager
'
test_expect_success 'wait for fluxion to be ready' '
  time flux python -c \
    "import flux, json; print(flux.Flux().rpc(\"sched.resource-status\").get_str())"
'
test_expect_success 'create a set of 10 inactive jobs' '
	flux submit --cc=1-1000 --quiet \
	    -N 1 --exclusive \
		--requires="host:fake[1005]" \
		--progress --jps \
		--setattr=exec.test.run_duration=0.01s \
		hostname
'

test_expect_success 'create a set of 2 running jobs' '
[...]

PR #1207, opened 2024-05-17
Test repair

trws:
Found a couple of issues with our build and test setup, one is just we're using too old a compiler for the standard version we're targeting, the other is a mistake in the test setup in cmake.

I should mention, this fixes the CI failure over in #1206


grondo (approved):
LGTM

PR #1209, opened 2024-05-20
qmanager: search blocked jobs during pending removal

### LINKED ISSUE #1208: alloc requests are leaked for canceled jobs with queue-policy = "easy"
Problem: flux hangs at shutdown on el cap as noted in:
- flux-framework/flux-core#5964
- flux-framework/flux-core#5990

It appears that fluxion (when configured with queue_policy=easy)  is not responding when unfulfilled alloc requests are canceled. This demonstrates in a standalone instance:

```
$ cat fluxion.toml
[sched-fluxion-qmanager]
queue-policy = "easy"
$ flux start -s2 -o,--conf=fluxion.toml
$ flux resource drain 0
$ flux submit -N2 hostname
ƒ2VGsWxto
$ flux cancel $(flux job last)
May 20 09:48:27.112761 sched-fluxion-qmanager.err[0]: jobmanager_cancel_cb: remove job (3284324581376): No such file or directory
$ flux jobs
       JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
$ flux queue status -v
Job submission is enabled
Scheduling is started
0 alloc requests queued
1 alloc requests pending to scheduler
0 running jobs
$ exit
<hang>
```

trws:
problem: if a job is pending and moved into the blocked state, it isn't removed by cancel

solution: if the job is not found in pending, try to remove it from blocked

As a note, I remember writing this logic at some point in the previous PR that added m_blocked.  I don't know when it got lost, but probably when a few other similar searches that we actually don't want got removed.  The job is now cancelled appropriately in this case at least.

fixes #1208

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -894,18 +894,25 @@ protected:
         s = m_pending.erase ({static_cast<double> (job->priority),
                               static_cast<double> (job->t_submit),
                               static_cast<double> (job->t_stamps.pending_ts)});
-        if (s == 0) {
-            // job must be in m_pending_provisional in this case
-            s = m_pending_provisional.erase (
-                              {static_cast<double> (job->priority),
-                               static_cast<double> (job->t_submit),
-                               static_cast<double> (job->t_stamps.pending_ts)});
-            if (s == 0) {
-                errno = ENOENT;
-                return -1;
-            }
-            found_in_prov = true;
+        if (s == 1) {
+            return 0;
         }
+        s = m_blocked.erase ({static_cast<double> (job->priority),
+                              static_cast<double> (job->t_submit),
+                              static_cast<double> (job->t_stamps.pending_ts)});
+        if (s == 1) {
+            return 0;
+        }
+        // job must be in m_pending_provisional in this case
+        s = m_pending_provisional.erase (
+                {static_cast<double> (job->priority),
+                static_cast<double> (job->t_submit),
+                static_cast<double> (job->t_stamps.pending_ts)});
+        if (s == 0) {
+            errno = ENOENT;
+            return -1;
+        }
+        found_in_prov = true;
         return 0;
     }
 
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -47,7 +47,7 @@ struct match_perf_t {
     std::chrono::time_point<std::chrono::system_clock> graph_uptime
                                         = std::chrono::system_clock::now ();
     /* Time since stats were last cleared */
-    std::chrono::time_point<std::chrono::system_clock> time_since_reset
+    std::chrono::time_point<std::chrono::system_clock> time_of_last_reset
                                         = std::chrono::system_clock::now ();
     struct time_stats {
         void update_stats (double elapsed) {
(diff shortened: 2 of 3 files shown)


garlick (commented):
Change seems OK although I'm not too familiar with the code.
We should add a sharness test for this one IMHO since it's easy to reproduce.

trws:
Sure, I have it mostly converted. Will add it to the unreservable job script since this is a related issue.

trws:
Ok, test added.

t/t4013-unreservable.sh
```diff
@@ -67,6 +67,21 @@ test_expect_success 'ensure all three succeeded' '
...
+test_expect_success 'create an inactive job' '
+	flux submit --quiet \
+	  -N 1 --exclusive \
+		--requires="host:fake[4]" \
+		--setattr=exec.test.run_duration=0.01s \
+		hostname
+'
+test_expect_success 'ensure we can cancel a blocked job' '
+	flux cancel $(flux job last)
+'
```
garlick, line 79: We were able to cancel jobs before the fix because they don't wait for the alloc response to post the clean event.  The problem was that the job manager counts outstanding alloc requests and blocks `flux queue idle` (run during shutdown) until that count reaches zero, which it never does.  Probably something like the following should be added after the cancel:
```sh
test_expect_success 'check that we responded to sched.alloc request' '
    flux job wait-event $(flux job last) clean &&
    count=$(flux queue status -v | grep pending | cut -d" " -f1) &&
    test $count -eq 0
'
```

garlick: Hmm, although that could be a bit racy since the response might not have made it to the job manager yet.

Oh, I guess maybe this would be better actually (and simpler too)
```
flux queue idle --timeout 30s
```

trws: Sure, happy to add that.  I should clarify though that while we could cancel a job in the job manager, the job wasn't actually being cancelled in the resource module, that's what the ENOENT was coming from in the test output.  Makes sense to test both issues.

garlick: Yeah I just meant that this test is not necessarily proving that the issue is fixed since `flux cancel` will be successful even if the `sched.alloc` request was not answered by sched.
```
test_expect_success 'ensure we can cancel a blocked job' '
	flux cancel $(flux job last)
'
```

garlick: p.s. I did build this and test it manually and it works!

trws: Ah, gotcha, I made the poor assumption that the error back from qmanager would propagate out. Thanks for testing!

trws: Added in `queue idle`, is this what you had in mind?

garlick (approved):
LGTM!

resource/modules/resource_match.cpp
```diff
@@ -2290,6 +2290,7 @@ static void stat_clear_cb (flux_t *h, flux_msg_handler_t *w,
 {
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
+    ctx->perf.time_of_last_reset = std::chrono::system_clock::now ();
```
milroy, line 2293: This line got dropped in the last stats PR.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -894,18 +894,25 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
         s = m_pending.erase ({static_cast<double> (job->priority),
                               static_cast<double> (job->t_submit),
                               static_cast<double> (job->t_stamps.pending_ts)});
+        if (s == 1) {
+            return 0;
+        }
+        s = m_blocked.erase ({static_cast<double> (job->priority),
+                              static_cast<double> (job->t_submit),
+                              static_cast<double> (job->t_stamps.pending_ts)});
```
milroy, line 902: This may not be the right time, but you pointed out before that we should replace these vector-keyed maps with tuple-keyed versions.

milroy (approved):
Looks good to me. I did note that we want to replace vector-keyed maps, but getting this PR merged is probably more urgent.

qmanager/policies/base/queue_policy_base.hpp
trws: Yeah, it's on my list, current plan is to add a function or method that takes a job and returns the key, then replace the map type and fix up the key generating function.  Should cut down on duplication and help with performance.

PR #1211, opened 2024-05-29
qmanager: always process unblocked jobs and make bf reactive during sched loop

### LINKED ISSUE #1210: unreservable jobs consume queue_depth
As noted in flux-framework/flux-core#5990, some follow up is needed on why jobs weren't scheduled when the "alloc leak" occurred on el cap.

[Edit from @trws]
Currently, the loop over jobs to schedule up to `queue_depth` jobs or `max_reservations` reservations (whichever is first) counts jobs in the blocked queue toward ` queue_depth`.  In a system where jobs are flowing in and out, this is not a problem because the scheduling loop gets kicked on each of those events and doesn't always move the blocked into the pending, so it will eventually process everything.  In a system with many blocked jobs and relatively few new jobs, the scheduler may not re-enter enough times to process through all the unreservable jobs to get to the ones that can actually be processed.  As an example:

Script below, TLDR: Drain some nodes, enqueue 1000 jobs requiring one of those nodes, enqueue two jobs that can run on anything, wait for the two jobs, hang forever 😞 

<details>
<summary>full sharness test</summary>

```sh
#!/usr/bin/env bash
#

test_description='
'

. `dirname $0`/sharness.sh

rpc() {
  flux python -c \
}

test_under_flux 16384 system

'

	flux config load <<-'EOF'
	[[resource.config]]
	hosts = "fake[0-16383]"
	cores = "0-63"
	gpus = "0-3"

	[[resource.config]]
	hosts = "fake[0-9999]"
	properties = ["compute"]

	[[resource.config]]
	hosts = "fake[10000-16000]"
	properties = ["test"]

	[[resource.config]]
	hosts = "fake[16001-16383]"
	properties = ["debug"]

	[sched-fluxion-qmanager]
	queue-policy = "easy"

	[sched-fluxion-resource]
	match-policy = "firstnodex"
	match-format = "rv1_nosched"
	EOF
'

'
test_expect_success 'drain a few nodes' '
	flux resource drain 1-1000 test with drained nodes
'
'
  time flux python -c \
'
test_expect_success 'create a set of 10 inactive jobs' '
	flux submit --cc=1-1000 --quiet \
	    -N 1 --exclusive \
		--requires="host:fake[1005]" \
		--progress --jps \
		--setattr=exec.test.run_duration=0.01s \
		hostname
'

test_expect_success 'create a set of 2 running jobs' '
	time flux submit --progress --jps --quiet --cc=1-2 --wait-event=start -N1 \
		--flags=waitable \
		--requires=compute \
		--setattr=exec.test.run_duration=5m hostname
'
  time flux job wait -av
'
	flux resource undrain 1-1000 && \
	sleep 10 && \
	flux dmesg && \
    flux jobs -a && \
'
	flux module load sched-simple
'
test_done

```
</details>

Anyway, the core problem is that a bunch of jobs that can't possibly run are eating up the queue_depth, which is usually 32.  I have a tweak that allows the scheduling loop to continue until it processes 32 _meaningful_ jobs, which solves this but makes qmanager unresponsive for _even longer_ during a scheduling loop so I'm loathe to push that up by itself.

trws:
fixes #1210

This is a bit of a big one, and may benefit from having each commit reviewed separately.  If anyone prefers a stacked push I can do that.

The main changes here are:
* Jobs that are determined to be blocked, cannot be reserved, do not count toward the queue_depth.  This means if there are 1000 blocked jobs, and two at the end, we will consider 1002 jobs even if the queue_depth is only 32. The observed behavior is substantially better than otherwise however, because it means we should always make progress on work that can run, where before the two jobs at the end would block forever if the 1000 blocked jobs stay blocked forever.
* The qmanager's backfill implementation now re-enters the reactor after processing a certain number of matches while leaving the scheduling loop active. It's a relatively straightforward approach that stores the loop state and re-enters it in a prep/check pair as needed that mostly re-uses existing machinery.

I would really appreciate extra eyes on this, and testing if anyone has time.  It should not only fix the blocking issue, but also our responsiveness issue (at least up to a point, async would be better, still working on that).

### DIFF
--- a/qmanager/policies/queue_policy_bf_base.hpp
+++ b/qmanager/policies/queue_policy_bf_base.hpp
@@ -34,14 +34,6 @@ protected:
 private:
     int cancel_completed_jobs (void *h);
     int cancel_reserved_jobs (void *h);
-    std::map<std::vector<double>, flux_jobid_t>::iterator &
-        allocate_orelse_reserve (void *h, std::shared_ptr<job_t> job,
-                                 bool use_alloced_queue,
-                                 std::map<std::vector<double>,
-                                     flux_jobid_t>::iterator &iter);
-    std::map<std::vector<double>, flux_jobid_t>::iterator &
-        allocate (void *h, std::shared_ptr<job_t> job, bool use_alloced_queue,
-        std::map<std::vector<double>, flux_jobid_t>::iterator &iter);
     int allocate_orelse_reserve_jobs (void *h, bool use_alloced_queue);
     std::map<uint64_t, flux_jobid_t> m_reserved;
     unsigned int m_reservation_cnt;
--- a/qmanager/policies/queue_policy_bf_base.hpp
+++ b/qmanager/policies/queue_policy_bf_base.hpp
@@ -36,7 +36,9 @@ private:
     int cancel_reserved_jobs (void *h);
     int allocate_orelse_reserve_jobs (void *h, bool use_alloced_queue);
     std::map<uint64_t, flux_jobid_t> m_reserved;
-    unsigned int m_reservation_cnt;
+    int m_reservation_cnt;
+    int m_scheduled_cnt;
+    decltype (m_pending)::iterator m_in_progress_iter = m_pending.end();
 };
 
 } // namespace Flux::queue_manager::detail
--- a/qmanager/policies/queue_policy_bf_base_impl.hpp
+++ b/qmanager/policies/queue_policy_bf_base_impl.hpp
@@ -24,7 +24,7 @@ namespace detail {
  ******************************************************************************/
 
 template<class reapi_type>
-int queue_policy_bf_base_t<reapi_type>::cancel_completed_jobs (void* h)
+int queue_policy_bf_base_t<reapi_type>::cancel_completed_jobs (void *h)
 {
     int rc = 0;
     std::shared_ptr<job_t> job;
@@ -38,7 +38,7 @@ int queue_policy_bf_base_t<reapi_type>::cancel_completed_jobs (void* h)
 }
 
 template<class reapi_type>
-int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void* h)
+int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void *h)
 {
     int rc = 0;
     std::map<uint64_t, flux_jobid_t>::const_iterator citer;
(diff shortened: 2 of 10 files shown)


trws:
Ok, figured this one out. It was a loop ordering issue, but I think this is now fixed, and should be a significant improvement.  The one question I think we need to figure out is what a reasonable trip count is for this.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -52,26 +52,44 @@ namespace queue_manager {
...
-            for (; (iter != m_pending.end()) && (i < m_queue_depth); ++i) {
+            for (int i=0; (m_in_progress_iter != m_pending.end()) && (m_scheduled_cnt < m_queue_depth); ++m_scheduled_cnt, ++i) {
+                // TODO: tune magic number, or make configurable
+                if (i > 1) {
+                    // set schedulable so that the callback machinery gets us
+                    // back in with prep and check
+                    set_schedulability(true);
+                    errno = EAGAIN;
+                    return -1;
+                }
```
milroy, line 90: I can't easily determine the impact of the magic number on the runtime of @garlick's reproducer here: https://github.com/flux-framework/flux-sched/issues/1210#issue-2315882743. Prior to this PR, the reproducer takes 350s to complete on my laptop. With the condition set as above the runtime is 345s. If I set the condition to `(i > 10)` the runtime is 345s, `(i > 100)` the runtime is 350s, and if I set it to `(i > 1000)` the runtime is 424s.

milroy (commented):
This looks like good progress, but there are some formatting issues (especially indentation). I'm having trouble understanding the impact of the changes, especially on the runtime of the reproducer: https://github.com/flux-framework/flux-sched/issues/1210#issue-2315882743

It would be good to add a sharness test for the issue as well.

qmanager/policies/queue_policy_bf_base_impl.hpp
trws: Sounds like there isn't one, at least not a relevant one.  Taking a trip through the reactor is a non-negligible cost, so I expected the runtime to get worse with smaller numbers due to the overhead added to every iteration.  I'll add a separate comment about the original concern I added in the issue.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -15,195 +15,150 @@
...
+        int queue_policy_bf_base_t<reapi_type>::cancel_completed_jobs(void* h)
+        {
+            int rc = 0;
+            std::shared_ptr<job_t> job;
+
+            // Pop newly completed jobs (e.g., per a free request from
+            // job-manager as received by qmanager) to remove them from the
+            // resource infrastructure.
+            while ((job = complete_pop()) != nullptr)
+                rc += reapi_type::cancel(h, job->id, true);
```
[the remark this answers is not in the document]
trws: I'm adding the clang format file to this, and formatting the file.

qmanager/policies/queue_policy_bf_base_impl.hpp
trws: I'm setting it to 5 somewhat arbitrarily, we can revisit if that proves problematic.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -15,195 +15,150 @@
...
+                    }
+                } else if (errno != EBUSY) {
+                    // The request must be rejected. The job is enqueued into
+                    // rejected job queue to the upper layer to react on this.
+                    iter = to_rejected(iter,
+                        (errno == ENODEV) ? "unsatisfiable" : "match error");
+                } else {
+                    if (!try_reserve) {
+                        iter++;
+                        continue;
```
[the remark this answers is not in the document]
trws: I'm adding a comment for this.  It's part of the refactor to flatten the alloc-only and alloc-and-reserve functions, if an allocation-only request fails because not enough resources are available, it returns EBUSY.  That was actually the normal usage of EBUSY, and the cause of our previous consternation with handling errno, it was completely normal for the alloc-only case and really odd for the alloc-and-reserve case.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -144,10 +162,15 @@ namespace queue_manager {
...
-            return rc;
+            if (!is_sched_loop_active()) {
+                rc = cancel_completed_jobs(h);
+                if (rc != 0)
+                    return rc;
+                rc = cancel_reserved_jobs(h);
+                if (rc != 0)
+                    return rc;
+            }
+            return allocate_orelse_reserve_jobs(h, use_alloced_queue);;
```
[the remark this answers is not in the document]
trws: I fixed that in a later, commit, fixing up in place now.

trws:
@milroy this should be ready for another look.  To answer your question about the performance comment [here](https://github.com/flux-framework/flux-sched/issues/1210#issue-2315882743) in #1210, I did this in two pieces.  The first backfill commit here makes it so that we process jobs until we have considered `queue_depth` jobs that _are not blocked_.  The first version of it that I put up, using the 1000 jobs enqueued example I've been using for perf testing, took quite a while to become responsive because it wasn't breaking every 32 jobs, it considered 1032 instead.

I considered that unacceptable, so the next commit in here refactors our base backfill policy implementation to periodically leave the scheduling loop, while still in scheduling state, to re-enter the reactor.  This version still processes all 1032 jobs without needing to be poked, but instead of processing all 1032 before processing incoming messages, it now processes 5. This version should be both more "correct" in that it will run work that can run if it's available, and substantially more responsive than before, at the potential cost of running the scheduling loop slightly slower because of re-entering the sched loop.  Thanks to your tests, I'm not really concerned about that reactor loop delay anymore, so I set it to a small but not singular value so we should have good reactivity while we work out how to make this thing actually async.

Also adjusted the unreservable test to ensure we can't miss this case going forward.

trws:
The new test has intermittent failures, timeouts, on fedora38 and bookworm seemingly.  Trying to track this down but I haven't been able to repro in my local dev container yet.

grondo:
In case it is helpful: sometimes running the test under `taskset -c 0` or `FLUX_TEST_VALGRIND=t` can more easily reproduce failures that only occur in Github actions. Otherwise, I've had success using the  [action-tmate](https://github.com/mxschmitt/action-tmate) action to run on the actual builders when failures can only reproduce there..

trws:
Thanks @grondo, the `taskset` did it.  Now trying to figure out how this can happen.  Just to be sure, `flux job wait -av` doesn't require any specific _order_ of completion does it? Trying to work out what could be causing this hang.

grondo:
> flux job wait -av doesn't require any specific order of completion does it? 

No it should just wait on all waitable jobs AFAIK

trws:
Ok, that was valuable.  I'm pretty sure we ended up with a race condition where the signal to reconsider blocked jobs caused the in-progress iterator of the scheduling loop to be invalidated, causing a hang.  This should take care of it.  I really want a more general mechanism for deferring work during the sched loop, but we need this _now_ if we can stabilize it.

trws:
Ok, now I think this finally does it.  @milroy it turns out that crazy fcfs issue is because fcfs wasn't deterministically invoking post_sched_loop after the sched loop actually ended, but relying on it being invoked whenever the sched loop was started.  In protecting the post_sched_loop from processing a sched loop in progress, it broke that, so now this fixes fcfs as well.

I really, really want to refactor this whole mess, there are far too many non-local flags going on, but this now passes all the tests I've managed to come up with for it, at least locally.  We'll see what happens in CI. 🤞

trws:
@grondo, note the checks_run change here by the way.  If you're good with it I can port that over to core as well.  Probably doesn't matter as much there because it's a sub-make rather than ctest, but it makes a huge difference for live-output on CI here.

qmanager/policies/queue_policy_bf_base.hpp
```diff
@@ -36,7 +36,9 @@ class queue_policy_bf_base_t : public queue_policy_base_t
     int cancel_reserved_jobs (void *h);
     int allocate_orelse_reserve_jobs (void *h, bool use_alloced_queue);
     std::map<uint64_t, flux_jobid_t> m_reserved;
-    unsigned int m_reservation_cnt;
+    int m_reservation_cnt;
+    int m_scheduled_cnt;
+    decltype(m_pending)::iterator m_in_progress_iter = m_pending.end();
```
milroy, line 41: Nit: spacing on the parenthesis. Disregard if later commits fix this.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -49,74 +49,97 @@ int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void* h)
...
+            // example, EBUSY means the request failed because not enough
+            // resources are available right now.
+
+            // Regardless we want to move to the next item in the map
+            ++m_in_progress_iter;
+
+            // If we are trying to reserve, then EBUSY means not only can it not
+            // be allocated, but it cannot ever be reserved given current
+            // conditions. This can happen if there are "down" resources.
+            // The semantics of our backfill policies are to skip this
```
milroy, line 135: These comments are really helpful and will facilitate future debugging.

qmanager/policies/queue_policy_fcfs_impl.hpp
```diff
@@ -150,6 +150,9 @@ int queue_policy_fcfs_t<reapi_type>::handle_match_failure (int errcode)
         set_schedulability (true);
         m_queue_depth_limit = false;
     }
+    // whatever happened here, a job transition has occurred, we need to run the
+    // post_sched_loop
+    m_scheduled = true;
```
milroy, line 155: Nice catch! I don't understand why this works, but I can confirm it fixes the CI test failure.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -142,20 +165,26 @@ int queue_policy_bf_base_t<reapi_type>::apply_params ()
...
+    if (!is_sched_loop_active ()) {
+        rc = cancel_completed_jobs (h);
+        if (rc != 0)
+            return rc;
+        rc = cancel_reserved_jobs (h);
+        if (rc != 0)
+            return rc;
+    }
+    return allocate_orelse_reserve_jobs (h, use_alloced_queue);
+    ;
```
milroy, line 181: Nit: extra `:`
```suggestion
```

milroy (approved):
I found a few minor typos and nits, but the rest looks good and sorely needed. Let's get this merged!

qmanager/policies/queue_policy_fcfs_impl.hpp
trws: This is a big part of why I think we ought to refactor the control flow.  It's one of the two control variables for the prep watcher, this one sets `pls_post_sched_loop` so the post-loop responses get processed.  I think the way this PR works out we now wait until the sched loop is done before processing responses, which was required to fix the issues with the backfill loop.  The other one is controlled by the sched loop active variable, along with another set of post-loop actions that happen immediately when it's set to false.

PR #1212, opened 2024-05-31
avoid using std::vector as a map key

trws:
This isn't a _major_ performance issue, but we had been using a vector as a map key for a static set of 3 elements, casted to doubles for some reason, in qmanager.  This means every time we did a lookup, we did at least one, probably more like three, heap allocations and frees.  It made me twitch every time I saw it, so this refactors that out in favor of a simple struct with defaulted comparison operators and a method to construct it with the appropriate values in the job_t type.

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -11,6 +11,7 @@
 #ifndef QUEUE_POLICY_BASE_HPP
 #define QUEUE_POLICY_BASE_HPP
 
+#include <flux/core/job.h>
 extern "C" {
 #if HAVE_CONFIG_H
 #include "config.h"
@@ -21,7 +22,6 @@ extern "C" {
 #include <cassert>
 #include <map>
 #include <algorithm>
-#include <vector>
 #include <unordered_map>
 #include <string>
 #include <memory>
@@ -68,6 +68,19 @@ struct t_stamps_t {
     uint64_t canceled_ts = 0;
 };
 
+/*! Helper struct for use as a key in job state maps
+ */
+struct pending_key {
+    unsigned int priority = 0;
+    double t_submit = 0.0;
+    uint64_t pending_ts = 0;
+    auto operator<=>(const pending_key&) const = default;
+    bool operator==(const pending_key&) const = default;
+};
+
+using job_map_t = std::map<pending_key, flux_jobid_t>;
+using job_map_iter = job_map_t::iterator;
+
 /*! Type to store a job's attributes.
  */
 class job_t {
@@ -84,13 +97,21 @@ public:
     job_t& operator= (const job_t &s) = default;
 
     bool is_pending () { return state == job_state_kind_t::PENDING; }
+    pending_key get_key () {
+        return {
+            .priority = priority,
+            .t_submit = t_submit,
+            .pending_ts = t_stamps.pending_ts
+        };
+    }
+
 
     flux_msg_t *msg = NULL;
     job_state_kind_t state = job_state_kind_t::INIT;
     flux_jobid_t id = 0;
     uint32_t userid = 0;
     unsigned int priority = 0;
-    double t_submit = 0.0f;;
+    double t_submit = 0.0f;
     std::string jobspec = "";
     std::string note = "";
     t_stamps_t t_stamps;
@@ -168,7 +189,7 @@ public:
     void reconsider_blocked_jobs () {
         m_pending_reconsider = true;
         if (!is_sched_loop_active ()) {
-            process_provisional_reconsider();
+            process_provisional_reconsider ();
         }
     }
 
@@ -382,10 +403,7 @@ public:
         }
         job->state = job_state_kind_t::PENDING;
         job->t_stamps.pending_ts = m_pq_cnt++;
-        m_pending_provisional.insert (std::pair<std::vector<double>, flux_jobid_t> (
-            {static_cast<double> (job->priority),
-             static_cast<double> (job->t_submit),
-             static_cast<double> (job->t_stamps.pending_ts)}, job->id));
+        m_pending_provisional.emplace (job->get_key (), job->id);
         m_jobs.insert (std::pair<flux_jobid_t, std::shared_ptr<job_t>> (job->id,
                                                                         job));
         m_schedulable = true;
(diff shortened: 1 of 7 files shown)


qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -84,13 +97,21 @@ class job_t {
     job_t& operator= (const job_t &s) = default;
 
     bool is_pending () { return state == job_state_kind_t::PENDING; }
+    pending_key get_key() {
```
milroy, line 100: Nit: spacing
```suggestion
    pending_key get_key () {
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -371,10 +392,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
         }
         job->state = job_state_kind_t::PENDING;
         job->t_stamps.pending_ts = m_pq_cnt++;
-        m_pending_provisional.insert (std::pair<std::vector<double>, flux_jobid_t> (
-            {static_cast<double> (job->priority),
-             static_cast<double> (job->t_submit),
-             static_cast<double> (job->t_stamps.pending_ts)}, job->id));
+        m_pending_provisional.emplace (job->get_key(), job->id);
```
milroy, line 395: Nit:
```suggestion
        m_pending_provisional.emplace (job->get_key (), job->id);
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -556,8 +574,8 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
      *  resource API. When a match succeeds, this method is called back
      *  by reapi_t.
      */
-    virtual int handle_match_success (int64_t jobid, const char *status,
-                                      const char *R, int64_t at, double ov) {
+    int handle_match_success (flux_jobid_t jobid, const char *status,
+                                      const char *R, int64_t at, double ov) override {
```
milroy, line 578: This line extends beyond 80 characters.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -587,9 +605,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
         if (m_jobs.find (id) == m_jobs.end ())
             return nullptr;
         job = m_jobs[id];
-        m_pending.erase ({static_cast<double> (job->priority),
-                          static_cast<double> (job->t_submit),
-                          static_cast<double> (job->t_stamps.pending_ts)});
+        m_pending.erase (job->get_key());
```
milroy, line 608:
```suggestion
        m_pending.erase (job->get_key ());
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -862,52 +877,28 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
...
-                           std::pair<std::vector<double>, flux_jobid_t> (
-                               {static_cast<double> (job->priority),
-                                static_cast<double> (job->t_submit),
-                                static_cast<double> (job->t_stamps.pending_ts)},
-                               job->id));
-            if (!res.second) {
-                errno = EEXIST;
-                return -1;
-            }
+        auto & pending_map = into_provisional ? m_pending_provisional : m_pending;
```
milroy, line 880:
```suggestion
        auto &pending_map = into_provisional ? m_pending_provisional : m_pending;
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -862,52 +877,28 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
...
         return 0;
     }
 
     int erase_pending_job (job_t *job, bool &found_in_prov)
     {
         size_t s;
-        s = m_pending.erase ({static_cast<double> (job->priority),
-                              static_cast<double> (job->t_submit),
-                              static_cast<double> (job->t_stamps.pending_ts)});
+        s = m_pending.erase (job->get_key());
```
milroy, line 892: Yeah, that's way cleaner.

resource/reapi/bindings/c++/reapi_module_impl.hpp
```diff
@@ -102,7 +102,7 @@ void match_allocate_multi_cont (flux_future_t *f, void *arg)
             return;
         }
     } else {
-        adapter->handle_match_failure (errno);
+        adapter->handle_match_failure (rj, errno);
```
milroy, line 105: Can you change the `rj` variable name in `match_allocate_multi_cont` to be more descriptive like `jobid`? I realize you didn't define it, but it's not clear what the variable is.

milroy (dismissed):
I found a few spacing-related nits that will probably get corrected by clang-format, and a line that extends beyond 80 characters. There's also a variable with an ambiguous name (not defined in this PR, but used in it) that can be made clearer.

With those changes this is good to go!

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -21,7 +22,6 @@ extern "C" {
 #include <cassert>
 #include <map>
 #include <algorithm>
-#include <vector>
```
milroy, line 24: Removing the `vector` include causes the build to fail. You can fix the build failure by including `vector` in queue_policy_bf_base.hpp

milroy:
You can fix the PR build failures by including `vector` in queue_policy_bf_base.hpp and replacing all occurrences of `std::map<std::vector<double>, flux_jobid_t>::iterator` with `job_map_iter` in queue_policy_bf_base.hpp and queue_policy_bf_base_impl.hpp.

qmanager/policies/base/queue_policy_base.hpp
trws: Yeah, this should be fixed now, it was because I tried to rebase away the stack on top of the qmanager PR, should all pass this time.

trws:
Ok, I think I addressed all of that.  Assuming the docker registry will cooperate we should see much more green this time around.  The rj variables became either jobid, or where that was taken `res_jobid`.

trws:
Turns out the focal container wasn't going to work anyway, since the gcc in there doesn't support defaulting the spaceship operator.  Switched it to a tuple and everything should be happy now.

milroy (approved):
It will simplify the PR to make the last commit a fixup and squash it into the first commit, but it's up to you.

This looks good an is ready to go!

trws:
I thought about it, but kinda want to keep it this way. I would prefer to do it the original way because it's more self-documenting and compiles faster, keeping the evolution shows how we got here and what we could do if we drop base focal support.

PR #1213, opened 2024-05-31
test/match-formats: make jq command valid

trws:
problem: somehow we've had an invalid jq command in one of the sched tests for quite some time now

solution: give it the "." pattern rather than ""

This is actually in one of the commits in #1211, but it's driving me crazy not having this fixed in other branches.


grondo (approved):
:+1:

PR #1216, opened 2024-06-05
comment blocks: stop using full-star boxes

trws:
problem: formatters, such as clang-format, don't preserve the spacing between the end of the headline text and the star at the end

solution: switch to a different headline box comment form that doesn't put anything after the text

We don't have to go with this style, but I do like the consistent width.  Here's the script I applied to make this along with the command to invoke it:

```python
import re
import sys
for f in sys.argv[1:]:
    with open(f) as fd:
        txt = fd.read()
    after = (re.sub(r"^/\*{30,}$\n \* {30,}\*\n \* +(.*?) [ \*]+$\n \* {30,}\*\n^ \*{30,}/$", r"""
// \1
////////////////////////////////////////////////////////////////////////////////""", txt, flags=re.M))
    with open(f, "w") as of:
        of.write(after)
```

command: ` rg -l '^ \* +.*\*$' | xargs python3 ./replace_blocks.py`


milroy (approved):
The new format looks fine to me.  Having a way to automatically enforce formatting is more important than a headline box appearance anyway.

PR #1217, opened 2024-06-06
qmanager: reconsider blocked jobs on reprioritize

trws:
problem: if a job in reserved state, still pending in qmanager, gets reprioritized it's possible that it could unblock a job in the blocked queue and without doing a reconsideration it's possible we could get a priority inversion

solution: reconsider blocked jobs when a pending job is reprioritized

I have yet to make a reproducer that reliably needs this, but I'm pretty sure if we can do this sequence it will happen:

1. start a long-running job that uses a node
2. enqueue a high priority job that requires one of the resources used by the first so that it's reserved
3. enqueue a lower priority job that could run immediately, but wont because it's blocked by the reservation
4. reprioritize the reserved job to have a lower priority than the third

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -775,6 +775,9 @@ public:
             if (insert_pending_job (job, found_in_prov) < 0)
                 return -1;
             m_schedulable = true;
+            // in case this job is now lower priority than one it was blocking,
+            // reconsider blocked jobs
+            reconsider_blocked_jobs ();
         }
         return 0;
     }
@@ -877,6 +880,9 @@ protected:
                 m_schedulable = true;
             }
         }
+        // in case this job is now lower priority than one it was blocking,
+        // reconsider blocked jobs
+        reconsider_blocked_jobs ();
         m_pending_reprio_provisional.clear ();
         return 0;
     }
--- a/t/t1013-qmanager-priority.t
+++ b/t/t1013-qmanager-priority.t
@@ -4,6 +4,9 @@ test_description='Fluxion takes into account urgency and t_submit'
 
 . `dirname $0`/sharness.sh
 
+export TEST_UNDER_FLUX_QUORUM=1
+export TEST_UNDER_FLUX_START_MODE=leader
+
 hwloc_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/hwloc-data`
 # 1 node, 2 sockets, 44 cores (22 per socket), 4 gpus (2 per socket)
 excl_1N1B="${hwloc_basepath}/001N/exclusive/01-brokers-sierra2"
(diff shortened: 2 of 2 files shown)


qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -765,6 +765,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
             if (insert_pending_job (job, found_in_prov) < 0)
                 return -1;
             m_schedulable = true;
+            reconsider_blocked_jobs ();
```
milroy, line 768: Could you add a comment along the lines of the commit message to explain why blocked jobs should be reconsidered here?

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -866,6 +867,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
                 m_schedulable = true;
             }
         }
+        reconsider_blocked_jobs ();
```
milroy, line 870: Ditto above; an explanatory comment would be helpful.

milroy (commented):
This change is small enough I'm good with it as is. Your description of the steps needed to test this change sound reasonable, and it would be good to have that check in place.

trws:
Ok, finally figured out the test.  Turns out the test I was thinking of _does_ work, but only with conservative backfill.  May need to look into why it doesn't happen with easy, but this did catch the issue and is fixed by the patch.   Also new comments with the reconsider calls.

t/t1013-qmanager-priority.t
```diff
@@ -64,6 +67,74 @@ test_expect_success 'priority: cancel all jobs' '
...
+test_expect_success 'wait for fluxion to be ready' '
+  flux python -c \
+    "import flux, json; print(flux.Flux().rpc(\"sched.resource-status\").get_str())"
+'
+test_expect_success 'start a single long job to reserve a node' '
+	jobid1=$(flux submit -N1 \
+    -t 300s \
+		--flags=waitable \
+		--setattr=exec.test.run_duration=30s \
+		sleep 300) &&
```
milroy, line 112: The alignment is odd here. Likely an artifact of clang-format.

t/t1013-qmanager-priority.t
```diff
@@ -64,6 +67,74 @@ test_expect_success 'priority: cancel all jobs' '
...
+		--setattr=exec.test.run_duration=30s \
+		sleep 300)
+'
+test_expect_success 'submit a job with low priority that only needs one node, must end up blocked' '
+	jobid3=$(flux submit -N1 \
+		--flags=waitable \
+		--urgency=8 \
+    -t 400s \
+		--setattr=exec.test.run_duration=30s \
+		sleep 400) &&
```
milroy, line 127: Ditto formatting.

milroy (approved):
I tested the reproducer and confirmed that it works as expected. I also confirmed that without the changes from the first commit in this PR test `t1013-qmanager-priority.t` fails as expected.

There is some odd formatting likely caused by clang-format in `t1013-qmanager-priority.t`, but apart from that this is ready to merge. Good catch @trws!

t/t1013-qmanager-priority.t
trws: Bah, I was modifying it in a poorly configured vim, thanks for catching that.

trws: Oh god, almost all the other lines are indented with _TABS_ no wonder it looks strange.

PR #1218, opened 2024-06-06
Reformat

trws:
Two small tweaks to add this to the format script and remove the special-casing for pack/unpack, then apply formatting.  Once we're happy with this, I intend to add a .git-ignore-revs or similar file we can use to ignore the global reformat commit in blame, but it has to have the commit sha in full in the file.


trws:
Now that the release is together, I rebased the reformat here.  Short of adding CI to verify it, which might accidentally work because it's in the format script now that I think about it, this should be good to go.

trws:
@milroy, what do you think of just getting this in? I've rebased and reapplied the formatting.  I've also created a pull request to your rank-based-partial-release branch with a rebase of your branch on top of this with the reformatting applied to every commit to preserve all your changes.  If you have new changes that aren't in there, or just want to do it yourself, this script completely automates it assuming you have fetched my reformat branch and are currently on your "partial-release" branch, or any other branch for that matter:

```sh
export BEFORE_MASS_FORMAT=af44d3b53457240d511bd2a32a3485fbdb91246d
export MASS_FORMAT=78b80b195bc36186abb438f499adbee5bef4a025
git rebase \
  --strategy-option theirs \
  --empty=drop \
  --exec "$(echo './scripts/format' \
      '|| echo prettier failed its ok; ' \
      'git commit -a --amend --no-verify --no-edit' \
    )" \
  --onto $MASS_FORMAT $BEFORE_MASS_FORMAT;
```

This is also now enforced in CI in this PR.

trws:
Apologies if anyone looked at this during the churn after the last comment.  Had to switch our format infrastructure to use pre-commit like core does to ensure we get consistent versions and avoid formatting differences from that.  Now all straightened up, and using clang-format 18.1.6.

trws:
Reminder to self: When we're ready to pull the trigger here, I need to change the name of the action in the branch protection rules so this and other PRs with the new name can pass.

milroy:
Sorry for the delay. I'll review this PR tonight.

milroy (approved):
This will be very helpful for the long-term maintenance of Fluxion.

Let's get it merged.

milroy:
> I've also created a pull request to your rank-based-partial-release branch with a rebase of your branch on top of this with the reformatting applied to every commit to preserve all your changes. If you have new changes that aren't in there, or just want to do it yourself, this script completely automates it assuming you have fetched my reformat branch and are currently on your "partial-release" branch, or any other branch for that matter

Thanks a lot for providing an automated way to update my PR. I've made a few changes to the branch so I I'll run the command myself.

scripts/format
```diff
@@ -10,3 +10,9 @@ if command -v black 2>&1 > /dev/null ; then
 else
   echo "black not found, python left unformatted"
 fi
+
+find resource qmanager src/ -name '*.h' -o -name '*.c' -o -name '*.cpp' -o -name '*.hpp' |
+  grep -v yggdrasil |
+  grep -v 'json.hpp' |
+  grep -v libtap |
+  xargs -P 8 -n 1 clang-format -i
```
jameshcorbett: It looks like this is removed in a later commit?

scripts/check-format
```diff
@@ -5,10 +5,17 @@ if command -v black 2>&1 > /dev/null ; then
...
 else
     echo "black not found, failing"
     exit 1
 fi
+
+find resource qmanager src/ -name '*.h' -o -name '*.c' -o -name '*.cpp' -o -name '*.hpp' |
+  grep -v yggdrasil |
+  grep -v 'json.hpp' |
+  grep -v libtap |
+  xargs -P 8 -n 1 clang-format --dry-run --Werror
```
jameshcorbett: I think this too is removed in a later commit?

jameshcorbett (approved):
I noticed two things that could maybe be squashed out of existence. Otherwise looks great. I'm wondering whether it needs to be merged manually since it expects the 'python format' action which has been renamed? In which case I'm happy to do that

PR #1223, opened 2024-06-17
qpolicy/base: schedulable on reconsider

trws:
This is related to the hang when processing 100+ batch submissions on blocked resources. On my dev machine it consistently processes all jobs after this, but it means we sometimes backfill more aggressively than the q depth would imply. We need to decide which result is right. @milroy, @grondo.


trws:
Bah, messed up the branch, it's on my fork now, _sigh_. Will fix.

PR #1224, opened 2024-06-19
qpolicy/base: schedulable on reconsider

trws:
problem: we should always reconsider the queue, run the sched loop, when reconsidering blocked jobs. In some corner cases we did not because they are merged directly rather than being inserted individually.

solution: explicitly set schedulable.

PR redone from #1223 because of branching issue.


trws:
While this doesn't solve the performance issue that makes it look like we have a hang, it _does_ solve an actual potential hang case.  It would be good to get this one in soonish if that's feasible.

milroy (approved):
This looks good! Ship it.

PR #1225, opened 2024-06-20
Excise filter

trws:
The filtered_graph buys us basically nothing, but it costs order 30%-80% depending on the operation and how we measure.  I need to run a performance study with and without, but for production cases it should always be a win, and it reduces compile times.  It's most noticeable with the many-time-point test such that there's a _ton_ of matches, but it should make at least a small difference everywhere.

This is stacked on top of #1224 because without it, this change made the race condition much more common and frequently hung one or more tests.

As part of routing out the filtered-graph types, I removed a number of "shared_ptr" arguments.  Overall my intention is that no regular pointer type argument should ever _own_ storage, but we can pass things by pointer when just sharing a reference to state owned by something else.  Ideally these would become references, but that's more disruptive to the code.

### DIFF
--- a/resource/evaluators/expr_eval_vtx_target.cpp
+++ b/resource/evaluators/expr_eval_vtx_target.cpp
@@ -96,8 +96,7 @@ done:
 }
 
 void expr_eval_vtx_target_t::initialize (const vtx_predicates_override_t &p,
-                                         const std::shared_ptr<
-                                             const f_resource_graph_t> g,
+                                         const resource_graph_t *g,
                                          vtx_t u)
 {
     m_initialized = true;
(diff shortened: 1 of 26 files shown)


milroy (approved):
LGTM!

I was trying to understand the original motivation for introducing the filtered graph, but could only find this PR: https://github.com/flux-framework/flux-sched/pull/283.

Could there be any performance advantage to using a filtered graph when subsystems are working? I can't think of anything, but I want to pose the question.

trws:
I think the answer to that is no, because we still have to defensively check to make sure we only take the out-edges that match the subsystem we're currently exploring.  This way we don't do it twice. If it turns out that it is slower though, then the solution would be to make it so we only consider the edges we want, and I would do that by changing how we store out-edges to make that efficient rather than doing this.

PR #1226, opened 2024-06-22
batch consume messages, immediately process cancellation

trws:
This is in progress, but so far takes the 10,000 allocs and 10,000 cancels from some terrifyingly long amount of time to 36 and a half seconds.  It's using a combination of @garlick's batching suggestion and immediately processing cancels by interrupting the scheduling loop if necessary.  Pauses may be too long, probably needs tuning, but _clearly_ better.  Popping it up here before we hop on the plane back from sweden this morning, so I'll be in and out of communication today.


trws:
This design turned out to be a bit of a dead end for match performance reasons, new PR coming that makes all of our backfill queues asynchronous.

PR #1227, opened 2024-06-25
qmanager: Asynchronously communicate with resource in the scheduling loop to allow job updates while matching

### LINKED ISSUE #1222: submitting a large number of jobs and canceling them caused scheduler "hang"
On elcap a user submitted a large number of jobs, each requesting a specific node, then thought better of it and canceled them all. For the past few hours the scheduler has not scheduled any jobs and the "alloc pending" count is stuck at many thousands.

Attempting to stop the queues has not improved the situation.

On the recommendation of @trws I've grabbed a core of the rank 0 broker (available in my homedir).

Here's the resource module stats at the time of this failure:
```json
{
 "match": {
  "succeeded": {
   "njobs": 11940,
   "njobs-reset": 11940,
   "max-match-jobid": 429461867754015744,
   "max-match-iters": 52,
   "stats": {
    "min": 0.001353974,
    "max": 32.755777653999999,
    "avg": 0.29216099639715115,
    "variance": 3.2035476969734025
   }
  },
  "failed": {
   "njobs": 83916,
   "njobs-reset": 83916,
   "max-match-jobid": 429655446928752640,
   "max-match-iters": 328,
   "stats": {
    "min": 0.0073356740000000004,
    "max": 68.145720294,
    "avg": 0.29968277191067089,
    "variance": 0.886960191744261
   }
 }
```

The use case here is that a user is looking for bad/slow nodes and needs to run a test on each node in a queue. For now I've suggested they allocate all the nodes from the queue and then use bulksubmit in that subinstance, which should work better.

trws:
This is a replacement for #1226, though it doesn't yet have eager cancellation of the sched loop to immediately process job cancellations. The short version is that this makes our backfill queues completely asynchronous with respect to the resource module, and deals with the 10,000 allocs followed by 10,000 cancels test in about 45 seconds on my laptop. Assuming the numbers are comparable, that's a bit under 100x faster than the time @milroy reported for the test in #1222.

This also _completely_ changes the performance profile of flux under this kind of load.  During a flood of requests like that, match still shows up as something like 30%, but that's all.  The job-injest and qmanager modules rise to almost match it for consumption, and the yaml processing becomes a meaningful cost.  May look and see if some of that can be easily squashed out.

fixes #1222

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -406,7 +406,7 @@ public:
         m_pending_provisional.emplace (job->get_key (), job->id);
         m_jobs.insert (std::pair<flux_jobid_t, std::shared_ptr<job_t>> (job->id,
                                                                         job));
-        m_schedulable = true;
+        set_schedulability (true);
         rc = 0;
     out:
         return rc;
@@ -454,7 +454,7 @@ public:
                 errno = EEXIST;
                 goto out;
             }
-            m_schedulable = true;
+            set_schedulability (true);
         }
         rc = 0;
     out:
@@ -496,7 +496,7 @@ public:
             m_complete.insert (std::pair<uint64_t,
                                         flux_jobid_t> (job->t_stamps.complete_ts,
                                                        job->id));
-            m_schedulable = true;
+            set_schedulability (true);
             break;
         default:
             break;
@@ -758,30 +758,17 @@ public:
             return -1;
         }
 
-        if (is_sched_loop_active ()) {
-            // if sched-loop is active, the job's pending state
-            // cannot be determined. There is "MAYBE pending state" where
-            // a request has sent out to the match service.
-            auto res = m_pending_reprio_provisional.insert (
-                           std::pair<uint64_t,
-                                     std::pair<flux_jobid_t, unsigned int>> (
-                               m_reprio_cnt, std::make_pair (job->id, priority)));
-            m_reprio_cnt++;
-            if (!res.second) {
-                errno = EEXIST;
-                return -1;
-            }
-        } else {
-            bool found_in_prov = false;
-            if (erase_pending_job (job.get (), found_in_prov) < 0)
-                return -1;
-            job->priority = priority;
-            if (insert_pending_job (job, found_in_prov) < 0)
-                return -1;
-            m_schedulable = true;
-            // in case this job is now lower priority than one it was blocking,
-            // reconsider blocked jobs
-            reconsider_blocked_jobs ();
+        auto res = m_pending_reprio_provisional.insert (
+                       std::pair<uint64_t,
+                                 std::pair<flux_jobid_t, unsigned int>> (
+                           m_reprio_cnt, std::make_pair (job->id, priority)));
+        m_reprio_cnt++;
+        if (!res.second) {
+            errno = EEXIST;
+            return -1;
+        }
+        if (!is_sched_loop_active ()) {
+            return process_provisional_reprio ();
         }
         return 0;
     }
@@ -881,12 +868,12 @@ protected:
                 job->priority = priority;
                 if (insert_pending_job (job, found_in_provisional) < 0)
                     return -1;
-                m_schedulable = true;
+                set_schedulability (true);
+                // in case this job is now lower priority than one it was blocking,
+                // reconsider blocked jobs
+                reconsider_blocked_jobs ();
             }
         }
-        // in case this job is now lower priority than one it was blocking,
-        // reconsider blocked jobs
-        reconsider_blocked_jobs ();
         m_pending_reprio_provisional.clear ();
         return 0;
     }
--- a/qmanager/policies/queue_policy_bf_base.hpp
+++ b/qmanager/policies/queue_policy_bf_base.hpp
@@ -27,19 +27,25 @@ public:
     virtual int reconstruct_resource (void *h, std::shared_ptr<job_t> job,
                                       std::string &R_out);
     virtual int apply_params ();
+    virtual int handle_match_success (flux_jobid_t jobid, const char *status,
+                                      const char *R, int64_t at, double ov);
+    virtual int handle_match_failure (flux_jobid_t jobid, int errcode);
 
 protected:
     unsigned int m_reservation_depth;
     unsigned int m_max_reservation_depth = MAX_RESERVATION_DEPTH;
 
 private:
+    int next_match_iter ();
     int cancel_completed_jobs (void *h);
     int cancel_reserved_jobs (void *h);
     int allocate_orelse_reserve_jobs (void *h, bool use_alloced_queue);
     std::map<uint64_t, flux_jobid_t> m_reserved;
     int m_reservation_cnt;
     int m_scheduled_cnt;
+    bool m_try_reserve = false;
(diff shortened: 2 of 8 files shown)


grondo (commented):
This sounds great! Unfortunately, I'm not familiar enough with the code to do a proper review, but I didn't notice any glaring issues. It would probably be good to get @milroy to rerun his workload on this branch and report back (and do a code review).

But wow, great improvement.

trws:
I forgot to base this on the reconsider fix, that should fix the hangs in the unreservable test we got on bookworm and jammy before.

trws:
Found another reconsider bug that caused the qmanager to spin after making it async like this.  Added fix here.

trws:
Aaaand clearly that did not work. Trying to track down the core issue here.

trws:
Fun new facts, this _did_ work, but there's an ordering bug that's exposed by this linked above.  Once that lands, this should be all green, and no longer spins.  I would recommend any side testing with this branch to be done with the flux-core change in flux-framework/flux-core#6062 in case you're looking @milroy.

milroy:
I can confirm the 100x speedup on my test scripts. I'm seeing runtimes of 47-53s now.

milroy:
I can also confirm successful test results with the change here: https://github.com/flux-framework/flux-core/pull/6062.

However, `ctest` is taking 59s in comparison with 37s on Fluxion `master`.

trws:
That's interesting, I suppose it's not entirely surprising since there's longer latency between a response from resource and processing it, but that's much more of a hit than I expected. Will try to get a perf trace and see what's changed.

trws:
Are your measurements using the same version of flux-core @milroy? When I run a master fluxion vs this PR, I get the same perf for both, but they're both built against the flux-core PR.

trws:
Good news everyone! (if you didn't read this in Farnsworth's voice, feel free to do so now 😉 )

The merged PR upstream fixed all the failures here with no further touching at all.  @milroy is there anything further you'd like so see or tweak here?

grondo:
Excellent! Can you suggest a PR title that would translate to a good line in the release notes? (release notes are semi-automated with a script, and each PR ends up as a line noted by title).

trws:
Updated, feel free to tweak if you like.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -865,6 +865,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t
 
     int process_provisional_reprio ()
     {
+        bool reprioritezed_a_job = false;
```
jameshcorbett, line 868: typo in variable name: `reprioritized`

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -47,6 +49,38 @@ int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void *h)
...
+    std::unique_ptr<char, decltype (_free)>  jobs_str (
+            json_dumps (jobarray.get(), JSON_INDENT (0)),
+            _free
+            );
+    if ( !jobs_str) {
+        errno = ENOMEM;
+        return -1;
+    }
+    int res = reapi_type::match_allocate_multi (m_handle, m_try_reserve, jobs_str.get (), this);
+    return res;
```
jameshcorbett, line 81: combine these lines?

jameshcorbett (commented):
I don't know the qmanager code base very well so my comments are just two superficial things I noticed.

trws:
I just found that this needs another patch.  It's "correct" in that it stops the spinning and gives us the async behavior, but it doesn't process jobs past queue-depth because of fixing the spinning issue. Patch coming shortly.

trws:
Ok, the patch is up, but as you can see it changes the queue-depth behavior.  Turns out fixing this issue brings us back there.  I'm going to fix the test at this point, and will explain this in the commit message as well, but here's what's going on.  The code we have doesn't re-enter the sched loop unless an event has occurred, and the test encodes this behavior and expects that _exactly one_ sched loop will happen after a job is cancelled.  With the new code, because something has been scheduled and there are jobs that haven't been considered, we start a new scheduling loop to try and get more work done immediately, as a result we process and start a job that sched used to think was "past queue depth" because it isn't past the queue depth on the second loop.

trws:
Pushed, I'm sorry for the last minute churn here, but in working on the benchmark for my next PR (to help deal with the many-time-points issue) I realized that this one essentially hung trying to run it once I sped up the job submission enough.  Either way, this one should be ready to go, other one coming shortly.

milroy:
@trws I don't think I was using the same versions of flux-core in the two setups for my previous tests. I'm now seeing very similar execution times between Fluxion master and this PR. Note that the `ctest` times have increased by almost a factor of 2 recently.

trws:
I'll try to figure out why that is, clearly not a good thing.  As soon as I get the failed-job match time PR up I'll try to bisect for it.

qmanager/policies/queue_policy_bf_base.hpp
```diff
@@ -27,19 +27,25 @@ class queue_policy_bf_base_t : public queue_policy_base_t
...
 
 protected:
     unsigned int m_reservation_depth;
     unsigned int m_max_reservation_depth = MAX_RESERVATION_DEPTH;
 
 private:
+    int next_match_iter ();
     int cancel_completed_jobs (void *h);
     int cancel_reserved_jobs (void *h);
     int allocate_orelse_reserve_jobs (void *h, bool use_alloced_queue);
```
milroy, line 42: Looks like `use_alloced_queue` is vestigial here.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -47,6 +49,37 @@ int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void *h)
...
+    if ( !jobs_str) {
+        errno = ENOMEM;
+        return -1;
+    }
+    return reapi_type::match_allocate_multi (m_handle, m_try_reserve, jobs_str.get (), this);
+}
+
 template<class reapi_type>
 int queue_policy_bf_base_t<reapi_type>::allocate_orelse_reserve_jobs (void *h,
                                                                       bool use_alloced_queue)
```
milroy, line 92: Looks like `use_alloced_queue` is disused.

milroy (approved):
I found some formatting nits, and a disused variable.

Apart from that this looks good to me. Really impressive speedup!

qmanager/policies/queue_policy_bf_base.hpp
trws: It is, and as far as I know has been for a while.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -47,6 +49,37 @@ int queue_policy_bf_base_t<reapi_type>::cancel_reserved_jobs (void *h)
...
+    }
+    auto _free = [](char *p) { free (p); };
+    std::unique_ptr<char, decltype (_free)>  jobs_str (
+            json_dumps (jobarray.get(), JSON_INDENT (0)),
+            _free
+            );
+    if ( !jobs_str) {
+        errno = ENOMEM;
+        return -1;
+    }
```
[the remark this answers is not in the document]
trws: Good point, I've split that up to help it flow a little better.

PR #1228, opened 2024-06-25
pruner: don't track own type for pruning

trws:
problem: the ALL resource type in the pruner specification was including the type of the anchor vertex, so in common uses like "ALL:core" all core vertices allocate a planner to track the number of cores they contain

solution: if the anchor and the target type match, don't track it

This saved roughly 600MB with a 16k node graph with 64 cores per node.  Also kinda trivial.  Looking for other low-hanging fruit.


milroy (approved):
This is a simple and impactful change. Looks good!

PR #1232, opened 2024-07-02
dfu: add a quick-to-test feasibility precheck to match

trws:
problem: When a large number of possible time points exist, unreservable jobs cause N matches where N is the number of time points. This makes jobs that can't be matched potentially very expensive, and heavy use of constraints make the case common.

solution: iterate over just the node vertices checking for constraint matches, whether they are UP and whether it's possible to allocate them (either at the end of time, or in the case of allocate immediately). If all of those pass, the node is deemed "feasible" and counted. If less nodes are feasible than are required by the request, it's immediately rejected.

In a test with 300 time points, without this patch I get between 5 and 7.5 seconds for each failing match. With this patch the failed match stats in the same test report a minimum of 0.005 and a maximum of 0.008 seconds. This will not help with *successful* matches, but for matches that can't work because of drained nodes, constraints, queues, etc. it should reduce match times massively.

<details>
<summary>
many-point performance test
</summary>
This is the script I used, this is the particularly cruel version where all 302 jobs are constrained to individual nodes.

```sh
#!/usr/bin/env bash
#

test_description='
'

. `dirname $0`/sharness.sh

rpc() {
  flux python -c \
}

test_under_flux 16380 system

'

	flux config load <<-'EOF'
	[[resource.config]]
	hosts = "fake[0-16380]"
	cores = "0-63"
	gpus = "0-3"

	[[resource.config]]
	hosts = "fake[0-9999]"
	properties = ["compute"]

	[[resource.config]]
	hosts = "fake[10000-16000]"
	properties = ["test"]

	[[resource.config]]
	hosts = "fake[16001-16380]"
	properties = ["debug"]

	[sched-fluxion-qmanager]
	queue-policy = "easy"

	[sched-fluxion-resource]
	match-policy = "firstnodex"
	match-format = "rv1_nosched"
	EOF
'

'
test_expect_success 'drain a few nodes' '
	flux resource drain 1-100 test with drained nodes
'
'
  time flux python -c \
'
test_expect_success 'create a set of jobs with disparate end time points' '
  # time=300 && \
  # for i in $(seq 1 300) ; do \
		# echo submitting $time && \
	flux submit --cc=300-600 \
		--quiet \
		--requires="host:fake[{cc}]" \
		-N 1 --exclusive \
		-t "{cc}s" \
		--progress \
		--jps \
		--wait-event=start \
		--setattr=exec.test.run_duration="600s" \
		sleep 600
		# time=$((time + 1)) ; \
	# done
'
# perf record --call-graph=fp -p $(pgrep flux-broker-0) sleep 10 &
# PID="$!"
test_expect_success 'create a set of 2 unreservable jobs' '
flux jobs -a
flux resource info
flux jobs -a
	flux submit --progress --jps --quiet -N1 \
		--requires="host:fake[5]" \
		--flags=waitable \
		--setattr=exec.test.run_duration=2s \
		sleep 2
	flux submit --progress --jps --quiet -N1 \
		--requires="host:fake[5]" \
		--flags=waitable \
		--setattr=exec.test.run_duration=2s \
		sleep 2
'
# kill -2 $PID
# wait $PID
# cp perf.data /Users/scogland1/Repos/flux/flux-sched/build/mine

  flux module stats sched-fluxion-resource
  # time flux job wait -av || true
'
	flux resource undrain 1-100 && \
	(time flux job wait -av || true) && \
	flux dmesg && \
    flux jobs -a && \
'
	flux module load sched-simple
'
test_done

```
</details>

### DIFF
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -14,7 +14,6 @@ extern "C" {
 #endif
 }
 
-#include <iostream>
 #include <cstdlib>
 #include <cerrno>
 #include "resource/traversers/dfu.hpp"
@@ -33,10 +32,10 @@ extern struct match_perf_t perf;
 ////////////////////////////////////////////////////////////////////////////////
 
 int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
-                                     detail::jobmeta_t &meta, bool x,
+                                     detail::jobmeta_t &meta,
+                                     bool x,
                                      vtx_t root,
-                                     std::unordered_map<std::string,
-                                                        int64_t> &dfv)
+                                     std::unordered_map<std::string, int64_t> &dfv)
 {
     int rc = 0;
     std::vector<uint64_t> agg;
@@ -45,13 +44,12 @@ int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
 
     meta.alloc_type = jobmeta_t::alloc_type_t::AT_SATISFIABILITY;
     planner_multi_t *p = (*get_graph ())[root].idata.subplans.at (dom);
-    meta.at = planner_multi_base_time (p)
-              + planner_multi_duration (p) - meta.duration - 1;
+    meta.at = planner_multi_base_time (p) + planner_multi_duration (p) - meta.duration - 1;
     detail::dfu_impl_t::count_relevant_types (p, dfv, agg);
     errno = 0;
-    if ( (rc = detail::dfu_impl_t::select (jobspec, root, meta, x)) < 0) {
+    if ((rc = detail::dfu_impl_t::select (jobspec, root, meta, x)) < 0) {
         rc = -1;
-        errno = (!errno)? ENODEV : errno;
+        errno = (!errno) ? ENODEV : errno;
         detail::dfu_impl_t::update ();
     }
     m_total_preorder = detail::dfu_impl_t::get_preorder_count ();
@@ -60,6 +58,81 @@ int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
     if (!errno)
         errno = saved_errno;
     return rc;
+}
+
+int dfu_traverser_t::request_feasible (detail::jobmeta_t const &meta,
+                                       match_op_t op,
+                                       vtx_t root,
+                                       std::unordered_map<std::string, int64_t> &dfv,
+                                       const subsystem_t &dom)
+{
+    if (op == match_op_t::MATCH_UNKNOWN)
+        return 0;
+
+    const auto target_nodes = dfv["node"];
+    const bool checking_satisfiability =
+        op == match_op_t::MATCH_ALLOCATE_W_SATISFIABILITY || op == match_op_t::MATCH_SATISFIABILITY;
+
+    if ((!meta.constraint) && (target_nodes <= get_graph_db ()->metadata.nodes_up))
+        return 0;
+
+    // check if there are enough nodes up at all
+    if (target_nodes > get_graph_db ()->metadata.nodes_up) {
+        if (op == match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE || op == match_op_t::MATCH_ALLOCATE) {
+            errno = EBUSY;
+            return -1;
+        }
+        if (checking_satisfiability) {
+            // if we're checking satisfiability, only return here if
+            // it's actually unsatisfiable
+            errno = ENODEV;
+            return -1;
+        }
+    }
+
+    const auto by_type_iter = get_graph_db ()->metadata.by_type.find ("node");
+    if (by_type_iter == get_graph_db ()->metadata.by_type.end ())
+        return 0;
+    auto &all_nodes = by_type_iter->second;
+
+    auto &g = *get_graph ();
+    const int64_t graph_end =
+        std::chrono::duration_cast<std::chrono::seconds> (
+            get_graph_db ()->metadata.graph_duration.graph_end.time_since_epoch ())
+            .count ();
+    // only the initial time matters for allocate
+    const int64_t target_time = op == match_op_t::MATCH_ALLOCATE ? meta.at : graph_end - 1;
+    int feasible_nodes = 0;
+    for (auto const &node_vtx : all_nodes) {
+        auto const &node = g[node_vtx];
+        // if it matches the constraints
+        if ((!meta.constraint || meta.constraint->match (node))
+            // if it's up and not drained
+            && (checking_satisfiability || node.status == resource_pool_t::status_t::UP)
+            // if it's available
+            && planner_avail_resources_during (node.schedule.plans, target_time, meta.duration)
+                   == 1) {
+            ++feasible_nodes;
+            if (feasible_nodes >= target_nodes) {
+                break;
+            }
+        }
+    }
+    if (feasible_nodes < target_nodes) {
+        // no chance, don't even try
+        if (op == match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE || op == match_op_t::MATCH_ALLOCATE) {
+            errno = EBUSY;
+            return -1;
+        }
+        if (checking_satisfiability) {
+            // if we're checking satisfyability, only return here if
+            // it's actually unsatisfiable
+            errno = ENODEV;
+            return -1;
+        }
+    }
+
+    return 0;
 }
 
 int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
@@ -76,6 +149,10 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
     int saved_errno = errno;
     planner_multi_t *p = NULL;
     const subsystem_t &dom = get_match_cb ()->dom_subsystem ();
+
+    // precheck to see if enough resources are available for this to be feasible
+    if (request_feasible (meta, op, root, dfv, dom) < 0)
+        return -1;
 
     if ((rc = detail::dfu_impl_t::select (jobspec, root, meta, x)) == 0) {
         m_total_preorder = detail::dfu_impl_t::get_preorder_count ();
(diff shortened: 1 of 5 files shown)


resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
     planner_multi_t *p = NULL;
     const subsystem_t &dom = get_match_cb ()->dom_subsystem ();
 
+    // precheck to see if enough resources are available for this to be feasible
```
milroy, line 79: General comment: this change makes the `schedule` function pretty large and hard to understand. Could you put it in another function (maybe `schedule_precheck`)?

resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
+    if (op != match_op_t::MATCH_UNKNOWN) {
+        bool checking_satisfiability = op == match_op_t::MATCH_ALLOCATE_W_SATISFIABILITY || op == match_op_t::MATCH_SATISFIABILITY;
+        graph_duration_t graph_duration = get_graph_db ()->metadata.graph_duration;
+        auto &g = *get_graph ();
+        int64_t graph_end = std::chrono::duration_cast<std::chrono::seconds>
+            (graph_duration.graph_end.time_since_epoch ()).count ();
+        auto p_iter = g[root].idata.subplans.find (dom);
+        auto by_type_iter = get_graph_db ()->metadata.by_type.find ("node");
+        int64_t target_time = graph_end - 1;
+        int64_t target_duration = 1;
```
milroy, line 89: Considering the requested duration might be better:
```suggestion
        int64_t target_time = graph_end - meta.duration;
        int64_t target_duration = meta.duration;
```

resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
+            // only the initial time matters for allocate, and we want to check
+            // the full duration
+            target_time = meta.at;
+            target_duration = meta.duration;
+        }
+        if (p_iter != g[root].idata.subplans.end () && by_type_iter != get_graph_db ()->metadata.by_type.end ()) {
+            p = p_iter->second;
+            auto &all_nodes = by_type_iter->second;
+            int feasible_nodes = 0;
+            for (auto const&node_vtx : all_nodes) {
```
milroy, line 101: Nit:
```suggestion
            for (auto const &node_vtx : all_nodes) {
```

resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
+                auto const &node = g[node_vtx];
+                // if it matches the constraints
+                if ((!meta.constraint || meta.constraint->match(node))
```
milroy, line 104: I think it would be faster to pull the `meta.constraint` check out to the top-level `if` statement condition. Then it would only check for feasible nodes if there are constraints set.  

You could cache the counts of the `UP` nodes and just compare `target_nodes` with `UP` node counts for jobspecs without constraints in the top-level check.

resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
+                        // if it's up and not drained
+                        && (checking_satisfiability || node.status == resource_pool_t::status_t::UP)
+                        // if it's available at the end of time
+                        && planner_avail_resources_during (node.schedule.plans, target_time, target_duration) == 1) {
```
milroy, line 108: This line looks like it needs wrapping, too.

resource/traversers/dfu.cpp
```diff
@@ -77,6 +76,58 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
...
+                    ++feasible_nodes;
+                    if (feasible_nodes >= target_nodes) {
+                        break;
+                    }
+                }
+            }
+            if (feasible_nodes < target_nodes) {
+                // no chance, don't even try
+                if (op == match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE || op == match_op_t::MATCH_ALLOCATE) {
```
milroy, line 117: Wrapping needed here, too.

milroy (commented):
This is a much-needed change. I think there are a few ways to improve it before merging.

resource/traversers/dfu.cpp
milroy: It should be possible to cache the `UP` counts by queue, too.

resource/traversers/dfu.cpp
trws: Agreed, I was having some tooling issues, but yes I'll do that.

resource/traversers/dfu.cpp
trws: Thanks for this, I had tried using the same as the schedule function uses below (graph_end - meta.duration - 1) and it didn't pass, but this does.  Changed.

trws: I have to amend this a bit.  It has to use an at of `graph_end - 1` for cases other than ALLOCATE because duration may actually be the whole graph, except the literal last second.  We can use the meta.duration directly in all cases though so I'm now doing that.

resource/traversers/dfu.cpp
trws: Sadly getting it by queue is non-obvious, I don't think we set a "default" property for example, and there are some strange corner cases.  We could track the number up/down by property, but that by itself doesn't seem to quite work.  Catching all possible updates is somewhat non-obvious even just to get the marking down/up.

That said, I managed to get it seemingly working for overall up/down state within the db and metadata code.  It's now checking that there are enough nodes up before going on at all, whether there's a constraint or not (if the total is < requested, might as well).

trws:
Ok, @milroy I think that addresses everything.  I'm happy to look at per-queue UP tracking next cycle, but we need to work out how we track the queue or queues of a resource.

trws:
Just realized I missed pushing the update to the mark methods, added and re-pushed.  It's the few changes in dfu_impl_update.cpp, with those it accurately tracks total nodes up.

resource/traversers/dfu.cpp
```diff
@@ -62,6 +60,78 @@ int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
...
+        op == match_op_t::MATCH_ALLOCATE_W_SATISFIABILITY || op == match_op_t::MATCH_SATISFIABILITY;
+
+    // check if there are enough nodes up at all
+    if (target_nodes > get_graph_db ()->metadata.nodes_up) {
+        if (op == match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE || op == match_op_t::MATCH_ALLOCATE) {
+            errno = EBUSY;
+            return -1;
+        }
+        if (checking_satisfiability) {
+            // if we're checking satisfyability, only return here if
```
milroy, line 83: Typo: "satisfiability."

resource/traversers/dfu.cpp
```diff
@@ -62,6 +60,78 @@ int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
...
+    const auto target_nodes = dfv["node"];
+    const bool checking_satisfiability =
+        op == match_op_t::MATCH_ALLOCATE_W_SATISFIABILITY || op == match_op_t::MATCH_SATISFIABILITY;
+
+    // check if there are enough nodes up at all
+    if (target_nodes > get_graph_db ()->metadata.nodes_up) {
+        if (op == match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE || op == match_op_t::MATCH_ALLOCATE) {
+            errno = EBUSY;
+            return -1;
+        }
```
milroy, line 81: I think we can avoid the loop over nodes for jobspecs without constraints by adding a block like this:
```c++
if ((!meta.constraint) && (target_nodes <= get_graph_db ()->metadata.nodes_up))
    return 0;
```
right after the highlighted code. If you also want to check the availability at the end of time, you can run a single query of available node counts via the aggregate filter at the root vertex.

milroy (commented):
This is looking good. I think there's still an optimization or two to add to the start of `request_feasible`

resource/traversers/dfu.cpp
trws: Ah, that's a good idea, is the `g[root].schedule.plans` sufficient? I'm still trying to grok how the various planners work across the tree.

trws: Actually, I'm going with the check as you posted it.  We can add the other later if it looks like it might be profitable.

milroy: You could adapt this: https://github.com/flux-framework/flux-sched/blob/802f2ca1291d4f9521d090e4aa73b88dffbcaf92/resource/traversers/dfu_impl.cpp#L166)

To be something like this:
```c++
    std::vector<uint64_t> aggs;
    planner_multi_t *p = (*m_graph)[root_vtx].idata.subplans[s];
    count_relevant_types (p, resource.user_data, aggs);
    len = aggs.size ();
    if ((rc = planner_multi_avail_during (p, at, d, &(aggs[0]), len)) == -1) {
```

trws:
Thanks @milroy.  Updated, rebased on top of the other one that just merged, and pushed.

resource/traversers/dfu.cpp
milroy: > Ah, that's a good idea, is the g[root].schedule.plans sufficient? I'm still trying to grok how the various planners work across the tree.

You need to use the `idata.subplans` `planner_multi` like what I pasted above. You probably didn't see that since we posted at the same time.

trws: Yup, exactly. Thanks for that, it makes sense.

milroy (approved):
This should be a big help for cases we've seen in production!

trws:
Here's hoping! 🤞

PR #1233, opened 2024-07-02
Fix match average calculation

milroy:
The current calculation of average match times uses the total job count rather than the job count since last reset.

Use the correct count to compute the average (hence downstream calculations, too).


trws (approved):
LGTM!

milroy:
Setting MWP.

PR #1234, opened 2024-07-02
checks_run: work when no autogen.sh is present

trws:
Apparently we haven't been building the flux-sched arm64 image for a while, this should fix that.

problem: we're unconditionally invoking autogen.sh

solution: don't fail if it's missing


trws:
@grondo I think we talked about this while I was out of town and I forgot to circle back, would be great to get the arm images building again.

grondo (approved):
LGTM

PR #1235, opened 2024-07-02
github-actions: actually push the flux-sched manifest

trws:
This is a rough one, no clue how we got along this far without it, maybe because the same thing runs on core it's actually just refreshing the correct manifest, not sure. Either way this should be a priority CI fix.

problem: somehow we've been updating the core manifest rather than the sched one, and either not hurting anything or not noticing, not sure which.  Either way, not great.

solution: actually push flux-sched's manifest


grondo (approved):
Ouch! Good catch

PR #1239, opened 2024-07-05
planner/dfu: avoid dereferencing empty vector

trws:
problem: a few places were using `&(vec[0])` to get an address of the start of a potentially empty vector.  This is almost always actually safe in practice, but it's not allowed and technically could be treated like a null dereference as well.

solution: use the `vec.data()` method that exists specifically for this purpose

@milroy tagging you here since we happened to be talking about this code just the other day.  As far as I know this hasn't been touched recently, but maybe the hardening flags in the buildfarm did. 🤷

### DIFF
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -189,7 +189,7 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
         len = planner_multi_resources_len (p);
         duration = meta.duration;
         detail::dfu_impl_t::count_relevant_types (p, dfv, agg);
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Looks good! I'll apply the diff as a patch to the RPM.

jameshcorbett:
> maybe the hardening flags in the buildfarm did.

I certainly didn't change anything by hand but maybe something in the buildfarm changed? I don't know the buildfarm that well.

trws:
We pull the hardening flags from the platform specfiles, so it could have been something we did, or even an update from RHEL. 🤷

PR #1240, opened 2024-07-07
graph: use builtin bidirectional graph

trws:
problem: walking inward edges is difficult and slow because they're mixed in with the `out_edges` range (we explicitly add out-edges to every vertex matching each inward edge for some reason).  This means we have to filter them out during traversals, and maintain them manually. They show up almost like another subsystem (the "in" subsystem) in some readers, and aren't generated at all by others.

solution: switch from directed to bidirectional graph type, remove the explicit addition of "in" edges, and fix the graph test to match the actual number of edges.

I was very surprised how easy this was, would have sworn I'd tried it and failed at least once before.  The main possible downside is that bidirectional graphs have to store every edge twice, and thus should require twice as much storage for edges as directional graphs. Strangely, switching to this actually didn't increase memory usage _at all_.  It stayed identical, and I don't know why.  The even _better_ news is that after removing the explicit back-edges, this actually saved 500MB for a 16,000 node 64-core graph, and made running the test under heaptrack take a full second less with 5,000,000 less allocations (17.5s -> 16.5s and 75m allocations to 70m for just loading the graph and stopping).


jameshcorbett (approved):
Looks reasonable. Two questions:

1. I'm guessing the JGF reader is one of the readers that doesn't bother with in-edges?
2. Are there no logic changes or simplifications that can be done in traversals now that in-edges are gone?

milroy:
I'm impressed it took so few changes to get `bidirectionalS` to work!

A couple of things to consider: along the lines of @jameshcorbett's question, we should be able to compress the JGFs by removing the "in" edges. 

Also, I noticed that the changes in this PR also compile and pass CI for `undirectedS`. @trws can you run the tests you described with a `boost::undirectedS`?

trws:
I didn't find where "in" edged were being created in JGF, but i'll check again.  If they're there, they should definitely go away.

I could run them for undirected, but that brings me to the answer to @jameshcorbett's question 2. This PR is just about switching the base graph type, keep it simple, but if we keep directional edges, then we can distinguish between edges going up and down the containment hierarchy in O(1) time, if we use undirected then we'll be back to O(N) time. I haven't done the optimizations to take advantage of that for upwards walks yet, but the performance improvement I got is partly from removing the inward edges we no longer need to consider from the downward walk.  There are several other things we can now do because the difference between `out_edges` and `in_edges` exists, and is constant time.  I'm also planning to wrap a small data structure for use in place of `OutEdgeList` that will let us get edges by type in either direction in O(1) as well, then we can do subsystems efficiently.

trws:
Ok, this should be ready for another pass.  The JGF reader now no ignores containment edges of the "in" type, correctly resolves parent vertices in the reader for the remove test that was relying on old behavior, much more efficiently finds those parent vertices, and the JGF test files no longer contain either "in" or "draws_power" edges.  Despite some of the json files getting changed from single-line to formatted, it's a net win of almost 40,000 lines, which is a bit crazy.  JGF reader needs a rework, but this at least gets things cleaned up for this purpose.

resource/schema/resource_graph.hpp
```diff
@@ -14,6 +14,7 @@
 #include "resource/schema/resource_data.hpp"
 #include <boost/config.hpp>
 #include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/graph_selectors.hpp>
```
milroy, line 17: Fluxion builds and passes the testsuite without `graph_selectors`.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -1157,18 +1156,14 @@ int resource_reader_jgf_t::get_parent_vtx (resource_graph_t &g,
...
-                parent_vtx = next_vtx;
-                rc = 0;
-                break;
-            }
-        }    
-  }
+        if (g[*ei].name.contains ("containment")) {
+            parent_vtx = next_vtx;
+            rc = 0;
+            break;
```
milroy, line 1162: This looks good for containment, but I seem to remember there was a corner case that motivated @zekemorton to implement the check this way. Zeke, can you comment here?

milroy (commented):
Pending Zeke's input on one changed function this looks good and has the side effect of removing >30,000 of the 142,886 lines (!) in the power JGF.

resource/readers/resource_reader_jgf.cpp
trws: FWIW, when the graph was a `directedS` graph, the approach I'm using here would not have worked at all since there wouldn't be a way to get in_edges like this, and even if we iterated all vertices to get them there would have been in_edges from the children as well.  As long as all graphs follow the invariant that there's always a containment tree, and it's always single parent, this should be fine.  If we have a case where one of those is violated, it would be wrong, but I'd want to know what that is and why it's invalidating the invariant.

This reminds me by the way, you asked why it is that everything has to be in the containment hierarchy, every component I mean, and this is one example of why.  If we want there to be other non-containment hierarchies (things like networking can't be hierarchies, but they shouldn't be anyway, they should be extra links between things that exist in other hierarchies) we could do that, but it would mean having the hierarchy the thing is resident in be part of the vertex, or the vertex type perhaps, and then having some operations that need to traverse more than one.

milroy: >  If we want there to be other non-containment hierarchies (things like networking can't be hierarchies, but they shouldn't be anyway, they should be extra links between things that exist in other hierarchies) 

I think this comes up in the `power` JGF, where the power subsystem is also a hierarchy. The current implementation can return the parent in the power subsystem instead of the parent in the containment hierarchy, which actually might not be desirable.

I think returning the parent vertex in the containment hierarchy will work well, and it's lower complexity.

milroy:
Do the changes in this PR allow simplification of the `by_outedges` metadata?

trws:
I agree with that assessment.  Honestly I expected it to be much more significant, but 100MB out of 5.3GB... :shrug:



> Do the changes in this PR allow simplification of the `by_outedges` metadata?

I had to go look. It reduces the amount of work to maintain it slightly by reducing the outedges considered during planner updates, but the `by_outedges` metadata itself is independent of the number of outedges, it's actually by a weight calculated through the subplan.  Something I didn't realize, but seems really important now that I do, is that the heuristic works like this (comment from code):

```c
        // Set dynamic traversing order based on the following heuristics:
        //     1. Current-time (jobmeta.now) resource availability
        //     2. Last pruning filter resource type (if additional
        //        pruning filter type was given, that's a good
        //        indication that it is the scarcest resource)
```

This means that in our current configuration, the heuristic is calculated based on the `rack:node` plan I think? Unless it only counts the prune filters that can be relevant at a given level, in which case it would be `ALL:gpu` at the node level.  That seems... not ideal, as does the heuristic.  May have to give that some thought.  There is definitely an optimization that can be made there though, since it's actually removing (and deallocating) the map node, modifying the key, then re-inserting it.  We can fix that with `.extract` now.

resource/readers/resource_reader_jgf.cpp
trws: We could make it capable of doing both with the same complexity if we pass in which edge type to consider, or have it embedded in the vertex type.  Maybe something to consider for later.

milroy: > The current implementation...

In case it wasn't clear (I think it was), I meant current flux-sched `master`, not the change in the PR.

PR #1241, opened 2024-07-07
graph: remove unused subsystem selector class

trws:
problem: I accidentally left this code in when removing the filtered graph

solution: remove it


jameshcorbett (approved):
👍

PR #1246, opened 2024-07-16
String interner for subsystems and resource types and initial usage

trws:
This is a bit of a large patch.  Most of it is in the "add Catch2" commit because that's adding an external checked in so that all of our builds on systems that don't have catch2 installed will build.  I recommend installing catch2 version 3+ in the base distro for faster builds, but it's not required.  There will be a follow-up PR that adds appropriate entries in the install scripts and dockerfiles and so-forth to make that happen for CI and our testenv containers.

The actual sched changes here are the addition of an interned string type that looks like this:
```c++
// create an interned_string type named subsystem_t that is a uint8_t and has a key of zero
// embedded in the type signature
using subsystem_t = interned_string<0, uint8_t>;

// add a string to the subsystem_t table
subsystem_t s("node");
```

The type is comparable, sortable, and will return a constant reference to the std::string on `.get()`. You can also access the id through `.id()` and generate ranges of these things.  There is also a wrapper around `boost::small_vector` that uses interned_strings as keys to simulate a very small map without having to store the keys.

Why go through all of this? My many_point test, copied below, went from an end-to-end time of 49 seconds to 34 seconds.  Throughput of jobs ingested from `submit -cc... --wait-event=start` went from ~30 to ~50 jobs per second, and the `stop_explore` method went from 38% of resource cycles to 0.2%.  We got less of a memory usage benefit than I hoped, I think largely because there are still a number of `std::map<>` instances, where alignment causes the savings to be relatively small for now. Right now it's about 8-10% reduction in base memory usage for sched, no impact on per-allocation usage I don't think.

<details>
<summary>perf-test-many-point.sh</summary>

```sh
#!/usr/bin/env bash
#

test_description='
'

. `dirname $0`/sharness.sh

rpc() {
  flux python -c \
}

test_under_flux 16380 system

'

	flux config load <<-'EOF'
	[[resource.config]]
	hosts = "fake[0-16380]"
	cores = "0-63"
	gpus = "0-3"

	[[resource.config]]
	hosts = "fake[0-9999]"
	properties = ["compute"]

	[[resource.config]]
	hosts = "fake[10000-16000]"
	properties = ["test"]

	[[resource.config]]
	hosts = "fake[16001-16380]"
	properties = ["debug"]

	[sched-fluxion-qmanager]
	queue-policy = "easy"

	[sched-fluxion-resource]
	match-policy = "firstnodex"
	match-format = "rv1_nosched"
	EOF
'

'
#test_expect_success 'drain a few nodes' '
#	flux resource drain 1-100 test with drained nodes
#'
'
  time flux python -c \
'
perf record --call-graph=fp -p $(pgrep flux-broker-0) -o $PERF_OUT sleep 60 &
test_expect_success 'create a set of jobs with disparate end time points' '
  # time=300 && \
  # for i in $(seq 1 300) ; do \
		# echo submitting $time && \
	flux submit --cc=300-600 \
		--quiet \
		--requires="host:fake[{cc}]" \
		-N 1 --exclusive \
		-t "{cc}s" \
		--progress \
		--jps \
		--wait-event=start \
		--setattr=exec.test.run_duration="600s" \
		sleep 600
		# time=$((time + 1)) ; \
	# done
'
test_expect_success 'submit one more job that holds the node the next needs until after all other time points' '
	holding_job=$(flux submit \
    --job-name=holder \
		--requires="host:fake[5]" \
		-N 1 --exclusive \
		-t "600s" \
		--progress \
		--jps \
		--wait-event=start \
		--setattr=exec.test.run_duration="600s" \
		sleep 600 )
'
# PID="$!"
test_expect_success 'create a set of 2 currently unreservable jobs' '
flux jobs -a
flux resource info
flux jobs -a
	flux submit --progress --jps --quiet -N1 \
		--requires="host:fake[5]" \
		--flags=waitable \
		--setattr=exec.test.run_duration=0.1s \
		sleep 0.1
	flux submit --progress --jps --quiet -N1 \
		--requires="host:fake[5]" \
		--flags=waitable \
		--setattr=exec.test.run_duration=0.1s \
		sleep 0.1
'
# kill -2 $PID
# wait $PID
# cp perf.data /Users/scogland1/Repos/flux/flux-sched/build/mine

  flux module stats sched-fluxion-resource
  # time flux job wait -av || true
'

#	flux resource undrain 1-100 && \

	flux job cancel $holding_job && \
	(time flux job wait -av || true) && \
	flux dmesg && \
    flux jobs -a && \
'
	flux module load sched-simple
'
test_done

```

</details>


trws:
I'm not sure what's up with the CI for jammy and focal, it's currently building fine and then seems like it just sits when it should start the check.  Looking into it.

trws:
Unfortunately the focal and jammy failures are legitimate.  I'm digging into _why_ right now, but I have a feeling this is an old libstdc++ or old boost biting us.

trws:
More data here. While the failures are legitimate, it seems like they're being caused by a difference in how gcc is linking the libraries on these distros.  If I compile the exact same code with clang on jammy it works, with gcc 11 (the default on jammy) it fails.

trws:
I was making a mistake and linking the interner but not the so that actually initialized some of the interned strings into some of the objects.  Most of the time this was fine because the externed symbols are required to deduplicate and work out, but some corner case made it not happen. 🤷 

This is more correct, hopefully should pass everywhere.

trws:
@milroy, this is finally set, everything passes, even focal (though it needed a new compiler, at least we can kinda limp along with one still in the focal repos).

milroy:
@trws I noticed the updates and build failures. Should I hold off on a review until everything is green, or should I review this as is?

trws:
I'll fix it up this morning. The edit was meant to be to just shift from a vector to an unordered map to avoid a danger of over-allocation if someone chooses a very high group ID, but I accidentally backed out a change needed to tolerate focal's ancient compiler.

trws:
Ok, the build failures should be fixed.  As part of that this is now stacked on top of #1256, but should be ready for review.

resource/traversers/dfu_impl.cpp
```diff
@@ -65,15 +64,15 @@ bool dfu_impl_t::exclusivity (const std::vector<Jobspec::Resource> &resources, v
     // and it requested exclusive access, return true;
     bool exclusive = false;
     for (auto &resource : resources) {
```
milroy, line 66: Is it worth considering converting `Jobspec::Resource` member data to `resource_type_t`?

src/common/libutil/interner.cpp
```diff
@@ -73,9 +73,14 @@ view_and_id get_both (const interner_t group_id, const std::string_view s, char
 
     {  // shared lock scope
         auto sl = std::shared_lock (*storage.mtx);
+#if __GNUC__ < 11
+        // this awful workaround is required by focal support
+        auto it = storage.ids_by_string.find (std::string(s));
```
milroy, line 78: Nit:
```suggestion
        auto it = storage.ids_by_string.find (std::string (s));
```

milroy (approved):
This looks good! I have a few questions, but I don't think they'll block merging.

For the record: vendoring Catch2 is fine with me. I think @garlick's point about TOSS being the limiting factor holds.

src/common/libutil/interner.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <mutex>
+#include <shared_mutex>
+#include <tuple>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace intern {
+namespace detail {
+std::shared_mutex group_mtx;
```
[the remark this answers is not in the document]
trws: I'm glad to hear the performance difference is small.  That said, while it's subtle, it's already used in two threads.  Because libjobspec is used by both resource and qmanager there can be race conditions in some cases, especially during startup.  The whole thing is set with reader/writer locks (poorly named by the C++ spec as shared_mutex for some reason).

src/common/libutil/interner.hpp
```diff
@@ -0,0 +1,245 @@
...
+        obj = interned_string{tmp};
+        return os;
+    }
+};
+template<class InternT, class T, int likely_count = 2>
+struct interned_key_vec : boost::container::small_vector<T, likely_count> {
+    using base = boost::container::small_vector<T, likely_count>;
+    T &operator[] (InternT s)
+    {
+        if (s.id () >= base::size ()) [[unlikely]]
```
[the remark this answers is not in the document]
trws: It's something I don't use often, since it's also easy to have that go wrong if the assertion isn't right enough of the time, but since this should only very, very rarely need to expand it seemed appropriate.

src/common/libutil/interner.cpp
trws: Huh... that should have come in automatically with pre-check/clang-format.  Good catch! I'll fix it for now and try to figure out why it wasn't as well.

resource/traversers/dfu_impl.cpp
trws: For now I converted only the "type" member.  Honestly I'm mildly concerned about even doing that, because it's a potentially user-controlled field, and these strings are immortal. It was one of the major uses and items that gets compared though, so it's important.  Still considering what to do about that in the longer term.  It's possible to make the strings reference counted, at some cost, or to add a cross-compatible version that's reference-counted until it's promoted to immortal?

src/common/libutil/interner.cpp
trws: Also this may just go away, now that the new images are up, if we drop focal support this code block need not exist.

src/common/libutil/interner.cpp
```diff
@@ -0,0 +1,175 @@
+#include "interner.hpp"
```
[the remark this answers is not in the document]
trws: I wouldn't mind moving it there.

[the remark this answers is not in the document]
trws: It has been moved.

src/common/libutil/interner.cpp
trws: PR to get rid of the need for this is #1257.

trws: Ok, since that merged, this code no longer exists.

src/common/libutil/interner.cpp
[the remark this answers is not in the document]
milroy: Got it; I didn't notice it was used in qmanager.

PR #1248, opened 2024-07-18
Don't add pruning filter for leaf vertices

milroy:
When `ALL` is specified for prune-filters of a leaf vertex type, other leaf vertices maintain a pruning filter for that resource type. For example, the default `ALL:core` adds `core` filters to `gpu` vertices. Since a leaf vertex is the terminus of a depth-first search of a single-subsystem graph, the extra pruning filters increase memory consumption unnecessarily.

Add a check that exits the pruning-filter priming function for leaf vertices.

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -1103,6 +1103,14 @@ int dfu_impl_t::prime_pruning_filter (const subsystem_t &s,
 
     (*m_graph)[u].idata.colors[s] = m_color.gray ();
     accum_if (s, type, (*m_graph)[u].size, to_parent);
+
+    // Don't create and prime a pruning filter if this is
+    // a leaf vertex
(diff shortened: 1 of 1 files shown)


milroy:
When running the 10,000 job submission and cancellation test with the 16K simulated node cluster this commit reduces memory usage by about 1GB (18%)

milroy:
After further thought, I don't think the `m_graph_db->metadata.roots.size () == 1` check is desired. If the size is `> 1` and the vertex has no out edges we still don't want to add pruning filters.  For any size `> 0` if out edges exist we still want pruning filters.

trws (approved):
I agree, and had come to a similar conclusion overnight.  I'm considering working the subsystem support such that it will catch all leaves even with subsystems anyway. LGTM!

milroy:
Thanks for the review! Setting MWP.

Issue #1249, opened 2024-07-18
qmanager spams annotation updates for all jobs


grondo:
As noted in flux-framework/flux-core#6114, we recently discovered that a possible contributing factor to rank 0 broker RSS growth is massive amounts of annotation updates from Fluxion for jobs. Speaking with @trws and @milroy earlier today, it appeared that they discovered a possible bug in qmanager that could send unnecessary annotation updates for all jobs on each check callback (so on each event loop iteration).

garlick:
Yep, I think you found it.  Try this:
```diff
diff --git a/src/modules/sched-simple/sched.c b/src/modules/sched-simple/sched.c
index 0f08af251..614206d42 100644
--- a/src/modules/sched-simple/sched.c
+++ b/src/modules/sched-simple/sched.c
@@ -292,9 +292,6 @@ static void annotate_reason_pending (struct simple_sched *ss)
 {
     int jobs_ahead = 0;
 
-    if (!flux_module_debug_test (ss->h, DEBUG_ANNOTATE_REASON_PENDING, false))
-        return;
-
     struct jobreq *job = zlistx_first (ss->queue);
     while (job) {
         if (schedutil_alloc_respond_annotate_pack (ss->util_ctx,
@@ -326,17 +323,17 @@ static void check_cb (flux_reactor_t *r, flux_watcher_t *w,
                       int revents, void *arg)
 {
     struct simple_sched *ss = arg;
-    flux_watcher_stop (ss->idle);
+    //flux_watcher_stop (ss->idle);
 
     /* See if we can fulfill alloc for a pending job
      * If current head of queue can't be allocated, stop the prep
      *  watcher, i.e. block. O/w, retry on next loop.
      */
     if (try_alloc (ss->h, ss) < 0 && errno == ENOSPC) {
-        annotate_reason_pending (ss);
-        flux_watcher_stop (ss->prep);
-        flux_watcher_stop (ss->check);
+        //flux_watcher_stop (ss->prep);
+        //flux_watcher_stop (ss->check);
     }
+    annotate_reason_pending (ss);
 }
 
 static int try_free (flux_t *h, struct simple_sched *ss, json_t *R)
```
then 
```
flux start
flux resource drain 0
flux submit hostname
```
watch the RSS

garlick:
Oh just saw in the other bug that you made more progress.  Sorry for the noise!

grondo:
No, this is good, thanks!

[...]

PR #1250, opened 2024-07-18
qmanager: annotate each job only once on reservation

trws:
problem: an annotation is sent for each job that enters reserved state when its predicted start time changes.  That is controlled by when the `old_at` and `at` times associated with the job in qmanager are different.  The implementation fails to update `old_at` after this to prevent repeat messages.  This was somewhat hidden by the fact that the subsequent updates had little effect, except after recent performance improvments made it possible for the qmanager to flood the job-manager with annotations in some cases.

solution: set old_at to at after deciding whether to send the annotation


grondo (approved):
LGTM and proven on elcap.
The one missing line that ate 200G :grin: 

Might want to autoclose issue #1249 with a "Fixes" message

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -97,6 +97,7 @@ int qmanager_cb_t::post_sched_loop (
             // if old_at == at, then no reason to send this annotation again.
             if (job->schedule.at == job->schedule.old_at)
                 continue;
+            job->schedule.old_at = job->schedule.at;
```
milroy, line 100: As we discussed in person, it _**might**_ be a good idea to move this line after `schedutil_alloc_respond_annotate_pack` so the assignment is made if the annotation response succeeds. I can also see an argument for leaving the line where it is to prevent repeated annotation response failures.

milroy (approved):
Great catch! Pending final placement of the one-line change this is ready to merge.

qmanager/modules/qmanager_callbacks.cpp
trws: I've been mulling on this, and I think since it's an informational rather than functional message I'd prefer to have it skip it for now.

PR #1252, opened 2024-07-19
flux-sharness: sync with flux-core

trws:
problem: I've been using heaptrack a lot, and other tools that need to be wrapped around flux itself.  This is difficult without something to integrate with test_funder_flux

solution: add some env vars to do a heaptrack recording or add a generic wrapper

### DIFF
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -3,6 +3,9 @@
 #  project-local sharness code for Flux
 #
 
+# add scripts directory to path
+export PATH="${SHARNESS_TEST_SRCDIR}/scripts:$PATH"
+
 #
 #  Extra functions for Flux testsuite
 #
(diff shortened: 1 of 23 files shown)


grondo (commented):
I like FLUX_TEST_WRAP, and this seems useful. Shall we add it to flux-core as well, since I think flux-sharness.sh is shared between projects.

trws:
Yeah, I thought we'd do both or neither but kept needing it here so started that way.  Made a short attempt at updating our sharness.sh version as well, 1.2.1 is out and at least in sched we don't have the heredoc support and a few other things, but utterly failed I'm sorry to say.

trws:
Now that this is merged in flux-core, just syncing up sched in this PR.

trws:
Not sure what's up with codecov, the measure on their website is that this somehow increased coverage by 0.06%, and then on here it says it failed and decreased coverage by 0.2%... Everything else is passing, and this doesn't actually change any tests, or implementation so... Yeah, perplexed about that.

grondo:
The PR shows indirect changes in `ephemeral.cpp` of -22%, with the whole of `ephemeral_t::insert()` being missed on this PR. 

I wonder if compiler version or flags differences could be exposing lines that did not appear before?

grondo (approved):
LGTM!

PR #1256, opened 2024-07-26
matrix: add explicit python path for el8

trws:
This one is slightly urgent.  None of our CI runs will pass on EL8 until this is dealt with, fallout from rebuilding the fluxrm/testenv image on el8 is it picked up a newer valgrind, which changed the python environment on the system in a way that interacts badly with the older cmake in there.

problem: upstream el8 has added a new version of valgrind, which pulls in python3.11. The older cmake logic on el8 finds the newer python but fails to find any development dependencies for it and errors out.

solution: force cmake to use the system python3.6, it must be an absolute path as well for some reason


jameshcorbett (approved):
Looks reasonable.

PR #1257, opened 2024-07-29
actions/docker: add ~el9~,f40,noble drop focal

trws:
problem: we're using old images and out of date WRT core

solution: pull in current versions and update build matrix

### DIFF
--- a/src/test/docker/fedora40/Dockerfile
+++ b/src/test/docker/fedora40/Dockerfile
@@ -1,0 +1,36 @@
+FROM fluxrm/flux-core:fedora40
+
+LABEL maintainer="Tom Scogland <scogland1@llnl.gov>"
+
+ARG USER=flux
+ARG UID=1000
+
+#  Install flux-sched dependencies
+RUN sudo dnf -y install \
+	boost-devel \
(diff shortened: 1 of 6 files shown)


trws:
NOTE: Dropping EL9 from this so it can proceed separately from the matrix repair upstream.  Will add that back in once flux-framwork/flux-core#6156 lands

src/test/docker/fedora40/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM fluxrm/flux-core:fedora40
+
+LABEL maintainer="Mark A. Grondona <mark.grondona@gmail.com>"
```
jameshcorbett, line 3: Is this right?

jameshcorbett (approved):
Looks reasonable, one question. He's listed as the maintainer on the other fedora dockerfiles, just checking that he's ok with continuing to be listed there.

src/test/docker/fedora40/Dockerfile
trws: I can change it at least here, good point.

PR #1265, opened 2024-08-08
qmanager: add get-stats and clear-stats callbacks

trws:
problem: it's hard to get information about the current state of queues withotu debugging fluxion

solution: add a stats callback that prints information about each queue including:
* The configuration and policy of the queue
* The counts of number of jobs transitioned into each state (total number ever added to pending, running, rejected, cancelled, etc.)
* For each pending queue an array of job-ids in F58 format ordered by their priority
* For each non-pending queue (running, rejected canceled) an array of job-ids ordered by job-id value.  These are transient queues that should never stay full for long, but just in case they're here anyway.

This does introduce a dependency on boost::json, which technically requires a higher version of boost. I think we already require a higher version for other reasons, but version 1.78 is already installed on TOSS4 from EPEL.  May need to tweak the rhel8 container, will see.  Here is some example output from a test where there are 10 jobs blocked on a drained node and two jobs that have completed:

```json
{
  "default": {
    "policy": "easy",
    "queue_depth": 32,
    "max_queue_depth": 1000000,
    "queue_parameters": {},
    "policy_parameters": {},
    "action counts": {
      "pending": 12,
      "running": 2,
      "reserved": 0,
      "rejected": 0,
      "complete": 2,
      "cancelled": 0,
      "reprioritized": 0
    },
    "pending_queues": {
      "pending": [],
      "pending_provisional": [],
      "blocked": [
        "fYDBt63",
        "fYEfsNP",
        "fYEfsNQ",
        "fYG9rej",
        "fYG9rek",
        "fYG9rem",
        "fYHdqw5",
        "fYHdqw6",
        "fYHdqw7",
        "fYK7qDR"
      ]
    },
    "scheduled_queues": {
      "running": [],
      "rejected": [],
      "canceled": []
    }
  }
}
```

There is plenty more information we could surface here, including the timestamps of jobs moving between states, current state each job considers itself to be in, etc. One thing I think we definitely want is something indicating partial release status, but we're currently not storing it so I'd rather do that after that gets worked out for final release.


CMakeLists.txt
```diff
@@ -89,7 +89,8 @@ pkg_check_modules(UUID REQUIRED IMPORTED_TARGET uuid)
 set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
-find_package(Boost 1.50 REQUIRED COMPONENTS
+find_package(Boost 1.75 REQUIRED COMPONENTS
```
garlick, line 92: If  this changes the build requirements it should probably be in a separate commit with a reason.

CMakeLists.txt
```diff
@@ -89,7 +89,8 @@ pkg_check_modules(UUID REQUIRED IMPORTED_TARGET uuid)
 set(Boost_USE_STATIC_LIBS OFF)
```
garlick, line 89: Commit message (and PR title): it's `stats-get` and `stats-clear` not the reverse.
Also, typo: `withotu`.

qmanager/modules/qmanager.cpp
```diff
@@ -652,6 +664,11 @@ int mod_start (flux_t *h, int argc, char **argv)
         qmanager_destroy (ctx);
         return rc;
     }
+    if ((rc = flux_msg_handler_addvec (h, statstab, (void *)ctx.get (), &ctx->stats_hndlr)) < 0) {
```
garlick, line 667: should be a corresponding `flux_msg_handler_delvec()` call in `qmanager_destroy()`

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -477,6 +512,28 @@ int qmanager_safe_cb_t::jobmanager_hello_cb (flux_t *h,
...
 
+void qmanager_safe_cb_t::jobmanager_stats_get_cb (flux_t *h,
+                                                  flux_msg_handler_t *w,
+                                                  const flux_msg_t *msg,
+                                                  void *arg)
+{
+    eh_wrapper_t exception_safe_wrapper;
+    exception_safe_wrapper (qmanager_cb_t::jobmanager_stats_get_cb, h, w, msg, arg);
+    if (exception_safe_wrapper.bad ())
+        flux_log_error (h, "%s: %s", __FUNCTION__, exception_safe_wrapper.get_err_message ());
```
garlick, line 523: `flux_log_error()` will append `strerror (errno)`.  Maybe you want `flux_log (h, LOG_ERR, ...)`?

qmanager/policies/queue_policy_fcfs.hpp
```diff
@@ -21,16 +21,19 @@ namespace detail {
...
-    virtual ~queue_policy_fcfs_t ();
-    virtual int run_sched_loop (void *h, bool use_alloced_queue);
-    virtual int reconstruct_resource (void *h, std::shared_ptr<job_t> job, std::string &R_out);
-    virtual int apply_params ();
-    virtual int handle_match_success (flux_jobid_t jobid,
-                                      const char *status,
-                                      const char *R,
-                                      int64_t at,
-                                      double ov);
-    virtual int handle_match_failure (flux_jobid_t jobid, int errcode);
```
garlick, line 33: Why no longer virtual?

qmanager/modules/CMakeLists.txt
```diff
@@ -11,4 +11,5 @@ target_link_libraries(sched-fluxion-qmanager PRIVATE
     flux::schedutil
     PkgConfig::JANSSON
     intern
+    Boost::json
```
garlick, line 14: It might be worth noting in the commit message why Boost::json is being used here when the rest of the project is using jansson.

garlick (commented):
I added a few quick comments.

If boost-json is now required as well as a newer version, the table in README.md should be updated.  Or actually what we did in flux-core was get rid of the table and add some scripts that just install the dependencies for rpm and deb based distros since updating the table became tedious:

https://github.com/flux-framework/flux-core/blob/master/scripts/install-deps-deb.sh

The debian/control file might need updating as well.

qmanager/modules/qmanager_callbacks.cpp
trws: Honestly I'm not completely sure.  The handler sets errno as part of catching a known exception, but then sets the error message based on the contents of the exception.  In some cases that will be duplicative, in others it might be useful to have both.  Will test out the output and see.

qmanager/policies/queue_policy_fcfs.hpp
trws: They are still virtual, this is an annoying corner of c++'s syntax.  If a base method is virtual, then all derived class methods of the same signature are also virtual, so having "virtual" on the declaration in the derived class can only ever hide errors. If you make a mistake copying the signature, argument types etc., then you get a _new_ virtual rather than the one you wanted to override. C++11 added the `override` keyword to mark that the method is meant to override a virtual from a base class, but to my immense annoyance decided to put it at the _end_ of the declaration, which is why that's added on these below.

As to why this is in this patch, it's because when I added `policy` as a new virtual, and correctly added `override` to the derived class versions, it activated the compiler warning for all the other methods that lacked `override`.

qmanager/modules/qmanager.cpp
trws: Thanks for that, I forgot that I had to add another of these and spaced on the destruction.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -324,6 +325,47 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
+        auto &pending_queues = obj["pending_queues"].emplace_object ();
+        add_queue (pending_queues["pending"].emplace_array (), m_pending);
+        add_queue (pending_queues["pending_provisional"].emplace_array (), m_pending_provisional);
+        add_queue (pending_queues["blocked"].emplace_array (), m_blocked);
+        auto &scheduled_queues = obj["scheduled_queues"].emplace_object ();
+        add_queue (scheduled_queues["running"].emplace_array (), m_running);
+        add_queue (scheduled_queues["rejected"].emplace_array (), m_rejected);
+        add_queue (scheduled_queues["canceled"].emplace_array (), m_canceled);
+    }
+
```
trws, line 368: This is sort-of a reply to the question of why boost::json.  I could rework this
to jansson, but to be honest it bugs me to be stuck using jansson pretty
regularly, and the choices for fixing that are either: 1) make a wrapper for it
that makes it work like nlohmann::json or boost::json or 2) use a library that
is less frustrating

To explain why it frustrates me:
* As we discussed the other day, `*_unpack ("{s:b}", "boolean", &b)` is a
  regular source of bugs, same deal for different int sizes and various other
  things because there's no static or dynamic checking of any kind for matching
  variadic argument sizes.
* Long sets, or nested sets, are really, really annoying to track.  Here's an
  example:
   ```c
   // formatted automatically
   json_pack ("{s:I s:I s:I s:{s:f s:f s:f s:f}}",
           "njobs",
           perf.succeeded.njobs,
           "njobs-reset",
           perf.succeeded.njobs_reset,
           "max-match-jobid",
           perf.succeeded.max_match_jobid,
           "max-match-iters",
           perf.succeeded.match_iter_count,
           "stats",
           "min",
           min,
           "max",
           perf.succeeded.max,
           "avg",
           avg,
           "variance",
           variance)
   // painstakingly manually formatted, not actually semantic
   json_pack ("{s:I s:I s:I s:I s:{s:f s:f s:f s:f}}",
           "njobs", perf.succeeded.njobs,
           "njobs-reset", perf.succeeded.njobs_reset,
           "max-match-jobid", perf.succeeded.max_match_jobid,
           "max-match-iters", perf.succeeded.match_iter_count,
           "stats",
               "min", min,
               "max", perf.succeeded.max,
               "avg", avg,
               "variance", variance)
   // C++ version, automatically formatted, structure is semantic
   value jv = {{"njobs", perf.succeeded.njobs},
               {"njobs-reset", perf.succeeded.njobs_reset},
               {"max-match-jobid", perf.succeeded.max_match_jobid},
               {"max-match-iters", perf.succeeded.match_iter_count},
               {"stats",
                {{"min", min},
                 {"max", perf.succeeded.max},
                 {"avg", avg},
                 {"variance", variance}}};
   ```
   Can you see the bug in either of the two C versions that would cause them to
   crash at runtime? I sometimes don't, while the nesting and type specification
   in the c++ version is completely structural and automated.
* Built-in support for common types, and built-in extensibility. I didn't do
  this for the maps in qmanager because we don't want to include the keys, but
  if we also wanted the running timestamp (key) from `m_running`, serializing
  the entire map is literally `obj["running"] = value_from(m_running)`
* jansson _requires_ dynamic memory allocations for absolutely every use-case
  unless you replace its allocators, which must be done globally in a
  non-thread-safe manner.  If I want to make parsing a message use stack space
  unless it can't fit, it's about 3 lines of code for one of the C++ native ones.


Now, all that said, we have choices.  These are varying amounts of work/pain but
they're all possible:
1. I can take the time to wrap jansson just enough to make it at least safe and
   vaguely non-painful to use.  It wouldn't be as nice, but at least give us
   something we can use to build up serializers, would probably have to re-read
   how to deal with `o` vs `O` yet again but it's possible.
2. We can use one of the C++ libraries, this has several sub-options:
   1. I could put nlohmann::json back in and we just use that, it's a single
      header file and doesn't change our requirements.  It doesn't have
      the nice memory management features but it has everything else, and is
      actually _nicer_ at the interface level. This one is about 30% faster than
      jansson for parsing.
   2. We use one that's actually massively faster than any other, that would
      probably be https://github.com/stephenberry/glaze, which is crazy fast and
      auto-generates serializers for structures. This is about 5x faster than
      nlohmann json, moreso if it uses compile-time layout for the
      deserialization structures.
   3. We use boost::json since it's already there on TOSS4, easy to get
      pre-built, has low impact on compile times, and is still 3 times
      faster than nlohmann json while being easier to use safely.
      We have sub-options here too:

      1. Use the pre-built version that requires it to be installed
      2. use the single header version so the requirments don't change at the
         cost of some build time and a non-trivial header being vendored


The current PR is based on doing 2.3.1.  Sorry for the book, but I've been
arguing with myself about where to fall in the choices list for a while, that's
part of why this took a bit to put together, and I'd appreciate input.

trws: @garlick, @grondo, @milroy, @jameshcorbett, would love to have your input here. It really has been bugging me how to balance benefits/costs here.

CMakeLists.txt
trws: Got it, fixed and thanks!

qmanager/modules/qmanager.cpp
trws: Fixed, I was glad to see that valgrind actually caught this, that's what the build failures were if you happened to see them.

qmanager/modules/qmanager_callbacks.cpp
trws: Here's what we get if I throw a `std::out_of_range("stuff and things")` exception under one of these:

```
Aug 08 21:45:11.801485 UTC sched-fluxion-qmanager.err[0]: jobmanager_stats_get_cb: ENOENT: stuff and things: No such file or directory
```

The message from the exception gets printed in addition to the translation of
the exception into an errno value.  That's probably not ideal since there is
likely to be a better way to do that in some cases, but until I get time to make
this more general and fix it for the rest I'm tempted to leave it like this so
we get both.

CMakeLists.txt
grondo: Just reporting here that this PR doesn't build on my jammy system due to this change in the Boost requirement:
```
CMake Error at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost: Found unsuitable version "1.74.0", but required is at
  least "1.75" (found /usr/include, found components: system filesystem graph
  regex)
Call Stack (most recent call first):
  /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:592 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.22/Modules/FindBoost.cmake:2360 (find_package_handle_standard_args)
  CMakeLists.txt:92 (find_package)

```

trws: Dang, yes, boost::json is added in 1.75, and jammy only offers 1.74. Will add a note to the other thread but this means having it pre-built in supported distros is not an option because of jammy, that might change the calculus somewhat.

qmanager/policies/base/queue_policy_base.hpp
trws: As @grondo points out, jammy lacks a sufficient boost version to assume boost::json is available in the upstream repos, which means 2.3.1 is probably not going to cut it, so either vendor it, or use one of the other options are looking more workable.

garlick: Great points.

I don't especially have a strong opinion.  As a sometimes maker of trivial patches to flux-sched, I'll have an easier go of it if we stick with jansson since it's what we are using in core, but that shouldn't really be a big factor.  Also minor, we do have some pack/unpack interfaces exposed in the core API so if you're not using jansson obviously you cannot pack/unpack objects directly.

Seems like you'd want to have a plan to migrate the project rather than let multiple libraries be used, or it could get a bit confusing.

If this breaks jammy (ubuntu 22.04) builds then that sucks though.

trws: Yeah breaking jammy is unacceptable, that needs to work, so it has to be one of the others.  I'll see if I can manage the "wrap jansson" yak shave with relatively low overhead and if so we can just do that.

trws:
Ok, this took far _far_ too long to rework to use jansson. I think I ended up spending 5 hours figuring out what was wrong with my format strings and working out the one-off jansson interfaces to make this work, it might have been more.  This at least gets the thing working, and starts to give us something to manage jansson values and build up serializers.  It also fixes the memory leaks in resource from unfreed jansson objects is taken care of.

Hopefully this is good enough for the release, and I'll spend a bit more time to make something that has the uniform interface of object and array-like things in C++, because this feels like a loss:

```diff
     {
-        using namespace boost::json;
-        auto &obj = jv.emplace_object ();
-        obj = {{"policy", this->policy ()},
-               {"queue_depth", m_queue_depth},
-               {"max_queue_depth", m_max_queue_depth},
-               {"queue_parameters", value_from (m_qparams)},
-               {"policy_parameters", value_from (m_pparams)},
-               {"action counts",
-                {{"pending", m_pq_cnt},
-                 {"running", m_rq_cnt},
-                 {"reserved", m_oq_cnt},
-                 {"rejected", m_dq_cnt},
-                 {"complete", m_cq_cnt},
-                 {"cancelled", m_cancel_cnt},
-                 {"reprioritized", m_reprio_cnt}}}};
+        json::value qparams;
+        to_json (qparams, m_qparams);
+        json::value pparams;
+        to_json (pparams, m_pparams);
         char buf[128] = {};
-        auto add_queue = [&] (::boost::json::array &a, auto &map) {
+        auto add_queue = [&] (json_t *a, auto &map) {
             for (auto &[k, jobid] : map) {
                 if (flux_job_id_encode (jobid, "f58plain", buf, sizeof buf) < 0)
-                    a.push_back (jobid);
+                    json_array_append_new (a, json_integer (jobid));
                 else
-                    a.push_back (buf);
+                    json_array_append_new (a, json_string (buf));
             }
         };
-        auto &pending_queues = obj["pending_queues"].emplace_object ();
-        add_queue (pending_queues["pending"].emplace_array (), m_pending);
-        add_queue (pending_queues["pending_provisional"].emplace_array (), m_pending_provisional);
-        add_queue (pending_queues["blocked"].emplace_array (), m_blocked);
-        auto &scheduled_queues = obj["scheduled_queues"].emplace_object ();
-        add_queue (scheduled_queues["running"].emplace_array (), m_running);
-        add_queue (scheduled_queues["rejected"].emplace_array (), m_rejected);
-        add_queue (scheduled_queues["canceled"].emplace_array (), m_canceled);
+        json::value pending;
+        pending.emplace_object ();
+        json::value pending_arr;
+        pending_arr.emplace_array ();
+        json_object_set (pending.get (), "pending", pending_arr.get ());
+        add_queue (pending_arr.get (), m_pending);
+        pending_arr.emplace_array ();
+        json_object_set (pending.get (), "pending_provisional", pending_arr.get ());
+        add_queue (pending_arr.get (), m_pending_provisional);
+        pending_arr.emplace_array ();
+        json_object_set (pending.get (), "blocked", pending_arr.get ());
+        add_queue (pending_arr.get (), m_blocked);
+
+        json::value scheduled;
+        scheduled.emplace_object ();
+        json::value scheduled_arr;
+        scheduled_arr.emplace_array ();
+        json_object_set (scheduled.get (), "running", scheduled_arr.get ());
+        add_queue (scheduled_arr.get (), m_running);
+        scheduled_arr.emplace_array ();
+        json_object_set (scheduled.get (), "rejected", scheduled_arr.get ());
+        add_queue (scheduled_arr.get (), m_rejected);
+        scheduled_arr.emplace_array ();
+        json_object_set (scheduled.get (), "canceled", scheduled_arr.get ());
+        add_queue (scheduled_arr.get (), m_canceled);
+
+        json_error_t err = {0};
+        jv = json::value (json::no_incref{},
+                          json_pack_ex (&err,
+                                        0,
+                                        // begin object
+                                        "{"
+                                        // policy
+                                        "s:s%"
+                                        // queue_depth
+                                        "s:I"
+                                        // max_queue_depth
+                                        "s:I"
+                                        // queue parameters
+                                        "s:O"
+                                        // policy parameters
+                                        "s:O"
+                                        // action counts
+                                        "s:o"
+                                        // pending queues
+                                        "s:O"
+                                        // scheduled queues
+                                        "s:O"
+                                        // end object
+                                        "}",
+                                        // VALUE START
+                                        // policy, str+length style
+                                        "policy",
+                                        this->policy ().data (),
+                                        this->policy ().length (),
+                                        // queue_depth
+                                        "queue_depth",
+                                        (json_int_t)m_queue_depth,
+                                        // max_queue_depth
+                                        "max_queue_depth",
+                                        (json_int_t)m_max_queue_depth,
+                                        // queue parameters
+                                        "queue_parameters",
+                                        qparams.get (),
+                                        // policy parameters
+                                        "policy_parameters",
+                                        pparams.get (),
+                                        // action counts
+                                        "action_counts",
+                                        json_pack ("{s:I s:I s:I s:I s:I s:I s:I}",
+                                                   "pending",
+                                                   m_pq_cnt,
+                                                   "running",
+                                                   m_rq_cnt,
+                                                   "reserved",
+                                                   m_oq_cnt,
+                                                   "rejected",
+                                                   m_dq_cnt,
+                                        "max_queue_depth",
+                                        (json_int_t)m_max_queue_depth,
+                                        // queue parameters
+                                        "queue_parameters",
+                                        qparams.get (),
+                                        // policy parameters
+                                        "policy_parameters",
+                                        pparams.get (),
+                                        // action counts
+                                        "action_counts",
+                                        json_pack ("{s:I s:I s:I s:I s:I s:I s:I}",
+                                                   "pending",
+                                                   m_pq_cnt,
+                                                   "running",
+                                                   m_rq_cnt,
+                                                   "reserved",
+                                                   m_oq_cnt,
+                                                   "rejected",
+                                                   m_dq_cnt,
+                                                   "complete",
+                                                   m_cq_cnt,
+                                                   "cancelled",
+                                                   m_cancel_cnt,
+                                                   "reprioritized",
+                                                   m_reprio_cnt),
+                                        // pending queues
+                                        "pending_queues",
+                                        pending.get (),
+                                        // scheduled queues
+                                        "scheduled_queues",
+                                        scheduled.get ()));
+        if (!jv.get ()) {
+            throw std::runtime_error (err.text);
+        }
     }
```

src/common/c++wrappers/jansson.hpp
```diff
@@ -0,0 +1,127 @@
+//
+// Created by Scogland, Tom on 8/11/24.
+//
```
milroy, line 3: Needs the LLNL block:
```suggestion
/*****************************************************************************\
 * Copyright 2024 Lawrence Livermore National Security, LLC
 * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
 *
 * This file is part of the Flux resource manager framework.
 * For details, see https://github.com/flux-framework.
 *
 * SPDX-License-Identifier: LGPL-3.0
\*****************************************************************************/
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -324,6 +326,121 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
+        scheduled_arr.emplace_array ();
+        json_object_set (scheduled.get (), "rejected", scheduled_arr.get ());
+        add_queue (scheduled_arr.get (), m_rejected);
+        scheduled_arr.emplace_array ();
+        json_object_set (scheduled.get (), "canceled", scheduled_arr.get ());
+        add_queue (scheduled_arr.get (), m_canceled);
+
+        json_error_t err = {0};
+        jv = json::value (json::no_incref{},
+                          json_pack_ex (&err,
```
milroy, line 377: This is pretty brutal, and I can only imagine the effort involved to get this to work. In the interest of getting the functionality out for the release, and in respect for the amount of work you put into wrapping jansson it's okay with me.

milroy (approved):
> Hopefully this is good enough for the release, and I'll spend a bit more time to make something that has the uniform interface of object and array-like things in C++, because this feels like a loss:

Improving the interface will certainly be helpful. I hope interface improvement isn't a rabbit hole but don't have enough experience with jansson to evaluate the risk.

I think we should test the other JSON libraries you listed to see if they have a substantial impact on Fluxion performance. I suspect partial cancel can be accelerated with a faster JSON library. I also wonder whether JSON manipulation will become a bottleneck with large JGFs.

With those reservations, I'm good to get this into the next release.

qmanager/policies/base/queue_policy_base.hpp
milroy: As I noted in the comment below, I think we should do a performance evaluation of the JSON library options, especially in light of the upcoming transition to JGF.

src/common/c++wrappers/jansson.hpp
trws: Ah, thanks.  I just went with the auto-generated one and completely forgot.

PR #1266, opened 2024-08-10
Support and handle optional "final" flag in .free RPC

milroy:
Recently, a large job on a large system was considered allocated by Fluxion, but was complete and released in flux-core (https://github.com/flux-framework/flux-core/issues/6179). The proposed solution was to amend RFC 27 to include an optional "final" boolean flag in the `.free` RPC. That flag can be used by Fluxion to determine if there is an allocation state discrepancy between flux-core and sched and take action based on that information.

This PR adds support to handle the optional "final" flag in the `.free` RPC. ~The current implementation only logs an error when a state discrepancy is detected, but doing a full cancellation of the job would be a straightforward addition to this PR.~ After further thought, executing a full cancel when a discrepancy exists between flux-core and -sched seems like a better approach since this state may go unnoticed by administrators.

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -621,14 +621,17 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
      *                   service module such as qmanager, it is expected
      *                   to be a pointer to a flux_t object.
      *  \param id        jobid of flux_jobid_t type.
+     *  \param final     bool indicating if this is the final partial release RPC
+     *                   for this jobid.
      *  \param R         Resource set for partial cancel
      *  \return          0 on success; -1 on error.
      *                       ENOENT: unknown id.
      */
-    int remove (void *h, flux_jobid_t id, const char *R)
+    int remove (void *h, flux_jobid_t id, bool final, const char *R)
     {
         int rc = -1;
         bool full_removal = false;
+        flux_t *flux_h = static_cast<flux_t *> (h);
 
         auto job_it = m_jobs.find (id);
         if (job_it == m_jobs.end ()) {
@@ -643,28 +646,55 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
             case job_state_kind_t::ALLOC_RUNNING:
                 // deliberately fall through
             case job_state_kind_t::RUNNING:
-                if ((rc = cancel (h, job_it->second->id, R, true, full_removal) != 0))
-                    break;
-                if (full_removal) {
+                if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
+                    flux_log_error (flux_h,
+                                    "%s: .free RPC partial cancel failed for jobid "
+                                    "%jd",
+                                    __FUNCTION__,
+                                    static_cast<intmax_t> (id));
+                    errno = EINVAL;
+                    goto out;
+                }
+                // We still want to run the sched loop even if there's an inconsistent state
+                set_schedulability (true);
+                if (full_removal || final) {
                     m_alloced.erase (job_it->second->t_stamps.running_ts);
                     m_running.erase (job_it->second->t_stamps.running_ts);
                     job_it->second->t_stamps.complete_ts = m_cq_cnt++;
                     job_it->second->state = job_state_kind_t::COMPLETE;
                     m_jobs.erase (job_it);
+                    if (final && !full_removal) {
+                        // This error condition indicates a discrepancy between core and sched.
+                        flux_log_error (flux_h,
+                                        "%s: Final .free RPC failed to remove all resources for "
+                                        "jobid "
+                                        "%jd",
+                                        __FUNCTION__,
+                                        static_cast<intmax_t> (id));
+                        // Run a full cancel to clean up all remaining allocated resources
+                        if (cancel (h, job_it->second->id, true) != 0) {
+                            flux_log_error (flux_h,
+                                            "%s: .free RPC full cancel failed for jobid "
+                                            "%jd",
+                                            __FUNCTION__,
+                                            static_cast<intmax_t> (id));
+                        }
+                        errno = EPROTO;
+                        goto out;
+                    }
                 }
-                set_schedulability (true);
                 break;
             default:
                 break;
         }
+
+        rc = 0;
+    out:
         cancel_sched_loop ();
         // blocked jobs must be reconsidered after a job completes
         // this covers cases where jobs that couldn't run because of an
         // existing job's reservation can when it completes early
         reconsider_blocked_jobs ();
-
-        rc = 0;
-    out:
         return rc;
     }
 
@@ -688,6 +718,26 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
     virtual int cancel (void *h, flux_jobid_t id, const char *R, bool noent_ok, bool &full_removal)
     {
         full_removal = true;
+        return 0;
+    }
+
+    /*! Remove a job whose jobid is id from any internal queues
+     *  (e.g., pending queue, running queue, and alloced queue.)
+     *  If succeeds, it changes the pending queue or resource
+     *  state. This queue becomes "schedulable" if pending job
+     *  queue is not empty: i.e., is_schedulable() returns true;
+     *
+     *  \param h         Opaque handle. How it is used is an implementation
+     *                   detail. However, when it is used within a Flux's
+     *                   service module such as qmanager, it is expected
+     *                   to be a pointer to a flux_t object.
+     *  \param id        jobid of flux_jobid_t type.
+     *  \param noent_ok  don't return an error on nonexistent jobid
+     *  \return          0 on success; -1 on error.
+     *                       ENOENT: unknown id.
+     */
+    virtual int cancel (void *h, flux_jobid_t id, bool noent_ok)
+    {
         return 0;
     }
 
--- a/t/issues/t6179-flux-core-housekeeping.sh
+++ b/t/issues/t6179-flux-core-housekeeping.sh
@@ -1,0 +1,164 @@
+#!/bin/bash
+#
+#  Ensure fluxion conforms to updated RFC 27 and behaves as desired if resource state
+#  discrepancies exit between flux-core and -sched.
+#
+
+log() { printf "flux-coreissue#6179: $@\n" >&2; }
+
+cat <<'EOF' >free.py
+import flux
+import json
+import sys
+import subprocess as sp
+from flux.job import JobID
+
+jobid = JobID(sys.argv[1])
+r_obj=json.loads(sp.check_output(['flux', 'job', 'info', sys.argv[1], 'R']).decode())
+obj = {'id': jobid, 'R': r_obj, 'final': True}
+flux.Flux().rpc('sched.free', obj)
+sys.exit(0)
+EOF
+
+cat <<'EOF' >incomplete-free.py
+import flux
+import json
+import sys
+import subprocess as sp
+from flux.job import JobID
+
+jobid = JobID(sys.argv[1])
+R_str = '{"version": 1, "execution": {"R_lite": [{"rank": "1", "children": {"core": "0-15", "gpu": "0-3"}}], "nodelist": ["node1"]}}'
+r_obj = json.loads(R_str)
+obj = {'id': jobid, 'R': r_obj, 'final': True}
+flux.Flux().rpc('sched.free', obj)
+sys.exit(0)
+EOF
+
+cat <<EOF >flux.config
+[sched-fluxion-resource]
+match-policy = "lonodex"
+match-format = "rv1_nosched"
+
+[resource]
+noverify = true
+norestrict = true
+
+[[resource.config]]
+hosts = "node[0-1]"
+cores = "0-15"
+gpus = "0-3"
+EOF
+
+log "Unloading modules..."
+flux module remove sched-simple
+flux module remove resource
+
+flux config load flux.config
+
+flux module load resource monitor-force-up
+flux module load sched-fluxion-resource
+flux module load sched-fluxion-qmanager queue-policy="easy"
+flux queue start --all --quiet
+flux resource list
+flux resource status
+flux module list
+
+log "Running test job 1"
+jobid1=$(flux submit --wait-event=alloc -N2 -t 1h --setattr=exec.test.run_duration=1m sleep inf)
+log "Sending final RPC for job 1"
+flux python ./free.py ${jobid1}
+state=$( flux ion-resource find "sched-now=allocated" )
+if [[ ${state} != *"null"* ]]; then
+    # retry since ./free.py isn't blocking
+    state=$( flux ion-resource find "sched-now=allocated" )
+    if [[ ${state} != *"null"* ]]; then
+        # .free didn't release all resources
+        exit 1
+    fi
+fi
+
+# Need to execute cancel to remove from job manager
+flux cancel ${jobid1}
+flux job wait-event -t 5 ${jobid1} release
+
+log "Running test job 2"
+jobid2=$(flux submit --wait-event=alloc -N2 -t 1h --setattr=exec.test.run_duration=1m sleep inf)
+log "Sending final RPC for job 2"
+flux python ./incomplete-free.py ${jobid2}
+state=$( flux ion-resource find "sched-now=allocated" )
+if [[ ${state} != *"null"* ]]; then
+    # retry since ./free.py isn't blocking
+    state=$( flux ion-resource find "sched-now=allocated" )
+    if [[ ${state} != *"null"* ]]; then
+        # .free didn't release all resources
+        exit 1
+    fi
+fi
+
+# Need to execute cancel to remove from job manager
+flux cancel ${jobid2}
+flux job wait-event -t 5 ${jobid2} release
+flux jobs -a
+
+log "reloading sched-simple..."
+flux module remove sched-fluxion-qmanager
+flux module remove sched-fluxion-resource
+flux module load sched-simple
+
+log "Unloading modules for FCFS test..."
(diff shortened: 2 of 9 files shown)


t/issues/t6179-flux-core-housekeeping.sh
```diff
@@ -0,0 +1,125 @@
...
+import json
+import sys
+import subprocess as sp
+from pprint import pprint
+
+jobid = int(sp.check_output(['flux', 'job', 'id', sys.argv[1]]).decode())
+r_obj=json.loads(sp.check_output(['flux', 'job', 'info', sys.argv[1], 'R']).decode())
+obj = {'id': jobid, 'R': r_obj, 'final': True}
+pprint(obj)
+flux.Flux().rpc('sched.free', obj)
```
milroy, line 19: Note that adding a `.get()` to this line causes the script to hang indefinitely. The action occurs as expected without `.get()`. I'm a bit concerned about the behavior but haven't figured out how to fix it.

Note that there's no real error checking in this test. I think the best way to check for errors would be to parse the  RPC output...

milroy:
@trws, this PR is ready for review. There are still two issues I'm trying to fix:

1. See the note about `.get()` hanging for the `.free` RPC in the testsuite test
2. `clang-format` is changing one or more files in this PR, and I can't reproduce it in the container in my dev environment

milroy:
@garlick or @grondo, it may be good for one of you to take a look at this, too.

t/issues/t6179-flux-core-housekeeping.sh
grondo: The `sched.free` RPC doesn't have a response (see the documentation of the [sched.free](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_27.html#free) request in RFC 27).

t/issues/t6179-flux-core-housekeeping.sh
```diff
@@ -0,0 +1,125 @@
...
+EOF
+
+cat <<'EOF' >incomplete-free.py
+import flux
+import json
+import sys
+import subprocess as sp
+from pprint import pprint
+
+jobid = int(sp.check_output(['flux', 'job', 'id', sys.argv[1]]).decode())
```
grondo, line 30: Instead of using a subprocess to run `flux job id`, just use the Python `JobID` class:

```python
from flux.job import JobID

jobid = JobID(sys.argv[1])
```

t/issues/t6179-flux-core-housekeeping.sh
```diff
@@ -0,0 +1,125 @@
...
+
+cat <<EOF >flux.config
+[sched-fluxion-resource]
+match-policy = "lonodex"
+match-format = "rv1_nosched"
+
+[resource]
+noverify = true
+norestrict = true
+path = "R"
```
grondo, line 51: Instead of using a manually created _R_, it might be easier to define `resource.config` in the config file, e.g.
```toml
[[resource.config]]
hosts = "node[0-1]"
cores = "0-15"
gpus = "0-3"
```

Only a suggestion or illustrative for next time since _R_ works fine here too.

t/issues/t6179-flux-core-housekeeping.sh
```diff
@@ -0,0 +1,125 @@
...
+flux module load sched-fluxion-qmanager queue-policy="easy"
+flux queue start --all --quiet
+flux resource list
+flux resource status
+flux module list
+
+log "Running test job 1"
+jobid1=$(flux submit -N2 -t 1h --setattr=exec.test.run_duration=1m sleep inf)
+log "Sending final RPC for job 1"
+flux python ./free.py ${jobid1}
```
grondo, line 71: There may be a small race condition that could cause this sequence to fail if the job submitted above doesn't get an `alloc` event before `free.py` is run. In that case the scheduler hasn't yet seen the job so `free.py` would be failing (though as you note, this currently isn't checked)

Add `--wait-event=alloc` to `flux submit` to ensure the scheduler has allocated resources before  proceeding with the test.

grondo: This probably goes for the other jobs submitted in the test as well.

t/issues/t6179-flux-core-housekeeping.sh
```diff
@@ -0,0 +1,125 @@
...
+# Need to execute cancel to remove from job manager
+flux cancel ${jobid1}
+flux job wait-event -t 5 ${jobid1} release
```
grondo, line 74: Between the free and the cancel is there some way to check that the job (eventually) reaches a consistent state? That is, could you send a query to Fluxion (perhaps `sched-fluxion-resource.resource-status`) to ensure all resources have been released with the `final=true` flag? (There might have to be a retry since `free.py` won't block until a response...)

grondo (commented):
Just adding a few suggestions to the test, since I 'm not qualified to fully review the other parts of the PR.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -504,14 +504,18 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
      *  \return          0 on success; -1 on error.
      *                       ENOENT: unknown id.
      */
-    int remove (void *h, flux_jobid_t id, const char *R)
+    int remove (void *h, flux_jobid_t id, bool final, const char *R)
     {
         int rc = -1;
         bool full_removal = false;
+        flux_t *flux_h = nullptr;
+        flux_h = static_cast<flux_t *> (h);
```
trws, line 518:
```suggestion
        flux_t *flux_h = static_cast<flux_t *> (h);
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -526,14 +530,34 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
-                    break;
-                if (full_removal) {
+                if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
+                    flux_log_error (flux_h,
+                                    "%s: .free RPC partial cancel failed for jobid "
+                                    "%jd",
+                                    __FUNCTION__,
+                                    static_cast<intmax_t> (id));
+                    errno = EPROTO;
+                    goto out;
```
trws, line 540: This logic looks really good, I like the flow.  One concern I have though is that the two new `goto out` lines skip the `cancel_sched_loop ();` and `reconsider_blocked_jobs ()` lines where originally the failure path used `break` to drop out of the switch and kept those.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -526,14 +530,34 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
-                if ((rc = cancel (h, job_it->second->id, R, true, full_removal) != 0))
-                    break;
-                if (full_removal) {
+                if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
+                    flux_log_error (flux_h,
+                                    "%s: .free RPC partial cancel failed for jobid "
+                                    "%jd",
+                                    __FUNCTION__,
+                                    static_cast<intmax_t> (id));
+                    errno = EPROTO;
```
trws, line 539: Is this a protocol error? What causes a cancel failure at this point, invalid R?  The other one definitely seems like EPROTO but I would have guessed EINVAL here.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -526,14 +530,34 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
+                        // Run a full cancel to clean up all remaining allocated resources
+                        cancel (h, job_it->second->id, true);
+                        flux_log_error (flux_h,
+                                        "%s: Final .free RPC failed to remove all resources for "
+                                        "jobid "
+                                        "%jd",
+                                        __FUNCTION__,
+                                        static_cast<intmax_t> (id));
+                        errno = EPROTO;
+                        goto out;
```
trws, line 559: I may be misreading the logic, but it seems like if the second cancel succeeds,
it should be logged but probably shouldn't return error? Feel free to push back
on this one since I'm not sure what effects are at the callsite outside this but
if the resources are freed that seems substantially less broken than if some of
them couldn't be removed.

```suggestion
                        flux_log_error (flux_h,
                                        "%s: Final .free RPC failed to remove all resources for "
                                        "jobid "
                                        "%jd",
                                        __FUNCTION__,
                                        static_cast<intmax_t> (id));

                        if (cancel (h, job_it->second->id, true) != 0) {
                            errno = EPROTO;
                            goto out;
                        }
```

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -526,14 +530,34 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
+                    }
                 }
                 set_schedulability (true);
```
trws, line 562: I think this line now needs to move up, maybe even to the start of this case? We should run the sched loop after any clearing of resources, even if something went slightly wrong, otherwise we might block for lack of incoming events.

trws (changes requested):
There are a couple bits to think about in the control flow, but this is a great improvement.  I think the `.get()` issue is just the lack of response as @grondo mentioned, so that should be fine.

t/issues/t6179-flux-core-housekeeping.sh
milroy: Good suggestion! Implemented.

t/issues/t6179-flux-core-housekeeping.sh
milroy: Also good idea! I'm not sure why I wrote it the other way originally...

t/issues/t6179-flux-core-housekeeping.sh
milroy: > Add --wait-event=alloc to flux submit to ensure the scheduler has allocated resources before proceeding with the test.

Thanks for the suggestion! Implemented.

t/issues/t6179-flux-core-housekeeping.sh
milroy: Good idea. I can do that with `flux ion-resource find "sched-now=allocated"`.

qmanager/policies/base/queue_policy_base.hpp
milroy: Oof, thanks for catching that.

qmanager/policies/base/queue_policy_base.hpp
milroy: I was going back and forth on this in part due to the use of `goto` if the job `id` isn't found in `m_jobs`: https://github.com/flux-framework/flux-sched/blob/6d2dc56d316c443216dc96ab6c03b46df33bc602/qmanager/policies/base/queue_policy_base.hpp#L517

But I suppose this error shouldn't prevent `cancel_sched_loop ()` and `reconsider_blocked_jobs ()` from getting called.

qmanager/policies/base/queue_policy_base.hpp
milroy: Yes, that's one possibility. A general traverser error can cause it, too. I'm actually not sure if we want to log this error here since it gets logged in the `qmanager_free_cb` and in `resource_match`. As in the code above I think the return code shouldn't be `0` if `cancel` experiences an error.

qmanager/policies/base/queue_policy_base.hpp
milroy: However, I think the return code shouldn't be `0` if `cancel` experiences an error, so I'll need to adjust this commit a bit to use a `break` and return `-1`.

qmanager/policies/base/queue_policy_base.hpp
milroy: How about moving `set_schedulability (true);` right before the return?

qmanager/policies/base/queue_policy_base.hpp
milroy: > I may be misreading the logic, but it seems like if the second cancel succeeds,
it should be logged but probably shouldn't return error? 

I didn't write this in a very clear fashion. There are actually two error conditions here, and the current implementation only logs one. I've updated the code to log both.

qmanager/policies/base/queue_policy_base.hpp
milroy: After more thought, I think logging an error if `cancel` fails provides useful information.

milroy:
> Take a look at https://github.com/flux-framework/flux-sched/pull/1267 and maybe rebase on there, I didn't realize we wouldn't get the format change on failure, and we definitely should so that adds it.

Yeah, that PR identified the formatting issue. Thanks!

t/issues/t6179-flux-core-housekeeping.sh
milroy: I'm not sure of the best way to perform a retry, though.

qmanager/policies/base/queue_policy_base.hpp
trws: That sounds like a good plan.  The main difference with the m_jobs check is that if there's no job, we don't actually do any canceling, there's no effect to cause the sched loop to need to update something.  If there's an error from cancel that means the same thing "no removal took place" then we should probably do the same, but if the set of free resources changed then the `set_schedulable` and the other two need to run.

qmanager/policies/base/queue_policy_base.hpp
trws: Ok, that seems reasonable, thanks @milroy.

qmanager/policies/base/queue_policy_base.hpp
trws: That would work too, as long as we're doing it in every case where something actually changes.  If we can avoid calling it when nothing changes, that keeps resource usage down, but setting it too often should still execute correctly.

qmanager/policies/base/queue_policy_base.hpp
milroy: > If there's an error from cancel that means the same thing "no removal took place" then we should probably do the same, but if the set of free resources changed then the set_schedulable and the other two need to run.

Unfortunately, I don't think `cancel` is atomic.

qmanager/policies/base/queue_policy_base.hpp
milroy: The guideline of setting schedulability only if the resource state changes does suggest moving`set_schedulability (true);` is the best approach with the understanding that `cancel` isn't atomic.

t/issues/t6179-flux-core-housekeeping.sh
milroy: I missed that the RPC doesn't have a response.

t/issues/t6179-flux-core-housekeeping.sh
milroy: I decided to just rerun the check.

milroy:
@trws and @grondo this PR is ready for another look.

qmanager/policies/base/queue_policy_base.hpp
trws: What do you mean by atomic in this context?

milroy: There are failure modes where it partially removes an allocation (in the case of full cancellation) or part of the set of resources designated to be deallocated (in the case of partial cancel). So if `cancel` fails the resource graph isn't necessarily in the same state it was in before the `cancel`.

trws: Gotcha, makes sense.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -526,28 +529,55 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
+                    goto out;
+                }
+                if (full_removal || final) {
                     m_alloced.erase (job_it->second->t_stamps.running_ts);
                     m_running.erase (job_it->second->t_stamps.running_ts);
                     job_it->second->t_stamps.complete_ts = m_cq_cnt++;
                     job_it->second->state = job_state_kind_t::COMPLETE;
                     m_jobs.erase (job_it);
+                    // We still want to run the sched loop even if there were errors
+                    set_schedulability (true);
```
trws, line 548: I think this now runs only on a full removal or a final release right? Maybe it should go just above the enclosing if statement to catch non-final partial releases as well?

trws (approved):
Ok, I'm happy with this pending the move of the set_schedulability call so we still start a loop on a successful partial release.  Good stuff @milroy!

qmanager/policies/base/queue_policy_base.hpp
milroy: You're right. I moved the statement right before the enclosing `if`.

milroy: Unfortunately, `t1001-qmanager-basic.t` fails with that change. Investigating...

milroy: It's the new check for qmanager stats:
```bash
expecting success: 
    flux module stats sched-fluxion-qmanager | tee stats.json &&
    jq -e ".queues.default.action_counts.rejected == 1" stats.json &&
    jq -e ".queues.default.action_counts.complete >= 4" stats.json

{
 "tx": {
  "request": 27,
  "response": 5,
  "event": 0,
  "control": 0
 },
 "rx": {
  "request": 10,
  "response": 22,
  "event": 0,
  "control": 0
 }
}
false
not ok 9 - qmanager: check stats
```

trws: It also looks like CI is passing, maybe a local rebuild issue?

milroy: Yeah, it's a local build issue.

milroy:
Thanks for the feedback @trws and @grondo! Setting MWP.

PR #1267, opened 2024-08-12
pre-commit: show diff on failure

trws:
problem: in CI we can't see what formatting was changed by the hooks

solution: add the `--show-diff-on-failure` flag so we can see what happened in case it doesn't repro easily elsewhere


grondo (approved):
Fwiw, LGTM!

milroy (approved):
LGTM! Let's get this merged.

PR #1268, opened 2024-08-13
policy: expose internals of resource module policy factory to allow custom policies

wihobbs:
Problem: the current fluxion policy matching factory doesn't allow for customizable policies.

Add a custom option and refactor existing policies to behave similarly. Add unit testing for the string parsing convenience functions.

Update documentation for existing and new custom match policies.

Outstanding to-do:

- [x] add an option in the scheduler config to actually pass a custom string in, and probably some validation too

### DIFF
--- a/doc/man5/flux-config-sched-fluxion-resource.rst
+++ b/doc/man5/flux-config-sched-fluxion-resource.rst
@@ -102,6 +102,45 @@ hinodex
 first
     Select the first matching resources and stop the search
 
+firstnodex
+    A node-exclusive scheduling whose behavior is identical to
+    ``first`` except each compute node is exclusively allocated.
+
+
+CUSTOM MATCH POLICIES
+=====================
+Match policies are initialized as collections of individual attributes
+that help the scheduler to select matches. For convenience, these 
+attributes are exposed to users such that they can write custom policies.
+Below is a list of match attributes which can be selected by users.
+
+policy
+    Allowed options are ``low`` or ``high``. If only ``policy=low``
+    or ``policy=high`` is specified, the behavior of the match policy is the
+    same as if ``match-policy=low`` or ``match-policy=high`` were selected,
+    respectively.
+
+node_centric
+    ``true`` or ``false`` are allowed options. Evaluate matches based on the
+    ID of the compute node first. 
+
+node_exclusive
+    ``true`` or ``false`` are allowed options. Exclusively allocate compute
+    nodes when a match is found.
+
+set_stop_on_1_matches
+    ``true`` or ``false`` are allowed options. When a match is found, take
+    it, without evaluating for potentially more optimal matches.
+
+::
+
+    [sched-fluxion-resource]
+
+    # system instance will use node-exclusive
+    # scheduling first-match (with nodes of high node IDs
+    # selected first).
+    match-policy = "policy=high node_exclusive=true set_stop_on_1_matches=true"
+
 
 EXAMPLE
 =======
--- a/resource/policies/base/test/matcher_policy_factory_test02.cpp
+++ b/resource/policies/base/test/matcher_policy_factory_test02.cpp
@@ -1,0 +1,93 @@
+/*****************************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+/*
+ * Test for the dfu_match_policy class(es). Compares shared pointers
+ * expected values to string inputs to these classes. Strings can be
+ * specified in config for custom policies, or some are provided.
+ */
+
+extern "C" {
+#if HAVE_CONFIG_H
+#include <config.h>
+#endif
+}
+#include <iostream>
+#include <string>
+#include <boost/lexical_cast.hpp>
+#include "resource/policies/dfu_match_policy_factory.hpp"
+#include "src/common/libtap/tap.h"
+
+using namespace Flux;
+using namespace Flux::resource_model;
+
+int test_parsers ()
+{
+    std::map<std::string, bool> container;
+    std::string e = "";
+    bool first = Flux::resource_model::parse_custom_match_policy (
+        "policy=high node_centric=true node_exclusive=true stop_on_1_matches=false blah=4",
+        container,
+        e);
+    ok (first == false, "blah is an unrecognized policy option");
+    ok (e == "invalid policy option: blah\n", "correct error message for invalid option");
+
+    std::map<std::string, bool> container2;
+    std::string e2 = "";
+    bool second = Flux::resource_model::parse_custom_match_policy (
+        "policy=1 node_centric=true node_exclusive=true stop_on_1_matches=false", container2, e2);
+    ok (second == false, "1 is an invalid option, must be high or low");
+    ok (e2 == "policy key within custom policy accepts only low or high\n",
+        "correct error message for high/low");
+
+    std::map<std::string, bool> container3;
+    std::string e3 = "";
+    bool third = Flux::resource_model::
+        parse_custom_match_policy ("policy=high node_centric=true stop_on_1_matches=true",
+                                   container3,
+                                   e3);
+    ok (third == true, "first is a valid policy");
+
+    bool fourth = Flux::resource_model::parse_bool_match_options ("high", container3);
+    ok (fourth == true, "policy first uses option high");
+
+    bool fifth = Flux::resource_model::parse_bool_match_options ("node_centric", container3);
+    ok (fifth == true, "policy first uses option node_centric");
+
+    bool sixth = Flux::resource_model::parse_bool_match_options ("stop_on_1_matches", container3);
+    ok (sixth == true, "policy first uses option stop_on_1_matches");
+
+    bool seventh = Flux::resource_model::parse_bool_match_options ("low", container3);
+    ok (seventh == false, "policy first does not use option low");
+
+    std::map<std::string, bool> container4;
+    std::string e4 = "";
+    bool eighth = Flux::resource_model::
+        parse_custom_match_policy ("policy=high node_centric=true node_exclusive=4",
+                                   container4,
+                                   e4);
+    ok (eighth == false, "node_exclusive accepts only true or false");
+    ok (e4 == "Policy option node_exclusive requires true or false, got 4\n",
+        "correct error msg for options");
+
+    return 0;
+}
+
+int main (int argc, char *argv[])
+{
+    plan (11);
+    test_parsers ();
+    done_testing ();
+    return 0;
+}
+
+/*
+ * vi: ts=4 sw=4 expandtab
+ */
--- a/resource/policies/dfu_match_policy_factory.cpp
+++ b/resource/policies/dfu_match_policy_factory.cpp
@@ -13,59 +13,175 @@ extern "C" {
 #include <config.h>
 #endif
 }
-
-#include <string>
+#include <array>
+#include <boost/algorithm/string.hpp>
 #include "resource/policies/dfu_match_policy_factory.hpp"
 
 namespace Flux {
 namespace resource_model {
 
-bool known_match_policy (const std::string &policy)
-{
-    bool rc = true;
-    if (policy != FIRST_MATCH && policy != FIRST_NODEX_MATCH && policy != HIGH_ID_FIRST
-        && policy != LOW_ID_FIRST && policy != LOW_NODE_FIRST && policy != HIGH_NODE_FIRST
-        && policy != LOW_NODEX_FIRST && policy != HIGH_NODEX_FIRST && policy != LOCALITY_AWARE
-        && policy != VAR_AWARE)
-        rc = false;
+const std::map<std::string, std::string> policies =
+    {{"first", "policy=high node_centric=true stop_on_1_matches=true"},
+     {"firstnodex", "policy=high node_centric=true node_exclusive=true stop_on_1_matches=true"},
+     {"high", "policy=high"},
+     {"low", "policy=low"},
+     {"lonode", "policy=low node_centric=true"},
+     {"hinode", "policy=high node_centric=true"},
+     {"lonodex", "policy=low node_centric=true node_exclusive=true"},
+     {"hinodex", "policy=high node_centric=true node_exclusive=true"},
+     {"locality", ""},
+     {"variation", ""}};
 
-    return rc;
+const std::vector<std::string> policy_options = {"policy",
+                                                 "node_centric",
+                                                 "stop_on_1_matches",
+                                                 "node_exclusive"};
+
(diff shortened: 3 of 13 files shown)


wihobbs:
And I know the testing and the code should be in separate commits... sigh (will fix)

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -13,7 +13,7 @@ extern "C" {
 #include <config.h>
 #endif
 }
-
+#include <iostream>
```
cmoussa1, line 16: I'm sorry if I missed it somewhere, but are you using anything from this `#include` anywhere? Perhaps this is left over from some debugging.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -22,50 +22,98 @@ namespace resource_model {
...
+    } else {
+        return_opt = stoi (policy_options.substr ((start_pos + 1), (end_pos - start_pos - 1)));
+    }
+    return return_opt;
+}
 
-    resource_type_t node_rt ("node");
+std::shared_ptr<dfu_match_cb_t> create_match_cb (const std::string &policy_requested)
+{
+    std::string policy = policies.find (policy_requested)->second;
```
cmoussa1, line 76: Should this `.find ()` call have any sort of error checking after the lookup of `policy_requested`? Maybe this is always guaranteed to find something successfully. If not, maybe you can check the contents of `policy` before moving on.

```c++
if (policy == "")
    // return an error?
    // set matcher to something default?
    // sorry that idk what the right behavior should be here
```

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -22,50 +22,98 @@ namespace resource_model {
...
+                ptr->add_score_factor (node_rt, 1, 10000);
+            }
+
+            if (parse_bool_match_options ("node_exclusive", policy)) {
                 ptr->add_exclusive_resource_type (node_rt);
+            }
+
+            if (option_exists ("stop_on_k_matches", policy)) {
+                ptr->set_stop_on_k_matches (parse_int_match_options ("stop_on_k_matches", policy));
+            }
```
cmoussa1, line 113: Does this `if-else` branch need an `else` statement? Excuse my ignorance on this - I'm not sure if the addition of custom policies allows for something other than `low` or `high`

resource/policies/base/test/matcher_policy_factory_test02.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+int test_parsers ()
+{
+    std::string policy_opts = policies.find ("first")->second;
+    bool first = Flux::resource_model::parse_bool_match_options ("high", policy_opts);
+    ok (first == true, "policy first uses option high");
+
+    policy_opts = policies.find ("high")->second;
+    bool fourth = Flux::resource_model::option_exists ("stop_on_k_matches", policy_opts);
+    ok (fourth == false, "policy high does not have stop_on_k_matches as substring");
```
cmoussa1, line 38: super minor: `bool fourth` comes before `second` and `third`; not sure if this is by intention though! Feel free to ignore me

resource/policies/base/test/matcher_policy_factory_test02.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    policy_opts = policies.find ("firstnodex")->second;
+    bool second = Flux::resource_model::parse_bool_match_options ("node_exclusive", policy_opts);
+    ok (second == true, "policy first uses option node_centric");
+    bool third = Flux::resource_model::option_exists ("stop_on_k_matches", policy_opts);
```
cmoussa1, line 43: Perhaps place newlines in between each test like you do in between the first two tests?

wihobbs: My convention was a newline each time the policy changed, not in between each test. However, looking at the other test, it has a newline after every `ok` so I'll stick with that convention.

resource/policies/base/test/matcher_policy_factory_test02.cpp
wihobbs: duh. will fix

resource/policies/dfu_match_policy_factory.cpp
wihobbs: Oh, actually, `known_match_policy` is called when the config is validated in `resource/modules/resource_match_opts.cpp` so it's already been validated before it gets here.

This will need additional work to support `custom`.

resource/policies/dfu_match_policy_factory.cpp
wihobbs: AFAIK low and high are the only options, unless `locality` or `variation` are chosen.

resource/policies/dfu_match_policy_factory.cpp
cmoussa1: In that case, a comment making note that it is already validated by the time it gets to that lookup might be helpful here!

jameshcorbett (commented):
Just did a quick look-through, it looks reasonable to me. I think the commits could be broken up better though (as you noted already)

resource/policies/dfu_match_policy_factory.hpp
```diff
@@ -24,19 +25,28 @@
...
-const std::string LOW_NODE_FIRST = "lonode";
-const std::string HIGH_NODE_FIRST = "hinode";
-const std::string LOW_NODEX_FIRST = "lonodex";
-const std::string HIGH_NODEX_FIRST = "hinodex";
-const std::string LOCALITY_AWARE = "locality";
-const std::string VAR_AWARE = "variation";
-
 bool known_match_policy (const std::string &policy);
 
+const std::map<std::string, std::string> policies =
```
trws, line 30: I'll do a deeper look over this, but first thing this should be split.  Put the declaration here with `extern` and the definition in the cpp file so this variable doesn't end up getting defined in all includers.

trws: To be fair, that's what the old code did, but it shouldn't 😬

garlick:
PR title seems truncated?

wihobbs:
Well, `custom` is a new type of policy that this PR is enabling. Or will enable eventually. Maybe "resource: refactor policy options to allow a user-defined policy" is more descriptive?

Note that this is still WIP and needs some config work and testing for the custom options. Happy to incorporate any feedback on the initial draft (and thanks to those who have already reviewed!), but it's going to stay in draft state until the code can:

- [ ] accept a custom string with policy options
- [ ] reject certain strings that are requesting invalid custom options, such as `set_stop_on_k_matches=-100` and such
- [ ] test some custom options and make sure they behave as expected

wihobbs:
Per discussion with Tom:

- [ ] To handle custom policies, "don't validate, parse" -- parse the input on a delimiter first and then verify that each key is valid, give an error message if it isn't, possibly demote this to a warning based on a config option
- [ ] Fluxion should issue a fatal warning if a policy named is not valid -- don't fall back on first (might already be done?)
- [ ] set_stop_on_k_matches probably only takes a value of 1, however, we should test this
- [ ] we should also think about having "node_centric" have a score factor of 10000, maybe we don't want to allow users to configure this, but maybe that number doesn't cover all our bases

resource/policies/dfu_match_policy_factory.hpp
```diff
@@ -24,19 +25,24 @@
...
-const std::string HIGH_NODE_FIRST = "hinode";
-const std::string LOW_NODEX_FIRST = "lonodex";
-const std::string HIGH_NODEX_FIRST = "hinodex";
-const std::string LOCALITY_AWARE = "locality";
-const std::string VAR_AWARE = "variation";
+extern const std::map<std::string, std::string> policies;
 
 bool known_match_policy (const std::string &policy);
 
+bool get_match_options (const std::string match_option, std::map<std::string, bool> &options);
```
jameshcorbett, line 32: Should this be removed?

jameshcorbett (commented):
Looks pretty good! Just noticed one thing.

resource/policies/dfu_match_policy_factory.hpp
wihobbs: Oh yeah, that was holdover from the first pass.

wihobbs:
I just asked @jameshcorbett to do a first pass to get me started. Note this still is in draft state until it has working sharness tests and additional reviews.

wihobbs:
Once the automated tests pass I'll drop draft from this PR.

trws:
This is starting to look really good @wihobbs!  There's one tweak I'd like to see to this so we can keep extending it.  I like the way the options are handled, this is exactly what I was hoping for.  The one tweak I'd make is to have a `policy=*` option rather than separate  `low=` and `high=`, since the base policy object types are low, high, locality and variation.  That way, we get the "you can't set both low and high at the same time" for free, and can even collapse the two low and high branches to create a matcher based on which it is then set options on the resulting object after the condition.  We can also check that policy has been set, either with the bare name or by having `policy=whatever` to allow options.  Does that sound workable?

wihobbs:
I think I've incorporated this feedback @trws, though slightly differently than you laid out in your comment. Instead of a `policy` key which accepts `locality`, `variation`, `high`, and `low`, I added a `policy` key that will accept `low` or `high`, and you can still set `match-policy` to `locality`, `variation`, `high`, and `low`, in which case the latter two would result in a "vanilla" `high` or `low` policy without the extra options. Does that suffice for what you laid out? It turned out to be not a huge lift.

I haven't implemented allowing `set_stop_on_k_matches` to accept an integer yet, but I'm going to go ahead and request a few reviews on this, to get feedback on the code and start getting the commits polished for approval. If you're okay with kicking that down the road to another PR (apologies, I forgot what we decided on that in our conversation yesterday), we could do that when the option itself supports integers other than 1.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -13,59 +13,168 @@ extern "C" {
...
+                                                 "node_exclusive"};
+
+/* This takes a string of match policy options, deliniated by
+ * spaces and equals signs, and turns them into a std::map.
+ * It validates that each option is set to either "true" or "false"
+ * and puts the corresponding boolean in the map.
+ */
+bool parse_custom_match_policy (const std::string long_string, std::map<std::string, bool> &split)
+{
+    std::vector<std::string> temp;
```
cmoussa1, line 48: IMHO perhaps a variable name like `options` might be more descriptive of what this is going to be initialized to.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -13,59 +13,168 @@ extern "C" {
...
+    boost::split (temp, long_string, boost::is_any_of (" "));
+    /* After splitting based on term, validate the terms. If invalid,
+     * abort.
+     */
+    for (int i = 0; i < temp.size (); i++) {
+        std::vector<std::string> temp_two;
```
cmoussa1, line 54: Same suggestion here! Maybe something like `setting` or something?

cmoussa1 (commented):
Looks pretty good @wihobbs!! I've left some general feedback; I should probably defer to someone who knows the sched code better than I do to actually approve this but I think this looks pretty good to me!

wihobbs:
Thanks @cmoussa1! Fixed and pushed.

resource/policies/dfu_match_policy_factory.cpp
```diff
@@ -13,59 +13,167 @@ extern "C" {
...
+         * get caught later?
+         */
+        boost::split (settings, options.at (i), boost::is_any_of ("="));
+        /* Place the terms in a map for ease of access. Lots of gross
+         * optionsorary structures needed here, hopefully they'll get
+         * garbage collected soon.
+         */
+        if (std::find (policy_options.begin (), policy_options.end (), settings.at (0))
+            == policy_options.end ()) {
+            std::cerr << "invalid policy option: " << settings.at (0) << std::endl;
```
grondo, line 66: Actually, you should probably avoid writing directly to stdout/err from a broker module.

If there's a way to throw an exception and catch the error message at a higher level, which would then `flux_log (LOG_ERR, ...)` the result, that might be good. Otherwise you'd have to pass down a `flux_t *` handle to this function so it can log errors.

You could also return an error via a separate parameter when the result is `false`. Some of the other Fluxion developers may have a preference here.

trws: We had a chat about this yesterday after the meeting. Just above this in the call chain we're passing a std::string through by reference and appending messages to it as necessary, then returning it up to where we can handle it.  The trick is we can't always have a handle here, because this is used in both the module and in `resource_query`.  Essentially the separate parameter option.  Example of the plan on line 516 in resource_match_opts.

I have a separate issue about doing a better job of this in fluxion as a whole, but don't want it to block this PR, so plan is to pass that output string down through for now.

wihobbs: Yep, that's a hole in this PR. Will fix with what we discussed. Sorry, I should have done that before marking it as "ready." I got excited [and maybe a little ahead of myself] when the testsuite _finally_ passed 😄

wihobbs: One irritating thing about my implementation of this is that it takes a 3rd positional argument of `std::string &error_string` to return the error message, which, because this function is used for validation and parsing the extra options into a map, requires initializing a string in some places where we _know_ it won't be needed.

@trws's suggestion of using `std::expected<bool, std::string>` as the return type would probably be cleaner.

grondo: TIL about `std::expected`!

trws: Yeah, a version of that is what I want us to do.  My ideal solution would be if we could use `std::expected<RetType, std::status_code>` but `status_code` isn't standard yet, and `error_code` is... easy to misuse.  We also don't require a C++23 stdlib yet.  The `boost::outcome` with the experimental header to use `boost::outcome<RetCode, boost::status_code>` is probably best, it's very hard to misuse and much faster to compile, but it requires just a bit of setup I haven't found time to do yet.

If you're up for it @wihobbs, it's not a lot of stuff and shouldn't be complicated and would really help out other places in sched too.  That said, I'd rather that be a separate PR from this since it's pretty orthogonal.

wihobbs: I opened #1330 for the task mentioned above.

doc/man5/flux-config-sched-fluxion-resource.rst
```diff
@@ -102,6 +102,45 @@ hinodex
...
+    it, without evaluating for potentially more optimal matches.
+
+::
+
+    [sched-fluxion-resource]
+
+    # system instance will use node-exclusive
+    # scheduling (with nodes of low node IDs
+    # selected first).
+    match-policy = "policy=high node_exclusive=true set_stop_on_1_matches=true"
```
trws, line 142:
```suggestion
    # scheduling first-match (with nodes of high
    # node IDs selected first).
    match-policy = "policy=high node_exclusive=true set_stop_on_1_matches=true"
```

trws (approved):
Assuming the one documentation fix, this looks really good to me. Thanks @wihobbs!

wihobbs:
Thanks @trws!

PR #1273, opened 2024-08-19
qmanager-stats: fix jansson refcount issue

trws:
problem: The `to_json` function for serializing maps with jansson had a refcounting bug in it, which wasn't caught by address sanitizer, valgrind, or testing.  This is what caused the issue @jamescorbett saw in the buildfarm, but only reliably reproduced on alpine.

solution: Temporary solution is to fix the refcount problem, longer term solution is to use something that isn't so insanely hard to use correctly and debug.

### DIFF
--- a/src/common/c++wrappers/jansson.hpp
+++ b/src/common/c++wrappers/jansson.hpp
@@ -119,7 +119,7 @@ inline void to_json (value &jv, const MapT &m)
     for (auto &[k, v] : m) {
         value val;
         to_json (val, v);
-        json_object_set_new (jv.get (), k.c_str (), val.get ());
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Ok I think I see why this is necessary--because `val` is destructed and decrefs its reference, so stealing it (like `json_object_set_new` does) is wrong and leads to the json object being garbage collected because the refcount hits zero. I'm surprised this wasn't caught before.

trws:
Yup, you got it. As to being surprised, me too... It seems that jansson checks the reference count of objects it accesses or something, somehow it manages to avoid accessing it after it expires, which is why it was so darned hard to track it down.

PR #1272, opened 2024-08-16
Update build and ci to enforce compiler versions, bump to C++23, remove vestigial deps, and generate images matching core

trws:
This is meant to solve a variety of build and CI issues we're facing, and also bring our requirements more up to date.  We actually barely need any built boost components anymore, so I'm removing them.  We may go the rest of the way, will consider it, but for now it's regex, system and filesystem that are all removed. 

This also explicitly checks compiler versions, note that the default initial gcc and clang for jammy fell off the back _but_ both gcc12 and clang15 are available in standard jammy apt repos, and in fact we had clang-15 pre-installed, so no change to OS requirements.  Since cmake was already checking for minimal C++20 support, this actually checks for explicit versions of gcc and clang, with an escape hatch to build with whatever.  The code currently builds, barely, with gcc11 and clang-14.  The explicitly required versions are gcc12 and clang-15 are the versions where all the features we use are documented to actually work correctly.

fixes #1264 

There's also a tiny fix in here for older clangs that don't support the c++20 removal of need for typename in some cases, and I'm adding a build using clang-15 so we catch things like this in the future.

@vsoch, I think you're making the spack package match these versions, but after this you should get a much more informative/clear message about requirements.

PR #1274, opened 2024-08-19
Resolve issues with Alpine linux and build release containers for amd64 and arm64

trws:
This is stacked on top of #1272 and #1273 and adds support for alpine linux on amd64 and arm64, with the attendant ci updates.  This also actually surfaced a very, very hard to repro bug in the jansson code I added for qmanager, so I'm seeing this as pretty useful right now.

(note that most of the changes in the diff are due to the stacking, it's just the alpine commit that's new here)

### DIFF
--- a/cmake/config_testing.cmake
+++ b/cmake/config_testing.cmake
@@ -34,6 +34,7 @@ function(flux_config_test)
     "${options}" "${oneValueArgs}" "${multiValueArgs}")
   set_property(TEST ${ARG_NAME} PROPERTY ENVIRONMENT "${TEST_ENV}")
 endfunction()
+find_program(TEST_SHELL NAMES bash zsh sh)
 function(flux_add_test)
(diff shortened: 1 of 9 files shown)


jameshcorbett (approved):
Just reviewing the alpine commit--LGTM but I think it could be broken into multiple commits since it does a bunch of seemingly unrelated things.

trws:
Fair point, it's split and more documented by individual pieces now.

jameshcorbett (approved):
Looks great thank you!

garlick:
Might want to change the PR title to something more descriptive since we typically use those for release notes.

trws:
Good point, thanks @garlick. Reworded.

trws:
Same issue with mergify here, it's way out of date on merge protections... Not sure how to fix it, maybe if we can get to the mergify config file update it will get it moving again.

PR #1275, opened 2024-08-20
update satisfiability RPC to conform with RFC 27

grondo:
This simple PR just brings flux-sched into line with the recent addition of the `feasibility` service for handling satsifiability requests in RFC 27.

For now, both the old `sched.feasibility` and `feasibility.check` RPCs are supported so that Fluxion can continue to work with existing flux-core which doesn't yet support `feasibility.check`.

The long term plan I think is to run `feasibility` as a separate module, so this is just a stopgap until that time.


garlick (approved):
LGTM

PR #1276, opened 2024-08-20
move feasibility service into sched-fluxion-resource and fix associated test

grondo:
This PR fixes a typo identified in https://github.com/flux-framework/flux-core/pull/6223, changes feasibility service registration to occur before the module is forced to the running state, and fixes one test that used a deprecated `flux job-validator` option.

I apologize for the noise. This version has indeed had some testing against the pending flux-core changes. :facepalm:

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -493,18 +493,6 @@ static int enforce_options (std::shared_ptr<qmanager_ctx_t> &ctx)
     return rc;
 }
 
-static int register_feasibility (flux_t *h)
-{
-    flux_future_t *f;
-    int rc;
-
-    if (!(f = flux_service_register (h, "sched")))
-        return -1;
-    rc = flux_future_get (f, NULL);
-    flux_future_destroy (f);
-    return rc;
-}
-
 static int handshake (std::shared_ptr<qmanager_ctx_t> &ctx)
 {
     int rc = 0;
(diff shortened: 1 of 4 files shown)


garlick (approved):
LGTM

grondo:
I've updated this branch to move the `feasibility` service into the `resource` module as initially suggested by @garlick. This is a temporary measure until there's a separate `feasibility` module in flux-sched, and works around issues with detection of `sched-fluxion-qmanager` as a scheduler module in the meantime since we don't add an addition service to the output of `flux module list`. (it was also probably the right thing to do in the first place)

The check in `t3301` is also future-proofed to allow for multiple service names in `flux module list` output.

Sorry for all the noise here.

garlick (approved):
The changes LGTM!

grondo:
Ok, will set MWP (assuming mergify still isn't unsynchronized with github)

grondo:
@mergifyio refresh

grondo:
@mergifyio refresh

trws:
Thanks for this @grondo, I'm guessing this also takes care of the "failed to register feasibility service: File exists" we get during startup on current master?

trws:
Also, I'm ok with a manual merge if mergify stays stuck. I'm going to try re-adding the protection rules it thinks are still in place, refreshing, then removing them and refreshing again... not sure what else to try at this point.

PR #1277, opened 2024-08-20
qmanager: fix lifetime issue in partial release

trws:
problem: The shared pointer for the job_t gets released before the cancel call, causing a potential use-after-free if the job isn't held by another queue at the time.

solution: take a reference to the shared_ptr before erasing it from the m_jobs map


milroy (approved):
I should have noticed this; thanks for catching it!

trws:
This seems like a pretty sneaky problem. I'm not 100% sure but I think it hides because jobs in our tests are often held in m_canceled or something like that when this runs, so they aren't freed early. Thanks for the review!

PR #1278, opened 2024-08-20
mergify: update to new mergify.yml requirements

trws:
problem: mergify is changing their option structure

solution: update to match new requirements and core

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,13 +1,7 @@
 queue_rules:
   - name: default
-    conditions:
-      - base=master
-      - label="merge-when-passing"
-      - label!="work-in-progress"
-      - "approved-reviews-by=@flux-framework/core"
-      - "#approved-reviews-by>0"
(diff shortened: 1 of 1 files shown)


jameshcorbett (commented):
So we no longer set the merge-when-passing label?

trws:
We do, the `conditions` block under queues didn't actually do anything, it was the one under the `pull_requests` block that actually made it work (or so their docs and json schema claim)

grondo (approved):
This LGTM. I think this will require a manual merge when ready though since the mergify config changed.

grondo:
Hm, there are still some warnings under the mergify checks tab though:

>    The configuration uses the deprecated update_method attribute of the queue action in one or more pull_request_rules. It must now be used under the queue_rules configuration.
>    A brownout is planned on August 26th, 2024.
>    This option will be removed on September 23rd, 2024.
>    For more information: https://docs.mergify.com/configuration/file-format/#queue-rules

PR #1279, opened 2024-08-20
docker-run-checks: export INSTALL_ONLY

trws:
problem: the install-only checks were accidentally running checks instead of only installing

solution: export the INSTALL_ONLY variable so it propagates appropriately

This is a bit of a facepalm mistake.  It only actually causes a failure on the arm runners because they run out of time.


jameshcorbett (approved):
👍

PR #1280, opened 2024-08-20
alpine: use root to avoid nosuid issue in ci

trws:
problem: just like with el9 and noble, alpine requires user to be set to root in the dockerfile here to build right under qemu-user-static in CI

solution: add the user root line where required

Another mistake that only shows up once we get into master, and this one only sometimes. 🤦


jameshcorbett (approved):
👍

PR #1281, opened 2024-08-20
resource: make feasibility service registration optional

grondo:
Problem: There are many cases in the testsuite where the Fluxion resource module is loaded alongside sched-simple for testing, but this currently fails because these two modules now conflict due to the feasibility service.

Since feasibilty is an optional service, and may soon move to a separate module anyway, make feasibility service registration optional.


trws (approved):

grondo:
Thanks! Setting MWP.

PR #1282, opened 2024-08-21
deps: add install-deps.sh script, use in docker

trws:
problem: it's easy to accidentally miss adding or removing a dependency in the various dockerfiles and the readme

solution: much like core, encode them as a script, then use it

The main motivation for this is I noticed I missed some yesterday, even after @jameshcorbett's useful feedback on a few.


jameshcorbett (approved):
Looks like a nice simplification

PR #1283, opened 2024-08-21
scripts: add build-matrix script to ease matching ci

trws:
problem: we have good scripts for running in our images locally, but not always with all the same flags as in CI

solution: add a script, build-matrix, that takes a name from the actions build matrix and runs docker_run_checks with the same flags as that matrix entry

### DIFF
--- a/scripts/build-matrix
+++ b/scripts/build-matrix
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+from pathlib import Path
+import platform
+import subprocess as sp
+import shlex
+import sys
+
+# populate matrix
+my_path = Path(__file__).absolute()
+generate_matrix = my_path.parent.parent.joinpath("src/test/generate-matrix.py")
+includes = json.loads(
+    sp.check_output([generate_matrix], env={"GITHUB_REF": "refs/heads/master"}).decode()
+)["include"]
+matrix = {}
+for b in includes:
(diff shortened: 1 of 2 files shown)


scripts/build-matrix
```diff
@@ -0,0 +1,64 @@
...
+
+p = argparse.ArgumentParser()
+p.add_argument("-p","--platform", default=platform.machine())
+p.add_argument("--interactive", action="store_true")
+p.add_argument("name")
+args = p.parse_args()
+
+my_path = Path(__file__).absolute()
+generate_matrix = my_path.parent.parent.joinpath("src/test/generate-matrix.py")
+docker_run_checks = my_path.parent.parent.joinpath("src/test/docker/docker-run-checks.sh")
```
jameshcorbett, line 22: unused variable

scripts/build-matrix
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+import argparse
+from pprint import pprint
+import json
+import os
+from pathlib import Path
+from re import L
```
jameshcorbett, line 8: unused variable

jameshcorbett (approved):
👍

grondo (approved):
Great idea! I like that the script can also list the entries in the matrix.

trws:
Thanks both of you, cleaned up a little bit and added it to the list to be processed by pre-commit so the style checks run on it.

PR #1285, opened 2024-08-30
modules: Move feasibility/satisfiability checking into a new module

jacobtkeio:
Problem: Feasibility checking takes up a significant amount of sched-fluxion-resource's time that could be better spent on scheduling.
Solution: Move feasibility checking into a new module, sched-fluxion-feasibility, that can run on multiple ranks.

### DIFF
--- a/resource/modules/feasibility.cpp
+++ b/resource/modules/feasibility.cpp
@@ -1,0 +1,366 @@
+/*****************************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+#include "resource_match.hpp"
+
+MOD_NAME ("sched-fluxion-feasibility");
+
+////////////////////////////////////////////////////////////////////////////////
+// Request Handler Prototypes
+////////////////////////////////////////////////////////////////////////////////
+
+static void feasibility_request_cb (flux_t *h,
+                                    flux_msg_handler_t *w,
+                                    const flux_msg_t *msg,
+                                    void *arg);
+
+static const struct flux_msg_handler_spec htab[] =
+    {{FLUX_MSGTYPE_REQUEST, "feasibility.check", feasibility_request_cb, 0},
+     FLUX_MSGHANDLER_TABLE_END};
+
+////////////////////////////////////////////////////////////////////////////////
+// Module Initialization Routines
+////////////////////////////////////////////////////////////////////////////////
+
+static void set_default_args (std::shared_ptr<resource_ctx_t> &ctx)
+{
+    resource_opts_t ct_opts;
+    std::string e = "";
+    ct_opts.set_load_format ("rv1exec");
+    ct_opts.set_match_subsystems ("containment");
+    ct_opts.set_match_policy ("first", e);
+    ct_opts.set_prune_filters ("ALL:core,ALL:node,ALL:gpu,ALL:ssd");
+    ct_opts.set_match_format ("rv1_nosched");
+    ct_opts.set_update_interval (0);
+    ctx->opts += ct_opts;
+}
+
+static std::shared_ptr<resource_ctx_t> getctx (flux_t *h)
+{
+    void *d = NULL;
+    std::shared_ptr<resource_ctx_t> ctx = nullptr;
+    if ((d = flux_aux_get (h, "sched-fluxion-feasibility")) != NULL)
+        ctx = *(static_cast<std::shared_ptr<resource_ctx_t> *> (d));
+    if (!ctx) {
+        try {
+            ctx = std::make_shared<resource_ctx_t> ();
+            ctx->traverser = std::make_shared<dfu_traverser_t> ();
+            ctx->db = std::make_shared<resource_graph_db_t> ();
+        } catch (std::bad_alloc &e) {
+            errno = ENOMEM;
+            goto done;
+        }
+        ctx->h = h;
+        ctx->handlers = NULL;
+        set_default_args (ctx);
+        ctx->matcher = nullptr; /* Cannot be allocated at this point */
+        ctx->writers = nullptr; /* Cannot be allocated at this point */
+        ctx->reader = nullptr;  /* Cannot be allocated at this point */
+        ctx->m_resources_updated = false;
+        ctx->m_resources_down_updated = true;
+        ctx->m_resources_alloc_updated = std::chrono::system_clock::now ();
+        ctx->m_acquire_resources_from_core = false;
+    }
+
+done:
+    return ctx;
+}
+
+static int process_args (std::shared_ptr<resource_ctx_t> &ctx, int argc, char **argv)
+{
+    int rc = 0;
+    optmgr_kv_t<resource_opts_t> opts_store;
+    std::string info_str = "";
+
+    for (int i = 0; i < argc; i++) {
+        const std::string kv (argv[i]);
+        if ((rc = opts_store.put (kv)) < 0) {
+            flux_log_error (ctx->h, "%s: optmgr_kv_t::put (%s)", __FUNCTION__, argv[i]);
+            return rc;
+        }
+    }
+    if ((rc = opts_store.parse (info_str)) < 0) {
+        flux_log_error (ctx->h, "%s: optmgr_kv_t::parse: %s", __FUNCTION__, info_str.c_str ());
+        return rc;
+    }
+    if (info_str != "") {
+        flux_log (ctx->h, LOG_DEBUG, "%s: %s", __FUNCTION__, info_str.c_str ());
+    }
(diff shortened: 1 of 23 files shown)


resource/modules/feasibility.cpp
```diff
@@ -0,0 +1,385 @@
+/*****************************************************************************\
+ * Copyright 2014 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
```
milroy: Update the copyright year:
```suggestion
/*****************************************************************************\
 * Copyright 2025 Lawrence Livermore National Security, LLC
 * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
 ```

File: resource/modules/resource.cpp [hunk 2]
```diff
@@ -0,0 +1,1548 @@
+/*****************************************************************************\
+ * Copyright 2014 Lawrence Livermore National Security, LLC
```
milroy:
```suggestion
 * Copyright 2025 Lawrence Livermore National Security, LLC
 ```

File: src/cmd/flux-ion-resource.py [hunk 3]
```diff
@@ -113,7 +113,7 @@ def rpc_namespace_info(self, rank, type_name, identity):
 
     def rpc_satisfiability(self, jobspec):
         payload = {"jobspec": jobspec}
-        return self.handle.rpc("sched-fluxion-resource.satisfiability", payload).get()
+        return self.handle.rpc("feasibility.check", payload).get()
```
milroy, line 116: Let's revisit this RPC name. Adhering to the existing pattern yields `sched-fluxion-feasibility.satisfiability` which is clunky.

resource/modules/resource_match.hpp
```diff
@@ -0,0 +1,192 @@
+/*****************************************************************************\
+ * Copyright 2014 Lawrence Livermore National Security, LLC
```
milroy, line 2:
```suggestion
 * Copyright 2025 Lawrence Livermore National Security, LLC
```

t/data/resource/jobspecs/shrink/shrink2.yaml
```diff
@@ -0,0 +1,20 @@
...
+               count: 1
+attributes:
+  system:
+    duration: 300
+tasks:
+  - command: [ "hostlist" ]
+    slot: default
+    count:
+      per_slot: 1
+
```
milroy: Following the current directory structure convention, I think it's better to move `shrink2-4` into `t/data/resource/jobspecs/satisfiability`.

t/t4005-match-unsat.t
```diff
@@ -24,9 +25,11 @@ test_debug '
     echo ${jobspec2}
 '
 
-test_expect_success 'loading resource module with a tiny machine config works' '
+test_expect_success 'loading resource modules with a tiny machine config works' '
     load_resource load-file=${grug} load-format=grug \
-prune-filters=ALL:core subsystems=containment policy=high
+prune-filters=ALL:core subsystems=containment policy=high &&
+    load_feasibility load-file=${grug} load-format=grug \
```
milroy: I'm wondering if we should set `prune-filters` to `ALL:core,ALL:node` in the tests and in the module.

t/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ set(ALL_TESTS
   t4012-set-status.t
```
milroy, line 85: The commit message body should wrap at 72 characters.

milroy (commented):
This is really good work. I have a few nits, and a minor point for discussion regarding the RPC name.

Apart from that, I'd like to discuss changing the satisfiability pruning filter to be `ALL:core,ALL:node` and perhaps add `ALL:gpu` as well. Adding a `node` type to the pruning filter by default should make the `is_satisfiable` function in the traverser faster and more powerful.

t/t4005-match-unsat.t
jacobtkeio: We decided to do this by leaving out the prune-filters option completely in tests, which reverts them to the default "ALL:core,ALL:node,ALL:gpu[,ALL:ssd]"

t/data/resource/jobspecs/shrink/shrink2.yaml
jacobtkeio: Done! It may be worth removing these files completely and using a command like `flux submit -N[2-4] --dry-run hostname` to generate them on the fly though

src/cmd/flux-ion-resource.py
jacobtkeio: Here is the RFC I mentioned! https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_27.html#feasibility

jacobtkeio:
Are we still OK with leaving in the uncovered code in resource_match.cpp:select_subsystems?
https://github.com/jacobtkeio/flux-sched/blob/a80379f98716f28e623f9899b705c4232d04e469/resource/modules/resource_match.cpp#L1234

jacobtkeio:
I pushed these last two t4014 changes at the suggestion of @wihobbs. They cover a bit of feasibility checking through the job validator and flux submit (issue 1383)

milroy:
> Are we still OK with leaving in the uncovered code in resource_match.cpp:select_subsystems?

Yes, my preference is to leave it in so that when subsystems are working correctly we won't have to remember to re-add the check.

milroy (commented):
> Separating feasibility into a new module fixes this.

How did you test this? I added the change in this commit to `t1020` in current `master` and the test still passes.

t/t4014-match-feasibility.t
```diff
@@ -123,6 +123,10 @@ test_expect_success 'job-validator correctly returns satisfiable' '
     grep "{\"errnum\": 0}"
```
milroy, line 123: Please add the information in the commit message to the previous commit and squash the change in this commit into the previous commit.

wihobbs:
> How did you test this? I added the change in this commit to t1020 in current master and the test still passes.

Running feasibility _at all_ will reject 

> Edit: I accidentally left a review comment rather than commenting on the specific commit. Adding a link to the commit I'm referencing: [7da3e16](https://github.com/flux-framework/flux-sched/commit/7da3e162e1b80bf5a5e19afebe03ffd1eec88e4d)
> 
> > Separating feasibility into a new module fixes this.
> 
> How did you test this? I added the change in this commit to `t1020` in current `master` and the test still passes.

I'll chime in on this since I originally reported #1383.

Separating feasibility into a new module doesn't fix #1383, running feasibility in the validator _at all_ will fix that issue, as infeasible jobs will be rejected before the job reaches `DEPEND`. 

That issue references a bug I found while testing this PR, that if the job reaches `SCHED`, a large number of infeasible jobs will cause Fluxion to not schedule anything at all. For an example of that, I posted one in the issue.

After a quick look I think `t1020` is running with feasibility enabled in the validator, so perhaps adding an extra test in `t4014` isn't necessary. (If I gave you bad advice @jacobtkeio, sorry!) Or perhaps we should add a test of many infeasible jobs with the validator _disabled_ to make sure that the sched loop can reject the jobs if they reach that point (after the issue is solved).

jacobtkeio:
This last force push reduced the subject line of 137c67e below 50 characters.

grondo:
> This last force push reduced the subject line of https://github.com/flux-framework/flux-sched/commit/137c67ed8ffe76f7f0c6b88372848dea0aaf16b3 below 50 characters.

Just an FYI - the 50 character limit should  just be a warning -- sometimes it is impossible to make the subject line that small.

milroy (approved):
Thanks for the diligence and the hard work, @jacobtkeio! This PR LGTM and I'm approving it. Please rebase against upstream and repush.

@trws, do you want to give this another look or can we merge it?

trws (dismissed):
Lets do it!

t/t4014-match-feasibility.t
```diff
@@ -0,0 +1,146 @@
...
+    load_feasibility &&
+    test -z "$(flux dmesg -c | grep -q err)"
+'
+
+test_expect_success 'loading job-validator conf works' '
+{ cat >conf.tmp << 'EOF'
+[ingest.validator]
+plugins = ["jobspec", "feasibility"]
+EOF
+} &&
```
grondo, line 108: Just a suggestion here. If you prefix `EOF` with `-` then the heredoc will ignore leading tabs. This alllows the contents to be indented, e.g.

```suggestion
test_expect_success 'loading job-validator conf works' '
        cat <<-'EOF' >conf.tmp &&
        [ingest.validator]
        plugins = ["jobspec", "feasibility"]
        EOF
```

If you take this suggestion, then the rest of this test file should probably be indented with tabs. (This is why we've settled on a semi-standard of using tabs in sharness tests in flux-core, though I know that isn't necessary done here in Fluxion)

grondo (commented):
just happened to notice one minor thing while perusing how this works (so the module can be added to modprobe config). Just leaving a comment in case it is helpful in case this is helpful in the future (since the PR has been merged)

t/t4014-match-feasibility.t
jacobtkeio: Sorry I missed this! This suggestion is definitely an improvement. I will plan to include it in my next PR if there are no objections!

Issue #1284, opened 2024-08-27
Partial cancel not releasing rabbit resources (?)


jameshcorbett:
Snipped results of `flux dmesg` on hetchy:

```
2024-08-27T01:46:53.149076Z sched-fluxion-resource.err[0]: run_remove: dfu_traverser_t::remove (id=152883667495027712): mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149167Z sched-fluxion-resource.err[0]: ssd0.
2024-08-27T01:46:53.149175Z sched-fluxion-resource.err[0]: Success.
2024-08-27T01:46:53.149181Z sched-fluxion-resource.err[0]: mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149186Z sched-fluxion-resource.err[0]: ssd1.
2024-08-27T01:46:53.149190Z sched-fluxion-resource.err[0]: Success.
2024-08-27T01:46:53.149194Z sched-fluxion-resource.err[0]: mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149199Z sched-fluxion-resource.err[0]: ssd2.
2024-08-27T01:46:53.149204Z sched-fluxion-resource.err[0]: Success.
2024-08-27T01:46:53.149208Z sched-fluxion-resource.err[0]: mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149218Z sched-fluxion-resource.err[0]: ssd3.
2024-08-27T01:46:53.149225Z sched-fluxion-resource.err[0]: Success.
2024-08-27T01:46:53.149234Z sched-fluxion-resource.err[0]: mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149244Z sched-fluxion-resource.err[0]: ssd4.
2024-08-27T01:46:53.149251Z sched-fluxion-resource.err[0]: Success.
2024-08-27T01:46:53.149257Z sched-fluxion-resource.err[0]: mod_plan: traverser tried to remove schedule and span after vtx_cancel during partial cancel:
2024-08-27T01:46:53.149267Z sched-fluxion-resource.err[0]: ssd5.
2024-08-27T01:46:53.149279Z sched-fluxion-resource.err[0]: Success.
[...]

PR #1292, opened 2024-09-05
Resource: support partial cancel of resources external to broker ranks

milroy:
Issue #1284 identified a problem where rabbits are not released due to a traverser error during partial cancellation. The traverser should skip the rest of the `mod_plan` function when an allocation is found and `mod_data.mod_type == job_modify_t::PARTIAL_CANCEL`. ~This PR adds a `goto` statement to return `0` under this circumstance.~ This PR is significantly updated in scope to reflect more comprehensive understanding of the problem.

This line is causing some of the errors reported in the related issues:  https://github.com/flux-framework/flux-sched/blob/996f999c9bc398845f91ea58851e517de63ad677/resource/traversers/dfu_impl_update.cpp#L436 That error condition (`rc !=0`) occurs because a partial cancel successfully removes the allocations of the other resource vertices (especially `core`, which is installed in all pruning filters by default) because they have broker ranks. However, when the final `.free` RPC fails to remove an `ssd` vertex allocation the full cleanup cancel exits with an error when it hits the vertices it's already cancelled.

This PR adds support for a cleanup `cancel` post partial cancel that skips the inapplicable error check for non-existent `planner` spans in the error criterion for removal of `planner_multi` spans.

Two related problems needed to be solved: handling partial cancel for brokerless resources when default pruning filters are set (`ALL:core`) and pruning filters are set for the resources excluded from broker ranks (e.g., `ALL:ssd`). In preliminary testing, supporting both was challenging because with `ALL:core` configured, the final `.free` RPC frees all `planner_multi`-tracked resources, which prevents a cleanup, full `cancel`. However, tracking additional resources (e.g., `ALL:ssd`) successfully removes resource allocations on those vertices only with a cleanup cancel.

This PR adds support for rank-based partial cancel with resources that don't have a rank with the rv1_nosched match format. 

Updates: after further investigation, issue #1305 is related as well. This PR also aims to address issue #1309.

### DIFF
--- a/qmanager/policies/base/queue_policy_base.hpp
+++ b/qmanager/policies/base/queue_policy_base.hpp
@@ -646,14 +646,28 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
             case job_state_kind_t::ALLOC_RUNNING:
                 // deliberately fall through
             case job_state_kind_t::RUNNING:
-                if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
-                    flux_log_error (flux_h,
-                                    "%s: .free RPC partial cancel failed for jobid "
-                                    "%jd",
-                                    __FUNCTION__,
-                                    static_cast<intmax_t> (id));
-                    errno = EINVAL;
-                    goto out;
+                if (!final) {
+                    if (cancel (h, job_it->second->id, R, true, full_removal) != 0) {
+                        flux_log_error (flux_h,
+                                        "%s: .free RPC partial cancel failed for jobid "
+                                        "%jd",
+                                        __FUNCTION__,
+                                        static_cast<intmax_t> (id));
+                        errno = EINVAL;
+                        goto out;
+                    }
+                } else {
+                    // Run a full cancel to clean up all remaining allocated resources
+                    if (cancel (h, job_it->second->id, true) != 0) {
+                        flux_log_error (flux_h,
+                                        "%s: .free RPC full cancel failed for jobid "
+                                        "%jd",
+                                        __FUNCTION__,
+                                        static_cast<intmax_t> (id));
+                        errno = EPROTO;
+                        goto out;
+                    }
+                    full_removal = true;
                 }
                 // We still want to run the sched loop even if there's an inconsistent state
                 set_schedulability (true);
@@ -664,24 +678,17 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
                     job_it->second->state = job_state_kind_t::COMPLETE;
                     // hold a reference to the shared_ptr to keep it alive
                     // during cancel
-                    auto job_sp = job_it->second;
                     m_jobs.erase (job_it);
-                    if (final && !full_removal) {
-                        // This error condition indicates a discrepancy between core and sched.
+                    if (full_removal && !final) {
+                        // This error condition can indicate a discrepancy between core and sched,
+                        // specifically that a partial cancel removed an allocation prior to
+                        // receiving the final .free RPC from core.
                         flux_log_error (flux_h,
-                                        "%s: Final .free RPC failed to remove all resources for "
+                                        "%s: removed allocation before final .free RPC for "
                                         "jobid "
                                         "%jd",
                                         __FUNCTION__,
                                         static_cast<intmax_t> (id));
-                        // Run a full cancel to clean up all remaining allocated resources
-                        if (cancel (h, job_sp->id, true) != 0) {
-                            flux_log_error (flux_h,
-                                            "%s: .free RPC full cancel failed for jobid "
-                                            "%jd",
-                                            __FUNCTION__,
-                                            static_cast<intmax_t> (id));
-                        }
                         errno = EPROTO;
                         goto out;
                     }
--- a/resource/traversers/dfu_impl_update.cpp
+++ b/resource/traversers/dfu_impl_update.cpp
@@ -535,6 +535,12 @@ int dfu_impl_t::mod_plan (vtx_t u, int64_t jobid, modify_data_t &mod_data)
         span = alloc_span->second;
         if (mod_data.mod_type != job_modify_t::PARTIAL_CANCEL) {
             (*m_graph)[u].schedule.allocations.erase (alloc_span);
+        } else {
+            // This condition is encountered when the vertex is
+            // not associated with a broker rank. We may need
+            // extra logic here to handle more advanced partial
+            // cancel in the future.
+            goto done;
         }
     } else if ((res_span = (*m_graph)[u].schedule.reservations.find (jobid))
                != (*m_graph)[u].schedule.reservations.end ()) {
--- a/resource/planner/c/planner_multi_c_interface.cpp
+++ b/resource/planner/c/planner_multi_c_interface.cpp
@@ -423,8 +423,13 @@ extern "C" int planner_multi_rem_span (planner_multi_t *ctx, int64_t span_id)
         goto done;
     }
     for (i = 0; i < it->second.size (); ++i) {
-        if (planner_rem_span (ctx->plan_multi->get_planner_at (i), it->second[i]) == -1)
-            goto done;
+        // If executed after partial cancel, depending on pruning filter settings
+        // some spans may no longer exist. In that case the span_lookup value for
+        // the resource type will be -1.
+        if (it->second[i] != -1) {
+            if (planner_rem_span (ctx->plan_multi->get_planner_at (i), it->second[i]) == -1)
+                goto done;
+        }
     }
     ctx->plan_multi->get_span_lookup ().erase (it);
     rc = 0;
--- a/t/CMakeLists.txt
+++ b/t/CMakeLists.txt
@@ -29,6 +29,7 @@ set(ALL_TESTS
   t1024-alloc-check.t
   t1025-rv1-reload.t
   t1026-rv1-partial-release.t
+  t1027-rv1-partial-release-brokerless-resources.t
   t3000-jobspec.t
   t3001-resource-basic.t
   t3002-resource-prefix.t
--- a/t/data/resource/jgfs/issue1284.json
+++ b/t/data/resource/jgfs/issue1284.json
@@ -1,0 +1,2333 @@
+{
+  "version": 1,
+  "execution": {
+    "R_lite": [
+      {
+        "rank": "0",
+        "children": {
+          "core": "0-9"
+        }
+      }
+    ],
+    "starttime": 0.0,
+    "expiration": 0.0,
+    "nodelist": [
+      "compute-01"
+    ]
+  },
+  "scheduling": {
+    "graph": {
+      "directed": true,
+      "nodes": [
+        {
+          "id": "0",
+          "metadata": {
+            "type": "cluster",
+            "basename": "compute",
+            "name": "compute0",
+            "id": 0,
+            "uniq_id": 0,
+            "rank": -1,
+            "exclusive": true,
+            "unit": "",
+            "size": 1,
+            "properties": {},
+            "paths": {
+              "containment": "/compute0"
+            }
+          }
+        },
+        {
+          "id": "1",
+          "metadata": {
+            "type": "rack",
+            "basename": "rack",
+            "name": "rack0",
+            "id": 0,
+            "uniq_id": 1,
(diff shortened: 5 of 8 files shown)


jameshcorbett (dismissed):
Confirmed it makes the error messages I was seeing go away!

milroy:
@jameshcorbett can I set MWP or do we need a test case to check for this in Fluxion?

jameshcorbett:
Yeah it would be good to have a test case somehow. I put this PR in flux-sched v0.38.0 via a patch so I don't think there's as much hurry to merge it. I'm also working on a test case in flux-coral2 environment but it's not working quite right yet and the tests are slow so it's taking a while.

I can provide a graph and jobspec that will hit the issue, but I don't know about the pattern of partial-cancel RPCs.

milroy:
After more thought, I think we need to add a testsuite check for this issue.

milroy:
> Some of my flux-coral2 tests are suggesting to me that the rabbit resources aren't freed, even though the error message goes away.

Let's make a reproducer similar to this one: https://github.com/flux-framework/flux-sched/blob/master/t/issues/t1129-match-overflow.sh

What are the scheduler and queue policies set to in the coral2 tests?

jameshcorbett:
> What are the scheduler and queue policies set to in the coral2 tests?

Whatever the defaults are I think, there's no configuration done.

> Let's make a reproducer similar to this one: https://github.com/flux-framework/flux-sched/blob/master/t/issues/t1129-match-overflow.sh

Thanks for the pointer, I should be able to look into this tomorrow!

milroy:
@jameshcorbett this PR is almost ready for review. First, I'd like to integrate the tests you wrote that reproduced the behavior and helped me fix it. Can you make a PR to my fork on the issue-1284 branch (https://github.com/milroy/flux-sched/tree/issue-1284)?

Alternatively I can cherry pick your commits and push them to my fork.

jameshcorbett:
> @jameshcorbett this PR is almost ready for review. First, I'd like to integrate the tests you wrote that reproduced the behavior and helped me fix it. Can you make a PR to my fork on the issue-1284 branch (https://github.com/milroy/flux-sched/tree/issue-1284)?
> 
> Alternatively I can cherry pick your commits and push them to my fork.

Working on that now.

milroy:
Could you refactor the tests and put them under `t/issues/` so the issue test driver script executes them?

resource/traversers/dfu_impl_update.cpp
```diff
@@ -535,6 +539,13 @@ int dfu_impl_t::mod_plan (vtx_t u, int64_t jobid, modify_data_t &mod_data)
         span = alloc_span->second;
         if (mod_data.mod_type != job_modify_t::PARTIAL_CANCEL) {
             (*m_graph)[u].schedule.allocations.erase (alloc_span);
+        } else {
+            // Add to brokerless resource set to execute in follow-up
+            // vertex cancel
+            // Need to convert to c_str () because interner is finalized
+            if (std::string ((*m_graph)[u].type.c_str ()) == "ssd")
+                mod_data.brokerless_res.insert (u);
+            goto done;
```
trws, line 548: This certainly works, but I'd recommend adding another long-lived resource_type for ssd in resource_data.[hc]pp instead, then referencing it here.  It will save the compute and make the code look simpler as well.

For other cases like this, where the resource type isn't known ahead of time, currently it's possible to catch the exception thrown if it can't be created to know that the match couldn't possibly succeed.  I could also add a way to do that in a non-throwing way if that would be helpful.

milroy: > This certainly works, but I'd recommend adding another long-lived resource_type for ssd in resource_data.[hc]pp instead, then referencing it here. It will save the compute and make the code look simpler as well.

I was sure there was a better way to do this and was going to ask you for input. I updated the PR to implement your suggestion.

> For other cases like this, where the resource type isn't known ahead of time, currently it's possible to catch the exception thrown if it can't be created to know that the match couldn't possibly succeed. 

This will be useful when match becomes more flexible. The more immediate use case will probably be integrating new resource types into fluxion via `grow`.

milroy: Leaving a note based on our conversation that the interner is unlocked during `grow` or acquisition, so this won't be an issue.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -667,23 +667,26 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
...
-                                        "%s: Final .free RPC failed to remove all resources for "
-                                        "jobid "
-                                        "%jd",
-                                        __FUNCTION__,
-                                        static_cast<intmax_t> (id));
+                        // This error condition can indicate a discrepancy between core and sched,
+                        // but commonly indicates partial cancel didn't clean up resources external
+                        // to a broker rank (e.g., ssds).
+                        flux_log (flux_h,
+                                  LOG_WARNING,
```
jameshcorbett, line 674: Does this mean `flux dmesg` is going to fill up with these warnings on rabbit systems? (not sure what our log levels are commonly set to)

resource/traversers/dfu_impl_update.cpp
```diff
@@ -539,6 +539,12 @@ int dfu_impl_t::mod_plan (vtx_t u, int64_t jobid, modify_data_t &mod_data)
         span = alloc_span->second;
         if (mod_data.mod_type != job_modify_t::PARTIAL_CANCEL) {
             (*m_graph)[u].schedule.allocations.erase (alloc_span);
+        } else {
+            // Add to brokerless resource set to execute in follow-up
+            // vertex cancel
+            if ((*m_graph)[u].type == ssd_rt)
```
jameshcorbett, line 545: Does this force ssds to be a brokerless resource always? Like there could never be `ssd` resources that belong to brokers (as they do on Sierra I think)?

jameshcorbett (commented):
Looks reasonable, a couple of questions and comments.

qmanager/policies/base/queue_policy_base.hpp
```diff
@@ -667,23 +667,26 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
                     auto job_sp = job_it->second;
```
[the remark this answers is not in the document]
milroy: Thanks; fixed it.

qmanager/policies/base/queue_policy_base.hpp
milroy: Yes, unfortunately I think it does. Open issue #1309 has some pros and cons to warning vs error messages. The potentially large number of messages in `flux dmesg` is a good argument for making this a `DEBUG` message.

resource/traversers/dfu_impl_update.cpp
milroy: No, if this condition is met that means the vertex has an allocation for the `jobid` in question but it is not one of the vertices canceled via the JGF reader (experimental) or in the broker ranks specified in the partial release `.free` RPC.

I do wonder if we want to drop the check for `(*m_graph)[u].type == ssd_rt`. Will there be other resource types not attached to a broker rank, and if so do we just want to cancel them no matter the type?

qmanager/policies/base/queue_policy_base.hpp
jameshcorbett: Agreed yeah I think clogging the `dmesg` logs would be pretty bad :/

jameshcorbett (approved):
LGTM I think, also confirmed it fixes the error in flux-coral2 that https://github.com/flux-framework/flux-coral2/pull/208 adds a check for (which is pretty much the same as the test added in this PR, so nothing too new).

resource/traversers/dfu_impl_update.cpp
trws: I think there will be. Especially when we go to doing remote ephemeral Lustre that might be a separate resource that would potentially encompass many ranks. Assuming that tweak, I also approve.

milroy:
I realized the brokerless resource cancellation implemented in this PR currently will lead to highly undesirable behavior. All the brokerless vertices will get cancelled in the first partial cancel traversal. That means that they will be allocatable before the final `.free` which can lead to state inconsistency between sched and core.

The best approach may be to propagate the final free to the traverser, which can then execute a cleanup, full cancel upon detecting brokerless resources. That has the added bonus of not triggering the `if (full_removal || final)` condition in qmanager, which will allow that condition to remain an error state.

A simpler alternative would be to always execute a full cancel when the final `.free` RPC is received. The disadvantage of that approach is additional overhead if there are no brokerless resources.

Thoughts @trws and @jameshcorbett?

trws:
If propagating the final free state to the traverser is feasible, I think that's a great choice, if not doing the full cancel unconditionally would be a good stopgap while we figure things out.

milroy:
I was able to implement propagating the free state to the traverser. I ran performance comparisons between the two approaches (propagation vs unconditional full cleanup cancel) and observed that the unconditional full cleanup cancel is almost always faster than propagating the free state. The exception appears to be when processing more than 1 `free` RPC.

However, I also observed errors with the free state propagation implementation under certain pruning filter configurations. For example, with the `ALL:core,ALL:ssd` pruning filter configuration the partial removal exits with an error on the `rack1` vertex. The error occurs when attempting to remove `core` types from the pruning filter span because the broker rank allocation being released includes N cores. However, `rack1` only contains `ssd` vertices, so the attempt to remove N cores from the pruning filter `core` span fails because N > 0 (i.e., it attempts to remove more resources than exist).

I can't think of a way to fix that error without adding substantial complexity to the traverser and planner logic. Given these findings I think the best course is to do a full cancel unconditionally upon receipt of the final `.free`. @trws what do you think?

milroy:
> The exception appears to be when processing more than 1 free RPC for a single jobid.

I expect this will be the most common case in production.

In terms of performance difference, I timed the cancellation in qmanager for testsuite tests 1026 and 1027. For cancellations with a single `.free` RPC for test 1027, the full cleanup cancel takes 102.3 microseconds on average (20 tests) in comparison with the propagating the free state implementation, which takes 121.8 microseconds (19% difference). 

For test 1026 with a series of four `.free` RPCs, the average (across 20 tests) sum of the four cancels is 495.2 microseconds for the full cleanup cancel in comparison with 424.3 microseconds for the propagating the free state implementation (14% difference).

trws:
It sounds like the complexity favors doing the unconditional full cancel, and the performance doesn't really push away from that much either.  Would you agree @milroy?  If so, then lets do that.  I have a feeling we should keep this in mind as we think through what we want to do when evolving the planner setup.

milroy:
> Would you agree @milroy? 

Yes, that's my conclusion, too.

>  I have a feeling we should keep this in mind as we think through what we want to do when evolving the planner setup.

Agreed. There's extra complexity and brittleness related to dealing with the planners in the context of allocation and resource dynamism and it manifests in unexpected ways.

trws:
Looks like there's a failure on el8, specifically failing to find an error output that's there?

```
  [Oct24 23:36] sched-fluxion-qmanager[0]: remove: Final .free RPC failed to remove all resources for jobid 88030052352: Success
  test_must_fail: command succeeded: grep free RPC failed to remove all resources log.out
  not ok 8 - no fluxion errors logged
```

Maybe it's going to stdout instead of stderr and getting missed? Not sure, seems odd that it's only there and not on the other ones though.

Is this otherwise ready to go? Trying to plan my reviews etc. for the day.

milroy:
> Unfortunately this looks like an error in the job removal. I don't yet know why it's only happening on el8.

The final flag on the `.free` RPC isn't getting set for `t1027` on el8 for some reason.

milroy:
@trws I thought of a much simpler and cleaner way to check whether a planner_multi sub span was removed by a prior partial cancel: https://github.com/flux-framework/flux-sched/pull/1292/commits/a42de67ef8dcfe0e8bbdc9f49c37ed2d28f66f6d

The PR is ready for a final review.

milroy:
I should add that I don't think the PR needs additional testsuite tests for whether a planner_multi sub span was removed by a prior partial cancel. The change is very small and should have been included in the first partial cancel PR.

trws (approved):

trws:
Looks good to me @milroy, thanks for all the work on this!

milroy:
Thanks for the review! Setting MWP.

PR #1293, opened 2024-09-05
JGF simplification

jameshcorbett:
Posting this as a WIP so I can link to it / work against it in flux-coral2.

### DIFF
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -347,7 +347,6 @@ int resource_reader_jgf_t::fill_fetcher (json_t *element,
                                          json_t **properties)
 {
     int rc = -1;
-    json_t *p = NULL;
     json_t *metadata = NULL;
 
     if ((json_unpack (element, "{ s:s }", "id", &f.vertex_id) < 0)) {
@@ -364,7 +363,7 @@ int resource_reader_jgf_t::fill_fetcher (json_t *element,
         goto done;
     }
     if ((json_unpack (metadata,
-                      "{ s:s s:s s:s s:I s:I s:I s?:i s:b s:s s:I }",
+                      "{ s:s s:s s:s s:I s:I s:I s?:i s:b s:s s:I s:o s?o }",
                       "type",
                       &f.type,
                       "basename",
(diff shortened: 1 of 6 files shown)


jameshcorbett:
OK, tests are failing, but I think it's generally working properly--the flux-coral2 testsuite runs successfully with the JGF generated from this branch. So remaining work is eliminating the `path` fields (which will require looking at edges, not sure how to do this yet) and then getting the tests to pass.

jameshcorbett:
Ok, only `t1018-rv1-bootstrap2.t` is failing now.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -341,36 +342,83 @@ int resource_reader_jgf_t::remap_aware_unpack_vtx (fetch_helper_t &f,
...
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": value for key (uniq_id) could not be inferred from outer ";
+            m_err_msg += "'id' field " + std::string (f.vertex_id) + ".\n";
+            return -1;
+        }
+    }
+    if (f.id == -1) {
+        f.id = f.uniq_id;
+        // for nodes, see if there is an integer suffix on the hostname and use it if so
+        if (f.type == std::string{"node"} && name != NULL) {
```
trws, line 360: This PR looks really good, the only case that mildly worries me is I can't recall how stoll handles leading zeros.  Some of the conversion routines use a leading zero to mean octal, and reject anything with multiple leading zeros, so does this work with node names like `cab003`?

jameshcorbett: I think we should be OK:

> If the value of base is ​0​, the numeric base is auto-detected: if the prefix is 0, the base is octal, if the prefix is 0x or 0X, the base is hexadecimal, otherwise the base is decimal. 

From https://en.cppreference.com/w/cpp/string/basic_string/stol

jameshcorbett: Oh and the default `base` is 10, which is what we're using here. Also confirmed in a little test of my own: `std::stoi("0000045"): 45`

trws: Excellent, thanks for checking!

trws (approved):

jameshcorbett:
Thanks, rebased with master, setting MWP.

PR #1295, opened 2024-09-12
cmake: clarify git-push suggestion

wihobbs:
@garlick pointed out in flux-framework/flux-core#6276 that the suggestion to `git push --tags` is a wee bit confusing, so I wanted to clarify in this one to stay consistent.


jameshcorbett (approved):
👍

PR #1296, opened 2024-09-16
Flexible Scheduling with or_slot [WIP]

zekemorton:
This PR introduces flexible scheduling to the traverser. Flexible scheduling will allow the user to define several different acceptable resource configurations that can be selected from while the traverser is walking the resource graph. These configurations are specified as or_slots, where a single or_slot is an acceptable compute unit. These or_slots are selected from equally from any available options at a given point in the graph traversal. Any combination of the or_slots may be selected, and they should be thought of as interchangeable configurations. 

Implementation: 
or_slot:
or_slots are a new slot-like resource type. They searched for and treated similarly to slots, but in this case you can have sibling or_slots at the same level. or slots also have the added complexity of needing to select the best configuration of the or_slots for available resources. At time of submitting this WIP PR, this is done by taking the union of all resources specified in all the or slots and completing the traversal with all of those resources to get resource counts for those resources. With those resource counts the best configurations can be selected and scheduled as normally with slots.

slot configuration selection:
After getting resource counts for all possible resources in all of the or_slots, the slot configuration is determined. I was not able to find an efficient means of finding the best configurations in terms of score. A first attempt landed me on a greedy algorithm that that would do well interns of score, but that would on occasion not find a match when a match could be found. This PR show an example of a DP algorithm that optimizes over the number of or_slots that it can schedule with the given resources. How the configuration is selected could be configurable in a similar way to how match policies are selected, and there could be much effort to give flexible and features to this selection process. An example might be to use the same DP algo, but weight some configurations more than others. 

Caveats:
- Because we take the union of all possible resources based on type, this leads to errors when using or_slots at levels above leaves. If we have tow configurations that have sockets, one with 10 cores and the other with 20, we can only get counts for sockets with one of those numbers of cores. This issue stems from how we color the graph during a traversal, which may not be necessary. It may be pssobile that we can remove the coloring, and then travers the graph for each or_slot. This may require some adaptation to the DP algo, but will be an overall improvement. 
- Slots are scheduled exclusively. Because slots are scheduled exclusively, this results in some weird behavior if or_slots are under slots or if or_slots contain non leaf nodes. This behavior needs to be tested further 
- The calculation for which configuration to use is done during the traversal and only with the context of the resource counts at that particular stage of the traversal. This means that with this method, it would be impossible to find an optimal configuration over the entire resource graph instead of just that local vertex

To do Items: 
- [ ] code clean up and add comments as needed 
- [ ] write full suite of tests, including tests with non leaf nodes
- [ ] investigate changes to traversal without graph coloring 
- [ ] modularize the or config selection policy
- [ ] handle or_slot counts appropriately 

Example Job Spec:
```
version: 9999
resources:
    - type: cluster
      count: 1
      with:
        - type: rack
          count: 1
          with:
            - type: node
              count: 1
              with:
                - type: socket
                  count: 1
                  with:
                    - type: or_slot
                      count: 1
                      label: small
                      with:
                        - type: core
                          count: 8
                        - type: gpu
                          count: 1
                    - type: or_slot
                      count: 1
                      label: big
                      with:
                        - type: core
                          count: 10
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: [ "app" ]
    slot: default
    count:
      per_slot: 1
```


zekemorton:
Capturing the original greedy solution here since I am going to collapse that commit with the dp solution:
```
int dfu_impl_t::dom_or_slot (const jobmeta_t &meta,
                             vtx_t u,
                             const std::vector<Resource> &slots,
                             unsigned int nslots,
                             bool pristine,
                             bool *excl,
                             scoring_api_t &dfu)
{
    int rc;
    bool x_inout = true;
    unsigned int qual_num_slots = 0;
    std::vector<eval_egroup_t> edg_group_vector;
    const subsystem_t &dom = m_match->dom_subsystem ();
    std::unordered_set<edg_t *> edges_used;
    scoring_api_t dfu_slot;

    // collect a set of all resource types in the or_slots to get resource
    // counts. This does not work well with non leaf vertex resources because
    // it cannot distinguish beyond type. This may be resolveable if graph
    // coloring is removed during the selection process.
    std::vector<Resource> slot_resource_union;
    std::set<resource_type_t> resource_types;
    for (auto &slot : slots) {
        for (auto r : slot.with) {
            if (resource_types.find (r.type) == resource_types.end ()) {
                resource_types.insert (r.type);
                slot_resource_union.push_back (r);
            }
        }
    }

    if ((rc = explore (meta,
                       u,
                       dom,
                       slot_resource_union,
                       pristine,
                       &x_inout,
                       visit_t::DFV,
                       dfu_slot,
                       nslots))
        != 0)
        goto done;
    if ((rc = m_match->dom_finish_slot (dom, dfu_slot)) != 0)
        goto done;
    for (unsigned int i = 0; i < nslots; ++i) {
        std::unordered_set<edg_t *> remove_edges;
        eval_egroup_t edg_group;
        for (auto slot : slots) {
            auto slot_shape = slot.with;
            int64_t score = MATCH_MET;
            eval_egroup_t test_edg_group;
            std::unordered_set<edg_t *> test_edges;
            bool found = true;
            for (auto &slot_elem : slot_shape) {
                unsigned int j = 0;
                unsigned int qc = dfu_slot.qualified_count (dom, slot_elem.type);
                unsigned int count = m_match->calc_count (slot_elem, qc);
                dfu_slot.eval_egroups_iter_reset (dom, slot_elem.type);
                while (j < count) {
                    auto egroup_i = dfu_slot.eval_egroups_iter_next (dom, slot_elem.type);
                    if (egroup_i == dfu_slot.eval_egroups_end (dom, slot_elem.type)) {
                        found = false;
                        break;
                    }
                    if (edges_used.find (&(*egroup_i).edges[0].edge) == edges_used.end ())
                        test_edges.insert (&(*egroup_i).edges[0].edge);
                    else
                        continue;
                    eval_edg_t ev_edg ((*egroup_i).edges[0].count,
                                       (*egroup_i).edges[0].count,
                                       1,
                                       (*egroup_i).edges[0].edge);
                    score += (*egroup_i).score;
                    test_edg_group.edges.push_back (ev_edg);
                    j += (*egroup_i).edges[0].count;
                }
            }
            if (!found) {
                continue;
            }
            if (edg_group.score < score) {
                remove_edges = test_edges;
                test_edg_group.score = score;
                test_edg_group.count = 1;
                test_edg_group.exclusive = 1;
                edg_group = test_edg_group;
            }
        }
        edges_used.insert (remove_edges.begin (), remove_edges.end ());
        edg_group_vector.push_back (edg_group);
    }
    for (auto &edg_group : edg_group_vector)
        dfu.add (dom, or_slot_rt, edg_group);

done:
    return (qual_num_slots) ? 0 : -1;
}
```

zekemorton:
After a discussion with @milroy, we decided that it does not make sense to modularize the policy and selection algorithm for or slots yet, and that we can save it for when we have more than one policy or decide on how that can be expressed in the job spec. this is a similar point to how we want to deal with or slot counts, we do not yet have a great solution for that yet. 

Additionally, I have done some performance testing on or slots to see how this algorithm affects the scheduling time. I uploaded the csv with this comment. The columns in the CSV are `Graph,Policy,Jobs,Slots,Or-Slots,Resources,Time` where Graph refers to the resource graph used, policy is the policy used either first or high, Jobs is the number of jobs scheduled or revered in a particular run, Slots is the number of regular slots either 0 or 1, or-slots is the number of or-slots, resources is the number of resources in a slot, and time is the time it took to schedule a single job using resource query. I plan to evaluate the data more closely and make some graphs, but I figured I upload the csv now in case anyone else wants to take a look 
[or_slot_results.csv](https://github.com/user-attachments/files/17564233/or_slot_results.csv)

zekemorton:
After feedback from @milroy and @trws, I reworked the structure of this code. Instead of adding a new resource type, `or_slot` I added a new traverser that changes the way that slots behave. Now, if you load the flexible traverser, `slot` will have the same behavior as `or_slot` as described above. 

To achieve this, I modified the class structure of `dfu_traverser_t` to not inherit from `dfu_implt_t` but to instead contain a member that is a `dfu_impl_t`. This works because the new flexible traverser, `dfu_flexible_t` inherits from `dfu_impl_t`.

zekemorton:
@milroy I think this is ready for review

resource/modules/resource_match.cpp
```diff
@@ -1533,6 +1534,9 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = 0;
 
+    // Select thr appropriate traverser based on CLI policy.
+    ctx->traverser =
+        std::make_shared<dfu_traverser_t> (ctx->opts.get_opt ().get_traverser_policy ());
```
milroy, line 1539: Per our discussion, `std::make_shared<dfu_traverser_t> ()` returns `nullptr` on error. Please update the traverser creation to use the same error condition checking as `create_match_cb` below.

zekemorton: Completed these changes with the latest push

resource/traversers/dfu.cpp
```diff
@@ -44,15 +44,15 @@ int dfu_traverser_t::is_satisfiable (Jobspec::Jobspec &jobspec,
     meta.alloc_type = jobmeta_t::alloc_type_t::AT_SATISFIABILITY;
```
milroy, line 44: Nit: commit message typo: "memerber element"

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
+/*****************************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
```
milroy: Nit: update the copyright year to 2025.

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
+/*****************************************************************************\
```
milroy, line 1: This comment applies to the commit message. Please add a few sentences to the commit message body describing the dynamic programming algorithm implementation.

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
...
+                accum_if (subsystem,
+                          aggregate.first,
+                          resource.count.min * aggregate.second,
+                          to_parent);
+            }
+        }
+    }
+}
+
+std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_config (
```
milroy: My original thought was to leave this as you wrote it and allow the compiler to perform RVO. However, every instance of calling `select_or_config ()` is in an assignment. My understanding is RVO doesn't occur in that case. You might want to use `std::move ()` in the assignments below or have the function return a reference. @trws may have a better suggestion.

milroy (commented):
This is looking very good. I'm leaving a first set of comments to start discussion on RVO before I complete reviewing the algorithm in the "traverser: add flexible traverser with or slots" commit.

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
+/*****************************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
```
milroy, line 2: There are a lot of opportunities to avoid copies in `for` loops and "in-out" parameters. Also, it's generally a good idea to add the `const` qualifier when possible. Here's a diff with my suggested changes, which should allow all tests to pass:

```diff
diff --git a/resource/traversers/dfu_flexible.cpp b/resource/traversers/dfu_flexible.cpp
index 0219acc7..73f9ad47 100644
--- a/resource/traversers/dfu_flexible.cpp
+++ b/resource/traversers/dfu_flexible.cpp
@@ -40,7 +40,7 @@ int dfu_flexible_t::match (vtx_t u,
 {
     int rc = -1;
     bool matched = false, or_matched = false;
-    for (auto &resource : resources) {
+    for (const auto &resource : resources) {
         if ((*m_graph)[u].type == resource.type) {
             // Limitations of DFU traverser: jobspec must not
             // have same type at same level Please read utilities/README.md
@@ -48,7 +48,7 @@ int dfu_flexible_t::match (vtx_t u,
                 goto ret;
             *match_resource = &resource;
             if (!resource.with.empty ()) {
-                for (auto &c_resource : resource.with) {
+                for (const auto &c_resource : resource.with) {
                     if (c_resource.type == slot_rt) {
                         *slot_resources = &resource.with;
                         *nslots = m_match->calc_effective_max (c_resource);
@@ -164,14 +164,14 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
         // Or slots should use a minimum of values rather than an accumulation
         // otherwise possible matches may be filtered out
         if (resource.type == slot_rt) {
-            for (auto &aggregate : resource.user_data) {
+            for (const auto &aggregate : resource.user_data) {
                 min_if (subsystem,
                         aggregate.first,
                         resource.count.min * aggregate.second,
                         to_parent);
             }
         } else {
-            for (auto &aggregate : resource.user_data) {
+            for (const auto &aggregate : resource.user_data) {
                 accum_if (subsystem,
                           aggregate.first,
                           resource.count.min * aggregate.second,
@@ -183,7 +183,7 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
 
 std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_config (
     const std::vector<Resource> &slots,
-    std::map<resource_type_t, int> resource_counts,
+    std::map<resource_type_t, int> &resource_counts,
     unsigned int nslots,
     std::unordered_map<Key, std::tuple<std::map<resource_type_t, int>, int, int>, Hash> &or_config)
 {
@@ -192,7 +192,7 @@ std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_c
     Key index = Key (resource_counts);
 
     // if available, use precomputed result
-    auto it = or_config.find (resource_counts);
+    auto const it = or_config.find (index);
     if (it != or_config.end ())
         return it->second;
 
@@ -202,7 +202,7 @@ std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_c
         return or_config[index];
     }
 
-    for (auto slot : slots) {
+    for (const auto &slot : slots) {
         int test;
         ++i;
         bool match = true;
@@ -210,7 +210,7 @@ std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_c
         updated_counts = resource_counts;
 
         // determine if there are enough resources to match with this or_slot
-        for (auto slot_elem : slot.with) {
+        for (const auto &slot_elem : slot.with) {
             unsigned int qc = resource_counts[slot_elem.type];
             unsigned int count = m_match->calc_count (slot_elem, qc);
             if (count <= 0) {
@@ -263,8 +263,8 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
     // coloring is removed during the selection process.
     std::vector<Resource> slot_resource_union;
     std::map<resource_type_t, int> resource_types;
-    for (auto &slot : slots) {
-        for (auto r : slot.with) {
+    for (const auto &slot : slots) {
+        for (const auto &r : slot.with) {
             if (resource_types.find (r.type) == resource_types.end ()) {
                 resource_types[r.type] = 0;
                 slot_resource_union.push_back (r);
@@ -301,12 +301,12 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
         int64_t score = MATCH_MET;
 
         // use calculated index to determine which or_slot type to use
-        for (auto &slot_elem : slots[slot_index].with) {
+        for (const auto &slot_elem : slots[slot_index].with) {
             unsigned int j = 0;
             unsigned int qc = dfu_slot.qualified_count (dom, slot_elem.type);
             unsigned int count = m_match->calc_count (slot_elem, qc);
             while (j < count) {
-                auto egroup_i = dfu_slot.eval_egroups_iter_next (dom, slot_elem.type);
+                const auto &egroup_i = dfu_slot.eval_egroups_iter_next (dom, slot_elem.type);
                 if (egroup_i == dfu_slot.eval_egroups_end (dom, slot_elem.type)) {
                     m_err_msg += __FUNCTION__;
                     m_err_msg += ": not enough slots.\n";
@@ -329,7 +329,7 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
 
         current_config = or_config[Key (std::get<0> (current_config))];
     }
-    for (auto &edg_group : edg_group_vector)
+    for (const auto &edg_group : edg_group_vector)
         dfu.add (dom, slot_rt, edg_group);
 
 done:
@@ -343,7 +343,7 @@ int dfu_flexible_t::prune_resources (const jobmeta_t &meta,
                                      const std::vector<Jobspec::Resource> &resources)
 {
     int rc = 0;
-    for (auto &resource : resources) {
+    for (const auto &resource : resources) {
         if ((*m_graph)[u].type != resource.type && resource.type != slot_rt)
             continue;
         // Prune by exclusivity checker
```

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
...
+                edg_group.edges.push_back (ev_edg);
+                j += (*egroup_i).edges[0].count;
+            }
+        }
+        edg_group.score = score;
+        edg_group.count = 1;
+        edg_group.exclusive = 1;
+        edg_group_vector.push_back (edg_group);
+
+        current_config = or_config[Key (std::get<0> (current_config))];
```
milroy: You can avoid calling the `Key` constructor by making the first element of `current_config` a `Key`. Here are the changes needed to `dfu_flexible.cpp`:

```diff
diff --git a/resource/traversers/dfu_flexible.cpp b/resource/traversers/dfu_flexible.cpp
index 0219acc7..2d6118c4 100644
--- a/resource/traversers/dfu_flexible.cpp
+++ b/resource/traversers/dfu_flexible.cpp
@@ -181,7 +181,7 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
     }
 }
 
-std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_config (
+std::tuple<dfu_flexible_t::Key, int, int> dfu_flexible_t::select_or_config (
     const std::vector<Resource> &slots,
     std::map<resource_type_t, int> resource_counts,
     unsigned int nslots,
@@ -255,7 +255,7 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
     std::unordered_set<edg_t *> edges_used;
     scoring_api_t dfu_slot;
     std::unordered_map<Key, std::tuple<std::map<resource_type_t, int>, int, int>, Hash> or_config;
-    std::tuple<std::map<resource_type_t, int>, int, int> current_config;
+    std::tuple<Key, int, int> current_config;
 
     // collect a set of all resource types in the or_slots to get resource
     // counts. This does not work well with non leaf vertex resources because
@@ -327,7 +327,7 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
         edg_group.exclusive = 1;
         edg_group_vector.push_back (edg_group);
 
-        current_config = or_config[Key (std::get<0> (current_config))];
+        current_config = or_config[std::get<0> (current_config)];
     }
     for (auto &edg_group : edg_group_vector)
         dfu.add (dom, slot_rt, edg_group);
```
I'll indicate the necessary changes to `dfu_flexible.hpp` as well.

resource/traversers/dfu_flexible.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "resource/traversers/dfu_impl.hpp"
+
+using namespace Flux::resource_model::detail;
+namespace Flux {
+namespace resource_model {
+
+class dfu_flexible_t : public dfu_impl_t {
+    // struct to convert map of resources counts to an index
+    struct Key {
+        Key ();
```
milroy: To allow `current_config` to have a `Key` as its first element, you'll need to explicitly set the default constructor. Here's the full diff to enable the change:

```diff
diff --git a/resource/traversers/dfu_flexible.hpp b/resource/traversers/dfu_flexible.hpp
index b56e00fb..1e18570a 100644
--- a/resource/traversers/dfu_flexible.hpp
+++ b/resource/traversers/dfu_flexible.hpp
@@ -20,7 +20,7 @@ namespace resource_model {
 class dfu_flexible_t : public dfu_impl_t {
     // struct to convert map of resources counts to an index
     struct Key {
-        Key ();
+        Key () = default;
         Key (const std::map<resource_type_t, int> &c)
         {
             counts = c;
@@ -63,9 +63,9 @@ class dfu_flexible_t : public dfu_impl_t {
                                                 bool &prestine,
                                                 unsigned int &nslots,
                                                 match_kind_t &ko);
-    std::tuple<std::map<resource_type_t, int>, int, int> select_or_config (
+    std::tuple<dfu_flexible_t::Key, int, int> select_or_config (
         const std::vector<Jobspec::Resource> &slots,
```

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
...
+                          resource.count.min * aggregate.second,
+                          to_parent);
+            }
+        }
+    }
+}
+
+std::tuple<std::map<resource_type_t, int>, int, int> dfu_flexible_t::select_or_config (
+    const std::vector<Resource> &slots,
+    std::map<resource_type_t, int> resource_counts,
```
milroy: In addition to passing the `resource_counts` parameter by reference (I already mentioned that in another comment), you can use the `const` qualifier here:

```suggestion
    const std::map<resource_type_t, int> &resource_counts,
```

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
...
+    for (auto slot : slots) {
+        int test;
+        ++i;
+        bool match = true;
+        std::map<resource_type_t, int> updated_counts;
+        updated_counts = resource_counts;
+
+        // determine if there are enough resources to match with this or_slot
+        for (auto slot_elem : slot.with) {
+            unsigned int qc = resource_counts[slot_elem.type];
```
milroy: If you use the `const` qualifier above, you need to access the element via `.at ()`:

```suggestion
            unsigned int qc = resource_counts.at (slot_elem.type);
```

resource/traversers/dfu_flexible.cpp
```diff
@@ -0,0 +1,365 @@
...
+        or_config[index] = std::make_tuple (resource_counts, nslots, 0);
+        return or_config[index];
+    }
+
+    for (auto slot : slots) {
+        int test;
+        ++i;
+        bool match = true;
+        std::map<resource_type_t, int> updated_counts;
+        updated_counts = resource_counts;
```
milroy: Possibly just a style suggestion, but you can combine these two lines:

```suggestion
        std::map<resource_type_t, int> updated_counts = resource_counts;
```

milroy (commented):
I've read through the dynamic programming algorithm, and it looks good to me. As we discussed a few times, we can extend this to a global XOR slot as well.

@trws this PR is ready for your review when you're back.

resource/traversers/dfu_flexible.cpp
trws: This is a better idea even without const FWIW, it will avoid implicitly adding an element that didn't exist.  It does add the possibility of an exception being thrown however.

zekemorton:
@milroy thank you for the suggested changes, I think they help a lot! I went ahead and applied them and pushed the changes here

milroy (dismissed):
I looked at the updates you made to the code based on my feedback and they all look good. I'm approving this PR.

Excellent work, @zekemorton!

resource/modules/resource_match.cpp
```diff
@@ -1533,6 +1534,12 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = 0;
 
+    // Select thr appropriate traverser based on CLI policy.
+    if (!(ctx->traverser =
```
trws: This condition can never be true.  `std::make_shared` only ever returns non-null. If it would be null, it throws `std::bad_alloc`.

zekemorton:
@milroy I just rebased this branch and I am guessing that maybe some of the changes to the pruning filters, perhaps from [this commit ](https://github.com/flux-framework/flux-sched/commit/a92e2c975fdeccc0fb727ca40ee50b351fda4486) might have affected the behavior for or_slots. In the currently failing test case, it looks like it is pruning at the cluster level, likely because there are no more gpus. I wanted to add this note in to discuss a bit further next week

zekemorton:
@milroy I think I found a solution to the pruning counts issue, I put my changes and and a description in the most recent commit. Let me know what you think! I can squash this commit later, but just wanted to let you review the method and changes first

zekemorton:
I am not sure why this last test is failing, it passes when I run it on Dane. It looks like maybe it is triggering a different ERRNO from the match allocate? The difference in output is directly triggered by the boolean `matched` which is set here:

```
// check for match success
    if ((errno == ENODEV) || (errno == EBUSY) || (errno == EINVAL) || (errno == ENOENT)) {
        matched = false;
        rc = -1;
    }
```

The only thing I can think of is that it's triggering a different errno or somehow unsetting for the CI tests?

resource/traversers/dfu_traverser_policy_factory.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+const std::string SIMPLE = "simple";
+
+bool known_traverser_policy (const std::string &policy);
+
+/*! Factory method for creating a matching callback
+ *  object, representing a matching policy.
+ */
+std::shared_ptr<dfu_impl_t> create_traverser (const std::string &policy = "simple");
+std::shared_ptr<dfu_impl_t> create_traverser (std::shared_ptr<resource_graph_db_t> db,
```
milroy, line 48: Please add a short, commented description of the two different traverser creation functions. See: https://github.com/flux-framework/flux-sched/blob/master/resource/policies/dfu_match_policy_factory.hpp

milroy (approved):
This looks great! Thanks for the persistence @zekemorton! Please add descriptive comments in the factory per my comment and this is good to merge.

milroy (approved):
LGTM! Go ahead and set MWP.

milroy:
Noting that @zekemorton discussed this PR with @trws in the group meeting yesterday, and Tom said he was good with the changes provided that the PR doesn't alter the default behavior (which it doesn't).

trws:
Awesome, great work!

PR #1297, opened 2024-09-17
JGF: default edge subsystem

jameshcorbett:
PR #1293 is getting too much (locally anyway, I haven't pushed my changes recently), I need to start smaller.

### DIFF
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -990,7 +990,7 @@ int resource_reader_jgf_t::unpack_edge (json_t *element,
     json_t *metadata = NULL;
     const char *src = NULL;
     const char *tgt = NULL;
-    const char *subsys = NULL;
+    const char *subsys = "containment";
 
     if ((json_unpack (element, "{ s:s s:s }", "source", &src, "target", &tgt)) < 0) {
         errno = EINVAL;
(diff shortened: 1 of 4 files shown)


resource/readers/resource_reader_jgf.cpp
```diff
@@ -1007,20 +1007,17 @@ int resource_reader_jgf_t::unpack_edge (json_t *element,
...
+        if ((json_unpack (metadata, "{ s:s }", "subsystem", &subsys)) < 0) {
+            errno = EINVAL;
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": subsystem key not found in edge metadata.\n";
+            goto done;
+        }
+        subsystem = subsys;
+    } else {
+        subsystem = "containment";
     }
```
trws, line 1020: It looks like this says "if the metadata key exists, it must contain a key named subsystem with a string value". Is that the requirement we want here? I could see us having other metadata keys at some point, and wanting it to still default the subsystem key if there is a metadata key but no subsystem key inside it.

src/python/fluxion/resourcegraph/V1.py
```diff
@@ -105,7 +103,7 @@ def __init__(self, rv1):
         rv1 -- RV1 Dictorary that conforms to Flux RFC 20:
                    Resource Set Specification Version 1
         """
-        super().__init__()
+        super().__init__(directed=False)
```
trws, line 106: Where did this come from? The graph is always directed, in fact a bidigraph if supported.

jameshcorbett: But it gets rid of the
```
"directed": true,
```
entry in edges like
```
"source": "44",
"target": "45",
"directed": true,
"metadata": {
  "subsystem": "containment"
}
```

jameshcorbett: I added a comment to clarify

resource/readers/resource_reader_jgf.cpp
jameshcorbett: Fixed I think, now it should only error out if there's an error unpacking.

src/python/fluxion/resourcegraph/V1.py
trws: Ah, got it, so we're not changing the semantics just getting rid of the field.  Makes sense.

src/python/fluxion/jsongraph/objects/graph.py
```diff
@@ -79,7 +79,7 @@ def add_edge(self, edge, force_direction=False):#TODO check existence of node id
         if isinstance(edge, Edge):
             if self._directed:
                 if edge.is_directed() == None:
-                    edge.is_directed(True)
+                    edge.set_directed(True)
```
trws, line 82: Looks like this is never called in our tests, at least during coverage, is it firing when desired?

jameshcorbett: I only hit the issue while working on [4392e83](https://github.com/flux-framework/flux-sched/pull/1297/commits/4392e83e88ab5915b6565b2d399602ac5e66f87f). If you don't specify whether the edge is directed AND you set the graph to be directed, the error occurs. Before this PR we failed the first condition, now with this PR we fail the second, so this conditional probably will never be hit except by custom subclasses of the graph.

trws: Ok, seems reasonable. Thanks!

trws (approved):

PR #1299, opened 2024-09-24
Jgf writer simplification

jameshcorbett:
Now that #1293 has made the JGF reader smarter about defaults and the ion-R utility less verbose, the JGF match writer needs similar treatment.


jameshcorbett:
Ok @trws there is one failing test but I'm pretty sure I know what's wrong and it's just a matter of updating some JGFs after making sure everything looks sane (which is kind of slow since the JSON isn't formatted nicely so requires a lot of manual intervention). If the code looks good to you I could probably get the tests passing in time for the release tomorrow.

trws (approved):

jameshcorbett:
Thanks! Setting MWP

PR #1287, opened 2024-09-03
Dense resource type with DoS protection

trws:
I'm still debating this change to be honest.  The refcounted strings for resource types are showing up as a meaningful cost in comparisons and refcounting, it's not terrible but it's also not great.  This solves the issue by making it so that dense storage interners have a notion of being "final" or placed in a state where new strings can't be added anymore. It also adds an interface to "open" a finalized interner for updates, on a per-thread basis, so that we can do things like database updates. To my surprise, this passed all but one of our tests without re-opening the interners, and that last one by re-opening them in update_resource.

The upside is that after this change, and without using it to optimize any of the type maps or anything like that, we get another 20-30% improvement in throughput for some of my tests.  I've seen as much as 250 jobs/s.  The downside is that the interface is a little bit odd for this, and at the moment the only way we have to signal the thing is closed is to throw an exception.  Given that the only source of new resource types after init should be jobspec though, which already throws exceptions for any parse errors (which is basically what this would be, if the resource type doesn't exist it can't match) I'm not sure how bad that is. Would love input on this.  Might be easier to use if it worked with a lock/unlock pair so lock_guard would work, not sure.

I _think_ this
fixes #1298

PR #1302, opened 2024-09-30
NEWS.md: add release notes for v0.39.0

jameshcorbett:
Problem: there are no release notes for v0.39.0.

Add them.

Depends on:

- [x] #1293 
- [x] #1287 

@zekemorton @milroy @trws anything else we should get in by tomorrow?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,20 @@
+flux-sched version 0.39.0 - 2024-10-01
+--------------------------------------
+
+### New Features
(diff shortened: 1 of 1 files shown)


trws:
I think that's the list, will go over the JGF simplification shortly, I just set MWP on the Dense resource type PR, because we're actually testing the error report in an old test, will add more to that but I think we need that fix in ASAP and I'm not sure I can get to the addition in a reasonable time.

trws (approved):

PR #1307, opened 2024-10-04
readers: fix rv1exec cancel with multiple entries in `R_lite` array

grondo:
This PR fixes resource cancel with rv1exec when there is more than one entry in the Rv1 `R_lite` array. 

Currently for _every_ resource free request (whether a "partial" cancel or not) where not all ranks have the same set of cores or gpus allocated, users will see these nuisance error message in their logs:

```
sched-fluxion-qmanager[0]: remove: Final .free RPC failed to remove all resources for jobid 228417863981989888: Success
sched-fluxion-qmanager[0]: jobmanager_free_cb: remove (queue=default id=228417863981989888): Protocol error
```

Since my home cluster has a differing number of cores between nodes, I get these errors for almost every job.

The issue is that while `resource_reader_rv1exec_t::partial_cancel_internal()` loops over the entries in `R_lite`, it doesn't accumulate the visited ranks idsets, instead overwriting the `ranks` string each time and only processing the final one.

Also, there clearly wasn't a test for this case, so I added a new sharness test that contains a reproducer, along with some sanity testing with housekeeping that could later be expanded.

### DIFF
--- a/resource/readers/resource_reader_rv1exec.cpp
+++ b/resource/readers/resource_reader_rv1exec.cpp
@@ -943,6 +943,7 @@ int resource_reader_rv1exec_t::partial_cancel_internal (resource_graph_t &g,
     json_t *entry = nullptr;
     const char *ranks = nullptr;
     struct idset *r_ids = nullptr;
+    size_t len;
 
     // Implementing cancellation of rank subgraph
     // will require further parsing of nodelist,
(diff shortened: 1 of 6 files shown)


jameshcorbett (approved):
Looks like a nice fix to me! Thanks for the extra tests.

garlick:
This fixes it for me!

grondo:
Setting MWP.

PR #1311, opened 2024-10-31
matrix: add el8 tag and make it test-install

trws:
problem: we're not hooking up to the el8 container from core properly

solution: tag the container after the test, and match the core container with test-install

This is meant to match with #flux-core/6401 and ensure we propagate the el8 container all the way through.


milroy (approved):
Looks good; thanks for the quick PR.

PR #1312, opened 2024-11-04
Rework libintern

### LINKED ISSUE #1308: fluxion specific `libintern.so` should have package-specific prefix and version suffix
Fluxion currently installs the string interner library as `/usr/lib64/libintern.so`. This has a larger than zero probability of conflicting with some other package, so as suggested by @jameshcorbett it should probably get a `fluxion` somewhere in there like `libfluxion-intern.so`.

Also, the library is for internal use only, and appears to be Fluxion version specific. To avoid conflicting with itself and to prevent/discourage linking directly to the library outside of Fluxion, a version should be added, like `libfluxion-intern.so.0.39.0`. 

This will prevent failure to reload fluxion modules after a package update because the broker still has the previous, incompatible libintern.so loaded.

Or, `libintern.so` could just be linked into the Fluxion modules and not installed at all (the simpler approach)

trws:
intern: fold into fluxion-data and version the so
problem: the libintern so wasn't being versioned, causing problems on
upgrades, see https://github.com/flux-framework/flux-sched/issues/1308

solution: the real dependency should be on libfluxion-data anyway, which
was closer to how we wanted it.  The libintern library is now built
static, linked into libfluxion-data, and provided that way.

Additionally, the libfluxion-data library is installed as
`libfluxion-data.so.{version without git suffix}` and linked as
`libfluxion-data.so.{version major}.{version minor}` with no unsuffixed
link installed.

The other two commits here are small fixes that I ran into needing to fix warnings in the cmake build and load properly on my tumbleweed vm.

fixes #1308


jameshcorbett (approved):
Looks reasonable!

CMakeLists.txt
```diff
@@ -131,9 +131,16 @@ pkg_check_modules(UUID REQUIRED IMPORTED_TARGET uuid)
 set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
-find_package(Boost 1.66 REQUIRED COMPONENTS
+find_package(Boost CONFIG 1.66 COMPONENTS
```
jameshcorbett, line 134: Typo in commit message: `con the find_package call`?

trws:
@grondo, I think this covers the issue, but want to run it by you before we finalize it.

garlick:
This is what I get when I build it:
```
$ dpkg -c flux-sched_0.39.0-17-g58458db7_amd64.deb|grep libfluxion-data
-rw-r--r-- root/root     43544 2024-11-04 16:36 ./usr/lib/x86_64-linux-gnu/libfluxion-data.so.0.39.0
lrwxrwxrwx root/root         0 2024-11-04 16:36 ./usr/lib/x86_64-linux-gnu/libfluxion-data.so.0.39 -> libfluxion-data.so.0.39.0
```
That seems OK to me?

trws:
That's what I was shooting for, so we could update on patch if we want to but will multi-version on possibly breaking changes (which right now means minor versions).

milroy (approved):
This change looks good to me.

Issue #1260, opened 2024-07-31
Error when matching rabbits: by_subplan: planner_multi_avail_during returned -1


jameshcorbett:
With this [JGF](https://github.com/user-attachments/files/16435134/tioga-JGF.json) and the jobspec below, I get the following error even though the match succeeds:

```yaml
version: 9999
resources:
  - type: slot
    count: 2
    label: default
    with:
    - type: ssd
      count: 1
    - type: node
      exclusive: true
      count: 1
      with:
      - type: slot
        label: task
        count: 1
        with:
        - type: core
          count: 1
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: [ "app" ]
    slot: task
    count:
      per_slot: 1
```

```
ERROR: by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
Invalid argument.
by_subplan: planner_multi_avail_during returned -1.
[snip -- continues for many many lines]
```

milroy:
I am able to reproduce this issue with a slightly simpler jobspec (note slot count of `1`):

```yaml
version: 9999
resources:
  - type: slot
    count: 1
    label: default
    with:
    - type: ssd
      count: 1
    - type: node
      exclusive: true
      count: 1
      with:
      - type: slot
        label: task
        count: 1
        with:
        - type: core
          count: 1
# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: [ "app" ]
    slot: task
    count:
      per_slot: 1
```
And I have a fix. 

The only remaining problem is I can't reproduce this issue with the test JGF (`t/data/resource/jgfs/issue1284.json`), with the reduced slot count. @jameshcorbett can you make me a smaller test JGF based on the tioga graph attached to this issue? Also, do you have any ideas why this won't reproduce on `issue1284.json`?

PR #1314, opened 2024-11-05
Don't prune traversal by leaf vertex subplans

milroy:
Issue #1260 and logs from production clusters have reported match errors such as ```by_subplan: planner_multi_avail_during returned -1.``` for JGFs with `node` vertices with an edge to the `cluster` vertex. The error occurs because PR #1248 prevents creation of subplan filters on leaf vertices, causing subsequent checks against the planner to fail. 

This PR adds a check to `by_subplan` to bypass pruning based on null subplans.

Currently, the test case succeeds without the fix because the error isn't logged in `flux dmesg` for some reason.


resource/traversers/dfu_impl.cpp
```diff
@@ -176,6 +176,11 @@ int dfu_impl_t::by_subplan (const jobmeta_t &meta,
     int saved_errno = errno;
     planner_multi_t *p = (*m_graph)[u].idata.subplans[s];
 
+    if (!p) {
+        // Subplan is null if u is a leaf.
```
jameshcorbett, line 180: Could the subplan be NULL for any other reasons? Are there any more checks needed here?

jameshcorbett (commented):
Looks good although it would be good to make sure the test fails without the PR...

resource/traversers/dfu_impl.cpp
trws: We discussed this in person this afternoon. If it's null for any other reason, it's unintentional, but checking if the vertex is a leaf or not is an unfortunately tricky/expensive operation to use in something that gets called this much.

milroy:
After many attempts I was unsuccessful at getting the error to propagate to `flux dmesg`. I updated the testsuite test to use `resource-query` instead. Let me know if I should keep poking at the resource module-based test.

I think the PR is ready for review.

jameshcorbett (approved):
LGTM thanks for this!

resource/traversers/dfu_impl.cpp
milroy: Noting for posterity that I tested pruning filter configurations that skip intermediate vertices (e.g., `cluster:core, node:core`, to skip `rack` vertices) and observed that the `rack` vertex subplan is not null. Apparently the code creates the planners in the intermediate vertices anyway.

milroy:
Thanks for the reviews. Setting MWP.

PR #1315, opened 2024-12-01
t: update job.exception output checks

chu11:
Problem: In the near future, the job.exception output in flux-core will change, resulting in failures in tests that expect a specific format.

Update the job.exception parsing in t1006-qmanager-multiqueue.t and t1024-alloc-check.t to be more flexible to future output changes.

----

in preparation for change in https://github.com/flux-framework/flux-core/pull/6453


trws (approved):
Looks ok, is the plan to put something between these elements? Would it make more sense to grep for all of the tokens separately?

chu11:
> Looks ok, is the plan to put something between these elements?

Yeah, the job id will go in between.

>  Would it make more sense to grep for all of the tokens separately?

Hmmm I guess could do that.  But not sure worth the effort right now given it's such a minor change.

trws:
That begs the question of why it isn't being added at the end, which would have left these unchanged.  Is there a common format being mimicked?

PR #1316, opened 2024-12-04
Add add_subgraph and remove_subgraph functionality to REAPI

milroy:
This PR exposes `grow` functions in the C and C++ REAPI bindings.

The `grow` functionality passes a JGF subgraph including the path from the cluster vertex to the subgraph root. For example, a JGF subgraph with a new node (`newnode`) and subnode resources to `cluster0` at `rack0` includes the cluster and rack vertices as well as the induced edges. The disadvantage of this approach is that the vertex metadata in the JGF subgraph needs to be sufficiently specified to identify the vertices that already exist in the graph (e.g., `cluster0`).


vsoch:
This is tested (and the basics are working in fluxion-go)! :partying_face: 

We start with cluster [tiny](https://github.com/flux-framework/fluxion-go/blob/add-grow-support/cmd/test/data/tiny.json), one rack and two nodes (node0 and node1). We demonstrate that if we ask for 4 nodes, we cannot be satisfied:

```console
Asking to MatchSatisfy 4 nodes (not possible)

        ----Match Satisfy output---
satisfied: false
error: <nil>
```

We then ask fluxion to grow from 2 to 4 nodes. We do that with a [new nodes request](https://github.com/flux-framework/fluxion-go/blob/add-grow-support/cmd/test/data/grow/new-nodes.json) that includes an existing path in the graph, `tiny0->rack0` but then defines two new nodes, `node2` and `node3`.

```console
🍔 Asking to Grow from 2 to 4 Nodes
Grow request return value: <nil>
```

We can get some verification that the graph node has node0-3 (4 nodes) by asking for 4 nodes with satisfy again. We are satisfied!

```console
Asking to MatchSatisfy 4 nodes (now IS possible)

        ----Match Satisfy output---
satisfied: true
error: <nil>
```

We now want to test shrink. Shrink takes the node path that we should prune at. For this case, we just prune off one node. 
```console
🥕 Asking to Shrink from 4 to 3 Nodes
Shrink request return value: <nil>
```
When we have 3 nodes we ask for 4 again, and we are no longer satisfied.

```console
Asking to MatchSatisfy 4 nodes (again, not possible)

        ----Match Satisfy output---
satisfied: false
error: <nil>
```

The testing shown above is [here in GitHub CI](https://github.com/flux-framework/fluxion-go/actions/runs/12780828020/job/35627749377) (shown across OSes) and that full PR is [here](https://github.com/flux-framework/fluxion-go/pull/12) and will just need to be updated when the branch here is merged (or we can work from this branch if that isn't going to happen soon - I can build/deploy a custom container that has it. 

## How will shrink work in Kubernetes?

High level - I'm thinking through the shrink design for a cluster in Kubernetes, and I think we have two use cases (that warrant different design strategies):

### 1. A need to shrink down in increments of 1

This will work fine to prune single nodes.

### 2. A need to shrink down in unknown increments (>1)

This could be a lot of requests to fluxion, for example, if we want to shrink down by 10, 20, or more nodes at once. We have a few options, I think. I'll think through each one.

- We can expose a function in fluxion-go that takes a list of nodes, and then (within the same call to fluxion-go) makes multiple calls to fluxion. That mostly replaces the need to do many grpc requests (one per node) with just one (to handle many requests, internally). 

- If we have an understanding of the increments, we can design a cluster graph with abstract levels of racks, where each rack is a group of nodes that are intended to be brought down together. Maybe we would design that graph based on topology. That way, we can do one request to trim the rack, and it will cut off all the nodes.

I haven't looked at what fluxion is doing in terms of the actual shrink (happening during a traversal?) but if there is an operation that can handle multiple cuts at once (e.g., done during one traversal) that could be an idea. But based on my impression of infrequent scaling, I don't really think optimizing this up the wazoo right now is that necessary.

## How will grow design work in Kubernetes?

For the first case (a flat topology that has nodes added to it) we likely just need to get the highest node identifier in the graph, and then we generate the json request that adds 1 to that (and however many we need). That can be stored as a variable somewhere, and on restart it can be calculated again from the live cluster.

For the second case that has multiple racks each with children, we would likely apply the same strategy as above, but on the level of the rack, and then the number of children under a rack is constant. We would calculate the node indices based on the number of racks and expected nodes per each. In the case that we have different numbers of nodes per rack (e.g., imagine different applications requiring different sized increments) we could label the rack with metadata that says exactly the number of children that are there.

Anyway - there are many ways to skin an avocado! Thanks for finishing this up @milroy I'm super pumped to see it working, and (TBA) to get it merged and deployed and into some of our projects!

vsoch:
@trws when you are back in business :briefcase: could you review this PR? And no rush if you are still on travel or otherwise - I can continue building from the PR branch.

milroy:
> @trws when you are back in business 💼 could you review this PR? And no rush if you are still on travel or otherwise - I can continue building from the PR branch.

We'll need to discuss what to do about testing. These API changes look like dead code to coverage.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -722,6 +748,73 @@ int resource_query_t::remove_job (const uint64_t jobid, const std::string &R, bo
...
+    }
+    // No need to check if params.load_format != "jgf" since
+    // remove_subgraph takes a subgraph path, which is
+    // supported by any resource graph
+    if ((reader = create_resource_reader ("jgf")) == nullptr) {
+        m_err_msg = __FUNCTION__;
+        m_err_msg += ": ERROR: can't create JGF reader\n";
+        return rc;
+    }
+    if ((rc = reader->remove_subgraph (db->resource_graph, db->metadata, subgraph_path)) != 0) {
```
vsoch: Is `remove_subgraph` on the reader or the traverser? I am looking [here](https://github.com/flux-framework/flux-sched/blob/e1af864ee48d503314fd6acc075102151cbdc0ee/resource/traversers/dfu_impl.hpp#L611).

milroy: I moved `remove_subgraph` from the reader to the traverser in a PR that was merged before the rebase here. Good catch!

trws (dismissed):

resource/modules/resource.cpp
```diff
@@ -1030,6 +1045,62 @@ static void get_property_request_cb (flux_t *h,
...
+{
+    std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
+    std::string errmsg = "", R_s = "";
+    char *R_subgraph = NULL;
+
+    if (flux_request_unpack (msg, NULL, "{s:s}", "subgraph", &R_subgraph) < 0)
+        goto error;
+
+    R_s = R_subgraph;
+    if (run_add_subgraph (ctx, R_s) != 0) {
```
trws:
```suggestion
    std::string errmsg = "";
    char *R_subgraph = NULL;

    if (flux_request_unpack (msg, NULL, "{s:s}", "subgraph", &R_subgraph) < 0)
        goto error;

    if (run_add_subgraph (ctx, R_subgraph) != 0) {
```

resource/modules/resource.cpp
```diff
@@ -1030,6 +1045,62 @@ static void get_property_request_cb (flux_t *h,
...
+{
+    std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
+    std::string errmsg = "", s_path = "";
+    char *subgraph_path = NULL;
+
+    if (flux_request_unpack (msg, NULL, "{s:s}", "subgraph_path", &subgraph_path) < 0)
+        goto error;
+
+    s_path = subgraph_path;
+    if (run_remove_subgraph (ctx, s_path) != 0) {
```
trws:
```suggestion
    std::string errmsg = "";
    char *subgraph_path = NULL;

    if (flux_request_unpack (msg, NULL, "{s:s}", "subgraph_path", &subgraph_path) < 0)
        goto error;

    if (run_remove_subgraph (ctx, subgraph_path != 0) {
```

resource/modules/resource_match.cpp
```diff
@@ -1685,6 +1685,67 @@ int run_find (std::shared_ptr<resource_ctx_t> &ctx,
...
+    if ((reader = create_resource_reader ("jgf")) == nullptr) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "creating JGF reader failed: %s",
+                  ctx->traverser->err_message ().c_str ());
+        goto done;
+    }
+    // Could be adding new resource types, so open the interner storage before adding the subgraph
+    subsystem_t::storage_t::open ();
+    resource_type_t::storage_t::open ();
```
trws: This is the right logic, but should be using the RAII wrapper I added to the interner.  Not sure it existed when you wrote this, but it would increase our exception safety here.  Should be just change these for:


```suggestion
    auto subsystem_open = subsystem_t::storage_t::open_for_scope ();
    auto resource_open = resource_t::storage_t::open_for_scope ();
```

and then remove the finalizes below.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -739,6 +769,59 @@ int resource_query_t::remove_job (const uint64_t jobid, const std::string &R, bo
...
+    vtx_t v = boost::graph_traits<resource_graph_t>::null_vertex ();
+
+    if ((reader = create_resource_reader ("jgf")) == nullptr) {
+        m_err_msg = __FUNCTION__;
+        m_err_msg += ": ERROR: can't create JGF reader\n";
+        goto out;
+    }
+    // Could be adding new resource types, so open the interner storage before adding the subgraph
+    subsystem_t::storage_t::open ();
+    resource_type_t::storage_t::open ();
```
trws: Same as the module implementation.

trws:
I had a couple small things, but with those and resolving the conflicts in the two cmakefiles I think this is good to go.

resource/reapi/test/reapi_cli_test.cpp
```diff
@@ -0,0 +1,137 @@
+#define CATCH_CONFIG_MAIN
+
+#include <catch2/catch_test_macros.hpp>
+#include <resource/reapi/bindings/c/reapi_cli.h>
+#include <fstream>
+#include <iostream>
+
+namespace Flux::resource_model::detail {
```
jameshcorbett: this test needs to be added to the build system?

resource/modules/resource_match.cpp
```diff
@@ -1745,6 +1745,68 @@ int run_find (std::shared_ptr<resource_ctx_t> &ctx,
...
+                  "creating JGF reader failed: %s",
+                  ctx->traverser->err_message ().c_str ());
+        goto done;
+    }
+    // Could be adding new resource types, so open the interner storage before adding the subgraph
+    // Scoped so the goto out statements above don't cross initialization.
+    {
+        auto subsystem_open = subsystem_t::storage_t::open_for_scope ();
+        auto resource_open = resource_type_t::storage_t::open_for_scope ();
+    }
```
jameshcorbett: Since these are scoped though, won't they be destructed and closed immediately?

resource/reapi/test/reapi_cli_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+TEST_CASE ("Initialize REAPI CLI and test match, satisfy, and cancel",
+           "[initialize-match-satisfy-cancel]")
+{
+    std::string options = "{\"load_format\": \"grug\"}";
+    std::stringstream buffer, buffer2, buffer3;
+    std::ifstream resource_file ("../../../t/data/resource/grugs/tiny.graphml");
+    std::ifstream jobspec_file ("../../../t/data/resource/jobspecs/basics/test001.yaml");
+    std::ifstream jobspec_file2 ("../../../t/data/resource/jobspecs/basics/test011.yaml");
+    match_op_t match_op = match_op_t::MATCH_ALLOCATE;
```
jameshcorbett: Use `SHARNESS_TEST_SRCDIR` here instead?

resource/reapi/test/reapi_cli_test.cpp
jameshcorbett: It also fails for me when I ran it locally.

resource/modules/resource_match.cpp
```diff
@@ -1745,6 +1745,68 @@ int run_find (std::shared_ptr<resource_ctx_t> &ctx,
...
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "add subgraph by path: %s failed: %s",
+                  R_subgraph.c_str (),
+                  reader->err_message ().c_str ());
+        goto done;
+    }
+    // Update total counts:
+    rc = ctx->traverser->initialize ();
+    flux_log (ctx->h, LOG_DEBUG, "successfully added subgraph %s", R_subgraph.c_str ());
```
jameshcorbett, line 1766: delete this line? It's logged without even checking the value of `rc`.

resource/modules/resource_match.cpp
trws: Yes, they need to be at function scope rather than a sub-scope like that.

resource/modules/resource_match.cpp
trws: Probably copy up the equivalent block from `run_remove_subgraph`: https://github.com/flux-framework/flux-sched/pull/1316/changes/BASE..408a2c01ce5270815c701964ba9821c6944320c4#diff-48628982be5211b25532174bc920cdedb26fbc5cee95e74f9f3288c1390b4f2eR1797-R1804

resource/reapi/bindings/c++/reapi.hpp
```diff
@@ -194,6 +194,38 @@ class reapi_t {
...
+     *                   service module, it is expected to be a pointer
+     *                   to a flux_t object.
+     *  \param R_subgraph JGF string of subgraph to attach to existing resources.
+     *                   Supports adding a JGF subgraph including the path from the
+     *                   cluster root to the subgraph root.
+     *  \return          0 on success; -1 on error.
+     */
+    static int add_subgraph (void *h, const std::string &R_subgraph)
+    {
+        return -1;
```
jameshcorbett, line 210: For not implemented methods @garlick set `errno = ENOSYS` elsewhere, I think maybe we should follow that pattern? E.g. commit [67952773d9a903f2a3582e19b19b1fe196fc9427](https://github.com/flux-framework/flux-sched/pull/1489/changes/67952773d9a903f2a3582e19b19b1fe196fc9427)

resource/modules/resource_match.cpp
milroy: Yes, this was an error that I realized shortly after pushing. I fixed it but didn't push the changes.

milroy:
I haven't finished updating this PR since it requires reworking the Catch2 test among several other things. I'll let you know when it's ready for a full review.

resource/modules/resource.cpp
milroy: Implemented.

resource/modules/resource_match.cpp
milroy: Good catch! Fixed.

resource/reapi/bindings/c++/reapi.hpp
milroy: The not-yet-implemented methods are set as you describe in the implementation header file (`reapi_cli_impl.hpp`). The ones with `return -1` are actually implemented.

resource/reapi/test/reapi_cli_test.cpp
milroy: I totally reworked the tests. I fixed this as part of the refactor.

resource/reapi/test/reapi_cli_test.cpp
milroy: I totally reworked the tests. I fixed this as part of the refactor.

resource/modules/resource.cpp
milroy: Implemented.

milroy:
I was in the midst of a refactor when @trws and @jameshcorbett looked at this, so the last version was pretty rough. I've implemented your suggestions, reworked the tests, and caught another bug (with the help of Claude).

This is ready for a final review.

resource/modules/resource_match.cpp
```diff
@@ -1745,6 +1745,46 @@ int run_find (std::shared_ptr<resource_ctx_t> &ctx,
...
+    // Could be adding new resource types, so open the interner storage before adding the subgraph
+    auto subsystem_open = subsystem_t::storage_t::open_for_scope ();
+    auto resource_open = resource_type_t::storage_t::open_for_scope ();
+
+    if ((reader = create_resource_reader ("jgf")) == nullptr) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "creating JGF reader failed: %s",
+                  reader->err_message ().c_str ());
+        goto done;
```
jameshcorbett, line 1746: `reader` is null here

jameshcorbett (approved):
One thing, otherwise this LGTM

resource/modules/resource_match.cpp
milroy: Good catch; fixed.

jameshcorbett:
Final review from @trws then merge?

milroy:
> Final review from @trws then merge?

Yes, that sounds right.

trws:
Looking it over now.

resource/reapi/bindings/c++/reapi.hpp
trws: these are somewhat strange.  Really they should be abstract methods, but they aren't methods.  They're in a strange limbo place because of the way reapi is set up.  In principle they should probably be just `throw std::logic_error("Function not implemented");` or something but since the rest of these are all like this, I'd say leave it as is right now.

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -203,9 +203,15 @@ extern "C" int reapi_cli_match (reapi_cli_ctx_t *ctx,
                                 int64_t *at,
                                 double *ov)
 {
-    uint64_t seq = ctx->rqt->get_job_counter ();
     int rc = -1;
 
+    if (!ctx || !ctx->rqt) {
+        errno = EINVAL;
+        return -1;
+    }
```
trws, line 211: Good catch!

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -393,13 +417,15 @@ extern "C" int reapi_cli_stat (reapi_cli_ctx_t *ctx,
...
+    std::string err_buf;
     if (ctx->rqt)
-        err_buf = ctx->rqt->get_resource_query_err_msg () + reapi_cli_t::get_err_message ()
-                  + ctx->err_msg;
-    else
-        err_buf = reapi_cli_t::get_err_message () + ctx->err_msg;
+        err_buf = ctx->rqt->get_resource_query_err_msg ();
+    err_buf += reapi_cli_t::get_err_message () + ctx->err_msg;
 
     return strdup (err_buf.c_str ());
```
trws, line 430: This is right given the contract we had, but I'm wondering if we should look at making this not return a string the caller has to free.

trws (approved):
Looks great, thanks for the quick rework @milroy!

resource/reapi/bindings/c/reapi_cli.cpp
milroy: Yeah, that makes sense. What about adding a commit that does something like this:

```C++
struct reapi_cli_ctx {
    resource_query_t *rqt;
    std::string err_msg;
    std::string err_out;   // return buffer for reapi_cli_get_err_msg
};

extern "C" const char *reapi_cli_get_err_msg (reapi_cli_ctx_t *ctx)
{
    if (!ctx) {
        errno = EINVAL;
        return "ERROR: reapi_cli context is null\n";   // static storage; not freed
    }
    ctx->err_out.clear ();
    if (ctx->rqt)
        ctx->err_out = ctx->rqt->get_resource_query_err_msg ();
    ctx->err_out += reapi_cli_t::get_err_message () + ctx->err_msg;
    return ctx->err_out.c_str ();
}
```

trws: How about we make that a new PR? I think this one is ready to go as is.

milroy:
Thanks for the helpful feedback and improvements @vsoch, @jameshcorbett, and @trws! Setting MWP.

PR #1320, opened 2024-12-10
qmanager: track API changes in core

garlick:
Problem: the raw RPC interfaces are changing to use a `size_t `instead of `int` for payload size.

Update scheduler usage.

flux-framework/flux-core#6467 must be merged before this will pass CI.


grondo (approved):
LGTM

PR #1321, opened 2024-12-17
qmanager: send partial-ok with sched.hello

garlick:
This changes qmanager to invoke the `sched.hello` request with the flag that tells it to allow partial allocations through.  schedutil then diminishes _R_ in the hello responses by the `free` set (if present) before handing it to the hello callback in qmanager.   We were wondering if fluxion would perhaps just handle this.

flux-framework/flux-core#6445 must be merged before it can work with flux-core master, however, preliminary testing with that branch appears to show a problem.  The sharness test added here fails when I reload the scheduler with one node in housekeeping, and fluxion says:
```
Dec 17 00:52:30.860604 UTC sched-fluxion-resource.err[0]: parse_R: json_unpack
Dec 17 00:52:30.860614 UTC sched-fluxion-resource.err[0]: run_update: parsing R: Invalid argument
Dec 17 00:52:30.860616 UTC sched-fluxion-resource.err[0]: update_request_cb: update failed (id=46825209856): Invalid argument
Dec 17 00:52:30.860730 UTC sched-fluxion-qmanager.err[0]: jobmanager_hello_cb: reconstruct (id=46825209856 queue=default): Invalid argument
Dec 17 00:52:30.860842 UTC sched-fluxion-qmanager.err[0]: error raising fatal exception on f2ELnSXR: job is inactive: No such file or directory
```
So there is a little more work to do here to run that down.  Marking this as a WIP.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -551,9 +551,14 @@ static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
             ctx = nullptr;
             goto done;
         }
+        int schedutil_flags = 0;
+#ifdef SCHEDUTIL_HELLO_PARTIAL_OK
+        // flag was added in flux-core 0.70.0
+        schedutil_flags |= SCHEDUTIL_HELLO_PARTIAL_OK;
+#endif
         if (!(ctx->schedutil =
                   schedutil_create (ctx->h,
-                                    0,
+                                    schedutil_flags,
                                     &ops,
                                     std::static_pointer_cast<qmanager_cb_ctx_t> (ctx).get ()))) {
             flux_log_error (ctx->h, "%s: schedutil_create", __FUNCTION__);
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -553,7 +553,7 @@ static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
         }
         if (!(ctx->schedutil =
                   schedutil_create (ctx->h,
-                                    SCHEDUTIL_FREE_NOLOOKUP,
+                                    0,
                                     &ops,
                                     std::static_pointer_cast<qmanager_cb_ctx_t> (ctx).get ()))) {
             flux_log_error (ctx->h, "%s: schedutil_create", __FUNCTION__);
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -280,7 +280,6 @@ out:
 
 static void status_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_t *msg, void *arg)
 {
-    size_t len = 0;
     const char *payload;
     flux_future_t *f = NULL;
 
@@ -288,12 +287,12 @@ static void status_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
         flux_log_error (h, "%s: flux_rpc (sched-fluxion-resource.status)", __FUNCTION__);
         goto out;
     }
-    if (flux_rpc_get_raw (f, (const void **)&payload, &len) < 0) {
-        flux_log_error (h, "%s: flux_rpc_get_raw", __FUNCTION__);
+    if (flux_rpc_get (f, &payload) < 0) {
+        flux_log_error (h, "%s: flux_rpc_get", __FUNCTION__);
         goto out;
     }
-    if (flux_respond_raw (h, msg, (const void *)payload, len) < 0) {
-        flux_log_error (h, "%s: flux_respond_raw", __FUNCTION__);
+    if (flux_respond (h, msg, payload) < 0) {
+        flux_log_error (h, "%s: flux_respond", __FUNCTION__);
         goto out;
     }
     flux_future_destroy (f);
@@ -310,25 +309,19 @@ static void feasibility_request_cb (flux_t *h,
                                     const flux_msg_t *msg,
                                     void *arg)
 {
-    size_t size = 0;
     flux_future_t *f = nullptr;
     const char *data = nullptr;
 
-    if (flux_request_decode_raw (msg, nullptr, (const void **)&data, &size) < 0)
+    if (flux_request_decode (msg, nullptr, &data) < 0)
         goto error;
-    if (!(f = flux_rpc_raw (h,
-                            "sched-fluxion-resource.satisfiability",
-                            data,
-                            size,
-                            FLUX_NODEID_ANY,
-                            0))) {
+    if (!(f = flux_rpc (h, "sched-fluxion-resource.satisfiability", data, FLUX_NODEID_ANY, 0))) {
         flux_log_error (h, "%s: flux_rpc (sched-fluxion-resource.satisfiability)", __FUNCTION__);
         goto error;
     }
-    if (flux_rpc_get_raw (f, (const void **)&data, &size) < 0)
+    if (flux_rpc_get (f, &data) < 0)
         goto error;
-    if (flux_respond_raw (h, msg, (const void *)data, size) < 0) {
-        flux_log_error (h, "%s: flux_respond_raw", __FUNCTION__);
+    if (flux_respond (h, msg, data) < 0) {
+        flux_log_error (h, "%s: flux_respond", __FUNCTION__);
         goto error;
     }
     flux_log (h, LOG_DEBUG, "%s: feasibility succeeded", __FUNCTION__);
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1528,8 +1528,8 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
     int rc = 0;
     int version = 0;
     int saved_errno;
-    uint64_t st = 0;
-    uint64_t et = 0;
+    double tstart = 0;
+    double expiration = 0;
     json_t *o = NULL;
     json_t *graph = NULL;
     json_error_t error;
@@ -1541,23 +1541,25 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
         errno = EINVAL;
         goto out;
     }
-    if ((rc = json_unpack (o,
-                           "{s:i s:{s:I s:I} s?:o}",
-                           "version",
-                           &version,
-                           "execution",
-                           "starttime",
-                           &st,
-                           "expiration",
-                           &et,
-                           "scheduling",
-                           &graph))
+    if ((rc = json_unpack_ex (o,
+                              &error,
+                              0,
+                              "{s:i s:{s:F s:F} s?:o}",
+                              "version",
+                              &version,
+                              "execution",
+                              "starttime",
+                              &tstart,
+                              "expiration",
+                              &expiration,
+                              "scheduling",
+                              &graph))
         < 0) {
         errno = EINVAL;
-        flux_log (ctx->h, LOG_ERR, "%s: json_unpack", __FUNCTION__);
+        flux_log (ctx->h, LOG_ERR, "%s: json_unpack: %s", __FUNCTION__, error.text);
         goto freemem_out;
     }
-    if (version != 1 || st < 0 || et < st) {
+    if (version != 1 || tstart < 0 || expiration < tstart) {
         rc = -1;
         errno = EPROTO;
         flux_log (ctx->h,
@@ -1565,8 +1567,8 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
                   "%s: version=%d, starttime=%jd, expiration=%jd",
                   __FUNCTION__,
                   version,
-                  static_cast<intmax_t> (st),
-                  static_cast<intmax_t> (et));
+                  static_cast<intmax_t> (tstart),
+                  static_cast<intmax_t> (expiration));
         goto freemem_out;
     }
     if (graph != NULL) {
@@ -1585,8 +1587,8 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
         format = "rv1exec";
     }
 
-    starttime = static_cast<int64_t> (st);
-    duration = et - st;
+    starttime = static_cast<int64_t> (tstart);
+    duration = static_cast<uint64_t> (expiration - tstart);
 
 freemem_out:
     saved_errno = errno;
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1559,7 +1559,8 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
         flux_log (ctx->h, LOG_ERR, "%s: json_unpack: %s", __FUNCTION__, error.text);
         goto freemem_out;
     }
-    if (version != 1 || tstart < 0 || expiration < tstart) {
+    if (version != 1 || tstart < 0 || expiration < tstart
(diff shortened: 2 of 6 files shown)


qmanager/modules/qmanager.cpp
```diff
@@ -560,7 +560,7 @@ static std::shared_ptr<qmanager_ctx_t> qmanager_new (flux_t *h)
         }
         if (!(ctx->schedutil =
                   schedutil_create (ctx->h,
-                                    SCHEDUTIL_FREE_NOLOOKUP,
+                                    SCHEDUTIL_HELLO_PARTIAL_OK,
```
grondo, line 563: Suggestion: if this coe is modified to use the `SCHEDUTIL_HELLO_PARTIAL_OK` macro only if it exists, then flux-sched won't need to have a new dependency on an as-yet unreleased version of flux-core.

grondo (commented):
You may have already done this, but I modified Fluxion's `parse_R` to print the `json_unpack` error and got:

```
Dec 17 16:40:58.375707 UTC sched-fluxion-resource.err[0]: parse_R: json_unpack: Expected integer, got real
```

grondo:
[RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html) allows optional microsecond precision in R `starttime` and `expiration` (and I'm guess that's what Fluxion is tripping over here), so I think `parse_R` needs a simple fix to accept real/float for `starttime` and `expiration`.

(FYI - I did make that simple change and your new test passes)

garlick:
Nice!  Thanks!  I will add those changes to the PR.

garlick:
OK pushed the following
- @grondo's fix for `parse_R()`
- modification to previous raw RPC changes to avoid dependency on flux-core 0.70
- now that flux-core has changed the schedutil flags to preprocessor defines, use the PARTIAL_OK one conditionally
- make the test conditional also

I'm not sure if we need a test that covers partial release with JGF as well.  I'll leave this a WIP that's figured out.

garlick:
Another push:
- fixed formatting
- fixed race between qmanager hello handshake and tests

garlick:
Strange.  The new test is failing on _some_ builders.  The test output includes indicates that the scheduler successfully set the partial-ok flag:

```
job-manager.debug[0]: scheduler: hello +partial-ok
```
but then later, the job manager behaves as though it's not set:
```
job-manager.err[0]: housekeeping: fv-az1945-857 (rank 0) from f2Mt8nf9 will be terminated because scheduler does not support restart with partially released resources
```
Bah, I've been staring at this for a while and think I'll step away and try again tonight or tomorrow.

garlick:
Reran the failing tests and they all work now!?!

garlick:
That _must_ be it, but the `+partial-ok` message in the logs is printed by the job manager, added by flux-framework/flux-core#6445 (the new flux-core version).   The only way I can think of that happening and then the test failing with the job manager message indicating that the scheduler doesn't support partial-ok is if, when the test reloads the fluxion modules, it's loading an older version that doesn't set the flag.  But that seems impossible.

I'm somewhat inclined to just move on from this since it is working now, but it is disconcerting!

grondo:
I was thinking it was a delay in available of the switch to preprocessor defines for schedutil flags  (flux-framework/flux-core#6520), but since the `+partial-ok` message was present in the logs, I'm also mystified.

garlick:
Just pushed a test with `match-format=rv1`, which fails.

The test does the following:
- runs a 4 node job, arranging for 1 node get stuck in housekeeping
- confirms job completed, housekeeping launched, and one node stuck
- fluxion thinks only 1 node is allocated (CORRECT)
- reloads the scheduler with the node still stuck
- fluxion thinks 4 nodes are allocated (WRONG)
- kills the stuck node
- fluxion thinks 0 nodes are allocated (CORRECT)

My guess is that the JGF reader is just reading the `scheduling` key and ignoring the ranks in Rv1 object, but I have not confirmed this.

garlick:
@milroy if you have any hints on this one, much appreciated!

milroy:
@garlick the reason for the failure is that the `free` key isn't getting unpacked in `jobmanager_hello_cb` and a partial free isn't getting run on the extracted ranks. This change allows all tests to pass in the bookworm image based on your branch:

```diff
diff --git a/qmanager/modules/qmanager_callbacks.cpp b/qmanager/modules/qmanager_callbacks.cpp
index 56251f83..920126ca 100644
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -142,6 +142,7 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg, const
     int rc = -1;
     std::string R_out;
     char *qn_attr = NULL;
+    const char *free_ranks = NULL;
     std::string queue_name;
     std::shared_ptr<queue_policy_base_t> queue;
     std::shared_ptr<job_t> running_job = nullptr;
@@ -150,6 +151,7 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg, const
     unsigned int prio;
     uint32_t uid;
     double ts;
+    std::string R_free;
     json_t *jobspec = NULL;
     flux_future_t *f = NULL;
 
@@ -158,7 +160,7 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg, const
      * evolves to include it.  If it is not set, it must be looked up.
      */
     if (flux_msg_unpack (msg,
-                         "{s:I s:i s:i s:f s?o}",
+                         "{s:I s:i s:i s:f s?s s?o}",
                          "id",
                          &id,
                          "priority",
@@ -167,6 +169,8 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg, const
                          &uid,
                          "t_submit",
                          &ts,
+                         "free",
+                         &free_ranks,
                          "jobspec",
                          &jobspec)
         < 0) {
@@ -216,6 +220,17 @@ int qmanager_cb_t::jobmanager_hello_cb (flux_t *h, const flux_msg_t *msg, const
               "requeue success (queue=%s id=%jd)",
               queue_name.c_str (),
               static_cast<intmax_t> (id));
+    if (free_ranks) {
+        R_free = "{\"version\":1,\"execution\":{\"R_lite\":[{\"rank\":\"" + std::string (free_ranks) + "\"}]}}";
+        if (queue->remove (static_cast<void *> (h), id, false, R_free.c_str ()) < 0) {
+            flux_log_error (h,
+                            "%s: partial free (id=%jd queue=%s)",
+                            __FUNCTION__,
+                            static_cast<intmax_t> (id),
+                            queue_name.c_str ());
+            goto out;
+        }
+    }
     rc = 0;
 out:
     flux_future_destroy (f);
```

garlick:
Progress!  That does get the tests passing alright!

Any worries about the following:

- The _R_ object built here using the free ranks is techncally invalid because it doesn't include the required `chidren` key (see [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html))
- In the `rv1_nosched` case `queue->remove()` would be attempting to free ranks that aren't allocated, because they have already been subtracted from _R_ passed in to the callback and then to `queue->reconstruct()`

I would have thought the correct solution would be to find where `queue->reconstruct()` is allocating resources for the `rv1` (as opposed to `rv1_nosched`) case, and have that code free the intersection of the JGF and rv1 `rank` key.

However, I'm fine with this  if you are!

garlick:
I just pushed that change, but I realized there may be a problem if any of the job's partially released resources had been reallocated to other jobs before the scheduler reload.

Then one or more jobs would be attempting to allocate the same resources during hello, and fail.

milroy:
> The R object built here using the free ranks is techncally invalid because it doesn't include the required chidren key (see [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html))

All the reader is doing in this case is extract the ranks to be canceled from _R_ (so _R_ doesn't have to be valid). I could add the `children` key to the `R_free` string, or I could also modify the reader to accept a string with only a range of ranks. The latter would require more complex modifications.

> * In the rv1_nosched case queue->remove() would be attempting to free ranks that aren't allocated, because they have already been subtracted from R passed in to the callback and then to queue->reconstruct()

That could be a problem depending on how the partial cancel fails. I don't think it would be too hard to handle those failures, though.

> I would have thought the correct solution would be to find where queue->reconstruct() is allocating resources for the rv1 (as opposed to rv1_nosched) case, and have that code free the intersection of the JGF and rv1 rank key.

> I just pushed that change, but I realized there may be a problem if any of the job's partially released resources had been reallocated to other jobs before the scheduler reload.
Then the partially released job would be attempting to allocate the same resources during hello, and fail.

These two observations are closely related and are valid concerns. There currently isn't a way to perform the intersection of the JGF and rv1 rank key, but I think adding support for the intersection would address these problems.

garlick:
Just pushed an enhancement to the test that demonstrates the failure just described (where patially freed nodes have been reallocated to another job.  I see this:
```
Dec 19 18:45:27.085776 UTC sched-fluxion-resource.err[0]: run: dfu_traverser_t::run (id=59877883904): update_vtx_plan: can't add span into core0.
Dec 19 18:45:27.085783 UTC sched-fluxion-resource.err[0]: run_update: run: Invalid argument
Dec 19 18:45:27.085786 UTC sched-fluxion-resource.err[0]: update_request_cb: update failed (id=59877883904): Invalid argument
Dec 19 18:45:27.085867 UTC sched-fluxion-qmanager.err[0]: jobmanager_hello_cb: reconstruct (id=59877883904 queue=default): Invalid argument
Dec 19 18:45:27.085955 UTC sched-fluxion-qmanager.err[0]: error raising fatal exception on f2aEComh: job is inactive: No such file or directory

```
and later when core tries to free resources it thinks are still allocated to the partially released job
```
Dec 19 18:45:27.414696 UTC sched-fluxion-qmanager.err[0]: jobmanager_free_cb: can't find queue for job (id=59877883904): No such file or directory
```

milroy:
@garlick I'll work on the JGF and rv1 `rank` intersection and send an update when it's ready.

On a related note, I don't seem to be able to fork your fork. I used to have a fork but must have deleted it and now GitHub thinks it still exists. Any ideas how I might make a PR to your fork?

garlick:
Thanks @milroy !

When working with someone else on a PR I usually add their fork as a remote, pull down their working branch, do some work,  then push it to my own repo.  Like
```
git remote add garlick  https://github.com/garlick/flux-sched.git
git fetch garlick
git checkout partial_ok
git push origin partial_ok
```
(if origin is the remote for your fork)

Then you can just tell me to cherry pick from your copy of my branch when you have something for me to try.

t/t1026-rv1-partial-release.t
```diff
@@ -174,20 +174,26 @@ test_expect_success HAVE_PARTIAL_OK 'run a job and wait for node to get stuck' '
...
+test_expect_success HAVE_PARTIAL_OK 'fluxion still shows 2 nodes allocated' '
+	test $(fluxion_allocated nnodes) -eq 2
 '
 test_expect_success HAVE_PARTIAL_OK 'kill housekeeping' '
 	flux housekeeping kill --all
 '
-test_expect_success HAVE_PARTIAL_OK 'fluxion shows 0 nodes allocated' '
+test_expect_success HAVE_PARTIAL_OK 'fluxion shows 1 node allocated' '
 	hk_wait_for_running 0 &&
 	test $(fluxion_allocated nnodes) -eq 0
```
milroy, line 198: I've updated the code to do the JGF intersection with the `free` ranks, and it passes everything but this check. Should this line be:
```suggestion
        test $(fluxion_allocated nnodes) -eq 1
```

garlick: oops, correct!

milroy:
> When working with someone else on a PR I usually add their fork as a remote, pull down their working branch, do some work, then push it to my own repo. Like

@garlick thanks for the reminder, I'm pretty sure that's what I did last time and just forgot.

I've pushed my changes here: https://github.com/milroy/flux-sched/tree/partial_ok. Note that I included a revert commit for f27b290b5ff02ea9ec609805c013b682cd3006a4 because those changes shouldn't be included. You can just cherry pick the other commits and drop f27b290b5ff02ea9ec609805c013b682cd3006a4 from your branch. I also included a fixup for the testsuite fixup to correct the allocated node equality check.

garlick:
I did just, plus reordered the commits in the PR to avoid breaking `git bisect`.  I'll drop the WIP from the PR title since this seems complete now.

Thanks!  Nice work @milroy!

t/t1026-rv1-partial-release.t
```diff
@@ -96,16 +116,91 @@ test_expect_success 'attempt to ensure dmesg buffer synchronized' '
...
+'
+test_expect_success HAVE_PARTIAL_OK 'reload fluxion modules with match-format=rv1' '
+	remove_qmanager &&
+	reload_resource match-format=rv1 &&
+	load_qmanager_sync &&
+	flux resource list &&
+	FLUX_RESOURCE_LIST_RPC=sched.resource-status flux resource list
+'
+test_expect_success HAVE_PARTIAL_OK 'run a job and wait for node to get stuck' '
+	flux run -N4 true
```
milroy, line 170: This was causing testsuite failures for the `fedora40 - clang18` image for me. Adding `&&` fixed it:
```suggestion
        flux run -N4 true &&
```

garlick: oops! Thanks!  Pushing that.

garlick:
Anything left to do here?  It'd be good to get this merged.

resource/modules/resource_match.cpp
```diff
@@ -1541,32 +1541,34 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
...
+                              "scheduling",
+                              &graph))
         < 0) {
         errno = EINVAL;
-        flux_log (ctx->h, LOG_ERR, "%s: json_unpack", __FUNCTION__);
+        flux_log (ctx->h, LOG_ERR, "%s: json_unpack: %s", __FUNCTION__, error.text);
         goto freemem_out;
     }
-    if (version != 1 || st < 0 || et < st) {
+    if (version != 1 || tstart < 0 || expiration < tstart) {
```
milroy, line 1562: There needs to be an additional check for numeric limits:

```suggestion
    if (version != 1 || tstart < 0 || expiration < tstart || expiration > static_cast<double> (std::numeric_limits<int64_t>::max ()) {
```

milroy (changes requested):
This looks really good, except there's an additional numerical check needed in `parse_R`. It's a quick fix.

I think we need an additional fluxion developer to take a look. @jameshcorbett or @trws, can you provide feedback?

trws:
This looks good to me. I'm vaguely nervous about the fact we have yet more code that only exists in the JGF reader, but that's a longer-term concern where we really need to clean out cruft rather than something to deal with here.  I'm good with taking this once the extra check you noted goes in @milroy.

resource/modules/resource_match.cpp
trws: @garlick, if you're good with this I think this is otherwise GTG.

garlick: OK by me!

Note I cannot approve this since I originally proposed it.  I'll leave it to you two to approve and set merge-when-passing when it's ready.

trws (approved):

trws:
Ok, I applied the last check, reformatted, rebased and set MWP.  @milroy if you'll approve I think this is good to go.

milroy (approved):
LGTM! I'll set MWP.

milroy:
Oh, right, MWP is already set.

PR #1322, opened 2024-12-18
Find by jobid and output aggregate filter data

milroy:
Errors in partial cancel are causing inability to schedule resources in production. While evidence points to orphaned (i.e., not cleared) vertex data such as aggregate filters, it is difficult to investigate and debug without being able to inspect the state of the vertices.

The PR adds capability to print vertex-specific data by `jobid` and an `agfilter` boolean. The state can be printed based on whether the jobid is in the `allocations`, `reservations`, `job2span`, and `tags` maps.

This PR is WIP because I need to add support for all writers.

### DIFF
--- a/resource/evaluators/expr_eval_api.cpp
+++ b/resource/evaluators/expr_eval_api.cpp
@@ -309,13 +309,14 @@ int expr_eval_api_t::validate (const std::string &e, const expr_eval_target_base
 {
     int rc = -1;
     pred_op_t op;
-    std::size_t next, at;
+    std::size_t next, at, last_before_space;
 
     if ((rc = validate_paren (e, target, 0, next)) < 0)
         goto done;
     at = next;
 
-    while (at <= e.find_last_not_of (" \t")) {
+    last_before_space = e.find_last_not_of (" \t");
+    while (at <= last_before_space) {
         if ((op = parse_pred_op (e, at, next)) == pred_op_t::UNKNOWN) {
             rc = -1;
             goto done;
@@ -339,14 +340,15 @@ int expr_eval_api_t::evaluate (const std::string &e,
 {
     int rc = -1;
     pred_op_t op;
-    std::size_t next, at;
+    std::size_t next, at, last_before_space;
     bool result1, result2;
 
     if ((rc = evaluate_paren (e, target, 0, next, result1)) < 0)
         goto done;
     at = next;
 
-    while (at <= e.find_last_not_of (" \t")) {
+    last_before_space = e.find_last_not_of (" \t");
+    while (at <= last_before_space) {
         if ((op = parse_pred_op (e, at, next)) == pred_op_t::UNKNOWN) {
             rc = -1;
             goto done;
--- a/resource/evaluators/expr_eval_api.cpp
+++ b/resource/evaluators/expr_eval_api.cpp
@@ -234,6 +234,29 @@ done:
     return rc;
 }
 
+int expr_eval_api_t::extract_leaf (const std::string &e,
+                                   const expr_eval_target_base_t &target,
+                                   std::vector<std::pair<std::string, std::string>> &predicates)
+{
+    int rc = -1;
+    std::size_t delim;
+    std::string key, value;
+
+    if ((delim = e.find_first_of ("=")) == std::string::npos) {
+        errno = EINVAL;
+        goto done;
+    }
+    key = e.substr (0, delim);
+    value = e.substr (delim + 1);
+    if ((rc = (&target)->extract (key, value, predicates)) < 0) {
+        errno = EINVAL;
+        goto done;
+    }
+
+done:
+    return rc;
+}
+
 int expr_eval_api_t::evaluate_pred (pred_op_t op, bool result2, bool &result1) const
 {
     int rc = 0;
@@ -248,6 +271,33 @@ int expr_eval_api_t::evaluate_pred (pred_op_t op, bool result2, bool &result1) c
             rc = -1;
             errno = EINVAL;
     }
+    return rc;
+}
+
+int expr_eval_api_t::extract_paren (const std::string &e,
+                                    const expr_eval_target_base_t &target,
+                                    size_t at,
+                                    size_t &nx,
+                                    std::vector<std::pair<std::string, std::string>> &predicates)
+{
+    int rc = -1;
+    std::size_t tok;
+    std::size_t len;
+
+    if (!is_paren (e, at)) {
+        if ((rc = parse_expr_leaf (e, at, tok, len)) < 0)
+            goto done;
+        if ((rc = extract_leaf (e.substr (tok, len), target, predicates)) < 0)
+            goto done;
+    } else {
+        if ((rc = parse_expr_paren (e, at, tok, len)) < 0)
+            goto done;
+        if ((rc = extract (e.substr (tok + 1, len - 2), target, predicates)) < 0)
+            goto done;
+    }
+    nx = tok + len;
+    rc = 0;
+done:
     return rc;
 }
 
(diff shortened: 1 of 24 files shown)


resource/planner/c/planner_c_interface.cpp
```diff
@@ -628,9 +628,8 @@ extern "C" int planner_reduce_span (planner_t *ctx,
         update_mintime_resource_tree (ctx, list);
         rc = 0;
     } else {
-        // Error
+        // Error; rc already -1
         errno = EINVAL;
-        rc - 1;
```
jameshcorbett, line 633: How does this cause incorrect behavior? I mean, it looks like it should be removed but also like it has no effect on anything?

jameshcorbett (commented):
I just gave it a brief look-through, how are you feeling about getting this in by Tuesday which is I think the deadline for getting the RPM into the next TOSS version?

resource/planner/c/planner_c_interface.cpp
milroy: You're right. There's no lvalue assignment here.

milroy:
> how are you feeling about getting this in by Tuesday which is I think the deadline for getting the RPM into the next TOSS version?

I'd lie to have @trws feedback before merging this PR.

milroy:
I updated the PR to include an alternative method of extracting `(p, x)` pairs of strings (in the spirit of what you suggested, @trws, and included that approach as a series of fixup commits. Those fixups can be squashed depending on which approach is better. 

This PR is no longer WIP and is ready for review.

resource/evaluators/expr_eval_api.cpp
```diff
@@ -315,6 +365,34 @@ int expr_eval_api_t::evaluate (const std::string &e,
...
+{
+    int rc = -1;
+    pred_op_t op;
+    std::size_t next, at;
+
+    if ((rc = extract_paren (e, target, 0, next, predicates)) < 0)
+        goto done;
+    at = next;
+
+    while (at <= e.find_last_not_of (" \t")) {
```
trws, line 380: Since `e` is static, may as well save the search each time.

```suggestion
    auto last_before_space = e.find_last_not_of (" \t");
    while (at <= last_before_space) {
```

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -47,7 +47,18 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
         rc = (lcx == "reserved" || lcx == "free") ? 0 : -1;
-    else
+    else if (p == "jobid-alloc" || p == "jobid-span" || p == "jobid-tag" || p == "jobid-reserved") {
+        try {
+            std::stoul (lcx);
+            rc = 0;
+        } catch (std::invalid_argument) {
+            errno = EINVAL;
+        } catch (std::out_of_range) {
+            errno = EINVAL;
```
trws, line 57: Why `EINVAL` rather than `ERANGE`?

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -85,6 +98,32 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
             result =
                 !m_overridden.sched_future_reserved && (*m_g)[m_u].schedule.reservations.empty ();
         }
+    } else if (p == "jobid-alloc") {
+        jobid = std::stoul (lcx);
+        result = !((*m_g)[m_u].schedule.allocations.find (jobid)
+                   == (*m_g)[m_u].schedule.allocations.end ());
```
trws, line 104:
```suggestion
        result = (*m_g)[m_u].schedule.allocations.contains (jobid);
```

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -85,6 +98,32 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
...
+    } else if (p == "jobid-reserved") {
+        jobid = std::stoul (lcx);
+        result = !((*m_g)[m_u].schedule.reservations.find (jobid)
+                   == (*m_g)[m_u].schedule.reservations.end ());
```
trws, line 108:
```suggestion
        result = (*m_g)[m_u].schedule.reservations.contains (jobid);
```

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -85,6 +98,32 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
...
+    } else if (p == "jobid-span") {
+        jobid = std::stoul (lcx);
+        if ((*m_g)[m_u].idata.job2span.find (jobid) == (*m_g)[m_u].idata.job2span.end ()) {
```
trws, line 111:
```suggestion
        if (!(*m_g)[m_u].idata.job2span.contains (jobid)) {
```

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -85,6 +98,32 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
...
+            goto done;
+        }
+        result = true;
+    } else if (p == "jobid-tag") {
+        jobid = std::stoul (lcx);
+        if ((*m_g)[m_u].idata.tags.find (jobid) == (*m_g)[m_u].idata.tags.end ()) {
```
trws, line 117:
```suggestion
        if (!(*m_g)[m_u].idata.tags.contains (jobid)) {
```

resource/traversers/dfu_impl.cpp
```diff
@@ -829,12 +833,46 @@ int dfu_impl_t::dom_find_dfv (std::shared_ptr<match_writers_t> &w,
     } else if (!result && !nchildren) {
         goto done;
     }
+    if (agfilter) {
+        // Check if there's a pruning (aggregate) filter initialized
+        if ((filter_plan = (*m_graph)[u].idata.subplans[dom]) == NULL)
+            goto done;
```
trws, line 839: Do you want to skip emitting the vertex entirely if the `agfilter` option is true and there isn't a filter initialized on this vertex? I feel like we talked about this but it's completely swapped out of cache.

trws (commented):
One question and a few stylistic things in comments on this review.  If the agfilter emit skip thing is intentional, then just a bit of cleanup and I'm good with this.

milroy:
Thanks for the feedback @trws! I'll implement it shortly.

High level question: do you prefer the original implementation of extraction, or the second one (in the `fixup!` commits)? I figured the latter because they it's closer to your initial feedback in December, but I want to be sure before I squash the commits.

trws:
Just glanced again, and the generic predicates rather than the explicit pair look better to me. 👍 on the fixups

resource/evaluators/expr_eval_api.cpp
milroy: Good idea. I've added a separate commit to apply the change to the existing functions, too.

resource/evaluators/expr_eval_vtx_target.cpp
milroy: It should be `ERANGE`. Fixed!

resource/evaluators/expr_eval_vtx_target.cpp
milroy: That's a lot cleaner! I'll use `contains ()` in the future.

resource/evaluators/expr_eval_vtx_target.cpp
milroy: Ditto above.

resource/evaluators/expr_eval_vtx_target.cpp
milroy: Ditto above.

resource/evaluators/expr_eval_vtx_target.cpp
milroy: Ditto above.

resource/traversers/dfu_impl.cpp
milroy: I think we do want to skip it, because otherwise if you just do something like `find agfilter=true` the output will be the entire resource graph rather than only the vertices (and induced edges) with agfilters. You can get the entire resource graph with another query.

trws: Got it, thanks. Approving momentarily.

trws (approved):

milroy:
Thanks for the reviews @jameshcorbett and @trws! Setting MWP.

PR #1323, opened 2024-12-18
Clear traverser before traversal operations in dfu

milroy:
Recent debugging revealed that the traverser error message wasn't getting cleared after some traversal operations, leading to spurious or error logging that can be unrelated to the error cause.

Wrap each traverser function call with `clear_err_message ()` to ensure error messages are correctly attributed to the cause.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1201,7 +1201,6 @@ static int mark_now (std::shared_ptr<resource_ctx_t> &ctx,
                         "%s: traverser::mark: %s",
                         __FUNCTION__,
                         ctx->traverser->err_message ().c_str ());
-        ctx->traverser->clear_err_message ();
         goto done;
     }
     flux_log (ctx->h,
@@ -2685,7 +2684,6 @@ static int run_find (std::shared_ptr<resource_ctx_t> &ctx,
                             "%s: %s",
                             __FUNCTION__,
                             ctx->traverser->err_message ().c_str ());
-            ctx->traverser->clear_err_message ();
         }
         goto error;
     }
@@ -2956,7 +2954,6 @@ static void set_status_request_cb (flux_t *h,
                         "%s: traverser::mark: %s",
                         __FUNCTION__,
                         ctx->traverser->err_message ().c_str ());
-        ctx->traverser->clear_err_message ();
         errmsg = "Failed to set status of resource vertex";
         goto error;
     }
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
Looks reasonable and harmless. Did you verify this makes error messages more reasonable?

milroy:
> Did you verify this makes error messages more reasonable?

I'm looking at the test cases to see if there's a straightforward way to check that the message has been cleared.

trws (changes requested):
I like what this is doing, but I'd prefer to have the `clear_err_message` calls at the beginning of the mark, run and find functions rather than their callers so we get that consistently between the module and command callers. If I'm reading it right, it should work to put them at the start of the methods in the dfu_traverser_t type in dfu.cpp.

milroy:
> If I'm reading it right, it should work to put them at the start of the methods in the dfu_traverser_t type in dfu.cpp.

That makes sense. I'll update the commit shortly.

milroy:
@trws, I updated the PR to address your change request and added an additional commit to remove the error message clearing from the resource match callers.

Let me know if the PR needs additional changes.

milroy:
I just remembered that the error message is actually member data of `dfu_impl_t` and `dfu_traverser_t` calls methods that change the error message.

I think we want to clear the error message at the beginning of methods in `dfu_impl_t`.

trws (approved):
Looks great, thanks @milroy!

trws:
> I just remembered that the error message is actually member data of `dfu_impl_t` and `dfu_traverser_t` calls methods that change the error message.
> 
> I think we want to clear the error message at the beginning of methods in `dfu_impl_t`.

I see what you mean, but I _think_ it's already where we want it.  The error message accumulates errors or issues as it passes through `dfu_impl_t` methods (hence all the `+= err` stuff).  With the clears in the `dfu_traverser_t` we should get the full error from whatever `dfu_impl_t` operations it ran as part of the traverser operation, where if we clear at the start of each of the `impl` methods we'd only get the last one unless we ensure that we propagate any error eagerly.

milroy:
> With the clears in the dfu_traverser_t we should get the full error from whatever dfu_impl_t operations it ran as part of the traverser operation, where if we clear at the start of each of the impl methods we'd only get the last one unless we ensure that we propagate any error eagerly.

That's a good point. I'll leave the clear calls in the `dfu_traverser_t` public API.

milroy:
The PR is good to go. I'm setting MWP. Thanks for the feedback @trws and @jameshcorbett!

milroy:
> I'm looking at the test cases to see if there's a straightforward way to check that the message has been cleared.

As an update, I can't think of an easy way to add a check for an uncleared error message.

PR #1325, opened 2025-01-09
Qmanager: remove no-longer-supported schedutil_free_respond()

milroy:
The `.free` RPC no longer receives a response, and `schedutil_free_respond()` is a noop. `jobmanager_free_cb` currently contains a call to schedutil_free_respond, which misleadingly suggests that not sending the response could be an error condition in flux-core.

This short PR remove the call and associated error handling from the qmanager free callback.


milroy:
@trws and @jameshcorbett I planned to add this commit to PR #1321 but realized that PR isn't related to the `.free` RPC, so I made this small PR instead.

jameshcorbett (approved):
👍

trws (approved):
Works for me, thanks @milroy!

Issue #1326, opened 2025-01-15
CI: alpine-arm64 build fails


jameshcorbett:
The alpine-arm64 build (which only runs on master) is failing and has been for several months. This means that the docker manifest is not generated and so multi-architecture images (such as `el9`) are only tagged separately (like `el9-amd64` and `el9-arm64`) rather than together.

```
  #8 [3/5] RUN sudo /scripts/install-deps.sh
  #8 0.056 .buildkit_qemu_emulator: /bin/sh: Invalid ELF image for this architecture
  #8 ERROR: process "/dev/.buildkit_qemu_emulator /bin/sh -c sudo /scripts/install-deps.sh" did not complete successfully: exit code: 255
  ------
   > [3/5] RUN sudo /scripts/install-deps.sh:
  0.056 .buildkit_qemu_emulator: /bin/sh: Invalid ELF image for this architecture
  ------
  Dockerfile:9
  --------------------
     7 |     # Install extra buildrequires for flux-sched:
     8 |     COPY scripts/ /scripts
     9 | >>> RUN sudo /scripts/install-deps.sh
    10 |     
    11 |     # Add configured user to image with sudo access:
  --------------------
  ERROR: failed to solve: process "/dev/.buildkit_qemu_emulator /bin/sh -c sudo /scripts/install-deps.sh" did not complete successfully: exit code: 255
docker-run-checks.sh: docker build failed
```

jameshcorbett:
@trws do you know what the source of the issue is here?

trws:
That looks like it's either not setting up the emulator correctly or it's not running the right container architecture on the emulator.  Likely the latter since it's saying qemu was running and /bin/sh was the wrong architecture but I'm not 100% sure.  Will take a look now and see if I can find it quickly.

trws:
Not seeing it right off _but_ the arm64 native runners for linux on public repos are in public preview, so we may be able to switch all the arm builds to actually run _on arm_.  Will throw something together and see what we get.  If nothing else it should be substantially faster.

trws:
They're the fastest runners, by almost half... This is crazy.

The fluxrm/flux-core:alpine image does not work currently:

```
~/Repos/flux/flux-sched arm-runners:master ⇡34 *12 ?60                                                                               ✘ 2 ▼  env
❯ docker run fluxrm/flux-core:alpine --platform=linux/arm64 /bin/bash
sudo: runuser: command not found
[...]

PR #1327, opened 2025-01-15
ci: drop alpine

jameshcorbett:
Problem: the alpine-arm64 build is failing and stopping the generate-manifests step from running.

Disable the alpine build, to be brought back once the arm64 build can be fixed.

See #1326


grondo (approved):
LGTM!

PR #1328, opened 2025-01-17
add Arm runners and bring back alpine

trws:
This does a couple of things to enable new arm runners:

1. add the runner to the matrix generator so we can choose between x86 and arm
2. remove the emulator support since we should no longer need it
3. reintroduce alpine

At least on a temporary basis this changes the matrix to always run the arm builds for testing purposes.

### DIFF
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -20,9 +20,9 @@ def on_master_or_tag(matrix):
 
 DEFAULT_MULTIARCH_PLATFORMS = {
     "linux/arm64": {
-        "when": on_master_or_tag,
+        # "when": on_master_or_tag, # use this to only run on merge
+        "when": lambda _: True,
         "suffix": " - arm64",
-        "command_args": "--install-only ",
         "timeout_minutes": 90,
         "runner": "ubuntu-24.04-arm",
     },
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -202,16 +202,14 @@ matrix.add_multiarch_build(
         CHECK_RUN_SOURCE_ENV="/opt/rh/gcc-toolset-13/enable",
     ),
 )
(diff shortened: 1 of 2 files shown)


src/test/generate-matrix.py
```diff
@@ -200,14 +204,14 @@ def __str__(self):
 )
 # Disabled because the arm64 build is failing and preventing the
 # generate-manifest step from running
```
jameshcorbett, line 207: This comment should be removed as well?

trws: Yup, I'll clean that out.  A couple other things that need to happen for this to work.   The testenv container doesn't exist for arm, and the core container is in some kind of strange state, so we need to wait a bit anyway.

trws:
I think this is all set now.  I'd like to re-enable the fedora40 arm setup, but we'd need to get that support back in core first.  As a bit of a postmortem, the problems and solutions were as follows:

* builds in sched failed because of broken fluxrm/flux-core:alpine arm tag because the arm build was busted upstream
* the builds in core were busted because the fluxrm/testenv:alpine tag only had an amd64 container and not an arm64 container attached to them, probably a mistake in a buildx --push or a docker push or similar
* pushed a new tag with both amd64 and arm64 manually for testenv:alpine
* core produced a working tag on the next merge to master
* updated runners here to use arm runners to make things faster, and actually run tests

I'm working on a small PR to update a readme in core, but it might also be really good to port over the use of the arm runners, they're working really well as far as I can tell.

src/test/generate-matrix.py
jameshcorbett: comment is still here?

jameshcorbett (approved):
LGTM just one little thing

trws:
Dang, I had that gone and must have brought it back in a rebase conflict. Thanks for the catch @jameshcorbett!

PR #1332, opened 2025-02-06
resource: remove property

jameshcorbett:
Problem: there is a use-case for removing properties from vertices, but there is no functionality for doing that via rpc.

Add a remove-property RPC.

The use-case is that compute nodes attached to down rabbits will have a property placed on them, and jobs that need to use rabbits will require that the property _not_ be present on nodes allocated to the job. However, when a rabbit comes up, the property will need to be removed from the node on its rack. That last part requires this PR.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -237,6 +237,11 @@ static void set_property_request_cb (flux_t *h,
                                      const flux_msg_t *msg,
                                      void *arg);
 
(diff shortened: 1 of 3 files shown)


trws (approved):
Looks really good to me, thanks @jameshcorbett!

jameshcorbett:
Thanks! force-pushed to add some additional tests, will set MWP.

PR #1333, opened 2025-02-07
evaluators: find by name and by property

jameshcorbett:
Problem: there is no way to find resources by name and by property.

Add `names=HOSTLIST` and `property=PROP` operators to the evaluators and document them in resource-query.

Example usage:
```
resource-query> find hostlist=tuolumne[1,1001-1019] and sched-now=free
      ------tuolumne1[1:x]
      ------tuolumne1001[1:x]
      ------tuolumne1002[1:x]
      ------tuolumne1003[1:x]
      ------tuolumne1004[1:x]
      ------tuolumne1005[1:x]
      ------tuolumne1006[1:x]
      ------tuolumne1007[1:x]
      ------tuolumne1008[1:x]
      ------tuolumne1009[1:x]
      ------tuolumne1010[1:x]
      ------tuolumne1011[1:x]
      ------tuolumne1012[1:x]
      ------tuolumne1013[1:x]
      ------tuolumne1014[1:x]
      ------tuolumne1015[1:x]
      ------tuolumne1016[1:x]
      ------tuolumne1017[1:x]
      ------tuolumne1018[1:x]
      ------tuolumne1019[1:x]
      ---cluster0[1:x]
INFO: EXPRESSION="hostlist=tuolumne[1,1001-1019] and sched-now=free"
resource-query> f property=pdebug
      ------tuolumne1005[1:x]
      ------tuolumne1006[1:x]
      ------tuolumne1007[1:x]
      ------tuolumne1008[1:x]
      ---cluster0[1:x]
INFO: EXPRESSION="property=pdebug"
```

I haven't yet been able to figure out how to make the children of nodes show up in the output, but I don't think that's a blocker. That could always be added later if it's needed. However, until that's added, only the `--simple` and `jgf` output formats work, because the others expect cores or other vertices to be present in the output (at least this is my hypothesis as to why they fail).

### DIFF
--- a/resource/evaluators/expr_eval_vtx_target.cpp
+++ b/resource/evaluators/expr_eval_vtx_target.cpp
@@ -68,6 +68,8 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
             rc = 0;
             hostlist_destroy (hlist);
         }
+    } else if (p == "property") {
+        rc = (lcx.length () > 0) ? 0 : -1;
     } else
         errno = EINVAL;
 done:
@@ -144,6 +146,8 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
             result = false;
         }
         hostlist_destroy (hlist);
+    } else if (p == "property") {
+        result = (*m_g)[m_u].properties.contains (lcx);
     } else {
         rc = -1;
         errno = EINVAL;
--- a/resource/evaluators/expr_eval_vtx_target.cpp
+++ b/resource/evaluators/expr_eval_vtx_target.cpp
@@ -16,6 +16,7 @@ extern "C" {
 
 #include <algorithm>
 #include <cctype>
+#include <flux/hostlist.h>
 #include "resource/evaluators/expr_eval_vtx_target.hpp"
 
 namespace Flux {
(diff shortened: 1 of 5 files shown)


jameshcorbett:
OK @trws I think this is ready for review.

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -122,6 +136,23 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
...
+        } else {
+            result = false;
+        }
+        hostlist_destroy (hlist);
+    } else if (p == "property") {
+        for (p_it = (*m_g)[m_u].properties.begin (); p_it != (*m_g)[m_u].properties.end ();
+             p_it++) {
+            if (lcx.compare (p_it->first) == 0)
+                result = true;
+        }
```
trws, line 155: Is this equivalent?
```suggestion
        result = (*m_g)[m_u].properties.contains(lcx);
```

trws (approved):
This looks really good to me.  If the contains is equivalent, I think that would be more self-documenting, but up to you.

resource/evaluators/expr_eval_vtx_target.cpp
jameshcorbett: Yeah I think so. Good suggestion, thanks. Force-pushed to make that change and setting MWP.

PR #1335, opened 2025-02-08
Break visitation cycle in mod_dfv to correct cancellation behavior

milroy:
The `mod_dfv` recursion visits high-level graph vertices ~twice~ (see update below). For partial cancel removals of half the allocated ranks, if the removal request equals half the number of total allocated vertices for the corresponding jobid tracked by pruning filters, the double visitation will trigger the removal of the root's job tag for that jobid. (This can be generalized for $n$ cancellations each of $1/n$ of the allocation's resources. Subsequent full cancels will fail to remove the allocation because the root job tag for the jobid doesn't exist. Coloring the vertices based on visitation stops the repeated visitation, fixing the behavior in testing.

This PR also fixes a bug in the test JGF that assigned two nodes the same rank (`rank0`). This caused another traversal problem where `node1`s job tag would be removed when partially canceling `node0`. 

Props to @vsoch for the debugging assistance that helped identify this problem!

~This PR is WIP because I need to investigate why the cancel DFV didn't used to need graph coloring, and because I need to add more thorough testing.~

Update: the cycles can occur at every vertex that is not a leaf and depends on the number of children each vertex has. Beyond eliminating the incorrect behavior, this fix substantially speeds up both full and partial cancel.


milroy:
I suspected that this bug predates partial cancel, so I tested against the release prior to merging partial cancel (v36.0 eca89f0ad93760b0fe26203b2f67c3d8255ac6cf). Indeed, v36.0 exhibits multiple vertex visitation behavior, and the fix in this PR also fixes the behavior in that release. I confirmed that the fix in this PR allows all tests to pass in the testsuite for v36.0.

Based on the code, this bug has existed for quite a long time (at least 6 years) and wasn't exposed because full cancel is robust to multiple vertex visitation. This is a high-priority fix.

vsoch:
> Based on the code, this bug has existed for quite a long time (at least 6 years) and wasn't exposed because full cancel is robust to multiple vertex visitation. This is a high-priority fix.

Six years! :exploding_head: This bug is fermented and fragrant! :cheese:

vsoch:
If multiple vertices were being visited, depending on the size of the graph for the request, does that mean there will be some small improvement to latency for the operations? :thinking:

milroy:
> If multiple vertices were being visited, depending on the size of the graph for the request, does that mean there will be some small improvement to latency for the operations?

Yes, the cycle breaking will improve performance for both partial and full cancellation. I'm not sure by how much.

vsoch:
_Begone, bug fixers!_   :mage: 

image

milroy:
This PR is ready for review. Dropping [WIP].

milroy:
This bug existed at or before release v0.5.0 in 2018.

trws:
Oh wow, so we had coloring for allocate but not for mod... I'm utterly gobsmacked. Quick glance looked good, looking in detail now.

trws (approved):
Excellent find @milroy !

milroy:
> Oh wow, so we had coloring for allocate but not for mod... I'm utterly gobsmacked. 

I noticed and fixed the broken allocation coloring in 2021 (460dd3c0f942ccb5a29df3fbce5a26fff676ed74) but failed to notice the similar problem in the cancellation traversal.

milroy:
> Yes, the cycle breaking will improve performance for both partial and full cancellation. I'm not sure by how much.

After further investigation, the cycles occur everywhere above the leaves, and the multiplicity equals the number of children. That means the speedup can be exponential.

milroy:
@trws thanks for the quick review! Can you do another pass? I added more testing and updated previous test outputs to check for correct pre and postorder visit counts in `cancel`.

milroy:
Actually, hold off on the re-review for a bit. I found more missing functionality.

vsoch:
> After further investigation, the cycles occur everywhere above the leaves, and the multiplicity equals the number of children. That means the speedup can be exponential.

That would be :exploding_head: !! It would be fun if we could test that.

milroy:
> It would be fun if we could test that.

All you need to do is compare the cancel visit counts before and after this fix. Unfortunately, I can't reproduce the multiple visitation for the full cancel for some reason.

vsoch:
I think I'd need to put this counter as some kind of global variable (that sums over loops) but for this quick test I made a map counter just inside the function. Below I recreated the setup for the "before fix" variant. We can see, for example, `tiny0` being visited twice before the fix (and I think that's what we first saw after you noticed the second node tag was removed @milroy), and it stuck out as weird that it was showing up twice:

image

Actually, I think we can see that tiny0 _and_ rack0 are visited twice. After that it's entirely different! (And note to @trws this is a tiny.json derivative with 1 tiny, 1 rack, and 2 nodes. The setup:

1. Creates a job that takes up the entire graph
2. Does a partial cancel of one node (succeeds)
3. Confirms that we can schedule another job on one node (also succeeds)
4. Attempts to full cancel the original job that was partial cancelled (only one node) - fails because the jobid isn't found.

Point 4 is what first hinted at the issue, at least for me, because `full_cancel`  was coming back true when it shouldn't have. That's why the jobid wasn't found, the tag had been removed from `node0` _and_ `node1`! Then we saw that doubling, and @milroy did his magic! :magic_wand:  Here is the same function setup but with the fix:

image

To put the debugging technique we used for future reference, first this is in the vscode development container, so it's easy to clone random branches of flux-sched to different locations in opt and build them as we need. At the top we need these imports:

```cpp
#include <iostream>
#include <fstream>
```

And then in each function, I found it worked well to write a text file that is named based on the function, and then either write fresh (so old/previous content is clobbered) or append `std::ios::app` (so I can see the order of things). If you use append, you can use the same file in different functions (and see how they relate).

```cpp
    std::ofstream outfile("/tmp/mod-agfilter.txt", std::ios::app);
    outfile << "We are running mod-agfilter\n";
```

And then to make it fast, I usually do a one liner that will do the compile, install, and re run tests and show me the output. I would use append and remove the file between tests so each recursive call was written to the same file. E.g.,

```bash
cd /opf/flux-sched/build && make -j4 && sudo make install && cd /workspaces/fluxion-go && make test-v && cat /tmp/mod-agfilter.txt
```

I'll attest again - printing is still (imho) a really powerful debugging technique! At least I'm too dumb to use the other tools,  and definitely very green with c++, but I can print stuff and stare at it. :P 

It was really fun and fulfilling to watch this play out, and I learned a little bit of c++ so I really enjoyed it. Thank you for including me! And bravo @milroy :raised_hands: :partying_face:

vsoch:
I can show the output of the test that I described too - each :star:  corresponds with one of the steps I described above. Here is the test running. Notice that `total_removal` is true (it should not be) and then the error is that the jobid isn't found. Beep boop.

image

trws:
Oh god, of course it does, because of the upwalk. 😱 
I'm looking it over again now, might be worth a performance test on a big graph, just for fun...

trws (approved):
I really like the stats added on, and it all looks good to me.

jameshcorbett (approved):
Looks good to me, what do you think @trws and @milroy should we try to get this in the next TOSS release? Freeze is Tuesday I think, so it would mean another release right now.

milroy:
Ok, so one more update. I can reproduce the multiple visitations on full cancel only on tag v0.36.0 (and presumably before). Here are the counts for full cancelling a single job (`t/data/resource/jobspecs/cancel/test001.yaml`) on `t/data/resource/grugs/resv_test.graphml`

```bash
preorder count: 123
postorder count: 51
```

and after adding coloring:

```bash
preorder count: 73
postorder count: 51
```

Note the same post visit count. It will require more testing to determine if the speedup is exponential, but it is likely not. The equivalent postorder counts indicate that the non-coloring code is exiting before performing all the deallocation operations.

Reminder to future selves: we should also figure out why this difference in full cancel doesn't seem to manifest on more recent versions of flux-sched.

milroy:
Thanks for the reviews @trws and @jameshcorbett, and for the debugging in the fluxion-go @vsoch!

Yes, we should get this in the latest TOSS release if possible, which will require a 0.42.2 release.

Setting MWP.

milroy:
Noting for posterity: I confirmed the change to a bidirectional graph (PR #1240) fixed the multiple visitation in full cancel. Specifically, commit https://github.com/flux-framework/flux-sched/commits/742812db2402c42204610b46a12d1257b2d54005 (PR 1240) does not exhibit multiple visitation in full cancel, but the last commit from the previous merged PR (https://github.com/flux-framework/flux-sched/commits/2ad04a2771e7ab81f39df3392286315a9f528231) does exhibit the behavior.

vsoch:
@trws do you know why the change of data structure (or that commit) fixed the bug in full cancel?

trws:
I'd have to dig a bit to be 100% sure but most of our traversals use some version of `for (auto edge : out_edges(g, vtx))`, looping over all edges going out of a given vertex.  The code before that change used explicit out edges for both edges to children and an edge to the parent. Once we switched to `bidirectional_graph` it no longer did that, `out_edges` would list only the downward edges or edges to other subsystems and `in_edges` would need to be included to go up.  As a side-effect, un-colored traversals would only go downward on tree-like graphs after that.

vsoch:
Good call on bidirectional graph! :grapes:

PR #1338, opened 2025-02-12
yggdrasil: fix invalid template calls

trws:
problem: yggdrasil will not compile with clang++ 19+ because it adds an
invalid template parameter to a pair of calls to `insert_leaf_base`.
I'm not sure how this ever worked.

solution: remove the template parameter

I ran into this trying to test the recent issues on my Tumbleweed VM, all tests pass with the change locally and they are the only pair of calls that have this.

### DIFF
--- a/src/common/yggdrasil/rbtree.cpp
+++ b/src/common/yggdrasil/rbtree.cpp
@@ -513,12 +513,12 @@ RBTree<Node, NodeTraits, Options, Tag, Compare>::insert(
 		Node * parent = this->root;
 
 		if (parent == nullptr) {
-			this->insert_leaf_base<false>(node, parent);
+			this->insert_leaf_base(node, parent);
 		} else {
 			while (parent->NB::get_right() != nullptr) {
 				parent = parent->NB::get_right();
 			}
-			this->insert_leaf_base<false>(node, parent);
+			this->insert_leaf_base(node, parent);
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


src/common/yggdrasil/rbtree.cpp
```diff
@@ -513,12 +513,12 @@ RBTree<Node, NodeTraits, Options, Tag, Compare>::insert(
 		Node * parent = this->root;
```
wihobbs, line 513: Should lines 495 & 497 drop the `<false>`/`<true>` as well?

```diff
diff --git a/src/common/yggdrasil/rbtree.cpp b/src/common/yggdrasil/rbtree.cpp
index 640190ab..327f4d2c 100644
--- a/src/common/yggdrasil/rbtree.cpp
+++ b/src/common/yggdrasil/rbtree.cpp
@@ -492,9 +492,9 @@ RBTree<Node, NodeTraits, Options, Tag, Compare>::insert(Node & node,

        if (parent->NB::get_left() != nullptr) {
          parent = parent->NB::get_left();
-         this->insert_leaf_base<false>(node, parent);
+         this->insert_leaf_base(node, parent);
        } else {
-         this->insert_leaf_base<true>(node, parent);
+         this->insert_leaf_base(node, parent);
        }
        */
 }
 ```

wihobbs (approved):

PR #1341, opened 2025-02-14
jobspec: enforce restrictions on resource range keys

sam-maloney:
Problem: restrictions on resource range keys described in flux-framework/rfc#444 are not currently checked/enforced

This PR implements checks in libjobspec to ensure the correct combinations of keys are present and the values are self-consistent.

The initial impetus for this was a range using the exponentiation operator but having a min value of 1 would get stuck in an infinite loop, so I've added a check for that in matcher.cpp which short-circuits similarly to when an exponent or multiplier of 1 is specified. However, with the new checks in jobspec.cpp, hopefully this situation should now never be reached in the first place, and rather fail with a useful error message.

I've separated the checks for consistent values from the checks on the presence of all or none of the max/operator/operand keys into separate commits, in case you'd rather keep the behaviour of using default values.


sam-maloney:
force-pushed changes from clang-format; sorry I forgot to run that before the first round!

sam-maloney:
Pushed some new tests for the new constraint checks to address the Codecov report (and simplified the checks slightly). Also noticed that negative values (particularly for max) could pass validation because of integer overflow after being read into the unsigned variables, so I tweaked the checks on min/max values to hopefully catch that as well.

trws (approved):
This is great, I'm sorry it took me a while to get to these, delighted to have the improved verification!

PR #1342, opened 2025-02-14
WIP: add a check after load to verify graph structure

trws:
problem: We don't detect graphs with invalid upward edges in the containment tree

solution: Add a search after load for any non-tree edges of type containment, and blow up hard if we find any.

Some tests still do not pass.

@milroy, this is the WIP stuff we were discussing earlier.  The RV1 test that's blowing up is `t3027-resource-RV.t`, but it looks like it might be an out of range error, and there are some JGFs in use in the file so it's possible it isn't an actual rv1 problem. 🤞


milroy:
> The RV1 test that's blowing up is t3027-resource-RV.t, but it looks like it might be an out of range error

Test 20 (`t3027-resource-RV.t`) is failing because the command to generate the JSON results in an error:
```bash
$ flux R encode -r 0-1 -c 0-1 -H fluke[1-0] | jq " .execution.R_lite[0].rank |= \"0^2\" " > out20.json
flux-R: invalid hostlist 'fluke[1-0]'
```
and thus `out20.json` is empty. I'll bet the out of range error is related to running the BFS check on an empty graph. As a side note, I'm not sure if this test is working as designed.

PR #1343, opened 2025-02-17
libjobspec: ensure task count has exactly one key

sam-maloney:
Problem: the desired number of tasks is ambiguous if more than one key is set in tasks.count

Fail with an error message if multiple keys are found in the jobspec

This is based on the proposed amendment to RFC 14 in flux-framework/rfc#441


trws:
Thanks for this @sam-maloney, I'm good with this change, and think we've all agreed with what this checks in the RFC change so I'm going to approve.

trws (approved):

Issue #1344, opened 2025-02-18
support `shrink` key added to the RFC 28 scheduler acquisition protocol


grondo:
https://github.com/flux-framework/rfc/pull/447 adds a new `shrink` key in the [RFC 28](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_28.html) resource acquisition protocol. The key contains an idset of execution targets which have been removed from the instance and should no longer be considered for scheduling, feasibility etc.

This is one step in solving https://github.com/flux-framework/flux-core/issues/6641, in which nodes lost during a resilient batch job are still considered schedulable at some point in the future, when in reality they will never come back online, so a job can block in pending state indefinitely.

What would it take to support the `shrink` key in Fluxion @trws @milroy? If we can get this enabled by the next release that would be ideal.

grondo:
A WIP PR is now posted to flux-core (https://github.com/flux-framework/flux-core/pull/6652) which adds support to the resource acquisition protocol for the `shrink` key, along with corresponding functionality in `sched-simple ` to shrink the resource set available for scheduling.

Before moving on to testing and merging the PR before the next release, I'd like to know if it is possible to get similar functionality in Fluxion quickly. I'm willing to poke at it if someone could give me a pointer where to start looking.

The functionality we'll need is to remove or otherwise mark "permanently unavailable" execution targets (ranks) provided in the `shrink` key of the acquisition protocol response.

grondo:
This is my first stab at support for the `shrink` key - I _definitely_ got something wrong here, but could use some feedback:
However, in testing this seems to "work" (for some definitely thereof). There isn't even an error when a job running on `shrink` ranks is finally freed:

```diff
diff --git a/resource/modules/resource_match.cpp b/resource/modules/resource_match.cpp
index 9edda2d8..5e607b1e 100644
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1258,23 +1258,54 @@ done:
     return rc;
 }
 
+static int shrink_resources (std::shared_ptr<resource_ctx_t> &ctx,
+                             const char *ids)
+{
+    int rc = -1;
+    std::set<int64_t> ranks;
+
+    if (!ids) {
+        errno = EINVAL;
+        goto done;
+    }
+    if ((rc = decode_rankset (ctx, ids, ranks)) < 0)
+        goto done;
+    if ((rc = ctx->traverser->shrink (ranks))) {
[...]

PR #1345, opened 2025-02-21
WIP: support `shrink` key in resource acquisition protocol response

grondo:
This PR adds initial support for the `shrink` key in the RFC 28 resource acquisition protocol response.

As discussed in the most recent Fluxion meeting, a new resource status is added in addition to `UP` and `DOWN`, which indicates the resource is not expected to ever be available again. I've called this state `LOST`.

The `LOST` state is treated the same as `DOWN`, except in the case of pruning and `request_feasible()`, where `LOST` resources are never considered.

This seems to address the use case and the spirit of the `shrink` idset, which is sent to a scheduler when nodes in a non-configuration based instance are lost.

This is a WIP until we can get some true system tests for this feature, which will require support from flux-core (pending in flux-framework/flux-core#6652)

Feel free to review the code as-is here though, in case I've missed something in the existing changes.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1234,12 +1234,37 @@ static int mark (std::shared_ptr<resource_ctx_t> &ctx,
                                                : mark_lazy (ctx, ids, status);
 }
 
+// Subtract the idset string in b from the idset string in a.
+// Return result if non-empty in resultp (*resultp=NULL if no ids in result)
+// Returns 0 on success, -1 on failure.
+static int subtract_ids (const char *a, const char *b, char **resultp)
+{
+    int rc = -1;
+    char *result = NULL;
+    struct idset *idset = idset_decode (a);
+
+    *resultp = NULL;
+    if (!idset || idset_decode_subtract (idset, b, -1, NULL) < 0)
+        goto out;
+
+    // Success if idset is empty (result will be NULL) or non-empty
+    // idset successfully encoded:
+    if (idset_count (idset) == 0 || (*resultp = idset_encode (idset, IDSET_FLAG_RANGE)))
+        rc = 0;
+out:
+    idset_destroy (idset);
+    return rc;
+}
+
 static int update_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
                                json_t *resources,
                                const char *up,
-                               const char *down)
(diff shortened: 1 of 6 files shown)


grondo:
@trws, @milroy: Any opinions on this approach vs actually removing the subgraph using the process detailed by @milroy in #1344?

I'm happy to go whichever way you suggest.

trws:
I'm surprised to say I'm on the fence.  If we can do it with existing facilities and without having to add a new state to track, that would be nice, and it sounds like the partial cancel plus remove should do that.  That said, it seems like the lost state might be useful for long-term down nodes. Maybe worth getting some feedback from Ryan or someone on if that would be a thing admins would want?

grondo:
Let me try the subgraph removal again and propose a second PR.

Admins already have a couple ways to handle long-term down nodes, e.g. `drain`, and nodes can also be excluded from the scheduler.

If the other approach doesn't come together early next week we could perhaps go with this as a short-term solution, to be replaced by removal in the next release.

grondo:
> Let me try the subgraph removal again and propose a second PR.

I've hit a wall in development of that second PR. As discussed in #1344, getting the subgraph removal working will still require work in a few areas, and is turning out to be more complex than we'd perhaps initially thought.

Given the above, and the fact that we'd _like_ to support shrink in the next release (planned for 1 week from today), I'm going to propose we look again at merging this version of the PR, with the idea that we'd remove support for the LOST status once subgraph removal (i.e. true shrink) is actually working. Since the LOST status is internal only at this point (besides one log message), I don't think eventual removal would be a problem.

With that in mind, I've pushed a test (actually the same test used in flux-core) that ensures the shrink key has the intended effect: that ranks included in the `shrink` key are not used when determining satisfiability of a job request.

trws:
I'd be alright with this, you ok @milroy?

milroy:
@trws and @grondo, I can review this PR tonight, but I also think I can get a PR out to support `shrink` via subgraph removal in about a day: https://github.com/flux-framework/flux-sched/issues/1344#issuecomment-2683743101. Let me know which (or both) I should do.

grondo:
I spoke up in the issue, but I'd say go ahead with your PR. We can keep this one around just in case, and you can review it only if needed.

PR #1346, opened 2025-02-26
comply with RFC 20 when parsing Rv1 `starttime` and `expiration`

grondo:
I noticed some errors in handling of Rv1 in Fluxion while testing #1345. Most notably, Fluxion does not support an expiration of `0.` as specified in [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#expiration), either in the initial Rv1 object used to initialize the resource graph or as a result of `flux update JOBID duration=0`. Additionally, RFC 20 specifies both `starttime` and `expiration` are optional, but Fluxion requires both keys.

This PR makes the necessary simple changes to bring Fluxion into compliance. This came up when testing the launch of Fluxion scheduled subinstances under sched-simple, which uses an expiration of `0.` to indicate unlimited (sched-fluxion-resource aborted at this point) It also may also come up if the duration of a job is updated to `0.`, which should indicate "unlimited", but Fluxion currently just sets the graph end time to `0.` and subsequent job submissions all fail with a generic "match error"

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1293,12 +1293,18 @@ static void update_resource (flux_future_t *f, void *arg)
         flux_log_error (ctx->h, "%s: update_resource_db", __FUNCTION__);
         goto done;
     }
-    if (expiration >= 0.) {
+    if (expiration > 0.) {
         /*  Update graph duration:
          */
         ctx->db->metadata.graph_duration.graph_end =
             std::chrono::system_clock::from_time_t ((time_t)expiration);
         flux_log (ctx->h, LOG_INFO, "resource expiration updated to %.2f", expiration);
+    } else if (expiration == 0.) {
+        /*  An expiration of 0 should be treated as "no expiration":
+         */
+        ctx->db->metadata.graph_duration.graph_end =
+            std::chrono::system_clock::from_time_t (detail::SYSTEM_MAX_DURATION);
+        flux_log (ctx->h, LOG_INFO, "resource expiration updated to 0. (unlimited)");
     }
     for (auto &kv : ctx->notify_msgs) {
         if ((rc += flux_respond (ctx->h, kv.second->get_msg (), NULL)) < 0) {
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -627,6 +627,10 @@ static int unpack_resources (json_t *resobj,
     json_t *jgf = NULL;
     size_t index;
     json_t *val;
+    double max_clock = static_cast<double> (
+        std::chrono::duration_cast<std::chrono::seconds> (
+            std::chrono::time_point<std::chrono::system_clock>::max ().time_since_epoch ())
+            .count ());
 
     if (!(ids = idset_create (0, IDSET_FLAG_AUTOGROW)))
         return -1;
@@ -646,6 +650,9 @@ static int unpack_resources (json_t *resobj,
                          &jgf)
             < 0)
             goto inval;
+        // RFC 20 states that end == 0.0 means "unset". Set to max_clock if so:
+        if (end == 0.)
+            end = max_clock;
         // flux-core validates these numbers, but checking here
         // in case Fluxion is plugged into another resource manager
         if (version != 1)
@@ -1539,6 +1546,7 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
     int saved_errno;
     double tstart = 0;
     double expiration = 0;
+    double max = static_cast<double> (std::numeric_limits<int64_t>::max ());
     json_t *o = NULL;
     json_t *graph = NULL;
     json_error_t error;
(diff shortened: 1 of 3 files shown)


trws:
This looks good, but it makes me wonder if we should be using a parsing or validation lib from core for pulling these in.  Some of it is our interpretation of the values, but things like the actual parsing seems like it could have been shared, or is there not a reusable piece for that?

trws:
That seems like something we should fix. It may be partly the stressful week, but  having to reinvent and then fix/maintain things that `sched-simple` gets to reuse because we're forced to be out of tree is frustrating.  I seriously appreciate you hopping in and fixing this, I just wish you didn't have to and it feels like re-using the reading code so optional values are optional and so-forth as tested upstream seems like it would help.

Regardless this is a positive fix, so I'm happy to take it once the tests shake out.

grondo:
> That seems like something we should fix. It may be partly the stressful week, but having to reinvent and then fix/maintain things that sched-simple gets to reuse because we're forced to be out of tree is frustrating. I seriously appreciate you hopping in and fixing this, I just wish you didn't have to and it feels like re-using the reading code so optional values are optional and so-forth as tested upstream seems like it would help.

That is a valid complaint. However, in this case I'm not sure it would have helped much because sched-simple treats an expiration of 0. as unlimited, whereas the Fluxion graph needs to have a maximum value set. Unfortunately, librlist _is_ the sched-simple implementation. I guess you're correct that it would have solved the "optional" issue, but librlist happens to use `-1.` for that -- so not a great interface.

We should indeed think about how to fix this going forward, but I always figured that would be part of an Rv2 implementation, where the resource format and its internal representation and operations would be complex enough that it might even warrant its own project that is pulled in by core and sched etc.

grondo:
This is so strange. In the fedora40 container I'm getting the following error when fluxion modules are loaded, though it works from the command line: (Still looking into this but it will have to wait til later)

```
expecting success: 
	flux module remove sched-simple &&
	load_resource &&
	load_qmanager &&
	test_debug "flux dmesg -H | grep version"

flux-module: load sched-fluxion-resource: libboost_graph.so.1.74.0: cannot open shared object file: No such file or directory
not ok 2 - load fluxion modules in parent instance
#	
#		flux module remove sched-simple &&
#		load_resource &&
#		load_qmanager &&
#		test_debug "flux dmesg -H | grep version"
#	

```

grondo:
Ah, there were some old fluxion modules in `.libs` in my builddir :hand_over_mouth: and `t/sharness.d/sched-sharness.sh` still prepends the `.libs` path to `FLUX_MODULE_PATH_PREPEND`. I'm assuming that's not needed anymore?

Anyway, the tests pass locally, so I'm guessing this is a race condition. Let me add some synchronization to the test and retry.

trws:
Yup, the .libs thing is vestigial for fluxion itself, we still need it for loading things from a built but not installed core I guess but probably only on that part of the path.

grondo:
Ok, I fixed that test issue (it was actually a missing `&&`) Also added a couple commits - one that removes `HAVE_JQ` prereq from a few tests, since `jq` is now assumed to be available (and checked in `flux-sharness.sh`), and also the fix to remove `.libs` from `FLUX_MODULE_PATH_PREPEND`

trws:
Sounds great, I'll take another quick pass now.

resource/modules/resource_match.cpp
```diff
@@ -1568,8 +1575,11 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
         flux_log (ctx->h, LOG_ERR, "%s: json_unpack: %s", __FUNCTION__, error.text);
         goto freemem_out;
     }
-    if (version != 1 || tstart < 0 || expiration < tstart
-        || expiration > static_cast<double> (std::numeric_limits<int64_t>::max ())) {
+    // RFC 20 states: "An expiration of 0. SHALL be interpreted as "unset",
+    // therefore if expiration is 0 set it to int64_t::max:
+    if (expiration == 0.)
+        expiration = max;
```
trws, line 1581: Codecov picked out that we never actually poke this in a test, but otherwise this looks great so I'm approving.

trws (approved):

resource/modules/resource_match.cpp
grondo: Oh that's a good catch. I thought the change to one of the input files would have covered this check, but let me double check on that one.

grondo: Thanks for catching that, I had actually missed the other decode of R in `unpack_resources()`. Fixed that and pushed again. You might want to do another quick review, I'm not sure I got the right "maximum" value for expiration the first time (`int64_t` max seemed to be too large in `unpack_resources()`)

grondo:
I'll go ahead and set MWP here a bit later this morning unless I hear a NACK.

jameshcorbett (approved):
This LGTM as well, I'd say set MWP!

PR #1347, opened 2025-02-26
resource: fix leak in `status_request_cb()`

grondo:
Problem: A lead in the sched-fluxion-resource module is reported when running t2317-resource-shrink.t with FLUX_TEST_VAlGRIND=t set.

The leak is from the line

    ctx->m_r_alloc = json_deep_copy (R_alloc);

Other cached R representations in the module ctx object are saved with `json::value::take()`. Use this for `ctx->m_r_alloc` as well, which seems to fix the leak.

I have no idea if this is the right fix :shrug:


trws:
Yup, that's the right fix.  The `take` function constructs the value object without incrementing the reference count, so that's the right way to make one that needs to be decrefed at the end but should not be increfed first.  Great catch @grondo!

trws (approved):

grondo:
Thanks! I'll set MWP.

PR #1348, opened 2025-02-26
sched-sharness: stop looking in modules/.libs

trws:
problem: we're still looking for libraries in unused `.libs` subdirectories that used to be used by the autotools build system. This can find old copies of the modules in some cases.

solution: stop looking in those directories

@grondo found this in #1346, should be a simple fix since this is the only place we reference those subdirs in fluxion.


trws:
Bah, just noticed you added a commit for this to #1346, I'll close this one.

trws:
Looks like it, I'll reopen then.

jameshcorbett (approved):
👍 but maybe @grondo should comment since he found the issue.

grondo:
> Looks like it, I'll reopen then.

Oops! Sorry about that!

PR #1349, opened 2025-02-27
resource: make property RPCs return error message

jameshcorbett:
Problem: the `set_property`, `get_property`, and `remove_property` RPCs log an error message instead of returning one to the sender. That is a bad pattern.

Don't log the message, instead return it in the RPC response.

See also https://github.com/flux-framework/flux-coral2/issues/294.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -2497,7 +2497,7 @@ static void set_property_request_cb (flux_t *h,
                                      void *arg)
 {
     const char *rp = NULL, *kv = NULL;
-    std::string resource_path = "", keyval = "";
+    std::string resource_path = "", keyval = "", errmsg = "";
     std::string property_key = "", property_value = "";
     size_t pos;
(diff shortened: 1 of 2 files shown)


zekemorton (approved):
I think these changes make sense, it might be worth considering updating all of the functions to return any potential error handling though so that the behavior is consistent.

jameshcorbett:
Thanks @zekemorton ! I opened an issue to standardize the behavior. I think it's not very high priority though because I don't know if the other RPCs are used anywhere, let alone in production.

I also pushed a commit improving the tests, that checks that the expected error message is returned. Setting MWP.

jameshcorbett:
@zekemorton  since codecov is failing, can you hit the merge button?

jameshcorbett:
Ok, sorry for all the force-pushes, I was trying to get the coverage up for codecov, I think it's finally working.

jameshcorbett:
@mergifyio refresh

PR #1352, opened 2025-02-28
Partial cancel and shrink resource graph by ranks

milroy:
Issue #1344 describes the need for supporting the `shrink` key as part of the updated RFC 28. To enable the desired functionality, Fluxion must delete a subgraph defined by an idset of ranks and ensure the allocations, reservations, and filter states are correct.

This PR moves the existing `remove_subgraph()` functionality from the readers to the traverser and enables subgraph removal by idset. The PR also adds the capability to partially cancel all jobids residing on the ranks defined in the idset. This capability is needed to ensure the pruning/aggregate filters remain consistent in terms of their current allocation states.


milroy:
A couple of notes: 

- After more thorough testing, I discovered that partial cancel by idset prior to subgraph deletion is necessary to ensure correct scheduling behavior.
- `t2317-resource-shrink.t` fails on most of the test images. The behavior depends on whether `ALL:core` or `ALL:core,ALL:node` is specified when Fluxion is loaded. The failures started today when I pulled the latest bookworm image of flux-core. @grondo do you have insight into why that's happening?

milroy:
> The behavior depends on whether ALL:core or ALL:core,ALL:node is specified when Fluxion is loaded.

To be specific: if `ALL:core` is used (current default) the test hangs until the 30 minute timeout is reached. With `ALL:core,ALL:node` specified, the test fails 1 of the 18 checks.

grondo:
> To be specific: if ALL:core is used (current default) the test hangs until the 30 minute timeout is reached. With ALL:core,ALL:node specified, the test fails 1 of the 18 checks.

I see the problem with the 1 failing test. The test was written for the `LOST` state and is looking for a log message specific to that. I can fix that (and also a problem with `skip_all` when the `monitor-force-down` RPC fails) and will post a diff here.

If the test is hanging with `ALL:core`, then this means that Fluxion is not behaving as expected and that a job that should be unsatisfiable is not rejected and instead blocks in the queue forever (FWIW, this is the issue we're trying to solve with shrink, the scheduler should reject a job if there aren't enough resources instead of just hanging until the instance time limit)

I'll add timeouts where appropriate and add that to the fix for the test. Sorry about that!

grondo:
This diff fixes the skip_all behavior of the test and removes the check for a log message from fluxion about the LOST node(s).
```diff
diff --git a/t/t2317-resource-shrink.t b/t/t2317-resource-shrink.t
index 55f9a339..34d11b64 100755
--- a/t/t2317-resource-shrink.t
+++ b/t/t2317-resource-shrink.t
@@ -19,8 +19,8 @@ force_down () {
 
 # If force-down is not supported then this version of fux-core does not
 # support shrink, so skip all tests:
-if ! force_down ""; then
-       skip_all 'resource.monitor-force-down failed, skipping all tests'
+if ! force_down "" 2>/dev/null ; then
+       skip_all='resource.monitor-force-down failed, skipping all tests'
        test_done
 fi
 
@@ -94,9 +94,6 @@ test_expect_success 'a 3 node job is now unsatisfiable' '
 test_expect_success 'but a 2 node job runs' '
        flux run -N2 hostname
 '
-test_expect_success 'fluxion noted LOST nodes' '
-       flux dmesg -H | grep fluxion | grep LOST
-'
 test_expect_success 'unload fluxion' '
        flux module remove sched-fluxion-qmanager &&
        flux module remove sched-fluxion-resource &&
```

I have to do a bit more work to add timeouts. I'll get to that in the morning. However, just FYI I ran my original test script and it doesn't seem to work as expected on this branch.

BTW, I wouldn't expect this branch to pass these tests, since it appears that handling of the "shrink" idset in the resource acquisition response hasn't been added here yet. (this also makes me confused why it works for `ALL:core,ALL:node` :confused:)

I could be missing something simple though.

grondo:
@milroy here's the [commit](https://github.com/flux-framework/flux-sched/pull/1345/commits/9304ad91ec445e5654a79d5367bfe08b7ea1e11f) in #1345 that handles the `shrink` key in the resource acquisition protocol response. Note that `shrink` ranks will also be in the `down` key, so that commit also does some work to create a down-but-not-lost idset (which in the current case will always be empty), since it didn't make sense to mark nodes down then lost. Perhaps the same approach works here to avoid marking ranks down only to remove them.

milroy:
> @milroy here's the [commit](https://github.com/flux-framework/flux-sched/pull/1345/commits/9304ad91ec445e5654a79d5367bfe08b7ea1e11f) in https://github.com/flux-framework/flux-sched/pull/1345 that handles the shrink key in the resource acquisition protocol response.

I was confused for a while about why my other test branch was passing CI. I lost those changes in this branch in a rebase and force push.

milroy:
I updated the PR with the commit you linked in PR #1345 and adapted it to my implementation. I also updated `t2317-resource-shrink.t` based on your diff, @grondo, and all tests are passing.

I don't understand how CI was passing in my other branch with my implementation of `shrink` key handling in the resource acquisition protocol, but maybe I was just confused.

milroy:
> If the test is hanging with ALL:core, then this means that Fluxion is not behaving as expected and that a job that should be unsatisfiable is not rejected and instead blocks in the queue forever

I think what I observed was what you reported here: https://github.com/flux-framework/flux-sched/issues/1344#issuecomment-2673331942 that prompted me to start working on making `ALL:core,ALL:node` the default. I dropped the filter change to `t2317-resource-shrink.t` (so it's using the current default of `ALL:core` again) and it passes now. I'm not sure why the behavior changed, perhaps the skip_all behavior?

grondo (approved):
This LGTM (though I did not review the earlier commits here) and a set of tests I ran by hand all seem to be working! Thanks @milroy!

milroy:
> Yes, if you ran the tests before https://github.com/flux-framework/flux-core/pull/6652 then there was no support for the shrink key in the flux-core you were testing against so the test may have just been skipped. Except for my typo in skip_all which should have made the test fail 🤷

I think that explains it. The container I was running the tests in was more than 4 days old.

> BTW, I'm a little worried about the difference in behavior between ALL:core,ALL:node and ALL:core -- we may want to adapt the test to run once with each, if that's possible.

The current test succeeds with both of the filter settings. I can't reproduce the blocking behavior with `ALL:core`.

resource/modules/resource_match.cpp
```diff
@@ -1234,6 +1234,43 @@ static int mark (std::shared_ptr<resource_ctx_t> &ctx,
...
+        flux_log (ctx->h, LOG_ERR, "decode_rankset (\"%s\") failed", ids);
+        goto done;
+    }
+    if (ctx->traverser->remove (ranks)) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "partial_cancel_by_ranks (\"%s\") failed: %s",
+                  ids,
+                  ctx->traverser->err_message ().c_str ());
+        goto done;
```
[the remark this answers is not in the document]
milroy: I should probably remove the underscores, but the intention of the message is correct.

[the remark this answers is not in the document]
milroy: But the error check above is wrong!

resource/traversers/dfu_impl_update.cpp
```diff
@@ -675,6 +675,117 @@ int dfu_impl_t::cancel_vertex (vtx_t vtx, modify_data_t &mod_data, int64_t jobid
...
+            return -1;
+
+        if (remove_metadata_outedges (parent_vtx, root) != 0)
+            return -1;
+
+        for (auto &vtx : vertices) {
+            // clear vertex edges but don't delete vertex
+            boost::clear_vertex (vtx, *m_graph);
+            remove_graph_metadata (vtx);
+        }
```
zekemorton, line 831: Why is this for loop nested in the other one? It looks like it might not do anything different in each loop

resource/traversers/dfu_impl_update.cpp
```diff
@@ -675,6 +675,117 @@ int dfu_impl_t::cancel_vertex (vtx_t vtx, modify_data_t &mod_data, int64_t jobid
...
+    boost::tie (ei, ei_end) = boost::out_edges (vtx, *m_graph);
+
+    for (; ei != ei_end; ++ei) {
+        next_vtx = boost::target (*ei, *m_graph);
+
+        for (const auto &paths_it : (*m_graph)[next_vtx].paths) {
+            // check that we don't recurse on parent edges
+            if (paths_it.second.find ((*m_graph)[vtx].name) != std::string::npos
+                && paths_it.second.find ((*m_graph)[vtx].name)
+                       < paths_it.second.find ((*m_graph)[next_vtx].name)) {
```
zekemorton, line 739: Is this check still needed after the change to in and out edges? It looks like on line 710 a check for the containment subsystem is sufficient

zekemorton (approved):
LGTM just a couple of notes but nothing major!

resource/traversers/dfu_impl_update.cpp
milroy: It shouldn't be nested. Good catch.

resource/traversers/dfu_impl_update.cpp
milroy: I think we still need a way to ensure vertices aren't visited multiple times if a JGF is invalid and specifies both edges. I updated the recursion to use coloring rather than map searches for efficiency.

milroy:
@trws can you give this a look when you get a chance? I think it's near merge-able but it would be good to get your feedback.

resource/traversers/dfu_impl_update.cpp
trws: Best to leave it here until we get something in to verify graphs as we load/update them.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -675,6 +675,114 @@ int dfu_impl_t::cancel_vertex (vtx_t vtx, modify_data_t &mod_data, int64_t jobid
...
+    }
+    for (auto it = m_graph_db->metadata.by_rank[(*m_graph)[v].rank].begin ();
+         it != m_graph_db->metadata.by_rank[(*m_graph)[v].rank].end ();
+         ++it) {
+        if (*it == v) {
+            m_graph_db->metadata.by_rank[(*m_graph)[v].rank].erase (it);
+            break;
+        }
+    }
+}
```
trws, line 812: Since `by_type`, `by_name` and `by_rank` are all maps, it would be nice to not have to do the log(n) search twice per iteration when they're all loop invariant (compiler can't always hoist these because of potential side-effects).  One of those is evaluating `.end()`, swapping to a range-based for would take care of that, or storing it ahead of time.  The other one would be swap `m_graph_db->metadata.by_type[(*m_graph)[v].type]` for declaring `auto &target_by_type = m_graph_db->metadata.by_type[(*m_graph)[v].type]` or similar then using the reference.

trws (approved):
Wow, I like how this turned out.  It's nice to see that all move out of the readers too.  Up to you on the stylistic comment, it's not urgent unless we find it's a performance problem.

resource/traversers/dfu_impl_update.cpp
milroy: Good idea. I implemented your second suggestion and pushed the changes.

milroy:
Thanks for the reviews everyone! I'll set MWP once I verify all tests complete and no further changes are needed.

PR #1353, opened 2025-03-02
Update the pruning filter default to `ALL:core,ALL:node` and fix `update_node_stats()`

milroy:
The discussion in Issue #1344 revealed undesirable and counterintuitive behavior of satisfiability and feasibility checking due to the `ALL:core` default pruning filter value.

This PR updates the default value and fixes two bugs: incorrect node `UP/DOWN` count updates in `mark()`, and the lack of a `zone_rt` which causes `t3013-resource-unsat.t` to fail. The bug in `mark()` and incorrect `update_node_stats()` behavior may be causing problems in production.


milroy:
This PR is based on PR #1352, and depends on those changes. It should only be merged after #1352.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -1052,16 +1052,23 @@ int dfu_impl_t::mark (const std::string &root_path, resource_pool_t::status_t st
...
+    for (auto &v : vit_root->second) {
         (*m_graph)[v].status = status;
-    m_graph_db->metadata.update_node_stats (vit_root->second.size (), status);
+        vtx_set.insert (v);
+        get_subgraph_vertices (v, vtx_set);
+    }
+    for (const auto &v : vtx_set) {
+        if ((*m_graph)[v].type == node_rt)
+            m_graph_db->metadata.update_node_stats ((*m_graph)[v].size, status);
+    }
```
trws, line 1071: Oof.  Really good catch, is this how we were getting multiple applications to the node level?

resource/schema/data_std.cpp
```diff
@@ -15,6 +15,7 @@ namespace resource_model {
 
 subsystem_t containment_sub{"containment"};
 
+resource_type_t zone_rt{"zone"};
```
trws, line 18: I see the reason for this in the commit message, and I assume that means that the resource graph doesn't contain any "zone" resources.  I'm not sure why that wouldn't have come up before, but that should error out anyway shouldn't it?

trws (commented):
Curious about adding "zone" otherwise this looks really, really good.

resource/traversers/dfu_impl_update.cpp
milroy: Yes, it explains and should fix that problem. Is it reported in a flux-sched issue?

resource/schema/data_std.cpp
milroy: I'm not sure why it didn't come up before. The manifestation without adding `zone_rt` is that the match allocate hits an error rather than finding the request unsatisfiable.

I don't know if we want to support `zone_rt` just to allow this test to behave as intended. I contemplated not adding `zone_rt` and just including the expected output without the jobspecs that are malformed.

resource/traversers/dfu_impl_update.cpp
trws: Only as part of the adams discussion I think.  Possibly also in the context of the graph coloring PR for update?

resource/schema/data_std.cpp
trws: If you have the output, it would be good to know if it got the error for invalid jobspec or for something else.  If it failed because that resource doesn't exist in the graph, that's a very different problem from it existing and failing because this wasn't declared, the second one I would want to dig into fixing right away because that could cause much deeper issues.

trws: Ok, that's the one I would have expected.  The resource graph doesn't contain a "zone" vertex right?

milroy: Correct. There's no `zone` in the resource graph.

trws: Then I would prefer we leave zone out of the pre-populated list, and either make that error the correct output for a specific subtest or remove that subtest.  It's the correct behavior now to get an error when you ask for a nonsensical resource.

resource/traversers/dfu_impl_update.cpp
milroy: This change will only affect `mark` and the node stats when ranks are manually marked `UP` or `DOWN` via the target path. I think `mark` is only used with a target path manually at least. I don't think it's related to graph coloring, but I'm not sure I understand your question.

trws: True, but I think that's what the flux-coral2 project uses or used to mark nodes for drained rabbits.  May not be related, but that's a failure path I had never considered before, just made me think that would be a sneaky way to end up with some of the issues we saw.

resource/schema/data_std.cpp
milroy: Ok, I removed that commit and updated the test to check for the error.

milroy:
I updated the PR based on on your feedback @trws. It's ready for a final review.

resource/traversers/dfu_impl_update.cpp
jameshcorbett: flux-coral2 does send RPCs to mark vertices by path, but it only does it for the `ssd` vertices that represent the rabbit. It has never had functionality to mark nodes `UP` or `DOWN`.

jameshcorbett: Oh and the `ssd` vertices are leaf vertices in this case.

milroy: > it only does it for the ssd vertices that represent the rabbit. It has never had functionality to mark nodes UP or DOWN.

The problem is the current code subtracts (adds) one node from the `nodes_up` count for _every_ `mark ()` call that takes a path argument. So executing `mark (/cluster0/rack0/rabbit0`, DOWN)` subtracts one from the `nodes_up` count.

jameshcorbett (approved):
This looks good to me!

resource/traversers/dfu_impl_update.cpp
milroy: > Would the scheduler just silently refuse to match resources?

Depending on the filter settings and the value of `nodes_up` it would return unsatisfiable or just not match.

> I'm guessing it wouldn't show up in the output of FLUX_RESOURCE_LIST_RPC=sched.resource-status flux resource list?

I think it wouldn't show up in the output of `FLUX_RESOURCE_LIST_RPC=sched.resource-status` because the parent (of type node) vertex status would still be `UP/DOWN`.

milroy:
Thanks for the review and feedback @jameshcorbett and @trws! Setting MWP.

resource/traversers/dfu_impl_update.cpp
trws: I think it would show up as the aggregate at the rack vertex being artificially lower. It would show up with the new stats @milroy added in the last release, but not in any normal status output because it isn't actually part of the status, it's meant to just be a pre-computed cache of those states.

milroy: > I think it would show up as the aggregate at the rack vertex being artificially lower.

I don't think the `nodes_up` value has an influence on the planner counts. I think the only way you could figure this out would be to create a stats output entry for `get_graph_db ()->metadata.nodes_up`.

trws: Oh right, quite so. I was thinking of it feeding into the other counts, but this one is separate.  That's a bit odd now that I think about it, and probably means my feasibility code is why things wouldn't run when this happened.

milroy: Yeah, I'm having trouble tracking the code paths through `request_feasible`. We need to clarify the interactions between the planner state, `nodes_up`, `target_nodes`, and `feasible_nodes`, and make the `nodes_up` and `feasible_nodes` more visible for diagnostics.

PR #1359, opened 2025-03-12
Ensure `set-status` and `mark()` are idempotent

milroy:
Setting a vertex status (`UP`|`DOWN`) is not idempotent. Each time `mark()` is called via `set-status` on a path or set of ranks, the `nodes_up` value will be incremented or decremented for each rank in the rank set or node in the subgraph rooted at the specified path. This behavior creates incorrect matching and satisfiability check results.

This PR ensures that `mark()` and thus `set-status` are idempotent.


jameshcorbett (approved):
Nice, wish I'd caught this in #1353

milroy:
Likewise, I wish I'd thought of it when I was creating that PR.

I'm trying to find a way to test `mark (std::set<int64_t> &ranks, resource_pool_t::status_t status)`, but `flux resource drain <rank>` checks the down ranks and is idempotent. If I can't think of another way to test that function behavior I'll set MWP.

trws (approved):
Thanks @milroy!

milroy:
Thanks for the reviews @jameshcorbett and @trws ! I can't figure out how to coerce non-idempotent behavior (prior to the fix here) via `flux resource drain <rank>` so the current testing should be sufficient.

Setting MWP.

PR #1361, opened 2025-04-12
resource-query batch run

JaeseungYeom:
This PR enables batch processing of resource queries.

- When user provide `-b queries.txt` for `--batch=queries.txt`, it runs all the queries in queries.txt in a batch.
- If a line starts with a number followed by a query, it repeats a query as many as the given number.
- A line starting with `#` is treated as a comment in the query batch file.
- When all the queries in a batch are done, it goes into the usual interactive query mode.
- If stopping is desires, add `quit` as the last query in the batch.
- The query command `statm` prints the current content of `/proc/self/statm`

### DIFF
--- a/resource/utilities/command.cpp
+++ b/resource/utilities/command.cpp
@@ -16,6 +16,8 @@ extern "C" {
 }
 
 #include <sys/time.h>
+#include <unistd.h>
+#include <cstdio>
 #include "command.hpp"
 #include <readers/resource_reader_factory.hpp>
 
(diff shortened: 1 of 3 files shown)


trws:
I haven't delved into this too deeply, but I'm curious, why this over either using an expect script or piping input into resource_query? Did you add some extra functionality in the batch script, or is it more like the init-script option of gdb maybe? Can you specify multiple of them?

JaeseungYeom:
> I haven't delved into this too deeply, but I'm curious, why this over either using an expect script or piping input into resource_query? Did you add some extra functionality in the batch script, or is it more like the init-script option of gdb maybe? Can you specify multiple of them?

I have not thought about piping. piping will achieve more or less what I wanted. There are a couple of extras but the still can be handled with stdin I think. Those are 1 number of repetitions, commenting, and statm output.

Issue #1180, opened 2024-04-18
DOWN vertices not reconsidered after coming back UP


jameshcorbett:
On rzadams I marked a rabbit vertex as down, and submitted a job that required a compute node on the same rack as that rabbit. The job, as expected, was stuck in SCHED. I then marked the rabbit vertex as UP, but the job remained stuck in SCHED. A new job went through fine.

It seems to me that the original job should have had its resource request reconsidered at some point?

milroy:
Can you provide some of the job details and resource requests? Did the job have constraints? Also, which resource reader is Fluxion using on rzadams?

milroy:
> Also, which resource reader is Fluxion using on rzadams?

After a bit more thought, it has to be JGF given the use of rabbits.

milroy:
A better question is what is the configured match policy?

trws:
Having just gone through this, my best bet would be that somehow the resource never got marked as UP in resource.  We definitely reconsider jobs when resources change state, in fact we reconsider all jobs when a resource is set DOWN even, so it's probably a failure to propagate that state or an issue with the matching.

trws:
Is this still reproducible @jameshcorbett? We've made the reconsideration much more robust in the last month or so.

jameshcorbett:
@behlendorf has been seeing this issue repeatedly on Hetchy. He submits a rabbit job, i.e. one that has `ssd` entries in the jobspec. However, all `ssd` vertices are marked down, so his job is stuck in `SCHED`. After `ssd` vertices are marked `up` again, the job remains in `SCHED`; however, new rabbit jobs run through fine.


The match policy on Hetchy is `lonodex`:
```
[corbett8@hetchy1001:~]$ cat /etc/flux/system/conf.d/sched-fluxion-resource.toml 
[sched-fluxion-resource]
# node exclusive starting from low node ids
match-policy = "lonodex"
# prune resource graph based on these criteria
prune-filters = "ALL:core"

# use rv1 to match graph generated from JGF
match-format = "rv1"
```

PR #1362, opened 2025-04-17
resource: send notify response on `mark` RPC

jameshcorbett:
Problem: the receipt of a `mark` RPC does not trigger qmanager to reconsider jobs.

If a vertex is marked as UP, send a response to the `sched-fluxion-resource.notify` RPC, so qmanager can reconsider jobs.

Draft until tests are added.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -3088,6 +3088,23 @@ error:
 }
 
 /*
+ * Send a NULL resource.notify message to get qmanager to reconsider jobs
+ */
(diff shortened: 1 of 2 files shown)


trws:
This looks good, but would you mind pulling that loop out into a "notify_of_resource_change" function or similar? I have a feeling we may want to use it other places, like property updates (think rabbit jobs when everything matching is marked badrabbit).

trws:
Ping, it would be good to get this merged to take care of #1180.

trws (approved):

PR #1376, opened 2025-05-19
Rq2 - info, find, and cancel

zekemorton:
This PR implements the functions info, cancel, and find for the resource query cli, rq2. These functions are already implemented in the reapi cli so functionality just needed to be added to rq2. 

In addition to adding these features, a few bugs were identified in reapi cli and resolved
 - fix return value in cancel by implementing the noent_ok parameter 
 - unset errno after finding a match in match allocate


trws (dismissed):
I went over, built and tested this. It seems like a good change to me.  @milroy any concerns?

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -111,6 +111,7 @@ int reapi_cli_t::match_allocate (void *h,
     // Check for an unsuccessful match
```
milroy, line 111: Good catch!

Nits in commit message: "does relies on," "it is unclear is match."

milroy (approved):
There are a couple of typos in one commit message, but this looks great to me.

milroy (approved):
Thanks for fixing the typos!

PR #1378, opened 2025-06-27
Fix agfilter output in find

milroy:
The output of `find` is currently incorrect when inspecting the agfilters' states. The `at` time is hard-coded as `0`, which works for `resource-query` tests. When Fluxion is running, the `at`time will take the value of the current epoch time in seconds.

This PR computes the `at` time for the current epoch time in seconds and introduces a check for an `at` time of `0` to coincide with testsuite expected outputs.


resource/traversers/dfu_impl.cpp
```diff
@@ -838,10 +839,20 @@ int dfu_impl_t::dom_find_dfv (std::shared_ptr<match_writers_t> &w,
...
+            int64_t now_epoch =
+                std::chrono::duration_cast<std::chrono::seconds> (now.time_since_epoch ()).count ();
             for (size_t i = 0; i < planner_multi_resources_len (filter_plan); ++i) {
                 int64_t total_resources = planner_multi_resource_total_at (filter_plan, i);
-                int64_t diff =
-                    total_resources - planner_multi_avail_resources_at (filter_plan, 0, i);
+                int64_t usage_now = planner_multi_avail_resources_at (filter_plan, now_epoch, i);
+                // Need to check for the case where now = 0, which is used in the testsuite.
+                // At will never be 0 in production
+                if (usage_now == total_resources) {
```
trws: I get the workaround to do now=0 for tests, but I'm not sure I understand this condition for getting in here.  It looks like `usage_now` being equal to `total_resources`  could happen in production if everything is in use, then end up calculating based on zero? Maybe I'm missing how this follows.  If it's for testing, do we want to key off of something set during testing for this?

milroy: I'm not convinced the choice of conditions is the best, either. Here's my reasoning, though:

1.`now_epoch` will be a large `int64_t` value (e.g., `1751504912`), and `usage_now` should equal `total_resources` (i.e., all resources available) for `resource-query` tests that initialize the planners to an initial time of 0. (I really should change the `usage_now` name to `free_now`.)
2. If the current usage is 0, check again at time 0 (which should never have occupied resources for a running instance of Fluxion).

I'm open to suggestion for better approaches. I initially decided against putting some additional command line flag in to avoid the extra plumbing, but that would be the most robust approach.

trws: Ok, so if all resources are available, check again at time 0 which should also be all available barring utter insanity. That does seem like it should work.  I was considering if we should key off of something like running under sharness, or even adjusting the way the tests behave to not need the code path, but this does seem reasonable as a workaround for now.

resource/traversers/dfu_impl.cpp
```diff
@@ -838,10 +839,20 @@ int dfu_impl_t::dom_find_dfv (std::shared_ptr<match_writers_t> &w,
...
             goto done;
         if (jobid == 0) {  // jobid not specified; get totals
+            auto now = std::chrono::system_clock::now ();
+            int64_t now_epoch =
+                std::chrono::duration_cast<std::chrono::seconds> (now.time_since_epoch ()).count ();
             for (size_t i = 0; i < planner_multi_resources_len (filter_plan); ++i) {
                 int64_t total_resources = planner_multi_resource_total_at (filter_plan, i);
-                int64_t diff =
-                    total_resources - planner_multi_avail_resources_at (filter_plan, 0, i);
+                int64_t usage_now = planner_multi_avail_resources_at (filter_plan, now_epoch, i);
```
trws: I agree about making this free_now by the way, or similar, that would be the one thing I'd really like to see before this merges.

resource/traversers/dfu_impl.cpp
milroy: >  I was considering if we should key off of something like running under sharness

I tried using `getenv ()` to check for a sharness environment variable, but the problem is some of the sharness tests use the resource module which then requires resetting environment variables. That might also get complicated for people trying to use `resource-query`. Do you have another idea for detecting running under sharness?

Barring another idea, I'm just going to update the variable name.

PR #1380, opened 2025-07-22
add support for modprobe based startup and shutdown

grondo:
This PR adds the necessary bits for Fluxion to support the modprobe rc replacement prototype proposed in

https://github.com/flux-framework/flux-core/pull/6774

Even in the WIP PR, modprobe is only active when `FLUX_RC_USE_MODPROBE` is set to allow a transition period.
This PR prepares the Fluxion repo for this transition with the following fairly minor changes:

 1. adds `fluxion.toml` modprobe configuration file which defines Fluxion modules for `flux-modprobe(1)`
 2. adds `/etc/flux/modprobe/rc1.d/fluxion.py` to preserve the ability to set  `FLUXION_RESOURCE_OPTIONS` and `FLUXION_QMANAGER_OPTIONS` in the environment with a space-separated set of module options.
 3. disables the Fluxion rc scripts if `FLUX_RC_USE_MODPROBE` is defined, since currently `rc1.d/*` and `rc3.d/*` scripts are executed even under modprobe.
 4. Adapt the testsuite to work when `FLUX_RC_USE_MODPROBE` is set
 
 The process would be to merge this PR (after https://github.com/flux-framework/flux-core/pull/6774 is out of WIP or merged). Then swap flux-core to use modprobe, remove the temporary handling of `FLUX_RC_USE_MODPROBE` here and the rc1/3 based module loading, and finally drop use of `FLUX_RC_USE_MODPROBE` altogether.

I'm posting this here early to raise awareness of upcoming changes. Take a look at the flux-core PR and please post any comments on the design or implementation (note I'm working on docs now)


garlick (approved):
Since modprobe was merged in flux-core 0.77.0, seems like this can go in now?

etc/modprobe/modprobe.d/fluxion.toml
```diff
@@ -0,0 +1,15 @@
+[[modules]]
+name = "sched-fluxion-resource"
+provides = ["feasibility"]
+priority = 500
```
trws: @grondo, I'm guessing priority is the mechanism for ordering module loading right? If so, why are these both the same priority?

t/t3300-system-dontblock.t
```diff
@@ -5,13 +5,27 @@ test_description='Test fluxion does not block on R w/ no scheduling key
...
 unset FLUXION_QMANAGER_RC_NOOP
 export FLUXION_RESOURCE_OPTIONS="load-allowlist=node,core,gpu load-format=hwloc"
 
+if test -n "$FLUX_RC_USE_MODPROBE"; then
+    unset FLUX_MODPROBE_DISABLE
+    # modprobe is in use, and therefore failure to load Fluxion modules
+    # causes instance failure instead of falling back to sched-simple.
+    # Therefore test with flux-start(1) instead of test_under_flux() as
+    # below.
+    test_expect_success 'fluxion immediately fails to be loaded with hwloc' '
```
trws: Is this meant to run both tests when modprobe is turned on? The messages are identical except for this lacking "reader", and the second also runs with `FLUX_MODPROBE_DISABLE` unset.

etc/modprobe/modprobe.d/fluxion.toml
garlick: Snipped from `modprobe.toml` in flux core
```
#   priority  (optional) The relative priority of this module and its provides.
#              Since alternatives are sorted by priority then load order, a
#              module may be given a high priority to ensure it overrides any
#              alternatives for a given service name.
#              The default priority is 100.

#   requires  (optional) An array of services this module requires, i.e. if
#             this module is loaded then services/modules in requires will also
#             be loaded. If this module also has to be loaded after any
#             required modules, add them to the after array as well.
#
#   after     (optional) An array of modules for which this module must be
#             loaded after. If this module also requires the service or module
#             it is loaded after, then the module must also be added to
#             `requires`.
#
#   needs     (optional) An array of modules which are required for this module
#             to be loaded.
```

So priority is about which modules are loaded to provide the `sched` service.  `sched-simple` has a low priority so that fluxion can override it if present.

trws: Ok, so priority is there to ensure this gets loaded in place of sched-simple I guess? Makes sense, thanks @garlick.

t/t3300-system-dontblock.t
grondo: No, it just runs the one test. The message is the same because it is testing the same thing AFAIK.
In the `FLUX_RC_USE_MODPROBE` case, only the code in this `if` block is run because the block ends with `test_done` which terminates sharness.

e.g.

```console
$ FLUX_RC_USE_MODPROBE=t ./t3300-system-dontblock.t 
ok 1 - fluxion immediately fails to be loaded with hwloc
# passed all 1 test(s)
1..1
```

grondo:
> Since modprobe was merged in flux-core 0.77.0, seems like this can go in now?

The current version of this PR doesn't do anything different unless `FLUX_RC_USE_MODPROBE` is set. I wonder if we should set that variable in one of the builders for now, or maybe all of them like it is done in flux-core now?

In any event, the next steps might look something like this:

 - set `FLUX_RC_USE_MODPROBE` in the flux-sched testsuite (and ensure that works for all ci builders)
 - switch flux-core to modprobe  startup by default, drop support for `rc*.old` in flux-core or require an env var to enabled it by request
 - drop use of `FLUX_RC_USE_MODPROBE` in flux-core and flux-sched, remove old rc files and workarounds

However, this will tie the new flux-core and flux-sched releases together, so maybe more thought on the transition plan needed :thinking:

garlick:
Maybe we can discuss how this would impact fluxion R&D at the sched meeting tomorrow?
Meanwhile it seems like it would be safe to set FLUX_RC_USE_MODPROBE in test in this PR and get it merged?

t/t3300-system-dontblock.t
trws: Ah, I missed the test_done. Is there a downside to using this version of the test in both cases?

trws:
It shouldn't be a problem for fluxion R&D so long as we have a check in the configuration for the appropriate required flux version, and that we add similar requirements to the spack package etc.  Actually that's the one thing that's missing is a version check bump in the pkg-config query for flux-core.

t/t3300-system-dontblock.t
grondo: The test won't behave the same when not under modprobe. In that case, the failure to load Fluxion modules causes sched-simple to get loaded, so the Flux instance would succeed instead of failing.

There may be a different way to work around that, but since we'll be removing the older test once modprobe replaces the existing rc1, it may not be worth the effort (unless you can think of something simple that hasn't occurred to me)

garlick:
With FLUX_RC_USE_MODPROBE=t set for tests, flux-core 0.77 is required, but looking at one of the bookworm test failures, it appears we're picking up something less than:
```
2025-08-06T16:12:46.2443717Z -- Checking for module 'flux-core'
2025-08-06T16:12:46.2544841Z --   Found flux-core, version 0.76.0-14-g713dfd6ae
```

grondo:
Oh, thanks good catch. I wonder if that's because of our noble image issues...

grondo:
Ok, there's an actual issue that the failures here are uncovering, though I'm baffled why it is only on a couple of the builds. `flux modprobe` is not failing reliably on module load failure (and also in testing other task failures), so failure to load the Fluxion modules is not causing rc1 failure as it should.

This also exposes a testing gap in the modprobe tests in core, so I'm going to go back over there and add the testing and fix the issue hopefully.

trws:
On 7 Aug 2025, at 14:13, Mark Grondona wrote:

> grondo left a comment (flux-framework/flux-sched#1380)
>
> Ah, the difference could depend on if sched-fluxion-resource exits before sched-fluxion-qmanager is loaded. If not, then modprobe/rc1 will succeed, if so, then sched-fluxion-qmanager fails to load and rc1 fails.

I remember thinking this would be a good idea at least once before.  My only concern was that we may not want to do that if we’re doing a reload rather than an initial load.  Is there a way fluxion can report failure to load effectively without taking the instance with it?

grondo:
> My only concern was that we may not want to do that if we’re doing a reload rather than an initial load.  Is there a way fluxion can report failure to load effectively without taking the instance with it?

I may be confused what you're asking here, but the instance won't fail on a module reload. The instance would fail on a module load during rc1 because any failure in rc1 causes the rc1 script to fail, which then causes the broker to abort the instance and begin shutdown.

trws:
Out of curiosity, how can this happen?

> or it may succeed (without qmanager or the Fluxion resource module loaded). 

That seems like a _bad thing_™️

trws:
> > My only concern was that we may not want to do that if we’re doing a reload rather than an initial load.  Is there a way fluxion can report failure to load effectively without taking the instance with it?
> 
> I may be confused what you're asking here, but the instance won't fail on a module reload. The instance would fail on a module load during rc1 because any failure in rc1 causes the rc1 script to fail, which then causes the broker to abort the instance and begin shutdown.

Aren't we getting rid of rc1 scripts? Does initial modprobe similarly behave differently than module reload?

grondo:
> Out of curiosity, how can this happen?

The issue (I think at least!)  is that the Fluxion resource can fail _after_ it has returned from a successful module load. So neither `flux module load` or the equivalent task under modprobe detects failure and the startup moves on to subsequent tasks.  This can easily be seen by loading `sched-fluxion-resource` on the command line with the invalid args from `t3300-system-dontblock.t`:

```console
$ flux module load sched-fluxion-resource load-allowlist=node,core,gpu load-format=hwloc
Aug 12 14:53:27.651958 PDT 2025 sched-fluxion-resource.err[0]: Unknown service method 'resource.get-xml'
Aug 12 14:53:27.652022 PDT 2025 sched-fluxion-resource.err[0]: update_resource_db: grow_resource_db: Function not implemented
Aug 12 14:53:27.652040 PDT 2025 sched-fluxion-resource.err[0]: update_resource: update_resource_db: Function not implemented
Aug 12 14:53:27.652091 PDT 2025 sched-fluxion-resource.err[0]: populate_resource_db_acquire: update_resource: Function not implemented
Aug 12 14:53:27.652100 PDT 2025 sched-fluxion-resource.err[0]: populate_resource_db: loading resources using resource.acquire
Aug 12 14:53:27.652109 PDT 2025 sched-fluxion-resource.err[0]: init_resource_graph: can't populate graph resource database
Aug 12 14:53:27.652116 PDT 2025 sched-fluxion-resource.err[0]: mod_main: can't initialize resource graph database
Aug 12 14:53:27.652180 PDT 2025 sched-fluxion-resource.crit[0]: module exiting abnormally
$ echo $?
0
```

I agree this should probably fixed.

Stepping back a bit, I'll reiterate that the existing version of this test is relying on the behavior of the rc1 startup script which forces `sched-simple` to load if at the end of the script there is no `sched` service loaded. So, current behavior in this test is that Fluxion modules fail to load, but the instance starts successfully with the simple scheduler.

This is a difference in behavior for the modprobe solution, which determines which scheduler to load based on configuration and doesn't fall back to `sched-simple`.


> Aren't we getting rid of rc1 scripts? Does initial modprobe similarly behave differently than module reload?

Yes, the intent with modprobe (and the current rc1) is to fail if any step during initialization fails.

trws:
Got it.  I think we could fix that in fluxion, but it would mean making some of the init stuff synchronous in a way that was probably being avoided.  If I had to guess the main culprit is probably waiting for the first `resource.acquire` response, which we currently don't actually do.  It might be a bit of a messy refactor but we could probably do it.

grondo:
Ok. So shall we open a separate issue on getting async Fluxion module startup working, or is there something we can do in the modprobe implementation to ameliorate this? We could add a task to fall back to sched-simple if that's desired?

grondo:
This just needs an update to add support for the new `sched-fluxion-feasibility` module. 

I notice that module needs to be loaded or should be loaded with the same arguments as `sched-fluxion-resource`. This can be done, but there is potential for the settings via config or otherwise to get out of sync. If this is a potential problem, is there a way for `sched-fluxion-feasibilty` to get the current options from the `sched-fluxion-resource` module during acquire?

trws:
We could add such a mechanism, currently I think the answer is "no" though.

PR #1382, opened 2025-07-24
Add all leaf types to the cluster vertex aggregate pruning filter

milroy:
To enable more frequent traversal pruning at the cluster vertex and reduce the complexity involved with reducing `planner_multi`spans per this comment: https://github.com/flux-framework/flux-sched/blob/a5482f61e5521b2252c5e8091c034fd67f01ce40/resource/planner/c/planner_multi_c_interface.cpp#L499-L507

this PR detects and adds all leaf vertex types to the cluster vertex's aggregate pruning filter.


resource/traversers/dfu_impl.cpp
```diff
@@ -1151,7 +1157,7 @@ int dfu_impl_t::prime_pruning_filter (subsystem_t s,
     for (auto &aggr : dfv)
         accum_if (s, aggr.first, aggr.second, to_parent);
 
-    if (m_match->get_my_pruning_types (s, (*m_graph)[u].type, out_prune_types)) {
+    if (m_match->get_my_pruning_types (s, type, out_prune_types)) {
         for (auto &type : out_prune_types) {
```
trws: I know this is off by one from the change, but using `type` as both an alias for `(*m_graph)[u].type` and shadowing it as the prune type in this for loop is rather unfortunate.

milroy: Good point. I'll fix it and force push.

milroy:
>  if we enter prime_pruning_filter in a leaf vertex, and it's not a pruning type, this now adds the pruning type at the cluster vertex, and then propagates counts up through accum_if which hits all the agfilters between the leaf and the cluster right?

That's correct. You can see the behavior in the new types and counts tracked at the cluster vertex in the updated testsuite outputs.

milroy:
Thanks for the review @trws! Adding MWP.

PR #1387, opened 2025-08-07
Add match_without_allocating match option

jacobtkeio:
This PR adds a MATCH_WITHOUT_ALLOCATING match option to sched-fluxion-resource, flux ion-resource, the C and C++ REAPI, and resource-query. A request for MATCH_WITHOUT_ALLOCATING will return resources that a jobspec would have been matched with under MATCH_ALLOCATE, but it will not actually allocate those resources.
A caveat is that MATCH_WITHOUT_ALLOCATING will return the soonest-available match, even into the future, while MATCH_ALLOCATE will return failure if no match is possible starting exactly at the requested match time.
 
 Matching without allocating is a first step towards basic multi-cluster scheduling.
 A more useful next step is to find and report the longest available matching resources in a cluster, for which the likelihood that another cluster has available matching resources within the same timeframe is highest. Then, users can broadcast requests for the longest available matching resources on many clusters (without actually allocating the requests) and then choose suitable times and resources on which to actually allocate their multi-cluster job.


jacobtkeio:
I have added a "within" parameter for the resource module, REAPI C/C++, resource-query, and test suite. The "within" parameter forces the traverser to return only matches that start within the given duration (i.e. start at $now + $within or earlier).

It only affects MATCH_ALLOCATE_ORELSE_RESERVE and MATCH_WITHOUT_ALLOCATING right now, since those are the only two match options that can return resource sets that begin in the future. However, I added the parameter to every match option so that future match options don't need to worry about adding it manually. Internally, it is implemented as a "latest" parameter to dfu_match_t::run(), which specifies the latest start time allowed for returned matches.

After a while I started wondering if it would be simpler to add a "within" parameter as a jobspec attribute instead, since this would cut down on the amount of option-handling code spread across resource, REAPI, and rq. We may still want the option-handling code in any case, but the implementation may ultimately be cleaner based off a jobspec attribute, perhaps including on the flux-core side of things.

jameshcorbett:
Where are we at on this PR? Do you need a review @jacobtkeio ? It looks like #1404 and #1405 still depend on this?

jacobtkeio:
Yes, both 1404 and 1405 depend on this (sorry for opening them so early!).
I have just a tiny bit more polishing to do on this PR before it's ready for review
This PR slightly depends on #1515 only because I peeled 1515 off of it, and I could probably rebase off the main branch instead if you like.

jacobtkeio:
This PR now also depends on #1534.

wihobbs:
This, plus your other two PRs that depend on it, would actually be cool candidates for the "PR stacking" feature being discussed in Slack [if you wanted to give that a try](https://github.github.com/gh-stack/getting-started/quick-start/).

resource/traversers/dfu.cpp
```diff
@@ -386,6 +394,10 @@ int dfu_traverser_t::run (Jobspec::Jobspec &jobspec,
         < 0)
         return -1;
 
+    // If matching without allocation, set meta.AT to prevent allocation in update
```
milroy: I think you mean this:
```suggestion
    // If matching without allocation, set alloc_type to prevent allocation in update
```

resource/modules/resource.cpp
```diff
@@ -383,7 +383,10 @@ static void match_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_t
                              &js_str)
         < 0)
         goto error;
-    if (is_existent_jobid (ctx, jobid)) {
+    if (std::string ("without_allocating") == cmd
+        || std::string ("without_allocating_future") == cmd) {
+        jobid = -1;
```
milroy, line 391: If I recall correctly, this is the sentinel value we agreed on. If so, please document it in a code comment and expand the commit message to indicate how it will be used.

t/t4017-match-without-allocating.t
```diff
@@ -0,0 +1,125 @@
...
+done
+
+test_expect_success 'removing resource works' '
+    remove_resource
+'
+
+test_expect_success 'loading resource module with policy=low works' '
+    load_resource \
+load-file=${grug} prune-filters=ALL:core \
+load-format=grug subsystems=containment policy=low
```
milroy: Nit: indentation
```suggestion
    load-file=${grug} prune-filters=ALL:core \
    load-format=grug subsystems=containment policy=low
```

t/t4017-match-without-allocating.t
```diff
@@ -0,0 +1,125 @@
...
+	match allocate ${jobspec}
+	match without_allocating ${jobspec}
+	match without_allocating_future ${jobspec}
+    quit
+	EOF
+    test $(grep MATCHED <rq.out | wc -l) -eq 2 &&
+    test $(grep ALLOCATED <rq.out | wc -l) -eq 4 &&
+    test $(grep "No matching resources found" <rq.out | wc -l) -eq 1 &&
+    grep "SCHEDULED AT" rq.out | sed -ne "6p" | grep 3600
+'
```
milroy: I think this test belongs in `t3001` in a dedicated `match_without_allocating` section.

resource/traversers/dfu.cpp
```diff
@@ -101,10 +102,11 @@ int dfu_traverser_t::request_feasible (detail::jobmeta_t const &meta,
             get_graph_db ()->metadata.graph_duration.graph_end.time_since_epoch ())
             .count ();
     // only the initial time matters for allocate
-    const int64_t target_time =
-        op == match_op_t::MATCH_ALLOCATE || op == match_op_t::MATCH_WITHOUT_ALLOCATING
-            ? meta.at
-            : graph_end - 1;
+    const int64_t target_time = op == match_op_t::MATCH_ALLOCATE
+                                        || op == match_op_t::MATCH_WITHOUT_ALLOCATING
+                                        || op == match_op_t::MATCH_WITHOUT_ALLOCATING_FUTURE
```
milroy: Caught with Claude's help:
```suggestion
```
`MATCH_WITHOUT_ALLOCATING_FUTURE` is semantically like `MATCH_ALLOCATE_ORELSE_RESERVE`, but the ternary makes it equivalent to `MATCH_ALLOCATE` here. So when a cluster is fully occupied now, the precheck probes `meta.at`, gets "unavailable" on every node, returns `EBUSY` — and the future-seeking loop that would have found the match at, say, `t=3600` is never reached.

Example reproducer:

```bash
$ ./resource/utilities/resource-query -L t/data/resource/grugs/tiny.graphml -S CA
> match allocate t/data/resource/jobspecs/match_format/t3040_only_node.yaml
> match allocate t/data/resource/jobspecs/match_format/t3040_only_node.yaml
# cluster now fully node-allocated
> match allocate_orelse_reserve <node jobspec with constraints: {...}>
  → RESERVED at 3600                    (correct)
> match without_allocating_future <same jobspec>
  → "No matching resources found" (EBUSY)   (should match at 3600)
```
Note the jobspec must use the `constraints:` key (plural — jobspec.cpp:337); a singular `constraint:` silently lands in optional and takes the fast path, masking the bug. That masking is also why `t4017`'s MWOAF test (unconstrained jobspec) passes.

resource/traversers/dfu_impl.cpp
```diff
@@ -133,7 +133,9 @@ int dfu_impl_t::by_excl (const jobmeta_t &meta,
         // available. Note: if Fluxion needs to support shared
         // resources at the leaf level this check will not catch
         // multiple booking.
-        if (meta.alloc_type == jobmeta_t::alloc_type_t::AT_ALLOC
+        if (at == meta.now
```
milroy, line 142: The correct-but-too-expensive way to do this is:

```C++
  if (exclusive_in || resource.exclusive == Jobspec::tristate_t::TRUE) {
      // Reject vertices holding an overrun job: its planner span has
      // expired but the job has not been freed, so its true release
      // time is unknown. The planner alone would report such a vertex
      // available (issue #1043). Applies at every probe time: an overrun
      // job must not be offered now or at any future point.
      if (meta.alloc_type == jobmeta_t::alloc_type_t::AT_ALLOC
          || meta.alloc_type == jobmeta_t::alloc_type_t::AT_NO_ALLOC) {
          planner_t *plans = (*m_graph)[u].schedule.plans;
          for (auto const &[jobid, span] : (*m_graph)[u].schedule.allocations) {
              int64_t start = planner_span_start_time (plans, span);
              int64_t dur = planner_span_duration (plans, span);
              if (start < 0 || dur < 0 || start + dur <= meta.now) {
                  errno = EBUSY;
                  return -1;
              }
          }
      }
```
I don't think that even buys much given that MWOAF will be used to estimate future availability anyway. I recommend extending the current comment to state the intended behavior explicitly: at `now`, results are exact (overrun-safe); at future probe times, MWOAF is deliberately optimistic with respect to overrunning jobs. This raises another issue: I think MWOA/F should be documented in a README.

t/t1032-without-alloc-check.t
```diff
@@ -0,0 +1,71 @@
+#!/bin/sh
+
+# Adapted from t1024
+
+test_description='
+Check that match without-allocating does not book resources when a previous
+allocation exceeds its walltime
+'
```
milroy, line 8: Claude ran this test with and without the previous commit, and the test passes regardless. Here's a summary:

The reason is that jobs are submitted as `flux submit -N1 -x hostname` with no -t limit → duration 0 → planner span extends to graph end → during the epilog the span is still active, so MWOA's EBUSY comes from the ordinary planner-busy path, not the new by_excl gate. The test never enters the overrun window this test claims to exercise.

Fix: mirror t1024 — `flux submit -N1 -x -t1s sleep 30`, wait for the exception `type=timeout` event (span now expired) then assert MWOA/MWOAF return 16 while the epilog runs. My probe confirmed this scenario distinguishes fixed from unfixed builds.

Also consider a mixed-cluster MWOAF test for ≥2 nodes, one overrunning, one with an active future span; assert MWOAF does offer the overrunning node at the other node's completion point.

milroy (commented):
The PR is in pretty good shape already, and should be good to merge as soon as a few items are corrected and implemented.

t/t4017-match-without-allocating.t
jacobtkeio: I've thrown in the same change for `t4001` as well!

resource/traversers/dfu_impl.cpp
jacobtkeio: I am having trouble understanding how this implementation differs in practice from the one in the PR. Would you mind explaining the issue this implementation solves?
I will definitely extend the comment to explain the optimism issue. Does allocate_orelse_reserve also behave in the same optimistic fashion? I think this case is equivalent to when a job right before an existing reservation runs over its wall time into the reservation time. Since we have no way to predict which jobs will overrun, aren't all future matches inherently optimistic?

t/t1032-without-alloc-check.t
jacobtkeio: Fixed, and I confirmed it fails before the fix. Sorry about that! I'll add the mixed cluster test

resource/traversers/dfu.cpp
jacobtkeio: I failed to reproduce this because I couldn't get the planner to report unavailable at now when all nodes were allocated, probably because I missed something obvious. Let me know if you would really like a reproducer and I'll keep working it out. Otherwise this is a simple fix!

milroy: You didn't miss anything obvious. Try this jobspec:

```yaml
# constrained_node.yaml — t3040_only_node.yaml plus a constraint
version: 9999
resources:
  - type: node
    exclusive: true
    count: 1
attributes:
  system:
    duration: 3600
    constraints:
      hostlist: [ "node[0-1]" ]
tasks:
  - command: [ "app" ]
    slot: default
    count:
      per_slot: 1
 ```
Then try this:
```bash
$ ./resource/utilities/resource-query -L t/data/resource/grugs/tiny.graphml -S CA
> match allocate t/data/resource/jobspecs/match_format/t3040_only_node.yaml
  → ALLOCATED at Now
> match allocate t/data/resource/jobspecs/match_format/t3040_only_node.yaml
  → ALLOCATED at Now                # both nodes of tiny.graphml now allocated
> match allocate_orelse_reserve constrained_node.yaml
  → RESERVED at 3600                # correct: future walk finds t=3600
> match without_allocating_future constrained_node.yaml
  → "No matching resources found"   # WRONG — should be MATCHED at 3600
```

resource/traversers/dfu_impl.cpp
milroy: We discussed this in detail during our 1:1, but here's an explanation for posterity:

With node A overrunning its allocation plus other nodes with active spans that create a future point `t_e` (NB: depends on resources tracked by aggregate filter) the MWOAF walk will inspect `t_e`, and at `t_e > meta.now` the `at == meta.now` gate is skipped on every vertex including A. MWOAF can then report node A available at `t_e` while its overrunning job may still hold it. For internal reservations this optimism is self-correcting (qmanager re-schedules), but MWOAF results are transmitted externally (no re-scheduling feedback loop). The recommendation is just to document the behavior.

t/t1032-without-alloc-check.t
milroy: No problem, that's what reviews are for!

resource/traversers/dfu.cpp
jacobtkeio: It worked like a charm! Thank you so much

milroy (approved):
Thanks for the patience and the diligent work on this feature, @jacobtkeio! This capability will be really useful for single- and multi-cluster workloads.

Also, credit to Claude for helping with the reproducer for MWOF feasible.

milroy:
The PR looks ready to merge to me, but I'd like either @trws or @jameshcorbett to weigh in before setting MWP.

trws (approved):
Agreed, let's get this in. =)

jameshcorbett (approved):
I looked through briefly and am on board, I set MWP.

PR #1388, opened 2025-08-08
ci: add debug tmate action and use `docker-manifest.py` in generate-manifest step

grondo:
This PR pulls a couple recent ci updates from flux-core:

 - switch the `generate-manfest` step to use `src/test/docker-manifest.py` (I don't think there was a problem here, but this does automatically keep manifest generation in sync with `generate-matrix.py`)
 - add the tmate action which is conditionally run when restarting builds with the "debug" box checked


grondo:
Thanks! This will need a manual merge due to changes under `.github`

PR #1389, opened 2025-08-13
ci: fix docker-manifest.py

grondo:
Problem: The `src/test/docker-manifest.py` script inadvertently contained a reference to flux-core in it, causing the manifest generation to occur in the wrong repo.

Update the script to specify flux-sched, not flux-core.


grondo:
I was thinking the same thing when I saw that mistake :facepalm: Sounds like a great idea to me!

PR #1390, opened 2025-08-16
testsuite: set broker.module-nopanic=1

garlick:
Problem: some tests rely on flux-core <= 0.77 broker module runtime error handling, where a module spuriously exiting with error is not fatal to the Flux instance.
    
flux-framework/flux-core#6969 added a check so that if a module exits with error on its own, the instance shuts down with an error so that this condition can be more easily detected.
    
Temporarily set the broker attribute `broker.module-nopanic=1` in failing fluxion tests to get the old behavior.  At some point,  perhaps after flux-core-0.78.0 is released, tests should be updated to expect the new behavior and this commit may be reverted.

p.s. unrelated: also added pop!OS to `install-deps.sh` in this PR since that's what I'm using for development.

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -10,7 +10,7 @@ fi
 
 PM=""
 case "$ID" in
-	ubuntu|debian)
+	ubuntu|debian|pop)
 		PM=apt
 		DEV_SUFFIX="-dev"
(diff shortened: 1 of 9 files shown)


grondo (approved):
LGTM!

garlick:
Thanks! I'll go ahead and set MWP

PR #1393, opened 2025-08-20
testsuite: don't load deprecated barrier module

garlick:
Problem: the flux-core barrier service is deprecated but its broker module is being loaded in fluxion tests.

It is not used by fluxion.  Drop it from rc scripts.


chu11 (approved):
LGTM

PR #1406, opened 2025-09-21
Fix partial cancel for allocations that span higher-level vertices

milroy:
## Problem summary and solution
Multiple occurrences of resource state divergence between core and sched have been reported on production systems when partial cancel is enabled and JGF is in use. A recent debugging session with @trws and @jameshcorbett on one such system revealed the cause: allocations spanning higher-level vertices (e.g., an allocation of 3 nodes spanning two racks) experience errors when reducing `planner_multi` spans during partial cancel. 

The source of the problem is that accumulated counts of lower-level resources are removed from all higher-level resources associated with the job regardless of whether those counts correspond to the number of job resources allocated in the subtree rooted at a vertex in question. 

For example, in a rack-spanning job with 3 nodes (1 in `rack0` and 2 in `rack1`), partially cancelling the 2 nodes in `rack1` causes `planner_multi` to attempt to remove 2 nodes from the job span on `rack0`. The attempt results in an error which leaves the vertices in an inconsistent, unallocatable state. Specifically, vertices can retain job tags and spans.

Along the course of investigating this problem, I discovered two more issues. First, the current rank-based removal (used prior to subgraph deletion) also accumulated counts without tracking counts associated with jobids or higher-level graph resources. That oversight can lead to even greater problems. Second, the JGF partial cancel logic was computing the number of resources removed by referencing an `at` time of `0` rather than the start time of the span. Fixing the second issue simply entails ~referencing the span start time~ getting the span's occupied resources with `planner_span_resource_count`.

Fixing partial cancel for higher-level, vertex-spanning allocations requires accumulating counts per rank and jobid and propagating those counts to the rank subgraph root's ancestors in the resource graph. That approach obviated a DFS cancellation. 

## Remaining items for discussion
When implementing tests using different pruning filter configurations and jobspecs, I noticed that one configuration did not result in a complete removal of all allocation vertex metadata: https://github.com/flux-framework/flux-sched/blob/992b8b18b7eb72be51ba72955f7aef1183b3494c/t/data/resource/expected/cancel/029.R.err. (Note the empty file: it means the partial cancel didn't remove all job metadata and thus the follow-up full cancel didn't error.) This is caused by partial cancel not removing the job tag due to this condition being met: https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L427-L430. The cause is a jobspec like this:
```yaml
version: 9999
resources:
  - type: slot
    count: 3
    label: default
    with:
      - type: node
        count: 1

# a comment
attributes:
  system:
    duration: 3600
tasks:
  - command: [ "app" ]
    slot: default
    count:
      per_slot: 1
```
The `slot` implicitly allocates children exclusively, however, the traverser does not accumulate their counts and set them in the agfilter upon allocation (i.e., `dfu` is empty): https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl_update.cpp#L75 

A follow-up, full cancel fixes this issue, but it warrants further discussion since implicit slot exclusivity causes problems elsewhere.

The second remaining discussion point is the necessity for a `sleep` here: https://github.com/milroy/flux-sched/blob/992b8b18b7eb72be51ba72955f7aef1183b3494c/t/t2317-resource-shrink.t#L46
This wasn't necessary before, and I'd like a better way to deal with the asynchronous behavior.


resource/readers/resource_reader_base.hpp
```diff
@@ -26,7 +26,9 @@ enum class job_modify_t { CANCEL, PARTIAL_CANCEL, VTX_CANCEL };
 
 struct modify_data_t {
     job_modify_t mod_type = job_modify_t::PARTIAL_CANCEL;
-    std::unordered_set<int64_t> ranks_removed;
+    std::unordered_set<int64_t> ranks;
```
trws, line 29: I may find out as I review later commits, but why an unordered set rather than a vector? Is there a chance of duplication or need to do checks on whether a rank exists in the set?

resource/readers/resource_reader_jgf.cpp
```diff
@@ -808,12 +808,14 @@ int resource_reader_jgf_t::cancel_vtx (vtx_t vtx,
...
     std::map<int64_t, int64_t>::iterator xspan_it;
+    uint32_t path_len = 0;
+    std::unordered_map<int64_t, uint32_t>::iterator rank_to_length_it;
 
     static subsystem_t containment_sub{"containment"};
     // remove from aggregate filter if present
     auto agg_span = job2span.find (update_data.jobid);
     if (agg_span != job2span.end ()) {
-        if ((subtree_plan = g[vtx].idata.subplans[containment_sub]) == NULL)
+        if ((subtree_plan = g[vtx].idata.subplans.at (containment_sub)) == NULL)
```
trws: My brain hitched on this tweak a bit.  The new version will throw an exception if the argument is out of bounds. Is the goal to support a null value in the structure and get an exception if we're past the end?

resource/readers/resource_reader_base.hpp
```diff
@@ -26,7 +26,9 @@ enum class job_modify_t { CANCEL, PARTIAL_CANCEL, VTX_CANCEL };
...
+    std::unordered_map<int64_t, vtx_t> rank_to_root;
+    std::unordered_map<int64_t, std::unordered_map<const char *, int64_t>> rank_to_counts;
```
trws: Why the char * here rather than using the interned type string type? I think that's what the vertex type member that gets converted to a c_str to go into this is.

resource/readers/resource_reader_jgf.hpp
```diff
@@ -30,7 +30,10 @@ struct jgf_updater_data {
     uint64_t duration = 0;
     bool reserved = false;
     // track counts of resources to be cancelled
-    std::unordered_map<const char *, int64_t> type_to_count;
+    std::unordered_map<int64_t, std::unordered_map<const char *, int64_t>> rank_to_counts;
+    // track length of paths to determine subgraph roots
+    std::unordered_map<int64_t, uint32_t> rank_to_length;
+    std::unordered_map<int64_t, vtx_t> rank_to_root;
```
trws: It looks like all three of these are usually used together, maybe better to have one map with a small struct?

resource/traversers/dfu_impl_update.cpp
```diff
@@ -1023,24 +1066,65 @@ int dfu_impl_t::remove (vtx_t root, const std::set<int64_t> &ranks)
...
+    std::unordered_map<vtx_t,
+                       std::unordered_map<int64_t, std::unordered_map<const char *, int64_t>>>
+        parent_jobid_counts;
+    for (const auto &rank_root : mod_data.rank_to_root) {
+        vtx_t subgraph_root_vtx = rank_root.second;
+        // Accumulate type_to_count for all vertices up to graph root
+        vtx_t parent_vtx = boost::graph_traits<resource_graph_t>::null_vertex ();
+        vtx_t curr_vtx = subgraph_root_vtx;
+        while (get_parent_vtx (curr_vtx, parent_vtx) == 0) {
+            // Add error check for rank_to_counts
```
trws: Vestigial comment?


```suggestion
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -1023,24 +1066,65 @@ int dfu_impl_t::remove (vtx_t root, const std::set<int64_t> &ranks)
...
+        for (const auto &jobid : v.second) {
+            modify_data_t mod_data_new;
+            mod_data_new.mod_type = job_modify_t::PARTIAL_CANCEL;
+            mod_data_new.type_to_count = jobid.second;
+            if ((rc = cancel_vertex (v.first, mod_data_new, jobid.first)) != 0) {
+                m_err_msg += __FUNCTION__;
+                m_err_msg += ": cancel_vertex failed.\n";
+                return rc;
+            }
+        }
```
trws, line 1123: Most of this update makes great sense, but I'm trying to work out the need for walking more than one jobid here.  Is there a case where a partial cancel is meant to apply to more than one jobid?  I would have thought we would need to filter out counts to jobids other than the one we're working on rather than needing to track multiple of them.

milroy:
Thanks for the early reviews and feedback @trws and @garlick!

I'll reply to the other comments later tonight, but will start here:

> I hadn't realized we were doing this, so an implicit exclusive allocation is just set at the root of the exclusive tree and we don't track the elements under it?

That's correct.

> I can see why for efficiency, but if we're removing all the markings how do we end up with things remaining tagged, or is it a case of cutting out too early so we still leave it on that root vertex?

This may be a bug. If the behavior were motivated by efficiency, I'd expect implicit and explicit exclusivity to behave the same. For partial cancel in the scenario above (implicit, exclusive node allocation), the reason all job tags associated with a jobid don't get removed is because (due to the implicit behavior) the `job2span` map doesn't get a `(jobid, span)` inserted here: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl_update.cpp#L93 Then during partial cancel, this condition gets met: https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L427-L430, which ends up skipping the `rem_tag`.

We could adjust the behavior of implicit `slot` exclusivity, but at that point we might as well revisit whether we actually want children of slots to be exclusive, as I believe there were rabbit use cases for allowing non-exclusive slots: https://github.com/flux-framework/flux-sched/discussions/1189#discussioncomment-9266199. Alternatively, I can drop the `!dfu.empty ()` condition here: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl_update.cpp#L75 That change allows all other CI tests to pass, but I'm not sure if there are side effects.

resource/readers/resource_reader_base.hpp
milroy: That's true, I could use a `resource_type_t`. The complexity comes from needing to handle `char *`s in the JGF reader: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/readers/resource_reader_jgf.cpp#L61 and `char *` required by planner: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl_update.cpp#L461

I could use `resource_type_t` in the `unordered_map` and then convert to `char *` where necessary. That would probably be faster for lookups and insertions.

resource/readers/resource_reader_jgf.hpp
milroy: Yeah, that would be cleaner!

resource/traversers/dfu_impl_update.cpp
milroy: Indeed. I'll get rid of that!

resource/readers/resource_reader_jgf.cpp
milroy: This was a mistake. I'll revert back to the current line.

resource/readers/resource_reader_base.hpp
milroy: > Is there a chance of duplication or need to do checks on whether a rank exists in the set?

Yes, there is duplication where the ranks are inserted in the JGF reader, and their existence is checked in `mod_agfilter`. However, if there's no need to call `mod_dfv` in partial cancel the existence check isn't necessary. After getting rid of the existence check I could put the ranks into a vector in `modify_data_t` in the JGF reader.

resource/traversers/dfu_impl_update.cpp
milroy: The need comes from shrinking the resource graph in `resource.acquire`: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/modules/resource_match.cpp#L1120 All jobids need to be partially cancelled from the resources corresponding to the decoded rankset derived from the `shrink` key. Then the subgraph corresponding to those ranks can be deleted from the resource graph.

There could be multiple jobids mapped to resources in a single rank, so to reduce spans at higher-level vertices you need a count of resources per jobid per rank.

t/data/resource/jobspecs/cancel/test026.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# a comment
+attributes:
+  system:
+    duration: 3600
+tasks:
+  - command: [ "app" ]
+    slot: default
+    count:
+      per_slot: 1
```
jameshcorbett, line 28: Would it be easy to add a test case with a jobspec that has a `rack` type above `node` rather than `slot`? It shouldn't matter, but that seems like a more natural case to me.

jameshcorbett (commented):
I've given this several looks through, and it seems reasonable! I'm still familiarizing myself with the partial cancel logic though.

resource/readers/resource_reader_base.hpp
milroy: Changing the key type to `resource_type_t` ends up being a simpler implementation as there's only one conversion to `char *`.

t/data/resource/jobspecs/cancel/test026.yaml
milroy: This jobspec is unsatisfiable (note that `slots` are not required):

```yaml
version: 9999
resources:
  - type: rack
    count: 1
    with:
      - type: node
        count: 17
        with:
          - type: socket
            count: 2
            with:
              - type: core
                count: 18
              - type: gpu
                count: 1
              - type: memory
                count: 8
...
```
But putting a `slot` with a count of 17 between `node` and `rack` behaves exactly like `test026.yaml`. I added the latter example to the tests, but let me know if you meant something else.

resource/readers/resource_reader_base.hpp
milroy: Good suggestion!

milroy:
I think I've addressed your edits and feedback @trws and @jameshcorbett. The PR is ready for another review.

milroy:
> We could adjust the behavior of implicit slot exclusivity...

I decided that adjusting implicit slot exclusivity behavior is beyond the scope of this PR after discussing with @trws and @jameshcorbett. I opened WIP PR #1408 to address that separate issue.

t/data/resource/jobspecs/cancel/test026.yaml
jameshcorbett: Sorry, I should have provided an example. I had in mind something more like
```yaml
version: 9999
resources:
  - type: rack
    count: 2
    with:
      - type: node
        count: 2
        with:
          - type: socket
            count: 1
            with:
              - type: core
                count: 1
              - type: gpu
                count: 1
...
```
Attempting to partial-cancel one node on each of the two resulting racks should result in the bug right?

I was just thinking that a jobspec like this would be the simplest reproducer.

milroy: Got it, that makes sense. I'll add that test case as well.

resource/planner/test/planner_test02.cpp
```diff
@@ -776,14 +777,17 @@ static int test_partial_cancel ()
           || rc != 0);
     ok (!bo, "series of partial removals fully removes span");
 
+    bo = (bo || used != 0);
+    ok (!bo, "partial reductions correctly report used resources of completely removed resource");
+
```
jameshcorbett: Given that `bo` is checked for false above on line 777/778, couldn't this be 

```
ok((used == 0), "partial reductions correctly...);
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -686,9 +695,18 @@ int dfu_impl_t::clear_vertex (vtx_t vtx, modify_data_t &mod_data)
...
-            planner_span_resource_count (plans, alloc_it.second);
+        if ((res_count = planner_span_resource_count (plans, alloc_it.second)) < 0) {
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": planner_span_resource_count failed.\n";
+            m_err_msg += strerror (errno);
+            m_err_msg += ".\n";
+            return -1;
+        }
+        counts += res_count;
+        mod_data.rank_to_counts[(*m_graph)[vtx].rank][(*m_graph)[vtx].type.c_str ()] += counts;
```
jameshcorbett: This looks like a change beyond just error checking. Before `type_to_count` was incremented by the return value of `planner_span_resource_count` but now `rank_to_counts` is incremented by the sum of all `planner_span_resource_counts` in the loop so far. I think the commit message might need updating but also can you explain this change for me?

resource/readers/resource_reader_base.hpp
```diff
@@ -26,8 +26,10 @@ enum class job_modify_t { CANCEL, PARTIAL_CANCEL, VTX_CANCEL };
 
 struct modify_data_t {
     job_modify_t mod_type = job_modify_t::PARTIAL_CANCEL;
-    std::unordered_set<int64_t> ranks_removed;
-    std::unordered_map<const char *, int64_t> type_to_count;
+    std::unordered_set<int64_t> ranks;
```
jameshcorbett, line 29: Nit but I don't think this commit will compile as it is, because I think there are still references to `ranks_removed` elsewhere. I noticed later commits change those references to just `ranks`. I wonder if you could do all that renaming in this commit?

jameshcorbett (approved):
Made another pass, things look good!

resource/traversers/dfu_impl_update.cpp
milroy: Yes, this commit includes changes intended for another commit. I'll fix that.

The need for `res_count` is to check the return value for an error state and exit under that condition. I should be updating the map outside the loop and will correct that in a fixup.

resource/planner/test/planner_test02.cpp
milroy: I updated the commit to remove `bo = (bo || used != 0);` and move the check to right after `used` is assigned.

resource/readers/resource_reader_base.hpp
milroy: You're right. I'm going to merge this commit with the dependent, later commit and do some reorganization.

milroy:
Thanks for the feedback, @jameshcorbett! I think I addressed all your comments. Can you take a quick look at the changes?

@trws, this is ready for a final review. Let me know if you have suggestions for improving the code coverage.

t/data/resource/jobspecs/cancel/test028.yaml
```diff
@@ -0,0 +1,31 @@
+version: 9999
+resources:
+  - type: rack
+    count: 1
+    with:
+      - type: slot
+        count: 17
+        label: default
+        with:
```
jameshcorbett, line 9: I don't understand why this is able to match actually. Isn't this effectively requesting 17 nodes on a rack?

milroy: Yes, it's asking for 17 nodes in one rack. The `power.json` JGF has 18 nodes per rack.

resource/traversers/dfu_impl_update.cpp
trws: Got it, so this isn't a partial cancel of a _job_ at all, it's a cancel of the entire rank.  It looks like the behavior is that all impacted jobs are shrunk, rather than cancelled, is that right? I would have intuitively expected it to be implemented as a loop over the jobids impacted by the rank invoking partial cancel once for each given that rank rather than down here. In the end it doesn't matter in terms of result, this version is probably faster actually, but it surprised me to see something I expected to operate on a single jobid handling multiple.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -809,56 +808,69 @@ int resource_reader_jgf_t::cancel_vtx (vtx_t vtx,
     std::map<int64_t, int64_t>::iterator span_it;
     std::map<int64_t, int64_t>::iterator xspan_it;
 
-    static subsystem_t containment_sub{"containment"};
+    const subsystem_t containment_sub{"containment"};
```
trws: This should be both const and static if possible.  No reason to look it up every call, the storage is too small to worry about it.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -983,41 +961,83 @@ int dfu_impl_t::remove (vtx_t root,
...
+        // Accumulate type_to_count for all vertices up to graph root
+        vtx_t parent_vtx = boost::graph_traits<resource_graph_t>::null_vertex ();
+        vtx_t curr_vtx = subgraph_root_vtx;
+        while (get_parent_vtx (curr_vtx, parent_vtx) == 0) {
+            const auto &rank_it = mod_data.rank_to_counts.find (rank_root.first);
+            if (rank_it == mod_data.rank_to_counts.end ()) {
+                m_err_msg += __FUNCTION__ + std::string (": ");
+                m_err_msg += std::to_string (rank_root.first) + " not found in rank_to_counts.\n";
+                return -1;
+            }
```
trws: This part seems to be loop invariant in the enclosing while loop, could this move to just before it?

t/t1028-rv1-partial-release-across-racks.t
```diff
@@ -0,0 +1,392 @@
...
+	flux cancel ${jobid1}
+'
+
+# Check job manager hello debug message for +partial-ok flag
+if flux dmesg | grep +partial-ok; then
+    test_set_prereq HAVE_PARTIAL_OK
+fi
+
+test_expect_success HAVE_PARTIAL_OK 'fluxion shows 1 node allocated' '
+	wait_for_allocated_count 1
```
trws, line 92: If the bug came back, this would become a hang right? Should we have a timeout here?

t/t1028-rv1-partial-release-across-racks.t
```diff
@@ -0,0 +1,392 @@
...
+'
+
+test_expect_success HAVE_PARTIAL_OK 'kill housekeeping' '
+	flux housekeeping kill --all
+'
+
+# After killing housekeeping the node should be released
+test_expect_success HAVE_PARTIAL_OK 'fluxion shows 0 nodes allocated' '
+	wait_for_allocated_count 0
+'
```
trws, line 257: These tests have a great deal of duplication, looking like they just swap policies and pruning filters right? Could we have the large chunk of tests in a shell function that's invoked after each config reload so it's easier to see what's different?

t/t1028-rv1-partial-release-across-racks.t
milroy: It will retry 5 times (once per 0.1s). If the count `!= 1` before the final retry the function exits with a return code of 1, and otherwise exits with a return code of 0.

t/data/resource/expected/cancel/024.R.out
```diff
@@ -104,8 +104,8 @@ INFO: RESOURCES=ALLOCATED
 INFO: SCHEDULED AT=Now
 INFO: =============================
 INFO: =============================
-INFO: PARTIAL CANCEL PREORDER COUNT="4"
-INFO: PARTIAL CANCEL POSTORDER COUNT="2"
+INFO: PARTIAL CANCEL PREORDER COUNT="0"
+INFO: PARTIAL CANCEL POSTORDER COUNT="0"
```
trws, line 108: How is it that both of these come out as zero?

t/data/resource/expected/cancel/026.R.err
```diff
@@ -0,0 +1 @@
+cancel: ERROR: nonexistent job 2
```
trws, line 1: Quite a few of these have this cancel error in their err output. I can't tell the way I'm looking at this, are these expected because the partial cancels already processed the full cancel before we get to the cancel command?

t/data/resource/expected/cancel/024.R.out
milroy: Since partial cancel now just directly removes allocation counts from the parents, there's no traversal and the counts should be zero. I could drop the `stats` option in the tests to suppress the counts.

t/data/resource/expected/cancel/026.R.err
milroy: Yes, now the partial cancels almost always (except for the implicit exclusivity cases) remove the full allocation where appropriate. The error indicates that the full cancel fails because the root jobid tag is removed.

resource/traversers/dfu_impl_update.cpp
milroy: > Got it, so this isn't a partial cancel of a job at all, it's a cancel of the entire rank.

Basically, yes, it's a rank cancellation that triggers partial cancellation of all jobs on the specified ranks.

> It looks like the behavior is that all impacted jobs are shrunk, rather than cancelled, is that right?

Correct, unless the ranks to be removed constitute all of a job's resources in which case this should completely remove the job, too.

t/t1028-rv1-partial-release-across-racks.t
milroy: So it won't hang if the bug comes back.

resource/traversers/dfu_impl_update.cpp
milroy: Good catch! I'll move this and the other instance of this block to within the outer `for` loop.

t/t1028-rv1-partial-release-across-racks.t
milroy: Yeah, I should be able to do that cleanly.

milroy: I reworked the test to call functions to load the new config and reload resource and qmanager. Let me know if the changes look alright.

resource/traversers/dfu_impl_update.cpp
milroy: A bit embarrassed I missed this originally, but it's fixed now.

t/data/resource/expected/cancel/026.R.err
milroy: I tested various methods of redirecting stderr to the `*.out` files to avoid creating the `*.err` files, and can't think of anything more reproducible than the output configuration I already have in this PR. I'm happy to consider other options, though.

milroy:
Thanks again for the suggestions @trws. I think I've addressed them all. This is ready for a hopefully final look.

resource/traversers/dfu_impl_update.cpp
trws: Not a problem, pretty sure it was correct as it was.

t/t1028-rv1-partial-release-across-racks.t
trws: Yeah that definitely helps, thanks!

t/data/resource/expected/cancel/024.R.out
trws: Got it.  We can do that but I'm not worried about it at the moment now that I know why it comes out this way.

t/data/resource/expected/cancel/026.R.err
trws: Don't worry about it.  I just wanted to be sure that was intentional, and couldn't follow some of the chains to be sure the partial cancel got everything.  Thanks!

trws (approved):
This looks great, thanks for all the work @milroy!

PR #1408, opened 2025-09-26
Change allocation and reservation behavior of implicit slot exclusivity

milroy:
### Problem summary
PR #1406 identified the unexpected and unintentional behavior of the implicit exclusivity of resources specified below a `slot` in a jobspec. In particular, while the vertices are allocated or reserved and their metadata entered appropriately, their counts are not propagated to parent aggregate filters. The missing propagation is due to `exclusive` being set to `false` here: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl.cpp#L238
which translates to `dfu` being empty here: https://github.com/flux-framework/flux-sched/blob/8e1718d1cd953454903f71f7a7bb25ec17e31fab/resource/traversers/dfu_impl_update.cpp#L75
hence the aggregate filter isn't updated. 

Beyond presenting difficulties for partial cancel, this behavior translates to aggregate filters tracking fewer than the actual allocated or reserved counts. This condition negates the benefit of aggregate filters and leads to searching far more of the graph than necessary.

### Solution
This PR introduces a jobspec traversal in `prime_jobspec`. The traversal sets all resources under each slot as exclusive. It also enables extension to allowing slots to contain resources requested non-exclusively. This PR also changes a large number of expected output. I've reviewed the output and made changes to commands as needed (e.g., changing `find jobid-alloc=1 agfilter=t` to `find jobid-span=1 agfilter=t` since higher-level vertices do not need to be tagged with an allocation corresponding of lower-level vertices).

~The extent of the changes means this PR should be WIP to allow further analysis.~ This PR is based on PR #1406.


milroy:
I rebased the PR off the latest `master`, and as expected one of the `t3038-resource-flexible.t` tests failed. @zekemorton, can you look at the output and let me know if the new output is correct?

zekemorton:
@milroy It looks like the new expected output is correct. Please feel free to update this test

milroy:
I tried various approaches, but haven't found a good way to mock the hardware setup and test process binding stability under the changes of this PR. The closest I've gotten is the implementation in `t1030-slot-exclusivity.t`.

With that caveat, this PR is ready for review.

milroy:
I'm trying to implement a test similar to https://github.com/flux-framework/flux-core/blob/master/t/t2604-job-shell-affinity.t, and have run into an obstacle. Running this test:
```bash
. $(dirname $0)/sharness.sh

hwloc_basepath=`readlink -e ${SHARNESS_TEST_SRCDIR}/data/hwloc-data`
# 1 brokers: 1 node, 2 sockets, 44 cores 4 gpus
excl_1N1B="${hwloc_basepath}/001N/exclusive/01-brokers-sierra2"

export FLUX_SCHED_MODULE=none

export FLUX_HWLOC_XMLFILE="${hwloc_basepath}/001N/exclusive/01-brokers-sierra2/0.xml"

test_under_flux 1

test_expect_success 'load test resources' '
        load_test_resources ${excl_1N1B} &&
        load_resource prune-filters=ALL:core,ALL:node subsystems=containment policy=first &&
        load_qmanager queue-policy=easy
'

test_expect_success 'submit a test job' '
        flux run -N1 -n1 -c2 -o cpu-affinity=dry-run true
'
```
results in this output:
```bash
0.030s: flux-shell[0]: ERROR: cpu-affinity: affinity: core42 not in topology
0.030s: flux-shell[0]:  WARN: cpu-affinity: unable to get cpuset for cores 42-43. Disabling affinity
```

Removing `load_test_resources ${excl_1N1B}` (with and without `FLUX_SCHED_MODULE=none`) causes flux to discover the physical test environment hardware, which suggests `FLUX_HWLOC_XMLFILE` isn't working. All this said, the `WARN` message does indicate that the expected cores are selected, so I could just use that message as the expected output. That approach seems clunky, though.

@grondo do you have ideas about what I'm doing wrong?

grondo:
This is due to this code in the shell affinity plugin:
```c
    /*  Attempt to get cpuset union of all allocated cores. If this
     *   fails, then it might be because the allocated cores exceeds
     *   the real cores available on this machine, so just log an
     *   informational message and skip setting affinity.
     */
    if (!(sa->cpuset = shell_affinity_get_cpuset (sa, sa->cores))) {
        shell_warn ("unable to get cpuset for cores %s. Disabling affinity",
                    sa->cores);
        return 0;
    }
```
The warning and return should perhaps be skipped if `dry-run` is set. Let me try to add a test for this case and see if this change fixes the undesired behavior.

Edit: Hm, but the cpuset member still needs to be created from the topology. Somehow the system topology is leaking in during this step. Will require more work than I thought.

grondo:
@milroy: This seems like a combination of bugs in both handling of `FLUX_HWLOC_XMLFILE` and `cpu-affinity=dry-run`.
See https://github.com/flux-framework/flux-core/issues/7401 for tracking.

This should be an easy fix, but I'll need to add a test case first.

resource/policies/base/matcher.cpp
```diff
@@ -297,15 +297,8 @@ bool matcher_util_api_t::get_my_pruning_types (subsystem_t subsystem,
...
-    if (m_x_resource_types.find (type) == m_x_resource_types.end ()) {
-        auto ret = m_x_resource_types.insert (type);
-        if (!ret.second) {
-            errno = ENOMEM;
-            rc = -1;
-        }
-    }
-    return rc;
+    auto ret = m_x_resource_types.insert (type);
+    return ret.second ? 0 : -1;
```
zekemorton, line 301: I could be wrong, but I think this might have a slight behavior change in the case that `type` already exists in `m_x_resource_types`. In the original, the return value would be 0 because the if conditional would be false and it would go straight to return. In the new version, it would do the insert and I think if the key already exists, ret.second will be set to false, which will change the return value to -1 

see https://cplusplus.com/reference/set/set/insert/

resource/policies/base/matcher.cpp
```diff
@@ -315,16 +308,9 @@ const std::set<resource_type_t> &matcher_util_api_t::get_exclusive_resource_type
 
 int matcher_util_api_t::reset_exclusive_resource_types (const std::set<resource_type_t> &x_types)
 {
-    int rc = 0;
     m_x_resource_types.clear ();
```
zekemorton, line 311: I don't know if the clear is still required since you are doing an assignment right after anyways.

resource/traversers/dfu_flexible.cpp
```diff
@@ -146,24 +146,28 @@ int dfu_flexible_t::min_if (subsystem_t subsystem,
...
         accum_if (subsystem, resource.type, resource.count.min, to_parent);
-        prime_jobspec (resource.with, resource.user_data);
+        prime_jobspec (resource.with, resource.user_data, js_trav);
 
         // Or slots should use a minimum of values rather than an accumulation
         // otherwise possible matches may be filtered out
         if (resource.type == slot_rt) {
+            js_trav.slot_resource = true;
+            js_trav.slot_level = js_trav.level;
             for (const auto &aggregate : resource.user_data) {
```
zekemorton, line 171: In the flexible traverser, it looks like you recurse **before** setting js_trav.slot_resource = true; 

I think this implies that it never gets set for any of the children of this slot. I tested it a test005 for the flexible traverser. This tests passes because there are two slots listed with the exact same resource types. I am guessing that the first slot does not set any exclusivity, but then the flag is triggered and the exclusivity is set in the next slot. Here is the original job spec for flexible test 005 

```
version: 9999
resources:
  - type: slot
    count: 2
    label: default
    with:
      - type: node
        count: 1
        with:
          - type: socket
            count: 2
            with:
              - type: core
                count: 5
              - type: gpu
                count: 1
              - type: memory
                count: 6
  - type: slot
    count: 2
    label: default
    with:
      - type: node
        count: 1
        with:
          - type: socket
            count: 2
            with:
              - type: core
                count: 5
              - type: gpu
                count: 1
              - type: memory
                count: 6
```
With this job spec, resource query correctly outputs that both nodes are full allocated. After removing one slot, with this as the job spec:
```
version: 9999
resources:
  - type: slot
    count: 2
    label: default
    with:
      - type: node
        count: 1
        with:
          - type: socket
            count: 2
            with:
              - type: core
                count: 5
              - type: gpu
                count: 1
              - type: memory
                count: 6
```
resource query is unable to find a match. I am not sure why this is the case, but if I reduce the number of slots, it can find a match. But it's clear that changes for implicit exclusivity don't take in effect.
```
version: 9999
resources:
  - type: slot
    count: 1
    label: default
    with:
      - type: node
        count: 1
        with:
          - type: socket
            count: 1
            with:
              - type: core
                count: 1
              - type: gpu
                count: 1
              - type: memory
                count: 1
```
```
resource-query> match allocate t/data/resource/jobspecs/flexible/test005.yaml
      ---------------core35[1:x]
      ---------------gpu1[1:x]
      ---------------memory7[2:x]
      ------------socket1[1:x]
      ---------node1[1:x]
      ------rack0[1:s]
      ---tiny0[1:s]
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
```

resource/traversers/dfu_impl.cpp
```diff
@@ -1239,23 +1239,36 @@ int dfu_impl_t::prime_pruning_filter (subsystem_t s,
...
+        // If the resource is a slot resource, set the flag and track the level.
+        // This sets the exclusivity of all child resources
+        if (resource.type == slot_rt) {
+            js_trav.slot_resource = true;
+            js_trav.slot_level = js_trav.level;
+        } else if (js_trav.slot_resource && js_trav.level <= js_trav.slot_level) {
+            // Reset slot resource tracking
+            js_trav.slot_resource = false;
+            js_trav.slot_level = std::numeric_limits<int>::max ();
+        }
```
zekemorton, line 1257: So I have thought through this logic here a bit, and I think it does do what we want it to do. I can't help but wonder if there is a simpler way? Is there a reason that we can just pass a simple boolean like `exclusive` into prime_jobspec instead of keeping track of depth? 
Maybe we can start with `exclusive`  set to false. Then if it's a slot resource, we call prime_jobspec for all the children resources with exclusive set to true. If it's not a slot, then call prime_jobspec with exclusive set to whatever was passed into it. I think that's generally how the resource graph traversal works, right?

zekemorton (commented):
Overall looks really good. I think there is an issue with the flexible traverser (see the note below), but aside from that just a few nit picks.

grondo:
@milroy See https://github.com/flux-framework/flux-core/pull/7402.

Once this is merged, you should be able to test using `cpu-affinity=dry-run` and `FLUX_HWLOC_XMLFILE` of another system without errors by also setting `FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM` in the environment of `flux start`. i.e. before you call `test_under_flux`.

Unfortunately, on current versions of Flux the test will fail as you've noted. One way to check if Flux supports `FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM` is to check if the core count is expected when using the XML file, so you could add something like this to the top of the test (before `test_under_flux`):

```sh
export FLUX_HWLOC_XMLFILE="${hwloc_basepath}/001N/exclusive/01-brokers-sierra2/0.xml"
export FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM=1

NCORES_EXPECTED=44
NCORES=$(flux start flux resource list -no {ncores})
if test $NCORES -ne $NCORES_EXPECTED; then
    test_debug "echo expected $NCORES_EXPECTED cores, got $NCORES"
    skip_all='skipping tests since FLUX_HWLOC_XMLFILE_NOT_THISSYTEM unsupported'
    test_done
fi 
```

milroy:
> Once this is merged, you should be able to test using cpu-affinity=dry-run and FLUX_HWLOC_XMLFILE of another system without errors by also setting FLUX_HWLOC_XMLFILE_NOT_THISSYSTEM in the environment of flux start. i.e. before you call test_under_flux.

I see that this PR was already merged, that was fast!

> so you could add something like this to the top of the test (before test_under_flux):

Thank you!

PR #1409, opened 2025-09-30
mergify: disable temporary PR branches

grondo:
Problem: Mergify is creating temporary PR branches and merging the main branch (plus possibly other PRs). The merge commit causes the pr-validator to fail, plus in the case of an up-to-date original PR, the checks are redundant.

This is occurring because the maximum parallel checks default was updated to 5 by mergify:

https://changelog.mergify.com/changelog/maxparallelchecks-is-now-5-by-default

Explicitly set max_parallel_checks to 1 to avoid the temporary branches. Also set batch_size to 1 in queue_rules for good measure, since we currently never want to batch up PRs for merging.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,7 +1,14 @@
 queue_rules:
   - name: default
+    batch_size: 1
     update_method: rebase
     merge_method: merge
+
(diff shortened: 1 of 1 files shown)


garlick (approved):
:+1:

PR #1412, opened 2025-10-08
Fix segfault when loading multiple subsystems

milroy:
Loading a graph and selecting multiple subsystems causes a segfault. The cause is accessing a `dfu_impl_t` member map that is a `nullptr`. The map isn't used anywhere else, and the graph metadata already contains the same map.

This small PR removes the vestigial map from `dfu_impl_t` and `find` the aux subsystem root from the graph metadata. It also adds a simple test for loading a graph and configuring both containment and power subsystems.

Note that this PR does not enable subsystem match functionality.


trws (approved):
Good catch.

PR #1417, opened 2025-11-11
ci: update typo checker version

dupgit:
Problem: The typo checker has more typos in their dictionary that we are missing.

Update the typo checker version to v1.39.0.


dupgit:
Thanks :-)

Issue #1414, opened 2025-10-21
Tracking issue for shrinking JGF


trws:
After a discussion between @milroy, @jameshcorbett  and I today this is the running plan for how we're going to shrink JGF and get rabbit scheduling working with JGF on larger systems.

1. [ ] @jameshcorbett Make it so that a subinstance can use R for loading while the parent is using RV1 with a sched key
2. [ ] @trws Add a versioning scheme and a way to identify partial and complete JGF
3. [ ] (deps: 1) @jameshcorbett Add an option to the JGF writer to only write out the root exclusive vertex of any exclusively allocated subtrees. Quick test on tuo's data showed this to reduce a full system job JGF from 21mb to 700k (uncompressed).
4. [ ] @milroy Work on template option for config so we can describe a node or a rack and re-use it to describe a system rather than having to write out every individual resource
5. [ ] all (deps: 3, 4) combine 3 and 4 to write out a templated version of the partial JGF so a child instance can use it to load as a full JGF instance
6. [ ] @trws factor out separate graph builder objects in readers to ensure we can support graph build time computations
7. [ ] @trws based on what we need for 3, prioritize adding easy traversal interface for containment (half done, just need to finish)

After this, we should also consider if we want to either:
- Switch to only storing an array of unique ids in the sched key. This is quite small/effiicient but would take a lot of work to rework the reader(s)
- Automatically apply templating where possible by calculating structural hashes of components of the containment tree. More work to implement, but potentially good compression and doesn't leave out information.

milroy:
Claude and I generated a prototype in Python for structural compression to address item 4 above. I'm pasting it here for discussion: https://gist.github.com/milroy/069f042e14f2fa21b6aab0f977b1453e

PR #1418, opened 2025-11-12
resource: add `rv1exec_force` load format

jameshcorbett:
Problem: in some cases it can be useful to make the Fluxion resource module
initialize itself from rv1exec rather than JGF, even when the `scheduling`
key of R is present. Currently however, if the `scheduling` key is populated,
the resource module initializes itself with JGF.

Add a load-format, `rv1exec_force`, that tells the module to ignore the
`scheduling` key.

Checks the first box in #1414. There would need to be some configuration done in the system instance so that all its subinstances load `sched-fluxion-resource` with the `load-format=rv1exec_force` option. After chatting with @grondo, it sounds like:

> The best way to do this might be to have ansible drop an extension in to /etc/flux/modprobe/rc1.d/ that forces the option via:
> ```
> def setup(context):
>     context.setopt("sched-fluxion-resource", "OPT=VAL")
> ```

Using `context.attr_get("instance-level", 0) > 0` to check whether or not the script was running against the system instance.

### DIFF
--- a/t/t3039-resource-force-load-format.t
+++ b/t/t3039-resource-force-load-format.t
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+test_description='Test the rv1exec_force load format'
+
+. $(dirname $0)/sharness.sh
+
+query="../../resource/utilities/resource-query"
+jobspec="${SHARNESS_TEST_SRCDIR}/data/resource/jobspecs/load_format/t3039.yaml"
+
+
+test_expect_success 'Generate rv1 with .scheduling key' '
+    flux R encode --hosts=compute[1-10] --cores=0-7 --gpu=0-1 | jq . > R.json &&
+    cat R.json | flux ion-R encode > R_JGF.json &&
+    jq -S "del(.execution.starttime, .execution.expiration)" R.json > R.norm.json
+'
+
+test_expect_success 'resource_query can be loaded with rv1exec_force' '
+    cat > match_jobspec_cmd <<-EOF &&
+    match allocate ${jobspec}
+    quit
+EOF
+    ${query} -L R_JGF.json -f rv1exec_force -F rv1_nosched -t R1.out -P high < match_jobspec_cmd &&
+    head -n1 R1.out | jq -S "del(.execution.starttime, .execution.expiration)" > match1.json &&
+    jq -S "del(.execution.starttime, .execution.expiration)" R.json > R.norm.json &&
+    test_cmp match1.json R.norm.json
+'
+
+test_expect_success 'resource_query can be loaded with rv1exec' '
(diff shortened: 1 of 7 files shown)


t/t3039-resource-force-load-format.t
```diff
@@ -0,0 +1,109 @@
...
+# a comment
+attributes:
+  system:
+    duration: 3600
+tasks:
+  - command: [ \"app\" ]
+    slot: default
+    count:
+      per_slot: 1
+" > jobspec.json
```
milroy: This method of creating the jobspec as part of the test via `echo` is uncommon in the testsuite. Is there a reason you aren't doing something like this: https://github.com/flux-framework/flux-sched/blob/563a4c14de62beb4af72165c60ca9a85df22d2b8/t/t3001-resource-basic.t#L18-L24

t/t3039-resource-force-load-format.t
```diff
@@ -0,0 +1,109 @@
...
+test_expect_success 'resource_query can be loaded with rv1exec_force' '
+    cat > match_jobspec_cmd <<-EOF &&
+    match allocate jobspec.json
+    quit
+EOF
+    ${query} -L R_JGF.json -f rv1exec_force -F rv1_nosched -t R1.out -P high < match_jobspec_cmd &&
+    head -n1 R1.out | jq -S "del(.execution.starttime, .execution.expiration)" > match1.json &&
+    jq -S "del(.execution.starttime, .execution.expiration)" R.json > R.norm.json &&
+    test_cmp match1.json R.norm.json
+'
```
milroy, line 26: The structure here is similar to what's done in the `3000`-series tests.

t/t3039-resource-force-load-format.t
```diff
@@ -0,0 +1,109 @@
...
+flux module remove -f sched-fluxion-qmanager &&
+flux module remove -f sched-fluxion-resource &&
+flux kvs get resource.R | jq ".scheduling = \"nonsense\"" > local_R &&
+flux kvs put resource.R="$(cat local_R)" &&
+flux module reload resource &&
+flux module load sched-fluxion-resource load-format="${1}" &&
+flux module load sched-fluxion-qmanager
+' > load_nonsense_scheduling_key.sh
+
+chmod +x load_nonsense_scheduling_key.sh
```
milroy: This structure is similar to what's found in the `1000`-series tests.

milroy (commented):
This PR is good overall, but the tests need some restructuring. I recommend extracting components of `t3039` that depend on `sched-fluion-resource` into a small `1000`-series test.

trws (approved):
Looks good to me.

PR #1419, opened 2025-11-20
resource: JGF writer variant that only writes non-exclusive vertices

jameshcorbett:
As described in #1414, there needs to be a way to 

> Add an option to the JGF writer to only write out the root exclusive vertex of any exclusively allocated subtree

This PR adds the writer. The PR still requires a reader that can read in these partial JGFs for jobs and "fill in" missing vertices.

Related to #1418.

### DIFF
--- a/t/t3040-jgf-shorthand-match-format.t
+++ b/t/t3040-jgf-shorthand-match-format.t
@@ -1,0 +1,106 @@
+#!/bin/sh
+
+test_description='Test the jgf_shorthand match format'
+
+. $(dirname $0)/sharness.sh
+
+jobspec_dir="${SHARNESS_TEST_SRCDIR}/data/resource/jobspecs/match_format"
+exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/match_format"
+query="../../resource/utilities/resource-query"
+
+test_expect_success 'Generate rv1 with .scheduling key' '
+    flux R encode --hosts=compute[1-10] --cores=0-7 --gpu=0-1 | jq . > R.json &&
+    cat R.json | flux ion-R encode > R_JGF.json &&
+    jq -S "del(.execution.starttime, .execution.expiration)" R.json > R.norm.json
+'
+
+test_expect_success 'resource_query can be loaded with jgf' '
+    cat > match_jobspec_cmd <<-EOF &&
+    match allocate ${jobspec_dir}/t3040.yaml
+    quit
+EOF
+    jq -S .scheduling R_JGF.json > JGF.json &&
+    ${query} -L JGF.json -f jgf -F rv1_nosched -t R1.out -P high < match_jobspec_cmd &&
+    head -n1 R1.out | jq -S "del(.execution.starttime, .execution.expiration)" > r_match.json &&
+    # the jobspec requests all resources in the graph, so the job R should be the same as
+    # the R for the graph
+    test_cmp r_match.json R.norm.json
+'
+
+test_expect_success 'jgf_shorthand does not emit cores or gpus with policy=lonodex' '
+    ${query} -L JGF.json -f jgf -F jgf_shorthand -t jgf1.out -P lonodex < match_jobspec_cmd &&
+    head -n1 jgf1.out | jq -S . > match1.json &&
+    test_must_fail grep core match1.json > /dev/null &&
+    test_must_fail grep gpu match1.json > /dev/null &&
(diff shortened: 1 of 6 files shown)


resource/traversers/dfu_impl_update.cpp
```diff
@@ -251,15 +251,18 @@ int dfu_impl_t::upd_sched (vtx_t u,
                            const jobmeta_t &jobmeta,
                            bool full,
                            const std::map<resource_type_t, int64_t> &dfu,
-                           std::map<resource_type_t, int64_t> &to_parent)
+                           std::map<resource_type_t, int64_t> &to_parent,
+                           bool excl_parent)
```
milroy, line 255: I'm a bit confused why `excl_parent` is necessary. Isn't `excl` sufficient? Does the value of `excl_parent` get modified?

jameshcorbett: The root vertex of an exclusive subtree is itself exclusive, but that vertex should still be emitted under this scheme. So just checking `excl` is not sufficient.

To make things more concrete, imagine a simple graph: a cluster with a bunch of nodes, and each node with a bunch of cores. If a user asks for some number of nodes exclusively, the writer should emit the cluster and node vertices and not the core vertices. If the writer only checked `excl`, it would only emit the cluster vertex, which is incorrect.

The way `excl_parent` works in this PR is that it starts out as `false` regardless of the value of `excl`, but for a given vertex `u` and its value of `excl` (call that value `B`), all of the children of `u` are passed `B` for their value of `excl_parent`. 

Does that make sense? I'll add some comments to the code and force-push.

resource/writers/match_writers.hpp
```diff
@@ -57,6 +57,14 @@ class match_writers_t {
     int compress_hosts (const std::vector<std::string> &hosts,
                         const char *hostlist_init,
                         char **hostlist);
+
+    /* Return a boolean indicating whether or not the writer should be invoked
+     * on vertices that form non-root parts of exclusive subtrees.
+     */
+    virtual bool emit_exclusive_subtrees (void)
```
milroy: Style comment: in other parts of Fluxion, the explicit `void` parameter isn't used. It looks like style preference in C++ is to use an empty parameter list: https://en.cppreference.com/w/cpp/language/virtual.html

resource/utilities/resource-query.cpp
```diff
@@ -146,7 +146,7 @@ To see cli commands, type in "help" in the cli: i.e.,
                                 (perf_class must be set using set-property).
                 (default=first).
 
-    -F, --match-format=<simple|pretty_simple|jgf|rlite|rv1|rv1_nosched>
+    -F, --match-format=<simple|pretty_simple|jgf|jgf_shorthand|rlite|rv1|rv1_nosched>
```
milroy, line 149: I like the name `jgf_shorthand`!

resource/writers/match_writers.hpp
```diff
@@ -174,6 +174,24 @@ class jgf_match_writers_t : public match_writers_t {
     json_t *m_eout = NULL;
 };
 
+/*! JGF_shorthand match writers class for a matched resource set
+ */
+class jgf_shorthand_match_writers_t : public jgf_match_writers_t {
+   public:
+    jgf_shorthand_match_writers_t () : jgf_match_writers_t ()
```
milroy: I think the modern best practice is to explicitly default the constructor:
```suggestion
    jgf_shorthand_match_writers_t () = default;
```
This change may require explicitly `default`ing the `jgf_match_writers_t` constructor.

resource/traversers/dfu_impl_update.cpp
milroy: The explanation helps, as do the code comments.

t/t3040-jgf-shorthand-match-format.t
```diff
@@ -0,0 +1,137 @@
...
+    count: 1
+attributes:
+  system:
+    duration: 3600
+tasks:
+  - command: [ \"app\" ]
+    slot: default
+    count:
+      per_slot: 1
+" > jobspec_only_node.yaml
```
milroy: Please take a look at https://github.com/flux-framework/flux-sched/blob/master/t/t3001-resource-basic.t for examples of `resource-query`-based tests.

Those tests add expected output and compare the generated output to expected, which would be easier to review. The downside is the ballooning size of the repo, but that hasn't stopped us in the past...

PR #1420, opened 2025-11-22
bindings: flux-sched python

vsoch:
This pull request will add flux-sched Python bindings for the resource API. We want to do this so we can use flux-sched in agentic contexts. I've reproduced the reapi cli, and added the start of the reapi module, but without complete tests, etc., because it's not of interest or priority now (can be worked on when needed). For this PR, I've added:

- Python bindings that are based on Cython
- A modern setup (e.g., pyproject.toml with python3 -m build over setup.py
- Tests with pytest that do match allocate, cancel, and info (and mimic what we test for the Golang bindings)
- Documentation in the README.md
- An entire metadata / package that uploads to pypi.

I am not building / testing across matrices because it's annoying to install Cython in each one. Instead I am using the .devcontainer as a CI environment. I will update the title to remove the WIP when that is working fully.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,40 @@ jobs:
     - name: format and linting checks
       run: pre-commit run --all-files --show-diff-on-failure
 
+  flux-sched-python:
+    name: flux-sched python
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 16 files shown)


vsoch:
#$#%# I just ran pre-commit on the entire flux-sched codebase...

ug.

jameshcorbett:
@vsoch do you have thoughts on how your PR compares to #1518 ? How would you like to proceed?

PR #1421, opened 2025-12-01
Fix or slot for first match policies 

zekemorton:
This MR introduces two bug fixes for the flexible traverser. 

The first bug fix was identified when developing xor slots. This issue causes the traverser to find matches that aren't necessarily a correct match for the job spec. This is because prime job spec takes minimum of potential counts and cannot be solely relied on to determine matches. For one of the optimizations in select_or_config, we rely on that to determine the number of matches. In the case that there are no matches, it might make an illogical match. This MR resolves this issues by calculating the number of matches and setting that value for the given job spec. This case would not normally be an issues for regular or slots, but it did come up for xor slots and it might be relevant with first match policy. 

The second bug fix is directly related to first match policy with or slots. This policy would have some differing results that would lead to either not enough resources being matched or no matches being found. This was cause by how the first match policy would prune the search after enough resources were found, and how the or slots determined what resources to search for. Or slots only use the type of resource when creating a union of or slots resources, so if that union contains a lower count than another or slot option, then a first match policy would never encounter enough recourses to match that high count option. This MR resolves this issue by ensuring that the resources union contains counts for the highest option of that resources type.


resource/traversers/dfu_flexible.cpp
```diff
@@ -289,16 +288,27 @@ int dfu_flexible_t::dom_slot (const jobmeta_t &meta,
...
+            auto it = pre_union.find (r.type);
+            if (it == pre_union.end ()) {
                 resource_types[r.type] = 0;
-                slot_resource_union.push_back (r);
+                pre_union.emplace (r.type, r);
+            } else {
+                // We need the maximum value for count here
+                // otherwise first match policies won't work
+                if (it->second.count.min < r.count.min)
+                    pre_union.insert_or_assign (r.type, r);
```
milroy, line 303: What's the advantage of using `insert_or_assign ()` instead of the `[]` operator?

milroy: Answered by @zekemorton and @trws in the Fluxion meeting: `r.type` doesn't have a default constructor and `insert_or_assign` requires objects with them.

milroy (approved):
@zekemorton and I have discussed these changes a few times and the PR looks good to me!

@trws if you have time to take a look, the added review would help but I don't think it's necessary.

PR #1422, opened 2025-12-03
readers: add jgf reader that can skip exclusive subtrees

jameshcorbett:
This PR addresses a bullet in #1414:

> Add an option to the JGF writer to only write out the root exclusive vertex of any exclusively allocated subtrees. Quick test on tuo's data showed this to reduce a full system job JGF from 21mb to 700k (uncompressed).

Stacked on top of #1419.

### DIFF
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -967,18 +967,40 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
     int rc = -1;
     unsigned int i = 0;
     fetch_helper_t fetcher;
+    std::vector<fetch_helper_t> additional_vertices;
+    std::map<std::string, vmap_val_t> empty_vmap{};
 
     for (i = 0; i < json_array_size (nodes); i++) {
         fetcher.scrub ();
+        additional_vertices.clear ();
         if ((rc = unpack_vtx (json_array_get (nodes, i), fetcher)) != 0)
             goto done;
         if ((rc = update_vtx (g, m, vmap, fetcher, update_data)) != 0)
             goto done;
+        if (fetch_additional_vertices (g, m, empty_vmap, fetcher, additional_vertices) != 0)
+            goto done;
+        for (auto &fetcher : additional_vertices) {
+            std::string vertex_id = std::to_string (fetcher.uniq_id);
+            fetcher.vertex_id = vertex_id.c_str ();
+            if ((rc = update_vtx (g, m, vmap, fetcher, update_data)) != 0) {
+                goto done;
+            }
+        }
     }
     rc = 0;
 
 done:
     return rc;
+}
+
+int resource_reader_jgf_t::fetch_additional_vertices (
+    resource_graph_t &g,
+    resource_graph_metadata_t &m,
+    std::map<std::string, vmap_val_t> &vmap,
+    fetch_helper_t &fetcher,
+    std::vector<fetch_helper_t> &additional_vertices)
+{
+    return 0;
 }
 
 int resource_reader_jgf_t::unpack_edge (json_t *element,
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -27,46 +27,6 @@ extern "C" {
 using namespace Flux;
 using namespace Flux::resource_model;
 
-class fetch_remap_support_t {
-   public:
-    int64_t get_remapped_id () const;
-    int64_t get_remapped_rank () const;
-    const std::string &get_remapped_name () const;
-    void set_remapped_id (int64_t i);
-    void set_remapped_rank (int64_t r);
-    void set_remapped_name (const std::string &n);
-    bool is_name_remapped () const;
-    bool is_id_remapped () const;
-    bool is_rank_remapped () const;
-    void clear ();
-
-   private:
-    int64_t m_remapped_id = -1;
-    int64_t m_remapped_rank = -1;
-    std::string m_remapped_name = "";
-};
-
-struct fetch_helper_t : public fetch_remap_support_t {
-    const char *get_proper_name () const;
-    int64_t get_proper_id () const;
-    int64_t get_proper_rank () const;
-    void scrub ();
-
-    int64_t id = -2;
-    int64_t rank = -1;
-    int64_t size = -1;
-    int64_t uniq_id = -1;
-    int exclusive = -1;
-    resource_pool_t::status_t status = resource_pool_t::status_t::UP;
-    const char *type = NULL;
-    std::string name;
-    const char *unit = NULL;
-    const char *basename = NULL;
-    const char *vertex_id = NULL;
-    std::map<std::string, std::string> properties;
-    std::map<subsystem_t, std::string> paths;
-};
-
 int64_t fetch_remap_support_t::get_remapped_id () const
 {
     return m_remapped_id;
--- a/resource/readers/resource_reader_jgf.hpp
+++ b/resource/readers/resource_reader_jgf.hpp
@@ -242,6 +242,11 @@ class resource_reader_jgf_t : public resource_reader_base_t {
                          std::map<std::string, vmap_val_t> &vmap,
                          json_t *nodes,
                          jgf_updater_data &updater_data);
+    virtual int fetch_additional_vertices (resource_graph_t &g,
+                                           resource_graph_metadata_t &m,
+                                           std::map<std::string, vmap_val_t> &vmap,
+                                           fetch_helper_t &fetcher,
+                                           std::vector<fetch_helper_t> &additional_vertices);
     int unpack_edge (json_t *element,
                      std::map<std::string, vmap_val_t> &vmap,
                      std::string &source,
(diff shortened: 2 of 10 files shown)


resource/readers/resource_reader_jgf.hpp
```diff
@@ -131,7 +170,7 @@ class resource_reader_jgf_t : public resource_reader_base_t {
      */
     virtual bool is_allowlist_supported ();
 
-   private:
+   protected:
```
trws, line 173: Did you mean to change all of these to protected, or just fetch_additional_vertices?

resource/readers/resource_reader_jgf_shorthand.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace resource_model {
+
+/*! JGF shorthand resource reader class.
+ */
+class resource_reader_jgf_shorthand_t : public resource_reader_jgf_t {
+   public:
+    virtual ~resource_reader_jgf_shorthand_t ();
+
+   private:
+    virtual int fetch_additional_vertices (
```
trws: This version is private with the base being marked protected, I'm vaguely surprised it compiles like this, but the privacy needs to match.  On a more stylistic note, the prefix `virtual` is unnecessary, especially with the `override` suffix and may fail some lints since this isn't introducing a new virtual.

resource/readers/resource_reader_jgf.hpp
jameshcorbett: I did mean to do this, but I'm not sure I stand by it. `resource_reader_jgf_shorthand_t` uses two methods from `resource_reader_jgf_t` that would otherwise have been private and inaccessible: `find_vtx` and `apply_defaults`. But I guess I could just make those two methods protected instead! Especially if the preference in C++ is to restrict access as much as possible.

resource/readers/resource_reader_jgf_shorthand.hpp
jameshcorbett: Fixed!

resource/readers/resource_reader_jgf.hpp
milroy: > But I guess I could just make those two methods protected instead! Especially if the preference in C++ is to restrict access as much as possible.

Yes, I think you should make just the two methods in question to `protected` and leave the rest as `private`.

resource/readers/resource_reader_jgf_shorthand.hpp
```diff
@@ -0,0 +1,37 @@
+/*****************************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
```
milroy: Nit: it's 2026, which reflects on how long it's taken me to get to this PR.

resource/readers/resource_reader_jgf_shorthand.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+#ifndef RESOURCE_READER_JGF_SHORTHAND_HPP
+#define RESOURCE_READER_JGF_SHORTHAND_HPP
+
+#include <string>
+#include <unordered_set>
+#include <jansson.h>
+#include "resource/schema/resource_graph.hpp"
```
milroy: You only need `"resource/readers/resource_reader_jgf.hpp"`:

```suggestion
```

resource/readers/resource_reader_jgf_shorthand.cpp
```diff
@@ -0,0 +1,22 @@
...
+}
+
+#include <map>
+#include <unordered_set>
+#include <unistd.h>
+#include <regex>
+#include <jansson.h>
+#include "resource/readers/resource_reader_jgf_shorthand.hpp"
+#include "resource/store/resource_graph_store.hpp"
+#include "resource/planner/c/planner.h"
```
milroy: You only need `"resource/readers/resource_reader_jgf_shorthand.hpp"`:

```suggestion
#include "resource/readers/resource_reader_jgf_shorthand.hpp"
```

resource/readers/resource_reader_jgf.cpp
```diff
@@ -967,20 +967,42 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
...
         if ((rc = unpack_vtx (json_array_get (nodes, i), fetcher)) != 0)
             goto done;
         if ((rc = update_vtx (g, m, vmap, fetcher, update_data)) != 0)
             goto done;
+        if (fetch_additional_vertices (g, m, empty_vmap, fetcher, additional_vertices) != 0)
+            goto done;
+        for (auto it = additional_vertices.begin (); it != additional_vertices.end (); ++it) {
+            std::string vertex_id = std::to_string (it->uniq_id);
+            it->vertex_id = vertex_id.c_str ();
+            if ((rc = update_vtx (g, m, vmap, *it, update_data)) != 0) {
```
milroy: More idiomatic, modern C++:

```suggestion
        for (auto &it : additional_vertices) {
            std::string vertex_id = std::to_string (it.uniq_id);
            it.vertex_id = vertex_id.c_str ();
            if ((rc = update_vtx (g, m, vmap, it, update_data)) != 0) {
```

t/t3041-jgf-shorthand-load-format.t
```diff
@@ -0,0 +1,81 @@
...
+            count: 2
+attributes:
+  system:
+    duration: 3600
+tasks:
+  - command: [ \"app\" ]
+    slot: default
+    count:
+      per_slot: 1
+" > jobspec.yaml
```
milroy: Similar comment to the ones I made on #1419 and #1418: please use, e.g., `jobspec_dir`.

milroy (commented):
This almost ready to merge. Good work!

resource/readers/resource_reader_jgf_shorthand.hpp
trws: Worth fixing for this one for consistency, but since the date requirement has finally become a thing of the past, we might want to re-evaluate how we mark all these files at some point.

t/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ set(ALL_TESTS
   t3038-resource-flexible.t
   t3039-resource-force-load-format.t
   t3040-jgf-shorthand-match-format.t
+  t3041-jgf-shorthand-load-format.t
```
milroy, line 76: After a bit more thought, it may be better to combine these into a single `t3040-jgf-shorthand.t`. @trws, what do you think?

t/data/resource/jobspecs/load_format/t3041.yaml
```diff
@@ -0,0 +1,21 @@
+version: 9999
```
milroy, line 1: Isn't this identical to `t/data/resource/jobspecs/load_format/t3039.yaml`?

milroy (commented):
I just have two remaining questions as to whether `t/data/resource/jobspecs/load_format/t3041.yaml` is needed, and whether the tests in `t3041` can be merged into `t3040` (with appropriate renaming).

t/data/resource/jobspecs/load_format/t3041.yaml
jameshcorbett: It is, yeah. I could combine them, but my inclination is to make them different--I'll force-push with that change.

t/CMakeLists.txt
jameshcorbett: Will be interested to see what Tom says, I don't feel strongly about it.

resource/readers/resource_reader_jgf_shorthand.cpp
```diff
@@ -0,0 +1,74 @@
...
+    if (v == boost::graph_traits<resource_graph_t>::null_vertex ()) {
+        return -1;
+    }
+
+    for (boost::tie (ei, ei_end) = boost::out_edges (v, g); ei != ei_end; ++ei) {
+        if (g[*ei].subsystem != containment_sub)
+            continue;
+        target = boost::target (*ei, g);
+
+        fetch_helper_t vertex_copy;
```
trws, line 53: You didn't add this @jameshcorbett, so I'm not asking for it to be fixed here, but we should seriously consider factoring out the vertex copy setup with this so we can just assign a vertex into it rather than having to get all this right in the future.

t/CMakeLists.txt
trws: I'm ok with them split.  If there were shared helpers or something that saved us duplication to merge them I'd want to do that but it looks like they're reasonably standalone.

trws (approved):
Looks good to me! If you're feeling up to it, a subsequent little refactor PR to make those fetch helpers a bit less C-like would be awesome but no need to block this for something we already had.

resource/readers/resource_reader_jgf_shorthand.cpp
jameshcorbett: I started to look at this, but this appears to be the only place where this vertex copy logic is used. I'd also be a bit worried because this logic grabs pointers to externally-managed data (with lines like `vertex_copy.type = g[target].type.c_str ()` It's fine in this context because the external data will be around at least as long as the pointer, but I'd be a bit worried that if we made a function or method for the copy logic, the fact that external data is being referenced would be obscured, making future uses of the function/method dangerous.

PR #1429, opened 2026-02-09
testsuite: set log-stderr-level on command line

garlick:
Problem: the log-stderr-level attribute is adjusted with flux-setattr(1) in several tests, but that interface may change as proposed in flux-framework/flux-core#7370.

Set it on the broker command line instead, by adding -Slog-stderr-level=N to test_under_flux.


jameshcorbett (approved):
👍

PR #1430, opened 2026-02-10
CMake: match hwloc version requirement to README

sam-maloney:
Problem: #1272 changed hwloc min version in README in but not CMakeLists.txt (Specifically commit https://github.com/flux-framework/flux-sched/commit/15c3c910972ab752b843908707fc39ef0143dc8f)

Solution: update CMakeLists.txt to match

This is partly a question of whether this was **intended** to be left as a _soft_ requirement? Or was it just a moot point since `hwloc` 2.0 came out in 2018?

The follow up question would be whether this requirement would apply across other Flux projects? I had asked an earlier question about dependency version mismatches here: https://github.com/flux-framework/flux-core/issues/7295


jameshcorbett (approved):
This looks good to me but I'm not sure if @trws intended it to be this way.

PR #1431, opened 2026-02-17
flexible traverser: add support for xor slots 

zekemorton:
This merge request adds support for a new logical slot type, xor_slot, in the flexible scheduler. An xor_slot allows a jobspec to express multiple alternative slot configurations that are acceptable for a workload, with the guarantee that only one of those alternatives will be selected for a given match. 

Key implementation details:

- Introduces a new resource type xor_slot_rt alongside slot_rt in the resource model schema.
- Extends prime_jobspec so that xor_slot aggregates are treated like slot aggregates when computing minimum values.
- Implements a new dfu_flexible_t::select override that:
  - Uses a new helper split_xor_slots to expand a jobspec containing xor_slot entries into multiple candidate ResourceList variants.
  - Iteratively attempts scheduling for each candidate resource variant via dom_dfv and resolve_graph.
  - On the first successful match, updates job spec resources to the chosen variant and returns success; if all options fail, returns an error.
- split_xor_slots performs a recursive depth first traversal of the jobspec resources, under the assumption that all xor_slot entries exist at the same level:
  - Separates xor_slot resources from the rest.
  - Recursively handles resources that are ancestors of xor_slot entries by cloning parents with each child option.
  - Flattens the resulting combinations and converts xor_slot resources to standard slot type in each expanded alternative.

Some ideas for improvement:

- Move most of the common logic for select into a helper functions that is called by both dfu_impl_t::select and dfu_flexible_t::select 
- add support for nested xor slots 
- move split xor slot outside of select so that it only needs to be called once for multiple schedule points.


resource/traversers/dfu_flexible.cpp
```diff
@@ -189,6 +191,89 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
...
+                option.type = slot_rt;
+                option.with = child_variant;
+                xor_options.push_back (option);
+            }
+
+            // Check if xor_options collection exceeds the limit
+            if (xor_options.size () > detail::SYSTEM_MAX_XOR_EXPANSION) {
+                errno = EOVERFLOW;
+                return {};
+            }
```
milroy: Claude and I agree that to avoid overshooting `SYSTEM_MAX_XOR_EXPANSION`, this logic should be pulled up into the `for` loop above. It's less important here than for the nested `for` loops below.

resource/traversers/dfu_flexible.cpp
```diff
@@ -189,6 +191,89 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
...
+                next.push_back (expanded);
+                next_variants.push_back (std::move (next));
+            }
+        }
+
+        // Check if expansion exceeds the limit
+        if (next_variants.size () > detail::SYSTEM_MAX_XOR_EXPANSION) {
+            errno = EOVERFLOW;
+            return {};
+        }
```
milroy: Same comment as above for overshoot, but this is the more important check given the nested `for` loop.

resource/traversers/dfu_flexible.cpp
```diff
@@ -189,6 +191,89 @@ void dfu_flexible_t::prime_jobspec (std::vector<Resource> &resources,
...
+            next.push_back (option);
+            results.push_back (std::move (next));
+        }
+    }
+
+    // Check if final expansion exceeds the limit
+    if (results.size () > detail::SYSTEM_MAX_XOR_EXPANSION) {
+        errno = EOVERFLOW;
+        return {};
+    }
```
milroy: Same as above. Also consider making a small function to reduce code duplication.

t/t4016-match-flexible.t
```diff
@@ -38,6 +40,42 @@ test_expect_success 'match-allocate works with flexible jobspecs' '
...
+
+test_expect_success 'match-allocate works with single-option nested xor-slot jobspecs' '
+    flux ion-resource match allocate ${jobspec_dir}/test011.yaml
+'
+
+test_expect_success 'reloading flexible traverser with tiny graph and first policy works' '
+    remove_resource &&
+    load_resource \
+load-file=${tiny_grug} prune-filters=ALL:core \
+load-format=grug subsystems=containment policy=first traverser=flexible
```
milroy: Is the lack of indentation deliberate?

resource/utilities/README.md
```diff
@@ -714,4 +809,3 @@ traversal types.
 If you are interested in our earlier discussions on the different classes
 of matching problems, please refer to
 [this issue](https://github.com/flux-framework/flux-sched/issues/247#issuecomment-310551638)
```
milroy, line 819: Claude suggested a table to compare OR and XOR slots:

| Feature | OR-slots | XOR-slots |
|---------|----------|-----------|
| Expansion timing | During traversal | Before traversal |
| Selection algorithm | Dynamic programming (optimal) | First-match |
| Use case | Same-level alternatives | Different hierarchy prefixes |
| Nesting support | No | Yes |

milroy (commented):
This PR is looking really good. Claude and I found a few minor issues that we already discussed. After they're fixed this will be good to merge.

resource/traversers/dfu_impl.cpp
```diff
@@ -1275,19 +1278,24 @@ int dfu_impl_t::select (Jobspec::Jobspec &j, vtx_t root, jobmeta_t &meta, bool e
...
         m_graph_db->metadata.v_rt_edges[dom].set_for_trav_update (needs, x_in, m_sequence_number);
     }
     return rc;
 }
 
+int dfu_impl_t::select (Jobspec::Jobspec &j, vtx_t root, jobmeta_t &meta, bool excl)
+{
+    return select_resources (root, j.resources, meta, excl);
+}
+
```
trws: Why move this into a new function rather than calling `dfu_impl_t::select` from the flexible select?  Is it used somewhere else?

zekemorton: I think I tried and I think the issue was that if the flexible traverser calls `dfu_impl_t::select` then `dfu_impl_t::select` will call `dfu_impl_t::dom_dfv` instead of `dfu_flexible_t::dom_dfv`. This would drop support for the or slots and might have some other drawbacks.  

Its been a little while though, so I if that doesn't sound right, I can the try change and see what happens.

trws:
Assuming dom_dfv is virtual, it should call the most derived version from flexible. If it wasn’t at the time you tested that definitely would have happened though.

Get Outlook for iOS<https://aka.ms/o0ukef>

zekemorton:
@milroy and @trws I believe I addressed all of the above feedback. 

for Toms comment about the select resources helper function, you are correct that making dom_dfv virtual does resolve that issue, thanks for that tip. I also opted to change the input parameters of select() so that it takes in a resource list rather than a job spec because that made some of the logic a bit simpler for the xor slots. I am happy to make adjustments though

milroy (dismissed):
I've read the updated changes and this looks ready to merge. Really nice work!

jameshcorbett:
What is this PR waiting on @zekemorton ? Review from Dan or Tom? Could you rebase with master? It sounds like Dan was good with this so I could maybe give a look over and we could merge it

zekemorton:
I think it was waiting on a final approval from Tom.

Dan had mentioned that I could make an update to the RFC to include `or` and `xor` slots, but mentioned that could pretty easily be a stand alone PR

trws (approved):

PR #1432, opened 2026-02-24
writers: add rv1_shorthand match format

jameshcorbett:
Problem: there is no match writer that emits both R and JGF
shorthand.

Add one.

Also, there is no way to distinguish shorthand JGF from complete
JGF.

Add a metadata object with two fields: 'version' and 'complete.'
'Version' is currently always 1. 'complete' is always true for the
regular JGF writer and is false for the shorthand writer when it
omits vertices.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -965,13 +965,28 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx, json_t *resou
     struct idset *grow_set = NULL;
     json_t *r_lite = NULL;
     json_t *jgf = NULL;
+    const char *writer_uri = NULL;
+    bool jgf_shorthand = false;
     auto guard = resource_type_t::storage_t::open_for_scope ();
 
     if ((rc = unpack_resources (resources, &grow_set, &r_lite, &jgf, duration)) < 0) {
         flux_log_error (ctx->h, "%s: unpack_resources", __FUNCTION__);
         goto done;
     }
-    if (jgf && ctx->opts.get_opt ().get_load_format () != "rv1exec_force") {
+    if (jgf) {
+        json_unpack (jgf, "{s?s}", "writer", &writer_uri);
+    }
+    // only complete JGF is useful here; for shorthand JGF we have to fall
+    // back to rv1exec. Assume JGF is complete by default
+    if (writer_uri) {
+        if (std::string (writer_uri) == "fluxion:jgf_shorthand")
+            jgf_shorthand = true;
+        else {
+            flux_log (ctx->h, LOG_ERR, "%s: unrecognized reader URI %s", __FUNCTION__, writer_uri);
+            goto done;
+        }
+    }
+    if (jgf && ctx->opts.get_opt ().get_load_format () != "rv1exec_force" && !jgf_shorthand) {
         if (ctx->reader == nullptr && (rc = create_reader (ctx, "jgf")) < 0) {
             flux_log (ctx->h, LOG_ERR, "%s: can't create jgf reader", __FUNCTION__);
             goto done;
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1369,7 +1369,28 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
         }
         R_graph_fmt = jgf_str;
         free (jgf_str);
-        format = "jgf";
+        const char *writer_uri = NULL;
+        if (json_unpack (graph, "{s?s}", "writer", &writer_uri) < 0) {
+            errno = EINVAL;
+            flux_log (ctx->h, LOG_ERR, "%s: json_unpack", __FUNCTION__);
+            rc = -1;
+            goto freemem_out;
+        }
+        if (writer_uri) {
+            if (std::string (writer_uri) == "fluxion:jgf_shorthand")
+                format = "jgf_shorthand";
+            else {
+                flux_log (ctx->h,
+                          LOG_ERR,
+                          "%s: unrecognized reader URI '%s'",
+                          __FUNCTION__,
+                          writer_uri);
+                rc = -1;
+                goto freemem_out;
+            }
+        } else {  // if URI not present, assume JGF
+            format = "jgf";
+        }
     } else {
         // Use the rv1exec reader
         R_graph_fmt = R;
@@ -1477,6 +1498,16 @@ static int run (std::shared_ptr<resource_ctx_t> &ctx,
             flux_log (ctx->h,
                       LOG_ERR,
                       "%s: create JGF reader (id=%jd)",
+                      __FUNCTION__,
+                      static_cast<intmax_t> (jobid));
+            goto out;
+        }
+    } else if (format == "jgf_shorthand") {
+        if ((rd = create_resource_reader ("jgf_shorthand")) == nullptr) {
+            rc = -1;
+            flux_log (ctx->h,
+                      LOG_ERR,
+                      "%s: create jgf_shorthand reader (id=%jd)",
                       __FUNCTION__,
                       static_cast<intmax_t> (jobid));
             goto out;
--- a/resource/readers/resource_reader_base.hpp
+++ b/resource/readers/resource_reader_base.hpp
@@ -30,6 +30,7 @@ struct modify_data_t {
     std::unordered_map<int64_t, vtx_t> rank_to_root;
     std::unordered_map<int64_t, std::unordered_map<resource_type_t, int64_t>> rank_to_counts;
     std::unordered_map<resource_type_t, int64_t> type_to_count;
+    uint64_t token = 0;
 };
 
 /*!  Base resource reader class.
--- a/resource/readers/resource_reader_base.hpp
+++ b/resource/readers/resource_reader_base.hpp
@@ -30,7 +30,7 @@ struct modify_data_t {
     std::unordered_map<int64_t, vtx_t> rank_to_root;
     std::unordered_map<int64_t, std::unordered_map<resource_type_t, int64_t>> rank_to_counts;
     std::unordered_map<resource_type_t, int64_t> type_to_count;
-    uint64_t token = 0;
+    uint64_t sequence_number = 0;
 };
 
 /*!  Base resource reader class.
@@ -77,8 +77,8 @@ class resource_reader_base_t {
      * \param at     start time of this job
      * \param dur    duration of this job
      * \param rsv    true if this update is for a reservation.
-     * \param trav_token
-     *               token to be used by traverser
+     * \param sequence_number
+     *               traversal token (trav_token) to be used by traverser
      * \return       0 on success; non-zero integer on an error
      */
     virtual int update (resource_graph_t &g,
@@ -88,7 +88,7 @@ class resource_reader_base_t {
                         int64_t at,
                         uint64_t dur,
                         bool rsv,
-                        uint64_t trav_token) = 0;
+                        uint64_t sequence_number) = 0;
 
     /*! Partial cancellation of jobid based on R.
      *
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -1165,7 +1165,8 @@ int resource_reader_jgf_t::update_tgt_edge (resource_graph_t &g,
     if (!found) {
         errno = EINVAL;
         m_err_msg += __FUNCTION__;
-        m_err_msg += ": JGF edge not found in resource graph.\n";
+        m_err_msg += ": JGF edge from " + source + " to ";
+        m_err_msg += target + " not found in resource graph.\n";
         goto done;
     }
     g[e].idata.set_for_trav_update (vmap[target].needs, vmap[target].exclusive, token);
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -978,7 +978,12 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
                 goto done;
             }
         }
-        if (fetch_additional_edges (g, m, vmap, fetcher, additional_vertices, update_data.token)
+        if (fetch_additional_edges (g,
+                                    m,
+                                    vmap,
+                                    fetcher,
+                                    additional_vertices,
+                                    update_data.sequence_number)
             < 0) {
             goto done;
         }
@@ -1003,7 +1008,7 @@ int resource_reader_jgf_t::fetch_additional_edges (resource_graph_t &g,
                                                    std::map<std::string, vmap_val_t> &vmap,
                                                    fetch_helper_t &root,
                                                    std::vector<fetch_helper_t> &additional_vertices,
-                                                   uint64_t token)
+                                                   uint64_t sequence_number)
 {
     return 0;
 }
@@ -1126,7 +1131,7 @@ int resource_reader_jgf_t::update_src_edge (resource_graph_t &g,
                                             resource_graph_metadata_t &m,
                                             std::map<std::string, vmap_val_t> &vmap,
                                             std::string &source,
-                                            uint64_t token)
+                                            uint64_t sequence_number)
 {
     if (vmap[source].is_roots.empty ())
         return 0;
@@ -1134,7 +1139,7 @@ int resource_reader_jgf_t::update_src_edge (resource_graph_t &g,
     for (const auto &kv : vmap[source].is_roots)
         m.v_rt_edges[kv.first].set_for_trav_update (vmap[source].needs,
                                                     vmap[source].exclusive,
-                                                    token);
+                                                    sequence_number);
 
     // This way, when a root vertex appears in multiple JGF edges
     // we only update the virtual in-edge into the root only once.
@@ -1147,7 +1152,7 @@ int resource_reader_jgf_t::update_tgt_edge (resource_graph_t &g,
                                             std::map<std::string, vmap_val_t> &vmap,
                                             std::string &source,
                                             std::string &target,
-                                            uint64_t token)
+                                            uint64_t sequence_number)
 {
     edg_t e;
     int rc = -1;
@@ -1169,7 +1174,7 @@ int resource_reader_jgf_t::update_tgt_edge (resource_graph_t &g,
         m_err_msg += target + " not found in resource graph.\n";
         goto done;
     }
-    g[e].idata.set_for_trav_update (vmap[target].needs, vmap[target].exclusive, token);
+    g[e].idata.set_for_trav_update (vmap[target].needs, vmap[target].exclusive, sequence_number);
     rc = 0;
 
 done:
@@ -1180,7 +1185,7 @@ int resource_reader_jgf_t::update_edges (resource_graph_t &g,
                                          resource_graph_metadata_t &m,
                                          std::map<std::string, vmap_val_t> &vmap,
                                          json_t *edges,
-                                         uint64_t token,
+                                         uint64_t sequence_number,
                                          jgf_updater_data &update_data)
 {
     edg_t e;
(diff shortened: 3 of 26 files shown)


resource/traversers/dfu_impl.cpp
```diff
@@ -927,7 +927,8 @@ int dfu_impl_t::dom_find_dfv (std::shared_ptr<match_writers_t> &w,
     // emitting the vertex, since data could be leftover from previous
     // traversals where the vertex was matched but not emitted
     (*m_graph)[u].idata.ephemeral.check_and_clear_if_stale (m_best_k_cnt);
-    if ((rc = w->emit_vtx (level (), *m_graph, u, (*m_graph)[u].size, agfilter_data, true)) < 0) {
+    if ((rc = w->emit_vtx (level (), *m_graph, u, (*m_graph)[u].size, agfilter_data, true, false))
+        < 0) {
```
milroy: It would be helpful to provide a commented explanation for the two final booleans in the `emit_vtx` call.

resource/writers/match_writers.hpp
```diff
@@ -113,8 +118,12 @@ class jgf_match_writers_t : public match_writers_t {
...
                           unsigned int needs,
                           const std::map<std::string, std::string> &agfilter_data,
-                          bool exclusive) override;
-    virtual int emit_edg (const std::string &prefix, const resource_graph_t &g, const edg_t &e);
+                          bool exclusive,
+                          bool excl_parent) override;
+    virtual int emit_edg (const std::string &prefix,
+                          const resource_graph_t &g,
+                          const edg_t &e,
+                          bool excl_parent);
```
milroy: Fedora 40 Clang-18 is reporting warnings:
```bash
   /usr/src/resource/writers/match_writers.hpp:124:17: warning: 'emit_edg' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    124 |     virtual int emit_edg (const std::string &prefix,
```

milroy: There are a lot more similar warnings due to missing overrides in this header, but I don't think you need to correct those for this PR.

resource/writers/match_writers.cpp
```diff
@@ -500,6 +500,38 @@ jgf_shorthand_match_writers_t &jgf_shorthand_match_writers_t::operator= (
     return *this;
```
milroy, line 511: Did you intend for this change and the header to be in the commit on removing `emit_exclusive_subtrees ()`?

resource/writers/match_writers.cpp
```diff
@@ -527,11 +544,20 @@ int jgf_shorthand_match_writers_t::emit_edg (const std::string &prefix,
...
 }
 
+int jgf_shorthand_match_writers_t::get_metadata (json_t **metadata)
+{
+    *metadata = json_pack ("{s:b}", "complete", (int)m_complete);
+    if (!(*metadata))
+        return -1;
+    return 0;
+}
+
```
milroy: This approach seems complex. Can you use `json_object_get (metadata, "complete")`, where `metadata` is `json_t *`?

resource/modules/resource_match.cpp
```diff
@@ -965,13 +965,19 @@ static int grow_resource_db (std::shared_ptr<resource_ctx_t> &ctx, json_t *resou
...
         flux_log_error (ctx->h, "%s: unpack_resources", __FUNCTION__);
         goto done;
     }
-    if (jgf && ctx->opts.get_opt ().get_load_format () != "rv1exec_force") {
+    if (jgf) {
+        json_unpack (jgf, "{s:{s:{s:b}}}", "graph", "metadata", "complete", &jgf_complete);
+    }
+    // only complete JGF is useful here; for incomplete JGF we have to fall
+    // back to rv1exec. Assume JGF is complete by default
+    if (jgf && ctx->opts.get_opt ().get_load_format () != "rv1exec_force" && jgf_complete) {
```
milroy: What about:
```suggestion
    if (jgf && ctx->opts.get_opt ().get_load_format () == "jgf" && jgf_complete) {
```

t/t1029-resource-force-load-format.t
```diff
@@ -43,6 +60,38 @@ test_expect_success 'without rv1exec_force, job fails' '
...
+'
+
+test_expect_success 'scheduler can be reloaded with jgf_shorthand jobs' '
+    jobid=$(flux submit -t15s -N1 flux start ./load_fluxion.sh sleep 10) &&
+    reload_resource load-format=rv1exec_force match-format=rv1_shorthand &&
+    reload_qmanager &&
+    flux cancel ${jobid} &&
+    flux job info ${jobid} R | jq -e ".scheduling.graph.metadata.complete == false" &&
+    flux job info ${jobid} R | jq .scheduling.graph | test_must_fail grep core
+'
```
milroy, line 95: It's probably good to check the vertex allocation and pruning filter states with `find` as well.

milroy (commented):
I have a few comments and small items to fix, but this is already in a pretty good state.

resource/writers/match_writers.cpp
jameshcorbett: Hmmm I'm not sure I understand. I think your comment would make sense to me if the function were `json_unpack` but it's actually `json_pack`.

resource/modules/resource_match.cpp
jameshcorbett: That would make sense but that would definitely be a change to behavior. And I think we want the graph to initialize with JGF if the `scheduling` key is populated but the load format was not specified.

resource/writers/match_writers.cpp
jameshcorbett: I can't actually tell what you're referring to by "this change" so I don't know how to respond unfortunately.

milroy: Ugh, I intended for the comment to be associated with the correct commit. 

Sorry, what I meant was: did you intend to add the definition and declaration of `jgf_shorthand_match_writers_t::emit_vtx` and `jgf_shorthand_match_writers_t::emit_edg` to the "writers: remove `emit_exclusive_subtrees` method" commit? If so, please add a description of the addition in the commit message.

resource/writers/match_writers.cpp
milroy: I should have read the call site more closely and I confused myself. You can disregard this comment.

resource/writers/match_writers.cpp
jameshcorbett: Oh yeah, I guess I did mean those to be all part of the same commit, but I didn't explain that in the commit message. Good catch thanks! I'll fix.

resource/traversers/dfu_impl.cpp
jameshcorbett: Done!

resource/writers/match_writers.cpp
milroy: Ok, my previous question didn't make a lot of sense. Let me try again.

Can `jgf_shorthand_match_writers_t::get_metadata` get the value of `m_complete`, and then pack that value once in `jgf_match_writers_t::emit_json` like this:

```C++
   if (!(*o = json_pack ("{s:{s:o s:o s?b}}",
                          "graph",
                          "nodes",
                          m_vout,
                          "edges",
                          m_eout,
                          "complete",
                          complete))) {
```
That should also simplify the `json_unpack` in `parse_R` in resource_match.cpp.

t/t1029-resource-force-load-format.t
milroy: You can do this `flux ion-resource find --format=jgf agfilter=true` and `flux ion-resource find --format=jgf jobid-alloc=${jobid}`.

milroy (commented):
I have a (hopefully) more sensible question about `jgf_shorthand_match_writers_t::get_metadata`, and a suggestion for two test outputs to check. Other than those items this PR is ready to merge.

resource/writers/match_writers.cpp
milroy: Update: after a discussion with @jameshcorbett, I think the current implementation with the `"metadata"` key is good because that key is part of the JGF spec.

t/t1029-resource-force-load-format.t
jameshcorbett: Done! Thanks for that.

jameshcorbett:
Aside from [229e25d](https://github.com/flux-framework/flux-sched/pull/1432/commits/229e25d56c126f8c473f72f0e9df0d4405fe677f), there are no code changes from @milroy 's last review. However, I did add some new tests in the last commit. The tests fail because the agfilter output is incorrect for cores: it considers fewer cores to be allocated than are in fact allocated. This discrepancy appears to have something to do with the match format and the way it skips vertices, because if the test is re-run with `rv1` match format instead of `rv1_shorthand`, or if the job is modified to only request nodes non-exclusively, the agfilter checks pass.

milroy:
So far I can't reproduce the behavior in the new tests using `resource-query`. That said, I don't understand some of the test checks. Why do you expect `used:0` for `cores` after the "run a rabbit job" test?

Also, I should note that checking the agfilter counts before reload yields expected counts, so the reload results in the reduced and incorrect counts.

jameshcorbett:
> So far I can't reproduce the behavior in the new tests using `resource-query`. That said, I don't understand some of the test checks. Why do you expect `used:0` for `cores` after the "run a rabbit job" test?

Because after that test the system is idle I believe.

> Also, I should note that checking the agfilter counts before reload yields expected counts, so the reload results in the reduced and incorrect counts.

Oh good catch so it must be something about the reader!

garlick:
Quick thought on the identifying metadata in the `scheduling` key.   Maybe since the key is scheduler-specific, we should use a URI that names the scheduler that that wrote it, then use path or a query option to provide more information, like for this PR:
```json
"scheduler": {
    "writer":"fluxion:rv1_shorthand"
}
```

or maybe
```json
"scheduler": {
    "writer":"fluxion:rv1_shorthand&version=1"
```

Since fluxion was the first to use the `scheduling` key, we could update  RFC 20 (Rv1) to say that if `writer` is missing,  it is assumed to be set to `fluxion` .  Then update RFC  40 (Fluxion Resource Set Extension) to say if the path is missing, it is assumed to be full JGF.

jameshcorbett:
> Quick thought on the identifying metadata in the `scheduling` key. Maybe since the key is scheduler-specific, we should use a URI that names the scheduler that that wrote it, then use path or a query option to provide more information, like for this PR:
> 
> ```json
> "scheduler": {
>     "writer":"fluxion:rv1_shorthand"
> }
> ```
> 
> or maybe
> 
> ```json
> "scheduler": {
>     "writer":"fluxion:rv1_shorthand&version=1"
> ```
> 
> Since fluxion was the first to use the `scheduling` key, we could update RFC 20 (Rv1) to say that if `writer` is missing, it is assumed to be set to `fluxion` . Then update RFC 40 (Fluxion Resource Set Extension) to say if the path is missing, it is assumed to be full JGF.

I don't have strong feelings but seems OK to me!

garlick:
Thanks - the RFC changes referenced above for `scheduling.writer` were approved, so I'll make a slightly stronger suggestion to implement that here.  If for some reason I didn't get it right,  we can change the RFC though.

trws:
The agfilter issue is a bit of a concern if we want to be able to get this onto a system and be able to reload fluxion without causing issues.  That's the main case where that will come up. Given how bad some of the issues related to that have been, I'm not comfortable making a true release where we have a known resource loss issue.  Could it be related to something going strange because now there are two things that can have cores?  The agfilter system was never designed to handle that.

If we can, including the updated "writer" key here would be good too.

milroy:
> Given how bad some of the issues related to that have been, I'm not comfortable making a true release where we have a known resource loss issue.

Yes, I agree that we need to at least understand the cause of the incorrect agfilter counts, and correct that behavior before merging this PR.

> Could it be related to something going strange because now there are two things that can have cores? 

I've been talking with @jameshcorbett and I think the cause is that the `jgf_shorthand` reader isn't updating the exclusive subtree edges after fetching and updating the subtree vertices: https://github.com/flux-framework/flux-sched/blob/c50400f79e9394b82e914ef0157b4b0e65c58642/resource/readers/resource_reader_jgf.cpp#L980

So this call: https://github.com/flux-framework/flux-sched/blob/c50400f79e9394b82e914ef0157b4b0e65c58642/resource/readers/resource_reader_jgf.cpp#L1302

should update the edges in the induced subgraph defined by the "additional vertices" discovered in the `fetch_additional_vertices` call.

I also agree about including the updated "writer" key.

jameshcorbett:
> > Given how bad some of the issues related to that have been, I'm not comfortable making a true release where we have a known resource loss issue.
> 
> Yes, I agree that we need to at least understand the cause of the incorrect agfilter counts, and correct that behavior before merging this PR.
> 
> > Could it be related to something going strange because now there are two things that can have cores?
> 
> I've been talking with @jameshcorbett and I think the cause is that the `jgf_shorthand` reader isn't updating the exclusive subtree edges after fetching and updating the subtree vertices:
> 
> https://github.com/flux-framework/flux-sched/blob/c50400f79e9394b82e914ef0157b4b0e65c58642/resource/readers/resource_reader_jgf.cpp#L980
> 
> So this call:
> 
> https://github.com/flux-framework/flux-sched/blob/c50400f79e9394b82e914ef0157b4b0e65c58642/resource/readers/resource_reader_jgf.cpp#L1302
> 
> should update the edges in the induced subgraph defined by the "additional vertices" discovered in the `fetch_additional_vertices` call.

I added a commit to a [branch](https://github.com/flux-framework/flux-sched/commit/495e040e9ac02f1aa6fcebc6e00856b1086f9075) in my fork that, as far as I can tell from the print statements, does in fact update the exclusive subtree edges. However, the same problem still results, so there must  be something else or perhaps additional missing.

milroy:
> I added a commit to a [branch](https://github.com/flux-framework/flux-sched/commit/495e040e9ac02f1aa6fcebc6e00856b1086f9075) in my fork that, as far as I can tell from the print statements, does in fact update the exclusive subtree edges.

It looks like that code will print the edges, but I don't think it will update them because the additional vertices aren't in `vmap`. Can you check if this line is getting called with sensible inputs: https://github.com/flux-framework/flux-sched/blob/81482c68a13995ef44697df4ab05be08707a9b67/resource/readers/resource_reader_jgf.cpp#L1166

jameshcorbett:
> > I added a commit to a [branch](https://github.com/flux-framework/flux-sched/commit/495e040e9ac02f1aa6fcebc6e00856b1086f9075) in my fork that, as far as I can tell from the print statements, does in fact update the exclusive subtree edges.
> 
> It looks like that code will print the edges, but I don't think it will update them because the additional vertices aren't in `vmap`. Can you check if this line is getting called with sensible inputs:
> 
> https://github.com/flux-framework/flux-sched/blob/81482c68a13995ef44697df4ab05be08707a9b67/resource/readers/resource_reader_jgf.cpp#L1166

The line is called but it turns out not with sensible inputs! The vertices (edges?) aren't marked as exclusive. Fixing that makes things work as expected and the tests pass! I'll be pushing up changes here shortly.

trws:
Looks like this is basically passing except for a missing `&&` chain on the python command to normalize json.  Out of curiosity though, why the python line rather than adding `-S` to the `jq` command it's piped into?

resource/readers/resource_reader_jgf.cpp
```diff
@@ -1323,7 +1337,7 @@ int resource_reader_jgf_t::partial_cancel (resource_graph_t &g,
     p_cancel_data.update = false;
     if ((rc = fetch_jgf (R, &jgf, &nodes, &edges, p_cancel_data)) != 0)
         goto done;
-    if ((rc = update_vertices (g, m, vmap, nodes, p_cancel_data)) != 0)
+    if ((rc = update_vertices (g, m, vmap, nodes, 0, p_cancel_data)) != 0)
         goto done;
```
jameshcorbett: @milroy I'm not sure if this is sensible...

jameshcorbett: Resolved in a later commit!

resource/modules/resource_match.cpp
```diff
@@ -1312,6 +1312,7 @@ static int parse_R (std::shared_ptr<resource_ctx_t> &ctx,
     int rc = 0;
     int version = 0;
     int saved_errno;
+    int jgf_complete = 1;
```
milroy: Where is this variable used?

resource/readers/resource_reader_jgf.cpp
```diff
@@ -955,6 +955,7 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
                                             resource_graph_metadata_t &m,
                                             std::map<std::string, vmap_val_t> &vmap,
                                             json_t *nodes,
+                                            uint64_t token,
```
milroy: I think it would be cleaner if you added `token` to `jgf_updater_data`. Then you wouldn't need to change the `update_vertices` signature.

resource/readers/resource_reader_jgf.cpp
```diff
@@ -978,6 +979,9 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
                 goto done;
             }
         }
+        if (fetch_additional_edges (g, m, vmap, fetcher, additional_vertices, token) < 0) {
```
milroy: If you add `uint64_t token` to `jgf_updater_data`, this line will be:

```suggestion
        if (fetch_additional_edges (g, m, vmap, fetcher, additional_vertices, update_data.token) < 0) {
```

resource/readers/resource_reader_base.hpp
```diff
@@ -97,13 +97,15 @@ class resource_reader_base_t {
...
+     * \param token  traversal token
      * \return       0 on success; non-zero integer on an error
      */
     virtual int partial_cancel (resource_graph_t &g,
                                 resource_graph_metadata_t &m,
                                 modify_data_t &mod_data,
                                 const std::string &R,
-                                int64_t jobid) = 0;
+                                int64_t jobid,
+                                uint64_t token) = 0;
```
milroy: I also suggest putting `token` into `modify_data_t`, but would like to know what @trws thinks.

milroy (commented):
I believe I found an unused variable. I also suggest putting `token` into `jgf_updater_data` to simplify the function signatures.

resource/readers/resource_reader_base.hpp
trws: That would be cleaner IMO, maybe rename to `sequence_number` or something too? The current name doesn't really evoke what it's used for.

jameshcorbett: Done!

jameshcorbett:
Thanks for the comments @milroy ! I've addressed & accepted them.

resource/readers/resource_reader_base.hpp
milroy: I think @trws is suggesting that you change `token` to `sequence_number` in both `modify_data_t` and `jgf_updater_data`. Unfortunately, the readers, traversers, and resource-query use `token` internally.

While I agree that `sequence_number` is much more descriptive, I think we need to change the name everywhere or not at all.

milroy:
@jameshcorbett the final issue to address is the question of the scope of the change of `token` to `sequence_number` (i.e., everywhere in the code, just the readers, or postpone the renaming). Once we settle that I'll approve the PR.

resource/readers/resource_reader_base.hpp
milroy: Actually, a compromise could be to use `sequence_number` in the readers and indicate with a comment that it's frequently called `token` outside of the readers.

jameshcorbett: Ok, I added two commits that standardizes the variable name to `sequence_number` everywhere.

milroy (approved):
The variable name change to `sequence_number` looks good! I'm approving this PR.

jameshcorbett:
Chatted with @trws offline, got go-ahead to set MWP, doing that now.

Issue #1425, opened 2026-01-29
Job Moldability in flux


kostiscpp:
In my effort to implement a custom job moldability scheduling algorithm, I examined the way fluxion currently handles the choice between different resource counts. 
As I understand it, the user submits a jobspec with ranges (count : {min, max, operator, operand}) and fluxion-resource decides for the maximum available resource count for each resource in its traverser + matcher component. 
After that, the appropriate R allocation is returned to flux-core and flux-shell tries to run the job. 
However, flux-core will try to run the job according to the jobspec in combination with the R allocation and it does not know how to parse a jobspec with ranges yet, as it supports only jobspecv1 submissions for the most part (correct me if I'm wrong). 
The solution for this problem would be for the scheduler to somehow annotate the "moldable" job with the count values it chose, in order for the flux-shell to run the job properly. 
In addition, the function jobspec_parse of flux-core should be altered to set the resource count of a range to a specific value according to the annotations provided by the scheduler.
This solution is agnostic to the way the scheduler chose the resource counts for each range, which will allow anyone to develop their custom moldability algorithms by defining a new traverser policy.
I would like to ask how feasible my idea is considering all the flux components and for direction on what could this annotation look like.
To add context to the conversation I aim to implement different moldability algorithms that take into account different job information that could be either user specified e.g. walltimes or queue/system related.

<img width="1322" height="852" alt="Image" src="https://github.com/user-attachments/assets/a843e889-b092-4531-8888-f021ee115c63" />

kostiscpp:
While it works for sched-simple, it doesn't seem to work for fluxion. 
I ran this example in the flux-sched dev container:
```
vscode@58ccaa27e95f:~$ flux --version
commands:               0.81.0-182-g8c14747ac
libflux-core:           0.81.0-182-g8c14747ac
libflux-security:       0.14.0
broker:                 0.81.0-182-g8c14747ac
FLUX_URI:               local:///tmp/flux-hGkCuL/local-0
build-options:          +hwloc.api==2.8.0+zmq==4.3.4
vscode@58ccaa27e95f:~$ cat nov1.toml 
[ingest.validator]
disable = true
vscode@58ccaa27e95f:~$ flux start -s 5 --config=nov1.toml
vscode@58ccaa27e95f:~$ flux resource list
     STATE NNODES NCORES NGPUS NODELIST
[...]

PR #1435, opened 2026-03-05
[WIP] Add nslots calculation and emit in R 

kostiscpp:
**Problem**
As discussed in #1425 currently jobspec requests with ranges are not fully supported by flux. 
The issue lies in the fact that when a job is submitted, flux-core does not have any information about how many tasks to run from the resulting R allocation, so it relies on its jobspec. However when a jobspec with ranges is submitted, core cannot really know what count the scheduler really decided for every resource and especially slots. 

**Solution (wip)**
The design choice to solve this problem at the moment in flux-core is to introduce a new parameter `nslots` that represents how many slots exist in a allocation. At the moment at tasks per slot is 1. This is the related rfc https://github.com/flux-framework/rfc/pull/466 and the pr for `sched-simple` integration https://github.com/flux-framework/flux-core/pull/6632. 
This PR introduces the `nslots` parameter to the rv1 and rv1_nosched files and implements a way to calculate it during traversal.
The idea is for the `scoring_api` to calculate nslots during the `dom_dfv()/select()` for each vertex and pass the result to its parent scaled by the level's chosen resource count if needed. At last the result is written to the job's metadata:
`dom_dfv snippet`
```cpp
    if (sm != match_kind_t::SLOT_MATCH) {
        std::vector<resource_type_t> types;
        dfu.resrc_types(dom, types);
        for (auto t : types) {
            unsigned int multiplier = dfu.best_i (dom, t);
            unsigned int count = dfu.nslots ();
            to_parent.set_nslots ((multiplier > 0 ? multiplier : 1) * count);
        }
    } else {
        dfu.set_nslots (dfu.best_i (dom, slot_rt));
        to_parent.set_nslots (dfu.best_i (dom, slot_rt));
    }
```
```cpp
int dfu_impl_t::select (Jobspec::Jobspec &j, vtx_t root, jobmeta_t &meta, bool excl)
{
    int rc = -1;
    scoring_api_t dfu;
    bool x_in = excl;
    subsystem_t dom = m_match->dom_subsystem ();
    tick ();
    m_preorder = 0;
    m_postorder = 0;
    rc = dom_dfv (meta, root, j.resources, true, &x_in, dfu);
    if (rc == 0) {
        unsigned int needs = 0;
        eval_edg_t ev_edg (dfu.avail (), dfu.avail (), excl);
        eval_egroup_t egrp (dfu.overall_score (), dfu.avail (), 0, excl, true);
        egrp.edges.push_back (ev_edg);
        dfu.add (dom, (*m_graph)[root].type, egrp);
        rc = resolve_graph (root, j.resources, dfu, excl, &needs);
        for (auto &r : j.resources) {
            unsigned int mult = dfu.best_i (dom, r.type);
            unsigned int nslots = dfu.nslots ();
            mult = mult > 0 ? mult : 1;
            nslots = nslots > 0 ? nslots : 1;
            meta.nslots = mult * nslots;
        }
        m_graph_db->metadata.v_rt_edges[dom].set_for_trav_update (needs, x_in, m_best_k_cnt);
    }
    return rc;
}
 ```
Then in `dfu_impl_update`, in the update method the nslots parameter is emitted to the writer just like the starttime, end time metadata.
A choice I made in the rv1, rv1_nosched writers is that if `nslots` is not set (equal to 0) then it is not written in the R file. I 

**Limitations/Topics for discussion**
1. All ctests pass except for:
   - `t3027-resource-RV.t` that expects allocations without the nslots parameter (easily fixable)
   -  `t1012-find-status.t`
       `t3022-resource-update.t`
       `t3025-resource-find.t`
       `t3039-resource-force-load-format.t`
       `t3040-jgf-shorthand-match-format.t`
       `t3041-jgf-shorthand-load-format.t`
       `t4009-match-update.t`  
        that compare the R returned by match allocate with the one retrieved by update allocate

   I think that having nslots required as a parameter in update allocate is not a great solution to this problem.
   We can: 
   - Accept that update allocate is not going to return a nslots count 
   or
   -  Embed it in some map from jobids to nslots
   or
   - Any other solution you propose

2. This solution works only for `nslots == ntasks`. 
   I think we should consider changing `nslots` semantics in the R file for `ntasks` . For calculating total tasks, we need to handle 3 cases according to the `count` section of  `jobspec.tasks`:
   - If `count` is set to `per_slot : x` then `ntasks = nslots * x`
   - If `count` is set to `per_resource: { rtype1 : x, rtype2 : y, ... }` then we need to add weights to each rtype mentioned under a slot accordingly 
   - If `count` is set to `total : x` then `ntasks = x`
   The design for this more complicated solution, especially for the `per_resource` case, needs to be discussed further.

### CODE REVIEW

File: resource/traversers/dfu_impl.cpp [hunk 1]
```diff
@@ -794,6 +794,18 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta,
         goto done;
     to_parent.set_avail (avail);
     to_parent.set_overall_score (dfu.overall_score ());
+    if (sm != match_kind_t::SLOT_MATCH) {
+        std::vector<resource_type_t> types;
+        dfu.resrc_types (dom, types);
+        for (auto t : types) {
+            unsigned int multiplier = dfu.best_i (dom, t);
+            unsigned int count = dfu.nslots ();
+            to_parent.set_nslots ((multiplier > 0 ? multiplier : 1) * count);
```
trws: This looks like it's potentially setting the nslots repeatedly for each type, overwriting each time. Is that right?

resource/traversers/dfu_impl.hpp
```diff
@@ -46,6 +46,7 @@ struct jobmeta_t {
     int64_t at = -1;
     int64_t now = -1;
     uint64_t duration = SYSTEM_DEFAULT_DURATION;  // will need config ultimately
+    unsigned int nslots = 0;
```
trws: This looks vestigial.

trws (commented):
This generally looks good while we're working with the `nslots` approach.  There's an unused variable, which the lints didn't catch for some reason, and I'm curious about the loop over types setting nslots.

resource/writers/match_writers.cpp
```diff
@@ -932,50 +932,104 @@ int rv1_match_writers_t::emit_json (json_t **j_o, json_t **aux)
         goto ret;
```
milroy, line 999: Commit b14e1d821820d06f012ccae9d714b7962375a08b improves commit 1ac620c6c642d734252acb18d0faa590b7dd94e9 by increasing code coverage. The later commit should be squashed into the earlier commit to create issues and then solve them in the same PR.

resource/traversers/dfu_impl.cpp
```diff
@@ -794,6 +794,18 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta,
...
+        std::vector<resource_type_t> types;
+        dfu.resrc_types(dom, types);
+        for (auto t : types) {
+            unsigned int multiplier = dfu.best_i (dom, t);
+            unsigned int count = dfu.nslots ();
+            to_parent.set_nslots ((multiplier > 0 ? multiplier : 1) * count);
+        }
+    } else {
+        dfu.set_nslots (dfu.best_i (dom, slot_rt));
+        to_parent.set_nslots (dfu.best_i (dom, slot_rt));
```
milroy: This block will execute for `OR` slots. If the traverser type is `flexible`, it's unclear if `dfu.best_i (dom, slot_rt)` correctly returns the count for the selected `OR` alternative. @zekemorton can you take a look?

resource/traversers/dfu_impl.cpp
```diff
@@ -1277,6 +1289,13 @@ int dfu_impl_t::select (Jobspec::Jobspec &j, vtx_t root, jobmeta_t &meta, bool e
         egrp.edges.push_back (ev_edg);
         dfu.add (dom, (*m_graph)[root].type, egrp);
         rc = resolve_graph (root, j.resources, dfu, excl, &needs);
+        for (auto &r : j.resources) {
+            unsigned int mult = dfu.best_i (dom, r.type);
+            unsigned int nslots = dfu.nslots ();
+            mult = mult > 0 ? mult : 1;
+            nslots = nslots > 0 ? nslots : 1;
+            meta.nslots = mult * nslots;
```
milroy: Similar comment to above: we discussed this a few weeks ago, but @kostiscpp can you comment on the repeated overwrites? For OR slots at the top level (multiple resources), which one determines the final nslots?

resource/writers/match_writers.cpp
```diff
@@ -932,50 +932,104 @@ int rv1_match_writers_t::emit_json (json_t **j_o, json_t **aux)
...
-                             "scheduling",
-                             jgf_o))) {
-            json_decref (rlite_o);
-            json_decref (jgf_o);
-            json_decref (rlite_aux_o);
-            rc = -1;
-            errno = EINVAL;
-            goto ret;
+        if (m_nslots) {
+            if (!(o = json_pack ("{s:i s:{s:o s:O s:O s:I s:I s:I} s:o}",
```
milroy: I made another, similar comment on the last commit, but f77670ff59fe2d4434f5256984ec2745e73964d4 decreases code coverage and corrects `"{s:i s:{s:o s:O s:O s:I s:I} s:o}"` to be `o*`. The second commit that touches this code should be squashed into the first commit as a fixup.

resource/evaluators/scoring_api.cpp
```diff
@@ -145,6 +145,14 @@ bool scoring_api_t::is_contained (subsystem_t s, resource_type_t const &r)
     return m_ssys_map[s].contains (r);
 }
 
+unsigned int scoring_api_t::nslots () {
+    return m_nslots;
+}
+
+void scoring_api_t::set_nslots (unsigned int nslots) {
+    m_nslots = std::max(m_nslots, nslots);
```
milroy: Should have a space before parenthesis per convention:
```suggestion
    m_nslots = std::max (m_nslots, nslots);
```

resource/traversers/dfu_impl.cpp
```diff
@@ -796,7 +796,7 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta,
     to_parent.set_overall_score (dfu.overall_score ());
     if (sm != match_kind_t::SLOT_MATCH) {
         std::vector<resource_type_t> types;
-        dfu.resrc_types(dom, types);
+        dfu.resrc_types (dom, types);
```
milroy: I believe this formatting issue was introduced in an earlier commit. The change needs to be applied there.

resource/evaluators/scoring_api.cpp
```diff
@@ -152,7 +152,7 @@ unsigned int scoring_api_t::nslots ()
 
 void scoring_api_t::set_nslots (unsigned int nslots)
 {
-    m_nslots = std::max (m_nslots, nslots);
+    m_nslots = nslots;
```
milroy: We need to consider the implications of this change.

With `max()` if multiple children set `nslots`, the parent gets the maximum, versus with assignment the last child to call `set_nslots()` wins.

resource/evaluators/scoring_api.cpp
milroy: Per Flux convention, there needs to be a space between the function name and the open-parenthesis:
```suggestion
    m_nslots = std::max (m_nslots, nslots);
```

resource/evaluators/scoring_api.cpp
milroy: Also, this is another instance of adding code in an earlier commit and then updating it in a later commit.

milroy (changes requested):
This is a good start. I have a few important formatting change requests that can be summarized as: a PR should not introduce changes (or issues) that it later updates or fixes in later commits (https://flux-framework.readthedocs.io/en/latest/contributing.html#developer-guidelines).

Along the lines of style, please take a look at our guidelines in RFC 7: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_7.html and commit ("patch") guidelines in RFC 1: https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#patch-requirements. In particular, the character widths of the commit bodies.

We'll also need to discuss the implications of the overwriting behavior and the impact to `OR` slots. I think `XOR` will be unaffected per my own testing. It would be good to get @zekemorton's feedback, too.

resource/evaluators/scoring_api.cpp
trws: I missed this one, good catch @milroy.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -868,10 +868,15 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
     if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu, emit_shadow, false)) > 0) {
         uint64_t starttime = jobmeta.at;
         uint64_t endtime = jobmeta.at + jobmeta.duration;
+        unsigned int nslots = jobmeta.nslots;
         if (writers->emit_tm (starttime, endtime) == -1) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned -1.\n";
         }
+        if (writers->emit_nslots (nslots) == -1) {
```
sam-maloney:
```suggestion
if (writers->emit_tm (starttime, endtime) == -1) {
m_err_msg += __FUNCTION__;
m_err_msg += ": emit_tm returned -1.\n";
}
        if (writers->emit_nslots (jobmeta.nslots) == -1) {
```
I think this is cleaner without the intermediate variable, and mitigates the possibility of inconsistent typing

resource/traversers/dfu_impl_update.cpp
```diff
@@ -934,10 +939,15 @@ int dfu_impl_t::update (vtx_t root,
     if ((rc = upd_dfv (root, writers, needs, x, jobmeta, false, dfu, emit_shadow, false)) > 0) {
         uint64_t starttime = jobmeta.at;
         uint64_t endtime = jobmeta.at + jobmeta.duration;
+        unsigned int nslots = jobmeta.nslots;
         if (writers->emit_tm (starttime, endtime) == -1) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned -1.\n";
         }
+        if (writers->emit_nslots (nslots) == -1) {
```
sam-maloney:
```suggestion
if (writers->emit_tm (starttime, endtime) == -1) {
m_err_msg += __FUNCTION__;
m_err_msg += ": emit_tm returned -1.\n";
}
        if (writers->emit_nslots (jobmeta.nslots) == -1) {
```
Same here, remove intermediate variable

resource/traversers/dfu_impl.hpp
sam-maloney: I think it's set in [L1297 of `resource/traversers/dfu_impl.cpp`](https://github.com/flux-framework/flux-sched/pull/1435/changes#diff-4b5845c0812a1c4cd78425de9cb1de1c495236f9562a4b598d204f7816d6aa41R1297) and then emitted in the changes to `resource/traversers/dfu_impl_update.cpp`

I would, however, suggest changing the type for consistency with https://github.com/flux-framework/flux-sched/pull/1457:
```suggestion
    int64_t nslots = 0;
```

resource/writers/match_writers.cpp
```diff
@@ -1055,6 +1059,12 @@ int rv1_match_writers_t::emit_tm (uint64_t start_tm, uint64_t end_tm)
     return 0;
 }
 
+int rv1_match_writers_t::emit_nslots (unsigned int nslots)
```
sam-maloney:
```suggestion
int rv1_match_writers_t::emit_nslots (int64_t nslots)
```

resource/writers/match_writers.cpp
```diff
@@ -1170,6 +1184,12 @@ int rv1_nosched_match_writers_t::emit_tm (uint64_t start_tm, uint64_t end_tm)
     return 0;
 }
 
+int rv1_nosched_match_writers_t::emit_nslots (unsigned int nslots)
```
sam-maloney:
```suggestion
int rv1_nosched_match_writers_t::emit_nslots (int64_t nslots)
```

resource/writers/match_writers.hpp
```diff
@@ -49,6 +49,10 @@ class match_writers_t {
     {
         return 0;
     }
+    virtual int emit_nslots (unsigned int nslots)
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots)
```

resource/writers/match_writers.hpp
```diff
@@ -241,6 +245,7 @@ class rv1_match_writers_t : public match_writers_t {
                           bool exclusive) override;
     virtual int emit_edg (const std::string &prefix, const resource_graph_t &g, const edg_t &e);
     virtual int emit_tm (uint64_t start_tm, uint64_t end_tm);
+    virtual int emit_nslots (unsigned int nslots);
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots);
```

resource/writers/match_writers.hpp
```diff
@@ -249,6 +254,7 @@ class rv1_match_writers_t : public match_writers_t {
     rlite_match_writers_t rlite;
     int64_t m_starttime = 0;
     int64_t m_expiration = 0;
+    unsigned int m_nslots = 0;
```
sam-maloney:
```suggestion
    int64_t m_nslots = 0;
```

resource/writers/match_writers.hpp
```diff
@@ -267,11 +273,13 @@ class rv1_nosched_match_writers_t : public match_writers_t {
                           const std::map<std::string, std::string> &agfilter_data,
                           bool exclusive) override;
     virtual int emit_tm (uint64_t start_tm, uint64_t end_tm);
+    virtual int emit_nslots (unsigned int nslots);
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots);
```

resource/writers/match_writers.hpp
```diff
@@ -267,11 +273,13 @@ class rv1_nosched_match_writers_t : public match_writers_t {
...
 
    private:
     rlite_match_writers_t rlite;
     int64_t m_starttime = 0;
     int64_t m_expiration = 0;
+    unsigned int m_nslots = 0;
```
sam-maloney:
```suggestion
    int64_t m_nslots = 0;
```

sam-maloney (changes requested):
Mostly just aligning the integer types with https://github.com/flux-framework/flux-sched/pull/1457

kostiscpp:
After internal conversations and testing, I changed my approach to solve this problem. 
The main issues with the previous logic were:
1. The multiplier logic works only for homogeneous allocations, so a moldable request would not be allocated properly in complex cases. (This is quite ironic considering this effort started as a way to support moldable requests - ranges to flux-core)
2. The nslots calculation happens during inside the `dom_dfv` method where all candidate matches are explored and it is not obvious how to differenceiate between different candidate matches. In general an unstable stage for calculating the final total slot count.
My new approach addresses these topics by changing the actual stage where the slot calculation happens to the `upd_dfv` stage (in this stage the final match subgraph is decided). However, in the update stage we have no information about slots. That's why we add a new edge metadata field that not only helps with counting slots but also could be the foundation for emitting RV2 whenever its implementation starts.
The field added is called `slot_role` and it has 3 possible values:
LEADER, MEMBER and NONE. 
Slots in general are expressed as edge groups, where all the resources under these edges are the resources inside that slots. In order to identify a slot, all we have to do is to assign a representative of an edge group associated with a slot (the leader) and all the sibling edges that are part of the slot as members of it.
In a nutshell:
1. In the `dom_dfv` stage we mark the slot - edge groups selected (in the `enforce ()` method)
2. In the update/`upd_dfv` stage we count how many slot edge group leaders there are according to the marked edges
3. At last we emit that count as nslots in Rv1
I decided to created 3 new writers for Rv1 with nslots to minimize changes in existing tests and to not interfere with new experimental tests e.g. [#1478](https://github.com/flux-framework/flux-sched/issues/1478).

resource/traversers/dfu_impl.hpp
```diff
@@ -597,6 +597,7 @@ class dfu_impl_t {
                  const jobmeta_t &jobmeta,
                  bool full,
                  std::map<resource_type_t, int64_t> &to_parent,
+                 uint64_t &nslots,
```
sam-maloney:
```suggestion
                 int64_t &nslots,
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -315,6 +315,7 @@ int dfu_impl_t::upd_dfv (vtx_t u,
                          const jobmeta_t &jobmeta,
                          bool full,
                          std::map<resource_type_t, int64_t> &to_parent,
+                         uint64_t &nslots,
```
sam-maloney:
```suggestion
                         int64_t &nslots,
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -925,14 +937,20 @@ int dfu_impl_t::update (vtx_t root,
     needs = static_cast<unsigned int> (m_graph_db->metadata.v_rt_edges[dom].get_needs ());
     m_color.reset ();
     bool emit_shadow = modify_traversal (root, false);
+    uint64_t nslots = 0;
```
sam-maloney:
```suggestion
    int64_t nslots = 0;
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -859,14 +865,20 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
     m_color.reset ();
 
     bool emit_shadow = modify_traversal (root, false);
+    uint64_t nslots = 0;
```
sam-maloney:
```suggestion
    int64_t nslots = 0;
```

resource/writers/match_writers.cpp
```diff
@@ -20,6 +20,7 @@ extern "C" {
 #include <cerrno>
 #include <algorithm>
 #include <iterator>
+#include <iostream>
```
sam-maloney: Is this left over from debugging, perhaps?

resource/writers/match_writers.cpp
```diff
@@ -1055,7 +1060,6 @@ int rv1_match_writers_t::emit_json (json_t **j_o, json_t **aux)
         }
     }
     *j_o = o;
-
```
sam-maloney, line 1065: Try and avoid extraneous whitespace changes

resource/writers/match_writers.cpp
```diff
@@ -1117,6 +1121,7 @@ int rv1_match_writers_t::emit_tm (int64_t start_tm, int64_t end_tm)
     return 0;
 }
 
+
```
sam-maloney: Same as above, avoid extraneous change

resource/writers/match_writers.cpp
```diff
@@ -1128,6 +1133,16 @@ jgf_match_writers_t &rv1_match_writers_t::get_jgf ()
     return jgf_writer;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// RV1 Nslots Writers Class Method Definitions
+////////////////////////////////////////////////////////////////////////////////
+
+int rv1_nslots_match_writers_t::emit_nslots (uint64_t nslots)
```
sam-maloney:
```suggestion
int rv1_nslots_match_writers_t::emit_nslots (int64_t nslots)
```

resource/writers/match_writers.cpp
```diff
@@ -1239,14 +1258,34 @@ int rv1_nosched_match_writers_t::emit_tm (int64_t start_tm, int64_t end_tm)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-// RV1 Nosched Writers Class Method Definitions
+// RV1 Nosched Nslots Writers Class Method Definitions
+////////////////////////////////////////////////////////////////////////////////
+
+int rv1_nosched_nslots_match_writers_t::emit_nslots (uint64_t nslots)
```
sam-maloney:
```suggestion
int rv1_nosched_nslots_match_writers_t::emit_nslots (int64_t nslots)
```

resource/writers/match_writers.cpp
```diff
@@ -1239,14 +1258,34 @@ int rv1_nosched_match_writers_t::emit_tm (int64_t start_tm, int64_t end_tm)
...
 jgf_match_writers_t &rv1_shorthand_match_writers_t::get_jgf ()
 {
     return jgf_writer;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// RV1 Shorthand Nslots Writers Class Method Definitions
+////////////////////////////////////////////////////////////////////////////////
+
+int rv1_shorthand_nslots_match_writers_t::emit_nslots (uint64_t nslots)
```
sam-maloney:
```suggestion
int rv1_shorthand_nslots_match_writers_t::emit_nslots (int64_t nslots)
```

resource/writers/match_writers.hpp
```diff
@@ -65,6 +68,10 @@ class match_writers_t {
     {
         return 0;
     }
+    virtual int emit_nslots (uint64_t nslots)
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots)
```

resource/writers/match_writers.hpp
```diff
@@ -279,6 +286,7 @@ class rv1_match_writers_t : public match_writers_t {
 
    protected:
     virtual jgf_match_writers_t &get_jgf ();
+    uint64_t m_nslots = 0;
```
sam-maloney:
```suggestion
    int64_t m_nslots = 0;
```

resource/writers/match_writers.hpp
```diff
@@ -290,6 +298,13 @@ class rv1_match_writers_t : public match_writers_t {
     jgf_match_writers_t jgf_writer;
 };
 
+/*! R Version 1 with "nslots" key match writers class
+ */
+class rv1_nslots_match_writers_t : public rv1_match_writers_t {
+  public:
+   virtual int emit_nslots (uint64_t nslots) override;
```
sam-maloney:
```suggestion
   virtual int emit_nslots (int64_t nslots) override;
```

resource/writers/match_writers.hpp
```diff
@@ -306,12 +321,22 @@ class rv1_nosched_match_writers_t : public match_writers_t {
                   bool excl_parent) override;
     virtual int emit_tm (int64_t start_tm, int64_t end_tm);
 
+   protected:
+    uint64_t m_nslots = 0;
```
sam-maloney:
```suggestion
    int64_t m_nslots = 0;
```

resource/writers/match_writers.hpp
```diff
@@ -306,12 +321,22 @@ class rv1_nosched_match_writers_t : public match_writers_t {
...
     rlite_match_writers_t rlite;
     int64_t m_starttime = 0;
     int64_t m_expiration = 0;
 };
 
+/*! R Version 1 with "nslots" key but no "scheduling" key match writers class
+ */
+class rv1_nosched_nslots_match_writers_t : public rv1_nosched_match_writers_t {
+   public:
+    virtual int emit_nslots (uint64_t nslots) override;
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots) override;
```

resource/writers/match_writers.hpp
```diff
@@ -322,6 +347,13 @@ class rv1_shorthand_match_writers_t : public rv1_match_writers_t {
     jgf_shorthand_match_writers_t jgf_writer;
 };
 
+/*! R Version 1 with JGF shorthand writer and "nslots" key
+ */
+class rv1_shorthand_nslots_match_writers_t : public rv1_shorthand_match_writers_t {
+   public:
+    virtual int emit_nslots (uint64_t nslots) override;
```
sam-maloney:
```suggestion
    virtual int emit_nslots (int64_t nslots) override;
```

sam-maloney (changes requested):
One extra include, and 2 minor formatting nits, but mostly just type pedantry to align with #1457

resource/writers/match_writers.cpp
kostiscpp: Oops, fixed now

jameshcorbett:
Just checking in --- where are we at on this? @kostiscpp are you waiting for further review?

kostiscpp:
@jameshcorbett Yes, kinda. There are many developments regarding slots in general (mainly xor slots and non exclusive slots) that need to be considered once they're ready and well defined (without conflicting eachother).

The one main change that needs to be done to my PR as I discussed with @milroy is to count only slots that are mapped to actual tasks. That would be easy if slot label was always directly associated with tasks, however in the case of xor slots this is not always the case. When this inconsistency is resolved I think my PR will be ready to go.

In any case, an early review would be useful if I've missed anything else.

resource/traversers/dfu_impl_update.cpp
```diff
@@ -859,14 +865,20 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
...
+        > 0) {
         int64_t starttime = jobmeta.at;
         int64_t endtime = jobmeta.at + jobmeta.duration;
         if (writers->emit_tm (starttime, endtime) == -1) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned -1.\n";
         }
+        if (writers->emit_nslots (nslots)) {
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": emit_tm returned -1.\n";
```
jameshcorbett, line 881:
```suggestion
            m_err_msg += ": emit_nslots returned -1.\n";
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -859,14 +865,20 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
...
-    if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu, emit_shadow, false)) > 0) {
+    if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu, nslots, emit_shadow, false))
+        > 0) {
         int64_t starttime = jobmeta.at;
         int64_t endtime = jobmeta.at + jobmeta.duration;
         if (writers->emit_tm (starttime, endtime) == -1) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned -1.\n";
         }
+        if (writers->emit_nslots (nslots)) {
```
jameshcorbett, line 879:
```suggestion
        if (writers->emit_nslots (nslots) < 0) {
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -925,14 +937,20 @@ int dfu_impl_t::update (vtx_t root,
...
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": emit_tm returned -1.\n";
+        }
```
jameshcorbett, line 954:
```suggestion
        if (writers->emit_nslots (nslots) < 0) {
            m_err_msg += __FUNCTION__;
            m_err_msg += ": emit_nslots returned -1.\n";
        }
```

resource/writers/match_writers.hpp
```diff
@@ -65,6 +68,10 @@ class match_writers_t {
     {
         return 0;
     }
+    virtual int emit_nslots (int64_t nslots)
+    {
+        return 0;
+    }
```
jameshcorbett, line 75: Add a short comment describing expected behavior? I think this method should be expected to return -1 on error, I suggested some changes to that effect.

resource/schema/infra_data.cpp
```diff
@@ -216,6 +216,21 @@ void relation_infra_t::set_weight (uint64_t weight)
...
+{
+    return m_slot_role;
+}
+
+void relation_infra_t::set_slot_role (relation_infra_t::slot_role_t slot_role)
+{
+    m_slot_role = slot_role;
+}
+
+bool relation_infra_t::is_slot_leader ()
```
jameshcorbett, line 217:
```suggestion
bool relation_infra_t::is_slot_leader () const
```

t/t3043-rv1-nslots.t
```diff
@@ -0,0 +1,68 @@
...
+
+test_description='Test nslots Calculation On Medium Machine Configuration in JGF'
+
+. $(dirname $0)/sharness.sh
+
+cmd_dir="${SHARNESS_TEST_SRCDIR}/data/resource/commands/nslots"
+exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/nslots"
+jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny.json"
+query="../../resource/utilities/resource-query"
+
```
jameshcorbett: t3043, t3044, and t3045 are pretty similar -- could be combined with some sort of helper function?

resource/writers/match_writers.hpp
```diff
@@ -32,8 +32,11 @@ enum class match_format_t {
     JGF_SHORTHAND,
     RLITE,
     RV1,
+    RV1_NSLOTS,
     RV1_NOSCHED,
+    RV1_NOSCHED_NSLOTS,
     RV1_SHORTHAND,
+    RV1_SHORTHAND_NSLOTS,
```
jameshcorbett: Adding three new whole writer types seems like overkill. Could the existing writers just be modified with a flag or something?

jameshcorbett (commented):
I didn't review this at a high level, for that we'll need @milroy it sounds like. But here is what I noticed during a quick review---

resource/writers/match_writers.hpp
kostiscpp: Yeah, this will never return -1. It should if there was some error state, but currently there isn't. I was just following the emit_tm design logic to be consistent with the rest of the code. The changes you suggested should be done for sure!

resource/traversers/dfu_impl_update.cpp
```diff
@@ -860,14 +866,20 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
...
+        > 0) {
         int64_t starttime = jobmeta.at;
         int64_t endtime = jobmeta.at + jobmeta.duration;
         if (writers->emit_tm (starttime, endtime) == -1) {
             m_err_msg += __FUNCTION__;
             m_err_msg += ": emit_tm returned -1.\n";
         }
+        if (writers->emit_nslots (nslots)) {
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": emit_tm returned -1.\n";
```
zekemorton, line 881: The error message references the wrong function (copy-paste from the `emit_tm` block
just above).

```suggestion
            m_err_msg += ": emit_nslots returned -1.\n";
```

resource/traversers/dfu_impl_update.cpp
```diff
@@ -926,14 +938,20 @@ int dfu_impl_t::update (vtx_t root,
...
```
zekemorton, line 953: Same copy-paste here.

```suggestion
            m_err_msg += ": emit_nslots returned -1.\n";
```

resource/writers/match_writers.cpp
```diff
@@ -1378,8 +1415,9 @@ match_format_t match_writers_factory_t::get_writers_type (const std::string &n)
 bool known_match_format (const std::string &format)
 {
     return (format == "simple" || format == "jgf" || format == "jgf_shorthand" || format == "rlite"
-            || format == "rv1" || format == "rv1_nosched" || format == "rv1_shorthand"
-            || format == "pretty_simple");
+            || format == "rv1" || format == "rv1_nslots" || format == "rv1_nosched"
+            || format == "rv1_nosched_nslots" || format == "rv1_shorthand"
+            || format == "rv1_shorthand_nslots" || format == "pretty_simple");
```
zekemorton, line 1420: These three new format strings (`rv1_nslots`, `rv1_nosched_nslots`,
`rv1_shorthand_nslots`) pass `known_match_format()` but have no matching case in
`get_writers_type()` (line 1395), so `--match-format=rv1_nslots` silently falls through
to the default `simple` writer. Since the feature is actually driven by
`--emit-nslots`/`-s` flag, these aliases look unnecessary.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -1031,6 +1036,7 @@ resource_query_t::resource_query_t (const std::string &rgraph, const std::string
     std::stringstream buffer{};
     std::shared_ptr<resource_reader_base_t> rd;
     match_format_t format;
+    int emit_nslots;
```
zekemorton, line 1039: This local is declared but never read or initialized — the real value flows through
`params.emit_nslots`. Safe to delete.

resource/reapi/bindings/c++/reapi_cli.hpp
```diff
@@ -52,11 +52,12 @@ struct resource_params_t {
     std::string matcher_name;     /* Matcher name */
     std::string matcher_policy;   /* Matcher policy name */
     std::string traverser_policy; /* Traverser policy name */
-    std::string o_fname;          /* Output file to dump the filtered graph */
-    std::string r_fname;          /* Output file to dump the emitted R */
-    std::string o_fext;           /* File extension */
-    std::string prune_filters;    /* Raw prune-filter specification */
-    std::string match_format;     /* Format to emit a matched resources */
+    int emit_nslots;
```
zekemorton, line 55: A default might be a good idea 

```suggestion
    int emit_nslots = -1;
```

resource/utilities/command.hpp
```diff
@@ -32,12 +32,13 @@ struct test_params_t {
     std::string matcher_name;     /* Matcher name */
     std::string matcher_policy;   /* Matcher policy name */
     std::string traverser_policy; /* Traverser policy name */
-    std::string o_fname;          /* Output file to dump the filtered graph */
-    std::ofstream r_out;          /* Output file stream for emitted R */
-    std::string r_fname;          /* Output file to dump the emitted R */
-    std::string o_fext;           /* File extension */
-    std::string prune_filters;    /* Raw prune-filter specification */
-    std::string match_format;     /* Format to emit a matched resources */
+    int emit_nslots;
```
zekemorton, line 35: Same as the reapi struct.

```suggestion
    int emit_nslots = -1;
```

resource/schema/infra_data.hpp
```diff
@@ -67,12 +68,16 @@ class relation_infra_t : public infra_base_t {
     uint64_t get_sequence_number () const;
     uint64_t get_weight () const;
     void set_weight (uint64_t);
+    slot_role_t get_slot_role () const;
+    void set_slot_role (slot_role_t);
+    bool is_slot_leader ();
```
zekemorton, line 73:
```suggestion
    bool is_slot_leader () const;
```

resource/traversers/dfu_impl.cpp
```diff
@@ -1091,10 +1096,22 @@ int dfu_impl_t::enforce (subsystem_t subsystem, scoring_api_t &dfu, bool enforce
...
                     (*m_graph)[e.edge].idata.set_for_trav_update (e.needs,
                                                                   e.exclusive,
                                                                   m_sequence_number);
+                    // if the egroup refers to a slot, set the edge group leader and members
+                    if (t == slot_rt && subsystem == dom && is_task_slot) {
+                        (*m_graph)[e.edge].idata.set_slot_role (
+                            eg_leader ? relation_infra_t::slot_role_t::LEADER
+                                      : relation_infra_t::slot_role_t::MEMBER);
+                        eg_leader = false;
+                    }
```
zekemorton, line 1114: I think `enforce()` only sets the slot role when `is_task_slot` is true, and never
clears it — cleanup only happens in `mod_dfv()`. So if a slot edge carries a stale
`LEADER` from a prior allocation and this pass doesn't overwrite it, `upd_dfv()`
(`dfu_impl_update.cpp:353`) will count it as a slot that isn't one.

Today slot exclusivity probably masks this, since two jobs won't share the same slot
edge. But since we're looking at allowing **non-exclusive slots**, that assumption is
about to go away — a shared slot edge could keep a role set by an earlier job. Worth
making this robust now rather than tracking down a miscount later.

Every edge `upd_dfv()` walks is one this same loop just stamped via
`set_for_trav_update()`, so resetting the role here for all slot edges (not just task
slots) should fully cover what gets counted

@milroy do you mind taking a look at this one to confirm?

zekemorton (commented):
A few things worth discussing beyond the inline notes:

**Task labels via `jobmeta_t`**
You mentioned in the slack channel that removing the job spec from the select () function was causing you problems by not having access to the task labels. The work around you found does seem to work, but I wonder if adding the tasks or task labels into job_meta_t might be a better option? Looks like job_meta_t get passed to most of the relevant functions (including select) and would just need to be added to a few more

Overall looks really good though! 

review assisted by claude Opus 4.8

resource/traversers/dfu_impl_update.cpp
kostiscpp: It was supposed to be fixed but it seems that some small corrections got lost during some rebase or merge process.

resource/writers/match_writers.cpp
kostiscpp: Same with that, it should have been corrected

resource/traversers/dfu_impl.cpp
kostiscpp: Cleaning up each edge directly after finally visiting it with `upd_dfv()` could work for this PR but I wanted to make edges retain per-job runtime info while the job is still in the RUN state. That would keep Rv2 consistent when an allocation for a specific job needs to be reconstructed by jobid.

Following that model, roles should only be cleaned after a job finishes or gets cancelled. I thought that `mod_dfv()` is called in both of those cases, which is why I did the cleanup process inside that function. Can you confirm that my assumption is correct?

Assuming cleanup is done reliably only when a job stops running, we can add a map attribute to slot edges that has job ids as keys and the slot role for said jobs as values. 
- Shared slots do not have ambiguous slot roles, but of course role lookup complexity increases. 
- We do not need to worry about resetting non-task slots during a later allocation pass, since one job cannot overwrite or inherit another job’s metadata . 
- Using the jobid of each job that leaves the `RUN` state, we can remove that job entry in each relevant slot edge map.

PR #1438, opened 2026-03-25
writers: use flux-core idset library

### LINKED ISSUE #1437: invalid idset is output in R
Over in flux-framework/flux-core#7486 a change in configuration resulted in Fluxion generating R with invalid idset in the `core` field:
```json
$ cat failing_R.json |  jq .execution
{
  "R_lite": [
    {
      "rank": "4",
      "children": {
        "core": "0,0-1,1-2,2-95",
        "gpu": "0-3"
      }
    }
  ],
  "nodelist": [
    "hetchy1004"
  ],
  "properties": {
    "pall": "4",
    "parrypeak": "4"
  },
  "starttime": 1774301723,
  "expiration": 1774308923
}
```

Note that ids `0-2` are duplicated in the rank `4` `core` set.

Fluxion should use libidset to encode idsets to ensure they are valid. If this isn't feasible, the internal Fluxion implementation needs to be fixed to generate valid idset strings.

jameshcorbett:
Problem: the match writers sometimes output duplicate IDs as part of an ID set.

Use flux-core's idset library, which eliminates duplication.

Fixes #1437


resource/writers/match_writers.cpp
```diff
@@ -33,31 +34,31 @@ namespace resource_model {
...
-                base = id;
-                runlen = 0;
-            }
+    char *encoded_idset = NULL;
+    struct idset *idset = idset_create (0, IDSET_FLAG_AUTOGROW);
+    if (!idset) {
+        return -1;
+    }
+    for (auto &id : ids) {
+        if (id != -1 && idset_set (idset, id) < 0) {
```
jameshcorbett: I don't understand why the `ids` vector would have `-1` as an element, but without this `id != -1` check, a bunch of test cases failed.

jameshcorbett: @milroy pointed out that ranks are often `-1`, but I don't see how this was handled in the previous implementation.

resource/writers/match_writers.cpp
```diff
@@ -33,31 +34,31 @@ namespace resource_model {
...
-        if (runlen)
-            o << "-" << (base + runlen);
+    }
+    if (!(encoded_idset = idset_encode (idset, IDSET_FLAG_RANGE))) {
+        idset_destroy (idset);
+        return -1;
+    }
+    idset_destroy (idset);
+    try {
+        o << std::string (encoded_idset);
```
trws: Why the string conversion here? IIRC ostringstream will take a null-terminated string as-is.  alternately if we know the length (not sure if idset's API can give us that off top of my head) a `std::string_view` would be more efficient than either of the others.

resource/writers/match_writers.cpp
trws: The only guess I have right now, is that in rlite at least there are only core and GPU IDs, which should never be -1.  That begs the question of why that would change with this patch, which it shouldn't, but it's probably part of the puzzle at least.

resource/writers/match_writers.cpp
jameshcorbett: good catch, fixed and force-pushed

resource/writers/match_writers.cpp
milroy: I don't, know, either. I assume the ids vector contains `-1` because of the `cluster` vertex? Which test cases fail?

milroy (commented):
Can you add a simple test for issue #1437?

resource/writers/match_writers.cpp
jameshcorbett: The following tests fail:
```
 36 - t1015-find-format.t (Failed)
	 49 - t1028-rv1-partial-release-across-racks.t (Failed)
	 74 - t3022-resource-update.t (Failed)
	 77 - t3025-resource-find.t (Failed)
	 96 - t4001-match-allocate.t (Failed)
	 97 - t4002-match-reserve.t (Failed)
	 98 - t4003-cancel-info.t (Failed)
	 99 - t4004-match-hwloc.t (Failed)
	100 - t4005-match-unsat.t (Failed)
	102 - t4007-match-var-aware.t (Failed)
	103 - t4008-match-jgf.t (Failed)
	107 - t4012-set-status.t (Failed)
	109 - t4014-match-feasibility.t (Failed)
	111 - t4016-match-flexible.t (Failed)
	115 - t6001-match-formats.t (Failed)
```
The `idset` type only accepts non-negative integers, so `idset_set` takes an unsigned integer, and when it receives `-1` it gets converted to the unsigned max value which throws an error.

jameshcorbett: I looked at `t6001`. Before this PR, it would generate files like 
```json
[
  {
    "rank": "-1",
    "children": {
      "core": "35"
    }
  }
]
```

After this PR, it generates files like 
```json
[
  {
    "rank": "",
    "children": {
      "core": "35"
    }
  }
]

```

jameshcorbett: I see two ways forward:
1) don't use the idset library, just change the existing implementation to prevent duplication of IDs
2) Change the tests (and/or some other part of the code, like the grug reader?) so that they don't have negative ranks, because core would throw an error if it was passed negative ranks.

jameshcorbett: I went with option 2 and now the PR is much larger.

jameshcorbett:
> Can you add a simple test for issue #1437?

I would love to but I'm not really sure how. Any pointers? If I can pass a vector like `< 0, 0, 1, 1, 2, 3, 4 >` to `compress_ids` I could reproduce #1437 but I'm not sure how to write a test that calls `compress_ids` directly and I'm also not sure how to set up a match such that it yields a vector like `< 0, 0, 1, 1, 2, 3, 4 >`. (That is probably a deeper issue actually--how the duplicates ended up there in the first place).

trws:
You could write a catch2 test that literally calls the function.  That's probably what I would do actually, there's no reason to make it an integration test.

jameshcorbett:
Ok, I added tests, I think this is good to go @trws @milroy

trws (approved):
LGTM

PR #1439, opened 2026-03-25
writers: treat rabbits as execution targets in R

jameshcorbett:
Problem: on El Cap systems, Flux brokers run on rabbit nodes, and jobs can run on them if users explicitly request them. However, they are not to be treated like normal compute nodes, and are not part of any queue. The compute part of each rabbit is part of the resource graph as a `rabbit` vertex. However, the r_lite writer does not emit rabbit resources, because it does not treat them the same as nodes.

Make the R match writer treat `rabbit` vertices like `node` vertices.

### DIFF
--- a/resource/schema/data_std.cpp
+++ b/resource/schema/data_std.cpp
@@ -23,6 +23,7 @@ resource_type_t node_rt{"node"};
 resource_type_t rack_rt{"rack"};
 resource_type_t slot_rt{"slot"};
 resource_type_t ssd_rt{"ssd"};
+resource_type_t storage_node_rt{"storage_node"};
 
 }  // namespace resource_model
 }  // namespace Flux
--- a/resource/writers/match_writers.cpp
+++ b/resource/writers/match_writers.cpp
@@ -506,6 +506,9 @@ rlite_match_writers_t::rlite_match_writers_t ()
     m_reducer[core_rt] = std::vector<int64_t> ();
     m_reducer[gpu_rt] = std::vector<int64_t> ();
     m_gatherer.insert (node_rt);
+    // as a temporary workaround to rabbit scheduling on el cap, add
+    // the storage node type here. To be removed when rv2 is ready.
+    m_gatherer.insert (storage_node_rt);
 }
 
 rlite_match_writers_t::rlite_match_writers_t (const rlite_match_writers_t &w)
--- a/t/data/resource/jobspecs/advanced/rabbit.yaml
+++ b/t/data/resource/jobspecs/advanced/rabbit.yaml
@@ -1,0 +1,34 @@
+version: 9999
+resources:
+  - type: chassis
+    count: 1
+    label: default
+    with:
+    - type: node
+      count: 1
+      exclusive: false
+      with:
+      - type: slot
+        label: task
+        count: 1
+        with:
+        - type: core
+          count: 1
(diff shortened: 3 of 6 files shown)


t/data/resource/jobspecs/advanced/rabbit.yaml
```diff
@@ -0,0 +1,34 @@
...
+          count: 1
+    - type: ssd
+      count: 1
+      exclusive: true
+    - type: rabbit
+      count: 1
+      with:
+      - type: core
+        count: 1
+        exclusive: true
```
milroy, line 25: You don't need to specify `exclusive` for a leaf vertex.

resource/schema/data_std.cpp
```diff
@@ -23,6 +23,7 @@ resource_type_t node_rt{"node"};
 resource_type_t rack_rt{"rack"};
 resource_type_t slot_rt{"slot"};
 resource_type_t ssd_rt{"ssd"};
+resource_type_t rabbit_rt{"rabbit"};
```
milroy: Per the discussion today, I recommend a more general type name like `storage_node_rt`.

resource/writers/match_writers.cpp
```diff
@@ -506,6 +506,7 @@ rlite_match_writers_t::rlite_match_writers_t ()
     m_reducer[core_rt] = std::vector<int64_t> ();
     m_reducer[gpu_rt] = std::vector<int64_t> ();
     m_gatherer.insert (node_rt);
+    m_gatherer.insert (rabbit_rt);
```
milroy: Please add a comment indicating that we intend to remove this one-off fix when we have a general solution. That will likely come with RV2.

milroy (commented):
I wish there were a better, more general fix for this issue right now, this is very simple and easy to revert if needed.

Based on the discussion today, please use a more general type name. Once that's done this PR is ready to merge.

t/data/resource/jobspecs/advanced/rabbit.yaml
jameshcorbett: I found that without `exclusive` the rabbit node and its associated core were not output in R. I think that may be because they aren't part of a `slot`?

resource/schema/data_std.cpp
jameshcorbett: Done!

resource/writers/match_writers.cpp
jameshcorbett: Done!

t/data/resource/jobspecs/advanced/rabbit.yaml
milroy: The cores should be allocated, but they aren't output because of the issue PR #1408 fixes. Can you double check that the cores are still allocated without `exclusive`?

jameshcorbett: I added a check, the third commit in this PR. However, the test fails. Assuming my test is correct, how concerning is this? I don't immediately see it as a problem.

jameshcorbett: After an offline discussion, I removed the test.

milroy (approved):
After further discussion and update to `storage_node_rt` this LGTM.

PR #1444, opened 2026-04-03
ci: add el10 build

grondo:
This PR adds an el10 build to go with the recent addition over in flux-core.

Some other minor fixes were required to get this working due to apparent PAM stack differences in the el10 image.

### DIFF
--- a/src/test/docker/el10/Dockerfile
+++ b/src/test/docker/el10/Dockerfile
@@ -1,0 +1,18 @@
+FROM fluxrm/flux-core:el10
+
+ARG USER=flux
+ARG UID=1000
+USER root
+
+# copy scripts into image
+COPY scripts/ /scripts
+# Install extra buildrequires for flux-sched:
+RUN /scripts/install-deps.sh -y --allowerasing \
+ && dnf clean all
+
+# Add configured user to image with sudo access:
+#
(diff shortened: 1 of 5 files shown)


src/test/docker/el10/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN /scripts/install-deps.sh -y --allowerasing \
+ && dnf clean all
+
+# Add configured user to image with sudo access:
+#
+RUN /scripts/add_docker_user.sh
+
+USER $USER
+WORKDIR /home/$USER
+ENV CHECK_RUN_SOURCE_ENV=/opt/rh/gcc-toolset-13/enable
```
trws: Do we need this? I thought you mentioned the other day that el10 is the first one where we don't need the toolkit @grondo.

grondo: Oh, good catch, I missed that when moving the el9 Dockerfile to el10. Will fix.

grondo: default GCC is 14.3.1 in the el10 image FYI

trws: Sounds good!

trws (approved):

grondo:
@mergifyio refresh

PR #1453, opened 2026-04-18
leave interner open after resource module unload

### LINKED ISSUE #1443: Updating Fluxion to v0.50.0 causes interner to abort
On Hetchy, I updated the `flux-sched` RPM and then ran the following:

```
	flux module remove -f sched-fluxion-qmanager &&
	flux module remove -f sched-fluxion-resource &&
	flux module reload resource &&
```

The broker crashed:

```
Apr 03 01:32:11 hetchy1 flux[1982448]: terminate called after throwing an instance of 'std::system_error'
Apr 03 01:32:11 hetchy1 flux[1982448]:   what():  This interner is finalized and must be open to add strings, found new string: 'storage_node': Cannot allocate memory
Apr 03 01:32:13 hetchy1 systemd[1]: flux.service: Main process exited, code=dumped, status=6/ABRT
Apr 03 01:32:13 hetchy1 systemd[1]: flux.service: Failed with result 'core-dump'.
```

When `systemd` restarted the broker a few seconds later, everything came up fine.

Similar issue: #1308

trws:
In @1443 we had a crash caused by the interner still being in a finalized state when the resource module was reloaded.  That should have been cleared by destructors, but on some platforms (like alpine) they do not run, and there must be another corner case where the data lib isn't fully reloaded.

Solution: Add an unfinalize for the dense interner and call it when the resource module is being unloaded, this way initialization on a reload can proceed without issue.  I also added a test that we do and do not get exceptions in appropriate conditions as part of the interner tests.

fixes #1443

### DIFF
--- a/resource/modules/feasibility.cpp
+++ b/resource/modules/feasibility.cpp
@@ -293,9 +293,9 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
         return -1;
     }
 
-    // prevent users from consuming unbounded memory with arbitrary resource types
-    subsystem_t::storage_t::finalize ();
-    resource_type_t::storage_t::finalize ();
+    // No finalize for string interners here, because the graph was initialized
+    // by resource, and this one has to match. Finalizing again here may work,
+    // but could cause a race condition in rare cases.
     return 0;
 }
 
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
LGTM!

PR #1454, opened 2026-04-18
traversers: `storage_node` constraints

### LINKED ISSUE #1447: Clusters with `storage_node` type vertices ignore queue constraints
On hetchy, users found that by running `flux run hostname` they could be allocated a rabbit node, even though rabbit nodes are independent of any queues. For example:

```
[corbett8@hetchy1001:~]$ flux resource list
     STATE QUEUE          PROPERTIES NNODES NCORES NGPUS NODELIST
      free bardpeak,pall                  1     64     8 hetchy1002
      free parrypeak,pall                11   1056    44 hetchy[1007-1013,1015-1018]
      free                rabbit          2    128     0 hetchy[201-202]
 allocated parrypeak,pall                 2    192     8 hetchy[1005-1006]
      down parrypeak,pall                 3    288    12 hetchy[1003-1004,1014]
[corbett8@hetchy1001:~]$ flux run hostname
flux-job: fxjZLrg9uH started                                                00:00:02
hetchy201
[corbett8@hetchy1001:~]$ flux run -qparrypeak hostname
flux-job: fxjjk8DuqZ started                                                00:00:02
hetchy201
```
Note `hetchy201` is a rabbit.

I think the bug is a result of the fact that rabbits are in the resource graph as the `storage_node` type. The job doesn't explicitly request the `node` type (the jobspec just has `slot` and `core` entries), so Fluxion is free to match the `storage_node` type. However, somehow the constraints check is bypassed.

jameshcorbett:
Fixes #1447.

### DIFF
--- a/resource/traversers/dfu.cpp
+++ b/resource/traversers/dfu.cpp
@@ -141,7 +141,7 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
                                std::unordered_map<resource_type_t, int64_t> &dfv)
 {
     int64_t t = 0;
-    int64_t sched_iters = 1;  // Track the schedule iterations in perf stats
+    int64_t sched_iters = 0;  // Track the schedule iterations in perf stats
     int rc = -1;
     size_t len = 0;
     std::vector<uint64_t> agg;
@@ -154,6 +154,7 @@ int dfu_traverser_t::schedule (Jobspec::Jobspec &jobspec,
     if ((rc = request_feasible (meta, op, root, dfv)) < 0)
         goto out;
 
+    sched_iters++;
     if ((rc = traverser->select (jobspec, root, meta, x)) == 0) {
         m_total_preorder = traverser->get_preorder_count ();
         m_total_postorder = traverser->get_postorder_count ();
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -225,9 +225,10 @@ int dfu_impl_t::by_status (const jobmeta_t &meta, vtx_t u)
 int dfu_impl_t::by_constraint (const jobmeta_t &meta, vtx_t u)
 {
     int rc = 0;
-    //  RFC 31 constraints only match against type == "node"
+    //  RFC 31 constraints only match against type == "node" and type == "storage_node"
     //  unspecified constraint matches everything
-    if (meta.constraint != nullptr && (*m_graph)[u].type == node_rt
+    if (meta.constraint != nullptr
+        && ((*m_graph)[u].type == node_rt || (*m_graph)[u].type == storage_node_rt)
         && !meta.constraint->match ((*m_graph)[u])) {
         rc = -1;
     }
--- a/t/CMakeLists.txt
+++ b/t/CMakeLists.txt
@@ -75,6 +75,7 @@ set(ALL_TESTS
   t3039-resource-force-load-format.t
   t3040-jgf-shorthand-match-format.t
   t3041-jgf-shorthand-load-format.t
+  t3042-storage-node.t
   t3300-system-dontblock.t
   t3301-system-latestart.t
   t4000-match-params.t
--- a/t/t3042-storage-node.t
+++ b/t/t3042-storage-node.t
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+test_description='Test matching the storage_node type'
+
+. $(dirname $0)/sharness.sh
+
+query="../../resource/utilities/resource-query"
+jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/rabbit.json"
+
+test_expect_success 'resource-query rejects jobspec asking for a node if only storage_node matches' '
+    flux run -N1 -n1 --requires=hosts:hetchy201 --dry hostname | jq . > node_requires_host.json &&
+    cat > query_cmd1 <<-EOF &&
+m allocate node_requires_host.json
+quit
+EOF
+    ${query} -L ${jgf} -f jgf < query_cmd1 > out1 &&
+    grep "No matching resources" out1
+'
+
+test_expect_success 'resource-query allocates storage node if jobspec only asks for cores' '
+    flux run -n1 --requires=hosts:hetchy201 --dry hostname | jq . \
+        > core_requires_host.json &&
+    cat > query_cmd2 <<-EOF &&
+m allocate core_requires_host.json
+quit
+EOF
+    ${query} -L ${jgf} -f jgf < query_cmd2 > out2 &&
+    test_must_fail grep "No matching resources" out2 &&
+    grep hetchy201 out2
+'
+
+test_expect_success 'resource-query rejects storage node if it fails property constraint' '
+    flux run -n1 --requires="hosts:hetchy201 and parrypeak" --dry hostname  | jq . \
+        > core_requires_host_and_parrypeak.json &&
+    cat > query_cmd3 <<-EOF &&
+m allocate core_requires_host_and_parrypeak.json
+quit
+EOF
+    ${query} -L ${jgf} -f jgf < query_cmd3 > out3 &&
+    grep "No matching resources" out3
+'
+
+test_expect_success 'resource-query accepts storage node with or-rabbit property constraint' '
+    flux run -n1 --requires="hosts:hetchy201 and (parrypeak or rabbit)" --dry hostname  | jq .\
+        > core_requires_host_and_parrypeak_or_rabbit.json &&
+    cat > query_cmd4 <<-EOF &&
+m allocate core_requires_host_and_parrypeak_or_rabbit.json
+quit
+EOF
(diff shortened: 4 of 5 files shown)


jameshcorbett:
@garlick I know you want this change reverted so I would like if possible to get your (begrudging) approval for this PR. Here is my argument for it:

- The code changes are simple, in total they amount to like four lines across this PR and the one that introduced the new vertex type. As far as hacks go, this one is pretty clean.
- A solution that uses some new kind of property matching with selective node exclusivity would not be trivial, jobspec might need to be extended to support it, and we have much higher priority things to tackle in Fluxion, like subsystems and a new match format. At the same time, we have a user (Marty) who really wants to try using the rabbits as compute resources and sell an important code team on it, and he wants it as soon as possible. We've had other users asking as well.
- Potential damage from rollout is minimal: we can try it out on Hetchy for a few weeks, and then enable it on rzadams or maybe even rzvernal if that would be enough for Marty. It's not something that is enabled just by upgrading the RPM; it requires configs to be changed.

trws (changes requested):
These changes look good as far as they go, but you missed the check in the "quick feasibility" function (typing on a phone and don't have the function name in front of me) that one is going to be a bit harder because it's a loop over nodes right now, and needs to be a loop over nodes and storage nodes for this to work. It probably solves part of the problem already because the storage nodes will be rejected above cores in the main traversal, but they'll be inappropriately accepted right now by the early check. We would also inappropriately reject jobs that require properties on storage_node vertices right now I think.

jameshcorbett:
I rebased on top of #1453 because otherwise the tests in https://github.com/flux-framework/flux-sched/pull/1454/commits/1a3244f70c95fd3d018ebca1c305145ee0685dbf failed when reloading a module with new resource types. However, the tests I added in that commit pass even without the changes introduced by https://github.com/flux-framework/flux-sched/pull/1454/commits/b22642f8c17cc5b100cd115565430476bf685af5. 

As I was working on this and trying to make sense of it, I started to wonder--do we need to change feasibility? The problem we saw was that users were submitting valid, feasible jobs but being allocated the wrong resources. I don't see feasibility really coming into it.

resource/traversers/dfu.cpp
```diff
@@ -88,10 +88,21 @@ int dfu_traverser_t::request_feasible (detail::jobmeta_t const &meta,
...
+        all_nodes.insert (all_nodes.end (), node_iter->second.begin (), node_iter->second.end ());
+    }
+    const auto storage_node_iter = get_graph_db ()->metadata.by_type.find (storage_node_rt);
+    if (storage_node_iter != get_graph_db ()->metadata.by_type.end ()) {
+        all_nodes.insert (all_nodes.end (),
+                          storage_node_iter->second.begin (),
+                          storage_node_iter->second.end ());
+    }
+
+    if (all_nodes.empty ())
```
trws: I think the vector might be replaceable with something like:

```c++
auto all_nodes = std::views::concat (node_iter->second, storage_node_iter->second);
```

If so, it would avoid having to actually materialize a new vector every time.

jameshcorbett: That's a C++ 26 feature I believe!

jameshcorbett:
Rebased with master since #1453 was merged, and force-pushed

resource/traversers/dfu.cpp
trws: Ah crap, sorry. 🤦
Occupational hazard I guess, this is part of rangev3 which became std::range/view in 20, I missed that this view didn't make it for so long.

It should be essentially equivalent to do something like:

```c++
std::array<std::vector<...> &, 2> node_vecs = {nodes, storage_nodes};
auto all_nodes = std::ranges::join_view(node_vecs);
```

It's not quite as nicely self-documenting, but this should give us the same benefit of avoiding the allocation in here.  If not we'll just eat it, but this is meant to be the tightest fast-path possible so if we can avoid a set of allocations it would be nice.

jameshcorbett: Ok, I force-pushed a change with the best Claude came up with--I don't know these view/no-copy functions and logic that well.

trws (approved):
Works for me, it's still an allocation but a much smaller single one rather than potentially several. I'm ok with trying this, but would like to hear from @garlick before we merge.

jameshcorbett:
I would still like your input @trws on my [earlier comment](https://github.com/flux-framework/flux-sched/pull/1454#issuecomment-4274949438)--why would the tests introduced in the fourth commit pass without the changes in the third commit? It doesn't make sense to me and it also means the code isn't tested properly. 

Without the changes in the _first_ commit, all but one of the tests in the _fourth_ commit pass. Specifically I get `not ok 26 - feasibility rejects storage_node with incorrect property constraint`.

Unfortunately it looks like the jammy builder hit a nasty compilation error :(

garlick:
Does representing rabbits as `storage_node` vs `node` have anything to do with whether the rabbit node can be an execution target (e.g. be drained, have a broker)?

If @trws is on board here, then I'm OK.  It seemed like he was not before, and I was raising opposition earlier in his absense to fixing the  first unintended consequence for something that it sound like he was unsure about as a long term solution.   I probably don't understand all the trade offs to have an informed opinion

jameshcorbett:
> Does representing rabbits as storage_node vs node have anything to do with whether the rabbit node can be an execution target (e.g. be drained, have a broker)?

No. `hetchy[201-202]` are rabbits and are represented as `storage_node` in fluxion:

```
[corbett8@hetchy1001:~]$ flux resource list
     STATE QUEUE          PROPERTIES NNODES NCORES NGPUS NODELIST
      free bardpeak,pall                  1     64     8 hetchy1002
      free parrypeak,pall                14   1344    56 hetchy[1004-1013,1015-1018]
 allocated parrypeak,pall                 1     96     4 hetchy1003
      down parrypeak,pall                 1     96     4 hetchy1014
      down                rabbit          2    128     0 hetchy[201-202]
[corbett8@hetchy1001:~]$ flux resource drain
TIME         STATE    REASON                         NODELIST
Apr20 10:20  drained* AUTO:PCR:2:PowerCycled         hetchy1014
Apr13 14:33  drained  Kalan: Draining to stop jobs   hetchy201
Apr13 14:33  drained* Kalan: Draining to stop jobs   hetchy202
```

trws:
> I would still like your input @trws on my [earlier comment](https://github.com/flux-framework/flux-sched/pull/1454#issuecomment-4274949438)--why would the tests introduced in the fourth commit pass without the changes in the third commit? It doesn't make sense to me and it also means the code isn't tested properly.
> 
> Without the changes in the _first_ commit, all but one of the tests in the _fourth_ commit pass. Specifically I get `not ok 26 - feasibility rejects storage_node with incorrect property constraint`.
> 
> Unfortunately it looks like the jammy builder hit a nasty compilation error :(

I'd like to talk to you about this one in voice if we can, I'm not sure I'm parsing the question, but I could see how something like this could happen if I'm understanding right.

trws:
> Does representing rabbits as `storage_node` vs `node` have anything to do with whether the rabbit node can be an execution target (e.g. be drained, have a broker)?

They're represented in core as normal ranks, in a sense this is currently fibbing to fluxion to make them a "node-like but non-node" type.  I'm not fond of this longer term, but the way the change is it shouldn't be able to impact any system that doesn't use that type, so I'm alright with it as essentially a feature-flagged option.

jameshcorbett:
Force-pushed to fix a compilation issue on jammy. @trws the way I intended this PR to be laid out was like

1. code changes A
2. Tests for A
3. code changes B
4. Tests for B

But the tests for B pass without the actual B code changes. So B isn't being tested in this PR as it stands, but I don't know why or what I should change, and I was hoping you would be able to help.

jameshcorbett:
After working on this more, trying to get feasibility to work properly and be tested, I got stuck in the weeds and started making more and more changes in the code base, which is what I wanted to avoid. I now propose that this PR go in without any changes to feasibility. Here is why:

1. As it stands, this PR resolves #1447, which is the primary objective.
2. The jobspecs causing the problems in #1447 were feasible and satisfiable, as is desired, the problem was that they were being matched with incorrect resources.
3. The jobspecs causing problems in #1447 did not request any specific number of nodes, so in any case, the `request_feasible` function would always pass for those jobspecs because the `target_nodes` variable in that function would be 0.
4. The only valid way to request the `storage_node` type under current configuration is to request it along with compute nodes. So feasibility could catch some of the more obvious mistakes there. Also, we don't expect many of these kinds of jobs in the near future.
5. The `request_feasible` function just checks whether there are enough feasible nodes. For this use cases it would need to be much more sophisticated, it would need to check whether there are enough feasible chassis where each chassis has one feasible storage node and some number of feasible compute nodes.

trws:
Are the tests added in [t: add tests for storage_node in request_feasible](https://github.com/flux-framework/flux-sched/pull/1454/commits/d4ede636e80d045623b72eca82215053f8561c11) still relevant?

jameshcorbett:
Force-pushed to combine the two testing commits and drop a reference to `request_feasible` in the tests!

trws (approved):

PR #1455, opened 2026-04-21
cmake: bump jansson requirement to v2.11

sam-maloney:
Problem: `resource/writers/match_writers.cpp` uses features from `jansson` that were added in v2.11, but only v2.10 is currently required in `CMakeLists.txt`

Bump required version to 2.11

Specifically, some of the `json_pack` format strings use the `o*` and `s*` specifiers, [which were added in version 2.11](https://jansson.readthedocs.io/en/latest/apiref.html#c.json_int_t:~:text=s*%20(string)%20%5Bconst%20char%20*%5D)

The new specifiers were added in https://github.com/flux-framework/flux-sched/pull/1299


trws (approved):
Since core merged the same, delighted to. Thanks @sam-maloney!

PR #1456, opened 2026-04-21
resource: match int types

sam-maloney:
Problem: a couple of values are converted from int64_t to uint64_t and then back again

Make them int64_t throughout

I was looking through #1435 and was wondering why these types seemed to be randomly mixed?

`m_starttime` and `m_expiration` are `int64_t` in https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.hpp#L287-L288
as is `jobmeta.at` in https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl.hpp#L46

So it seems like `jobmeta.at` is being converted to `uint64_t` in https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L864-L866
only to match the signature of `emit_tm (uint64_t, uint64_t)` in https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.cpp#L1113-L1118
but this function immediately stores its values back into the `m_starttime` and `m_expiration` variables of `int64_t` as noted above


sam-maloney:
Apparently I should not have tried to use the GitHub web IDE because I somehow managed to create this branch on the main repo instead of my own fork, mea culpa :upside_down_face:  Will close and re-open once I move the branch.

PR #1457, opened 2026-04-21
resource: match int types

sam-maloney:
Problem: a couple of values are converted from int64_t to uint64_t and then back again

Make them int64_t throughout

I was looking through #1435 and was wondering why these types seemed to be randomly mixed?

`m_starttime` and `m_expiration` are `int64_t` in [`rv1_match_writers_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.hpp#L287-L288) and [`rv1_nosched_match_writers_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.hpp#L311-L312), as is [`jobmeta.at`](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl.hpp#L46). The only class variable that is currently a `uint64_t` is [`jobmeta.duration`](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl.hpp#L48) and that discrepancy actually leads to a [type mismatch in this operation](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L865), and it is even initialized to `SYSTEM_DEFAULT_DURATION` [which is itself an `int64_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L18).

Overall it seems like [`jobmeta.at` is being converted to `uint64_t`](https://github.com/flux-framework/flux-sched/blob/master/resource/traversers/dfu_impl_update.cpp#L864) only to match the signature of [`emit_tm (uint64_t, uint64_t)`](https://github.com/flux-framework/flux-sched/blob/master/resource/writers/match_writers.cpp#L1113-L1118) but this function immediately stores its values back into the `m_starttime` and `m_expiration` variables of type `int64_t` as noted above, so it seems like just keeping everything as `int64_t` throughout would be more consistent?

### DIFF
--- a/resource/traversers/dfu_impl.hpp
+++ b/resource/traversers/dfu_impl.hpp
@@ -45,7 +45,7 @@ struct jobmeta_t {
     int64_t jobid = -1;
     int64_t at = -1;
     int64_t now = -1;
-    uint64_t duration = SYSTEM_DEFAULT_DURATION;  // will need config ultimately
+    int64_t duration = SYSTEM_DEFAULT_DURATION;  // will need config ultimately
     std::shared_ptr<Jobspec::Constraint> constraint;
 
     bool is_queue_set () const
--- a/resource/traversers/dfu_impl_update.cpp
+++ b/resource/traversers/dfu_impl_update.cpp
@@ -861,8 +861,8 @@ int dfu_impl_t::update (vtx_t root, std::shared_ptr<match_writers_t> &writers, j
     bool emit_shadow = modify_traversal (root, false);
     // Regardless of value of `x`, value for `excl_parent` parameter starts as `false`
     if ((rc = upd_dfv (root, writers, needs, x, jobmeta, true, dfu, emit_shadow, false)) > 0) {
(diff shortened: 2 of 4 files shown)


trws (approved):
These are probably relics, IIRC back in the distant past we used uint64_t for all of out time until we started to use chrono in more places and switched to signed so we could catch negative cases rather than having them be huge values.

Thanks for the cleanup!

trws:
Oh, I should have asked, I think we gave you enough access to be able to set `merge-when-passing` for yourself on this PR now that you have an approving review.  If something doesn't work with that, poke me and I'll set it and see what's going on with the permissions.

sam-maloney:
> Oh, I should have asked, I think we gave you enough access to be able to set `merge-when-passing` for yourself on this PR now that you have an approving review. If something doesn't work with that, poke me and I'll set it and see what's going on with the permissions.

Seems to have worked, thanks!

PR #1458, opened 2026-04-21
cmake preset: use colons to separate ASAN options

chu11:
Problem: Multiple ASAN options are separated by comma in the CMake Presets file.  They should be separated by colon.

Update ASAN_OPTIONS to use colons.

----

mirroring fix in flux-core https://github.com/flux-framework/flux-core/pull/7553


trws (approved):
Quite right, thanks @chu11!

PR #1461, opened 2026-04-24
NEWS.md: add release notes for v0.50.1

jameshcorbett:
There have been a bunch of important fixes since `0.50.0`, and for TOSS updates in the coming weeks it would be good to include those fixes if possible.

@garlick I think you had mentioned a preference for only having very targeted patches to the existing RPM but a lot of these seem either very useful or at worst harmless.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,18 @@
+flux-sched version 0.50.1 - 2026-04-24
+--------------------------------------
+
+### Fixes
+ * traversers: `storage_node` constraints (#1454)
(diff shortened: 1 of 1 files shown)


garlick:
"Very useful or at least harmless" falls short of the justification needed to bypass our normal test/soak cycles IMO.  It should be "Fixes a critical problem observed in production".

For such changes, what we would normally do is extract the minimum fix as a patch and roll a new SRPM with the release number (last digit) incremented.

Edit: an SRPM based on the current version targeted for production that is

PR #1463, opened 2026-04-28
resource: bulk set- and remove-property rpcs

### LINKED ISSUE #1460: change `set-` and `remove-property` to accept multiple vertices
Similar to #1201 , the `flux-coral2` project could send many fewer total RPCs if it could bundle multiple targets into a single `set-` or `remove-property` RPC.

jameshcorbett:
Fixes #1460.

### DIFF
--- a/resource/modules/resource.cpp
+++ b/resource/modules/resource.cpp
@@ -858,23 +858,58 @@ static void set_property_request_cb (flux_t *h,
                                      const flux_msg_t *msg,
                                      void *arg)
 {
-    const char *rp = NULL, *kv = NULL;
-    std::string resource_path = "", keyval = "", errmsg = "";
+    const char *kv = NULL;
+    json_t *paths_json = NULL;
+    std::string keyval = "", errmsg = "";
     std::string property_key = "", property_value = "";
+    std::vector<std::string> resource_paths;
     size_t pos;
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
     std::map<std::string, std::vector<vtx_t>>::const_iterator it;
     std::pair<std::map<std::string, std::string>::iterator, bool> ret;
     vtx_t v;
 
-    if (flux_request_unpack (msg, NULL, "{s:s s:s}", "sp_resource_path", &rp, "sp_keyval", &kv)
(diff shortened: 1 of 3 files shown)


trws:
I read through this but keep not having time to write up the response, this is the big thing:
Currently if any of the sets or unsets fail, all the remaining are silently skipped. We need at least the offset/item that failed in the error response to allow for ratcheting retry.

jameshcorbett:
> I read through this but keep not having time to write up the response, this is the big thing: Currently if any of the sets or unsets fail, all the remaining are silently skipped. We need at least the offset/item that failed in the error response to allow for ratcheting retry.

That makes sense. The RPC error message does give the first resource path that failed, but then whatever receives the response would have to parse that error message. Do you have a suggestion for how it could be improved?

trws:
Ok, I see what you mean in that it puts it in the message. That's not what I meant though.  The response should be a json payload that actually has the offset and/or the resource path as a key/value pair so the caller can retry or continue.  As of now I think the right solution is to use `flux_respond_pack` for partial success rather than going directly to encode_error, so we can include structured information.

As it sits this looks like it works, and I would be ok with taking this since we have a time crunch if you're willing to start a followup to make it so we can make this more reliable after.  What do you think?

I'm also starting a discussion on the flux-internal slack channel to discuss making this easier in the error case.  That's a really strange gap in our API.

jameshcorbett:
> Ok, I see what you mean in that it puts it in the message. That's not what I meant though. The response should be a json payload that actually has the offset and/or the resource path as a key/value pair so the caller can retry or continue. As of now I think the right solution is to use `flux_respond_pack` for partial success rather than going directly to encode_error, so we can include structured information.
> 
> As it sits this looks like it works, and I would be ok with taking this since we have a time crunch if you're willing to start a followup to make it so we can make this more reliable after. What do you think?


Sure, that sounds good. Let's do that as a follow-up -- I want to make sure this gets in the next TOSS release so that flux-coral2 can start leveraging it ASAP. I'd be happy to do that follow-up work and I don't think it would break anything.

I'll make an issue for the follow-up work, are you willing to approve this PR?

trws (approved):

trws:
Approved, and I saw the issue. Thanks!

PR #1465, opened 2026-04-29
[WIP] Add support for non-exclusive slot resources

milroy:
This PR adds the capability to specify non-exclusive resources under a slot. The use case supports rabbit allocations as well as other scenarios where a job might need exclusive node access but non-exclusive access to related pool vertices.

Further testing is needed to determine if complex rabbit allocation scenarios are supported by these changes, and how to handle combinations of `exclusive: true` and `exclusive: false` resources in the same slot subtree. The current status is WIP.


milroy:
I mistakenly created a branch on flux-sched rather than my fork. I'm deleting this PR and will open one with the changes made against my fork of flux-sched.

Issue #1033, opened 2023-05-16
tracking: locality aware scheduling in Fluxion


grondo:
This is a tracking issue for handling locality aware scheduling of on-node resources (currently only GPUs I think) in Fluxion.
This is a requirement for our Production Ready System Instance milestone, tracked in this project: 

https://github.com/orgs/flux-framework/projects/33

It may be that the desired effect is already achievable by adding some topology information to R or JGF for Fluxion.
We can discuss solutions at that level directly in this issue. 

Related issues:
 [ ] - #986

grondo:
Some notes from the meeting
 - Currently node local topology information is lost to Fluxion because we initialize from configured R, which has a flat list of core ids and gpu ids on a node.
 - @trws noted in the meeting that even if we can recover topology information in Fluxion for locality aware scheduling of cores and gpus, gpu logical ids may not be consistent between system restarts. We need a way to ensure a logical GPU id on a node is consistent between the scheduler and the job shell (or mpibind shell plugin for that matter).

grondo:
FYI - a user wants to run 2 jobs on a system, each using 4 cores and 4 gpus, but they are hitting a roadblock due to this issue. If they request two jobs each with 4 cpus and 4 gpus, then they each get different GPUs, but the cores are not necessarily adjacent to the allocated GPUs, and in fact the sets of cores will likely be sequential.

Until we fully resolve this issue, is there some easy way to rewrite the instance R in a batch job to include JGF with the full topology? Then we might be able to get this user going. @milroy @trws?

grondo:
This came up again in yesterday's dev meeting.

In production we currently rely on mpibind to choose the "best" mapping of tasks to cores based on a heuristic. However, mpbind is restricted to the cores handed to it by Flux, and in the case where mpibind doesn't have access to the whole node, it has to make bad choices.

What we're looking for here is a way to seed Fluxion with the on-node topology for batch jobs so that it is able to do better core selection given, e.g. NUMA node layout, GPU locality, etc.

Any ideas here would be appreciated.

cc: @ryanday36

PR #1470, opened 2026-05-02
resource: support amending R with JGF in a subinstance

grondo:
While investigating flux-framework/flux-core#7521 and #1033, I hit a roadblock because sched-fluxion-resource currently _requires_ that a subinstance job's R was originally encoded with JGF, even when the R has been amended with JGF before the Fluxion resource module has been loaded.

This PR works around that issue by skipping namespace remapping unless the parent R has JGF. If JGF is present R but missing in the parent's record of R, assume that the JGF doesn't need remapping.

This was enough to get my tests working (example use case added as a sharness test), so this could be merged now, but hopefully with Rv2 this double fetch of the job's R from parent can be removed.

### DIFF
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -883,7 +883,9 @@ done:
     return rc;
 }
 
-static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx, json_t **p_r_lite_p)
+static int unpack_parent_job_resources (std::shared_ptr<resource_ctx_t> &ctx,
+                                        json_t **p_r_lite_p,
+                                        bool *parent_has_jgf_p)
 {
     int rc = 0;
     int saved_errno;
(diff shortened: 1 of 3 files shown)


resource/modules/resource_match.cpp
```diff
@@ -930,6 +930,13 @@ static int grow_resource_db_jgf (std::shared_ptr<resource_ctx_t> &ctx, json_t *r
     char *jgf_str = NULL;
     vtx_t v = boost::graph_traits<resource_graph_t>::null_vertex ();
 
+    // Validate that jgf is a proper JGF object with a graph key
+    if (!jgf || !json_is_object (jgf) || !json_object_get (jgf, "graph")) {
+        errno = EINVAL;
+        flux_log_error (ctx->h, "%s: invalid JGF in scheduling section", __FUNCTION__);
+        goto done;
+    }
```
jameshcorbett, line 938: commit has a message about updating a test but that change isn't in this commit?

jameshcorbett (approved):
LGTM one question about a commit message

t/t5200-jgf-load-no-parent-jgf.t
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
+
+test_description='Test loading JGF when parent has only R_lite (no JGF)'
+
+. `dirname $0`/sharness.sh
+
+if ! command -v hwloc-ls >/dev/null; then
+	skip_all='Skipping JGF load tests because hwloc-ls not found'
+	test_done
```
jameshcorbett, line 9: Test name is incorrect in commit message (t5999 instead of t5200)

resource/modules/resource_match.cpp
grondo: Good catch, some commit message references were from a previous version of the branch.

grondo:
I've fixed those two issues and force pushed the result. Will set MWP now, thanks!

PR #1480, opened 2026-05-08
ci: fix fluxrm/flux-sched:latest docker image tag

grondo:
Problem: The docker-deploy.sh script pushes fluxrm/flux-sched:latest by tagging a single-arch jammy image, which does not work correctly with the multiarch manifest-based workflow.

Copy the fixed docker-deploy.sh and docker-manifest.py scripts from flux-core, modified for flux-sched. Choose noble as the distro for the `latest` tag by request.


jameshcorbett (approved):
LGTM

PR #1481, opened 2026-05-12
WIP: python: add reapi_cli bindings

garlick:
This adds a python class wrapper for `reapi_cli`.  The intent is to use this to implement a `ResourcePool`  class that would let the flux-core python schedulers use fluxion's resource graph, as discussed in

Marking as a WIP while I try to build the pool class and shake out any missing details, but any feedback would be welcome.

I tried to follow the same patterns used for cffi in flux-core, although I'm admittedly in unfamiliar territory and may have done something dumb!  Goodness knows claude did many dumb things while this was iterated, and with only me to call out cmake/python/cffi/fluxion issues, heh, maybe you should be afraid.


trws:
I’m excited to look this one over! I need to prioritize the non-exclusive slots PR this morning but I’ll get this one started today too.

Get Outlook for iOS<https://aka.ms/o0ukef>

garlick:
OMG, I just duplicated a lot of @vsoch's work over in
- #1420 

We'll have to decide if the CFFI vs cython approach is preferable.

garlick:
Please hold off on reviewing this - I'm redoing a few things after making a first pass on a pool class.

trws:
Ok, we should talk about the alternatives by the way. There’s a meaningful tradeoff between cython and cffi here and the answer could well be that we want to do some of each or one or the other. The short form version is this:

  *   cffi is less extra dependencies and makes it easier to interoperate with the core python bindings
  *   cython can directly bind to c++, including directly instantiating templates, so if we ever want to bind to anything that isn’t in the C-based API, this could be a meaningful benefit

garlick:
Now based on top of
- #1484

And rebased on master which now includes:
- #1483

garlick:
This has gone really stale.  I'll reopen once we've gotten further along with this PR's prerequisites which have been stalled for a while.

PR #1483, opened 2026-05-22
resource: fix errno violations

garlick:
This fixes some errno problems identified by @milroy and me in
- #1467
- #1481

I spent some time with claude spelunking various functions called out by Dan and came up with this in lieu of the `errno = 0` workarounds originally proposed.

`match_allocate()` was pretty convoluted and had several latent bugs in addition to the egregious one in which it would expect callers to test errno when it returned success.  I thought it best to restructure that function a bit rather than just do targetted fixes.


jameshcorbett (approved):
LGTM!

garlick:
Thanks @jameshcorbett !  We probably need approval from @milroy as well since I changed his original approach.

garlick:
P.s. there are more of this type of problem in the planner code.   I'll submit another batch later today if all goes well.

milroy (approved):
These changes look good to me, although I'm less familiar with the code in the qmanager commit. Is there some way to test the changes? Nothing comes to mind immediately but will think about it today.

milroy:
> P.s. there are more of this type of problem in the planner code.

I'm not surprised and suspect this problem exists in many more code locations. I'll keep digging as well.

garlick:
Good question about testing.   Maybe some libtap tests are possible here.  We have some tests in flux-core that set errno = 0, call a function in a way that makes it fail, and then verify that it failed and errno now holds the expected value.

trws:
Catch2, or tap if one must, tests would be the way yeah. Especially for this kind of API contract level stuff.

trws:
Wait to merge this until I can look over the  qmanager thing. It should be more correct this way, but I remember having a reason for adding that and want to be sure.

trws (approved):
Looks good, thank you both for these.  The errno propagation from `remove` on a queue policy may change, but there's no code path where that should be a negative.

garlick:
Thanks!  I'll go ahead and set MWP and get the next batch rebased.

PR #1484, opened 2026-05-22
fix more errno violations

garlick:
More claude spelunking for errno anti-patterns.

This is based on top of
- #1483

Claude's smmary of errno violations fixed:

✅ Planner NULL dereferences (3 instances)
✅ Planner type error inverting returns
✅ Planner missing errno in avail_during()
✅ Qmanager spurious errno = 0
✅ Pruning functions (by_avail, by_excl, by_subplan) - removed save/restore patterns
✅ Prime_pruning_filter - removed save/restore pattern
✅ Policy dom_finish callbacks - missing errno on -1 returns (4 functions)
✅ Policy set_stop_on_k_matches - missing errno
✅ Traverser resolve_graph - missing errno
✅ Traverser dom_dfv - removed defensive errno checks
✅ Traverser dfu_impl_update - missing errno
✅ Reapi cancel functions - removed save/restore patterns
✅ Writer emit_vtx_base - missing errno after json_object_set_new failures

  Legitimate patterns kept:
  - Save/restore in destructors/cleanup functions
  - Save/restore in utility functions returning bool (HostlistConstraint::match)
  - errno = 0 before C++ I/O operations
  - Readers/writers using m_err_msg instead of errno
  - CLI tools using std::cerr instead of errno 
  - Comparator functions returning -1/0/1 for ordering

### DIFF
--- a/resource/planner/c/planner_c_interface.cpp
+++ b/resource/planner/c/planner_c_interface.cpp
@@ -194,6 +194,7 @@ static bool avail_during (planner_t *ctx, int64_t at, uint64_t duration, const i
             ok = true;
             break;
         } else if (request > point->remaining) {
+            errno = EBUSY;
             ok = false;
             break;
         }
--- a/resource/planner/c/planner_c_interface.cpp
+++ b/resource/planner/c/planner_c_interface.cpp
@@ -58,6 +58,10 @@ static scheduled_point_t *get_or_new_point (planner_t *ctx, int64_t at)
     try {
         if (!(point = ctx->plan->sp_tree_search (at))) {
             scheduled_point_t *state = ctx->plan->sp_tree_get_state (at);
+            if (!state) {
+                errno = EINVAL;
+                return nullptr;
+            }
             point = new scheduled_point_t ();
             point->at = at;
             point->in_mt_resource_tree = 0;
--- a/resource/planner/c/planner_c_interface.cpp
+++ b/resource/planner/c/planner_c_interface.cpp
@@ -211,6 +211,10 @@ static scheduled_point_t *avail_resources_during (planner_t *ctx, int64_t at, ui
     }
 
     scheduled_point_t *point = ctx->plan->sp_tree_get_state (at);
+    if (!point) {
+        errno = EINVAL;
+        return nullptr;
+    }
     scheduled_point_t *min = point;
     while (point) {
         if (point->at >= (at + (int64_t)duration))
@@ -484,6 +488,8 @@ extern "C" int64_t planner_avail_resources_during (planner_t *ctx, int64_t at, u
         return -1;
     }
     min_point = avail_resources_during (ctx, at, duration);
+    if (!min_point)
+        return -1;
     return min_point->remaining;
 }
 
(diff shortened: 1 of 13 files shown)


resource/planner/c/planner_c_interface.cpp
```diff
@@ -194,6 +194,7 @@ static bool avail_during (planner_t *ctx, int64_t at, uint64_t duration, const i
             ok = true;
             break;
         } else if (request > point->remaining) {
+            errno = EBUSY;
```
trws, line 201: This is right, but the commit message is missing the `Solution:` marker.  The structure is otherwise there so not sure hoe much it matters.

resource/planner/c/planner_c_interface.cpp
```diff
@@ -501,6 +501,10 @@ extern "C" int64_t planner_avail_resources_at (planner_t *ctx, int64_t at)
         return -1;
     }
     state = ctx->plan->sp_tree_get_state (at);
+    if (!state) {
+        errno = EINVAL;
```
trws: I'm not sure this is the right errno code for this. Come to think of it I'm not sure the one above is either, except for invalid `ctx`. An error here would mean that somehow there's no representation of available resources at a time within the range. I guess that could be invalid, but ENOENT might be more descriptive? Also wondering if the `at` beyond either start or end should be ERANGE.

resource/traversers/dfu_impl.cpp
```diff
@@ -1278,6 +1278,11 @@ int dfu_impl_t::select (Jobspec::Jobspec &j, vtx_t root, jobmeta_t &meta, bool e
         egrp.edges.push_back (ev_edg);
         dfu.add (dom, (*m_graph)[root].type, egrp);
         rc = resolve_graph (root, j.resources, dfu, excl, &needs);
+        if (rc < 0) {
+            // resolve_graph() doesn't have an errno contract.
+            // Set errno = ENODEV (unsatisfiable) when it fails.
+            errno = ENODEV;
```
trws, line 1294: This seems more right, but makes me nervous.  It's an invalid state at this point, and I think will update the graph afterward, possibly leading to reserved or allocated resources being reported as unsatisfiable.

I think this one needs its own issue.  Could take this now because it's more right than what we're doing but I think you found a potentially deeper problem here.

resource/planner/c/planner_c_interface.cpp
garlick: Did you want Solution: markers in all the commit messages?  Commits in the history seem more likely than not to omit this.  In flux-core we require Problem: but tend to omit Solution:.  Absolutely NP to fix if you prefer that structure.

garlick:
Pushed a couple fixups (code change and unit test change) for the ENOENT/ERANGE review comment which can be squashed if accepted.

resource/planner/c/planner_c_interface.cpp
trws: We don't have to, just trying to make sure we keep things reasonably consistent and wanted to check. They're easy enough to parse for a human now.  Thanks!

trws (approved):
This looks great to me, thanks @garlick!  I put up an issue for the resolve_graph thing, may take some refactoring or reworking to be able to handle that case without a fatal error.

garlick:
Thanks @trws!  Fixups squashed and I'll set MWP.

PR #1487, opened 2026-05-26
reapi: add reapi_cli_clone()

garlick:
This adds clone capability to the reapi_cli.  The immediate use case is forward simulation in the python scheduling framework.  However, it seems like this could also be useful in future unit tests - for example, set up some graph state (laboriously), then try multiple tests on it that mutate the state, each on a fresh copy?

Anyway, I realize that it will be prohibitively expensive to deep copy a large graph on the scheduler critical path, but (a) maybe forward simulation doesn't always need to be on the critical path, and (b) maybe clone could be optimized in the future with copy-on-write or similar.

Peeled off of
- #1481

### DIFF
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -566,6 +566,77 @@ resource_query_t::~resource_query_t ()
 {
 }
 
+static void initialize_matchers_and_traversers (resource_query_t *rq)
+{
+    std::string tmp_err = "";
+
+    if (!(rq->matcher = create_match_cb (rq->params.matcher_policy))) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't create matcher\n";
+        throw std::runtime_error (tmp_err);
+    }
+    if (rq->set_subsystems_use (rq->params.matcher_name) != 0) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't set subsystems\n";
+        throw std::runtime_error (tmp_err);
+    }
+    if (rq->params.prune_filters != ""
+        && rq->matcher->set_pruning_types_w_spec (rq->matcher->dom_subsystem (),
+                                                  rq->params.prune_filters)
+               < 0) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't set pruning filters\n";
+        throw std::runtime_error (tmp_err);
+    }
+
+    // Wire a fresh traverser to the new db and matcher.
+    rq->traverser = std::make_shared<dfu_traverser_t> (rq->params.traverser_policy);
+    if (rq->traverser->initialize (rq->db, rq->matcher) != 0) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't initialize traverser\n";
+        throw std::runtime_error (tmp_err);
+    }
+
+    // Create fresh writers (stateless: only carry output format config).
+    match_format_t format = match_writers_factory_t::get_writers_type (rq->params.match_format);
+    if (!(rq->writers = match_writers_factory_t::create (format))) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't create match writers\n";
+        throw std::runtime_error (tmp_err);
+    }
+}
+
+resource_query_t::resource_query_t (const resource_query_t &o)
+{
+    std::string tmp_err = "";
+
+    params = o.params;
+    m_err_msg = o.m_err_msg;
+    jobid_counter = o.jobid_counter;
+    perf = o.perf;
+    jobs = o.jobs;
+    allocations = o.allocations;
+    reservations = o.reservations;
+
+    // Deep-copy the resource graph database; resource_graph_db_t's copy
+    // constructor rebuilds by_outedges with stable descriptors from the new
+    // graph, preserving all allocation state and up/down status.
+    db = std::make_shared<resource_graph_db_t> (*o.db);
+
+    // Reset vertex colors: deep-copied idata.colors hold old-epoch values
+    // from the source traverser.  initialize() bumps m_color_base, making
+    // the old black values appear black again and causing stop_explore to
+    // skip every vertex in prime_pruning_filter, zeroing subtree planners.
+    resource_graph_t &g = db->resource_graph;
+    vtx_iterator_t vi, vi_end;
+    for (boost::tie (vi, vi_end) = boost::vertices (g); vi != vi_end; ++vi)
+        g[*vi].idata.colors.clear ();
+
+    // Re-create the matcher from the stored policy rather than copying
+    // through the base-class pointer, which would slice the concrete subtype.
+    initialize_matchers_and_traversers (this);
+}
+
 resource_query_t::resource_query_t (const std::string &rgraph, const std::string &options)
 {
     m_err_msg = "";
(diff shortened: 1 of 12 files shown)


resource/reapi/bindings/c++/test/reapi_cli_clone_test.cpp
```diff
@@ -0,0 +1,898 @@
...
+            {"id": "0", "metadata": {"type": "cluster", "basename": "tiny", "name": "tiny0", "size": 1, "paths": {"containment": "/tiny0"}}},
+            {"id": "1", "metadata": {"type": "node", "basename": "node", "name": "node0", "size": 1, "rank": 0, "paths": {"containment": "/tiny0/node0"}}},
+            {"id": "2", "metadata": {"type": "core", "basename": "core", "name": "core0", "size": 1, "id": 0, "rank": 0, "paths": {"containment": "/tiny0/node0/core0"}}}
+        ],
+        "edges": [{"source": "0", "target": "1"}, {"source": "1", "target": "2"}]
+    }
+})";
+
+static const char *tiny_params =
+    "{\"load_format\": \"jgf\", \"matcher_policy\": \"high\", "
```
trws: Raw strings above and below, and escaping here seems a bit out of place.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -566,6 +566,86 @@ resource_query_t::~resource_query_t ()
...
+    params.elapse_time = o.params.elapse_time;
+    params.disable_prompt = o.params.disable_prompt;
+    params.flux_hwloc = o.params.flux_hwloc;
+    params.reserve_vtx_vec = o.params.reserve_vtx_vec;
+    m_err_msg = o.m_err_msg;
+    jobid_counter = o.jobid_counter;
+    perf = o.perf;
+    jobs = o.jobs;
+    allocations = o.allocations;
+    reservations = o.reservations;
```
trws: This made me sad, and convinced me to do a bit of digging.  The `r_out` value is never used except to close it. In fact I don't think it's ever even set after some of the refactoring @zekemorton did a while ago.  After #1488 this should be able to be `params = o.params`.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -566,6 +566,86 @@ resource_query_t::~resource_query_t ()
...
+        g[*vi].idata.colors.clear ();
+
+    // Re-create the matcher from the stored policy rather than copying
+    // through the base-class pointer, which would slice the concrete subtype.
+    if (!(matcher = create_match_cb (params.matcher_policy))) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't create matcher\n";
+        throw std::runtime_error (tmp_err);
+    }
+    matcher->set_matcher_name (params.matcher_name);
```
trws: This is covered by the next line.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -566,6 +566,86 @@ resource_query_t::~resource_query_t ()
...
+        throw std::runtime_error (tmp_err);
+    }
+
+    // Create fresh writers (stateless: only carry output format config).
+    match_format_t format = match_writers_factory_t::get_writers_type (params.match_format);
+    if (!(writers = match_writers_factory_t::create (format))) {
+        tmp_err = __FUNCTION__;
+        tmp_err += ": ERROR: can't create match writers\n";
+        throw std::runtime_error (tmp_err);
+    }
```
trws: This block is very close to identical to the corresponding one in the `(graph, options)` constructor, lines 586-645, but just different enough to make them potentially hard to keep in sync. It would make me much more comfortable if they were factored out into a helper function or private method.

resource/reapi/bindings/c/test/reapi_cli_test.c
```diff
@@ -0,0 +1,80 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include "resource/reapi/bindings/c/reapi_cli.h"
+#include "resource/policies/base/match_op.h"
+#include "src/common/libtap/tap.h"
+
+static const char *tiny_jgf =
+    "{"
+    "\"graph\": {"
```
trws, line 25: Raw string would be nice here.

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -85,6 +85,33 @@ extern "C" int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph, c
...
+    if (!ctx || !ctx->rqt) {
+        errno = EINVAL;
+        return nullptr;
+    }
+
+    reapi_cli_ctx_t *clone = nullptr;
+    try {
+        clone = new reapi_cli_ctx_t;
+        clone->rqt = new resource_query_t (*ctx->rqt);
+        return clone;
```
trws: This would be better written as:

```c++
auto clone = std::make_unique<reapi_cli_ctx_t>();
clone->rqt = new resource_query_t (*ctx->rqt);
return clone.release();
```

That way the `delete clone` line can be removed from all the catch branches, because the unique_ptr will clean it up on exception, while the release will give out the pointer and stop the unique_ptr from deleting it.

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -85,6 +85,33 @@ extern "C" int reapi_cli_initialize (reapi_cli_ctx_t *ctx, const char *rgraph, c
...
+    } catch (std::runtime_error &) {
+        delete clone;
+        errno = EPROTO;
+        return nullptr;
+    } catch (...) {
+        delete clone;
+        errno = EINVAL;
+        return nullptr;
+    }
+}
```
trws, line 121: This is a good start, but see the above function for saving off the error string from the `.what` member.
This should include a branch for system_error before the runtime_error branch like this:

```c++
} catch (std::system_error &e) {
  errno = e.code();
  ctx->err_msg += __FUNCTION__;
  ctx->err_msg += ": System error: " + std::string (e.what ()) + "\n";
  return nullptr;
}
```

I'm making a note to make a nice general catch for this file to use, having to redo this over and over is really rough.

resource/schema/infra_data.cpp
```diff
@@ -162,6 +162,7 @@ relation_infra_t::relation_infra_t (const relation_infra_t &o) : infra_base_t (o
     m_needs = o.m_needs;
     m_sequence_number = o.m_sequence_number;
     m_exclusive = o.m_exclusive;
+    m_weight = o.m_weight;
 }
```
trws: This is why I hate these things. 😭 

I'm trying to squash them as I find them, would you mind replacing this and the one below with this:
```suggestion
relation_infra_t::relation_infra_t (const relation_infra_t &o) = default;
```

Replacing the body of the method with `= default;` will make the compiler generate the copy and assignment constructors using the appropriate copy for all bases and members, and we'll never have this problem again.

trws (changes requested):

resource/reapi/bindings/c/reapi_cli.cpp
garlick: Hmm, but the error doesn't apply to the source `ctx` (it is presumably fine) so is storing it there OK?

trws: I'm thinking we treat it much the way errors are treated on a cuda queue for example.  While it isn't specific to the target object, it's the only way we have to persist that error data that's specific to this calling context without making a global or thread_local, which I'd rather avoid, so I think it's alright.  It's mainly important for the `runtime_error`, where without the message it could be any of at least a dozen things.

resource/reapi/bindings/c++/test/reapi_cli_clone_test.cpp
garlick: TIL! :joy:

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: Great point, will fix.

resource/reapi/bindings/c/test/reapi_cli_test.c
garlick: Yes they would (!) but it's a C test. :shrug:

Edit: hmm, well it could be converted I suppose.  However it's testing a C interface so it seems appropriate to do that with a C test.  Thoughts?

trws: Blarg, quite right.  Until we get on c23 so we can use `#embed` I guess that's the best we have.  Not sure why they refuse to take the raw string syntax but... 🤷

garlick:
Thanks for the fast turnaround on that review and the constructive feedback!  I've pushed the suggested changes as fixups - if they look good, LMK and I'll squash.  Also rebased on master after your `r_out` RP was merged.

trws (approved):
Thanks for the quick fixes, this looks great!

I've been looking into whether we can make the graph itself an immutable datastructure recently, if that pans out we could do the clone pretty cheaply.  Certainly doesn't hurt to wire it all through regardless.

garlick:
Merci!  Squashed fixups and I'll set MWP.

PR #1488, opened 2026-05-26
Remove r_out

trws:
in #1487 @garlick noted that `r_out` can't be copy constucted, which is true, but it's also unused. Excise it so we can clean some things up.

I also ran into a header resolution issue exposed by the new errno test as part of this, reapi_cli was inappropriately marking its dependence on flux::core as private.


garlick (approved):
LGTM!

minor: there's a loose CMakeLists.txt whitespace change in the r_out commit (could drop that)

PR #1489, opened 2026-05-27
reapi: add resource status get/set API

garlick:
Problem: reapi_cli doesn't offer interfaces for marking resources up|down, but this will be required for use in the flux-core python scheduler framework as discussed in

This was peeled off of
- #1481

New C++ methods (implemented in reapi_cli, stubbed in reapi_module) EDIT updated 4 June 2026:
```c++
    /*! Set resource status (up or down) for a resource at the specified path.
     *
     *  \param h         Opaque handle. How it is used is an implementation
     *                   detail. However, when it is used within a Flux's
     *                   service module, it is expected to be a pointer
     *                   to a flux_t object.
     *  \param resource_path    Path to resource (e.g., "/cluster0/rack0/node3")
     *  \param status    resource_pool_t::status_t::UP or resource_pool_t::status_t::DOWN
     *  \return          0 on success; -1 on error with errno set:
     *                       EINVAL: invalid parameters, resource path not found,
     *                               or unexpected internal error
     */
    static int set_status (void *h, const std::string &resource_path, resource_pool_t::status_t status);

    /*! Get resource status for a resource at the specified path.
     *
     *  \param h         Opaque handle. How it is used is an implementation
     *                   detail. However, when it is used within a Flux's
     *                   service module, it is expected to be a pointer
     *                   to a flux_t object.
     *  \param resource_path    Path to resource (e.g., "/cluster0/rack0/node3")
     *  \param status    Reference to receive status
     *  \return          0 on success; -1 on error with errno set:
     *                       EINVAL: invalid parameters, resource path not found,
     *                               or unexpected internal error
     */
    static int get_status (void *h, const std::string &resource_path, resource_pool_t::status_t &status);

    /*! Set resource status (up or down) for the subtree root (typically node) at each ranks in a
     * set. Unknown ranks are silently ignored.
     *
     *  \param h         Opaque handle. How it is used is an implementation
     *                   detail. However, when it is used within a Flux's
     *                   service module, it is expected to be a pointer
     *                   to a flux_t object.
     *  \param ranks     RFC 22 idset string representing a set of ranks, or "all".
     *  \param status    resource_pool_t::status_t::UP or resource_pool_t::status_t::DOWN
     *  \return          0 on success; -1 on error with errno set:
     *                       EINVAL: invalid parameters or unexpected internal error
     */
    static int set_rank_status (void *h, std::string_view ranks, resource_pool_t::status_t status);

    /*! Get resource status for the node at a rank.
     *
     *  \param h         Opaque handle. How it is used is an implementation
     *                   detail. However, when it is used within a Flux's
     *                   service module, it is expected to be a pointer
     *                   to a flux_t object.
     *  \param rank      Rank number as a string, e.g. "1"
     *  \param status    Reference to receive status
     *  \return          0 on success; -1 on error with errno set:
     *                       EINVAL: invalid parameters, rank not found,
     *                               or unexpected internal error
     */
    static int get_rank_status (void *h, std::string_view rank, resource_pool_t::status_t &status);
```
added a new C header `resource_status.h` to define a C style status enum:
```c
typedef enum resource_status_t {
    RESOURCE_UP = 0,
    RESOURCE_DOWN = 1
} resource_status_t;
```
and implemented these new API functions
```c
/*! Set resource status (up or down) for a resource at the specified path.
 *
 *  \param ctx           reapi_cli_ctx_t context object
 *  \param resource_path Path to resource (e.g., "/cluster0/rack0/node3")
 *  \param status        RESOURCE_UP or RESOURCE_DOWN
 *  \return              0 on success; -1 on error with errno set:
 *                           EINVAL: invalid parameters, resource path not found,
 *                                   or unexpected internal error
 */
int reapi_cli_set_status (reapi_cli_ctx_t *ctx,
                          const char *resource_path,
                          resource_status_t status);

/*! Get resource status for a resource at the specified path.
 *
 *  \param ctx           reapi_cli_ctx_t context object
 *  \param resource_path Path to resource (e.g., "/cluster0/rack0/node3")
 *  \param status        Pointer to receive status
 *  \return              0 on success; -1 on error with errno set:
 *                           EINVAL: invalid parameters, resource path not found,
 *                                   or unexpected internal error
 */
int reapi_cli_get_status (reapi_cli_ctx_t *ctx,
                          const char *resource_path,
                          resource_status_t *status);

/*! Set resource status (up or down) for the subtree root (typically node) of each rank in a set.
 *  Unknown ranks are silently ignored.
 *
 *  \param ctx           reapi_cli_ctx_t context object
 *  \param ranks         RFC 22 idset string representing rank set, or "all"
 *  \param status        RESOURCE_UP or RESOURCE_DOWN
 *  \return              0 on success; -1 on error with errno set:
 *                           EINVAL: invalid parameters or unexpected internal error
 */
int reapi_cli_set_rank_status (reapi_cli_ctx_t *ctx, const char *ranks, resource_status_t status);

/*! Get resource status for the node at a rank.
 *
 *  \param ctx           reapi_cli_ctx_t context object
 *  \param rank          Single rank number, as a string e.g. "2"
 *  \param status        Pointer to receive status
 *  \return              0 on success; -1 on error with errno set:
 *                           EINVAL: invalid parameters, rank not found,
 *                                   or unexpected internal error
 */
int reapi_cli_get_rank_status (reapi_cli_ctx_t *ctx, const char *rank, resource_status_t *status);
```

Note that the path versions of these interfaces are not required at this point for a python scheduler, but they could be handy in the future e.g. for marking rabbits up/down and they are more natural than the rank versions in fluxion space.

### DIFF
--- a/resource/reapi/bindings/c++/reapi_cli.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli.hpp
@@ -16,6 +16,7 @@ extern "C" {
 #include "config.h"
 #endif
 #include <flux/core.h>
+#include <flux/idset.h>
 #include <cstdint>
 #include <cerrno>
 }
@@ -174,6 +175,16 @@ class reapi_cli_t : public reapi_t {
                      double &min,
                      double &max,
                      double &avg);
+    static int set_status (void *h,
+                           const std::string &resource_path,
+                           resource_pool_t::status_t status);
+    static int get_status (void *h,
+                           const std::string &resource_path,
+                           resource_pool_t::status_t &status);
+
+    static int set_rank_status (void *h, std::string_view ranks, resource_pool_t::status_t status);
+    static int get_rank_status (void *h, std::string_view rank, resource_pool_t::status_t &status);
+
     static const std::string &get_err_message ();
     static void clear_err_message ();
 
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -29,8 +29,6 @@ namespace Flux {
 namespace resource_model {
 namespace detail {
 
-const int NOT_YET_IMPLEMENTED = -1;
-
 static double get_elapsed_time (timeval &st, timeval &et)
 {
     double ts1 = (double)st.tv_sec + (double)st.tv_usec / 1000000.0f;
@@ -171,7 +169,8 @@ int reapi_cli_t::update_allocate (void *h,
                                   double &ov,
                                   std::string &R_out)
 {
-    return NOT_YET_IMPLEMENTED;
+    errno = ENOSYS;  // not implemented
+    return -1;
 }
 
 int reapi_cli_t::match_allocate_multi (void *h,
@@ -179,7 +178,8 @@ int reapi_cli_t::match_allocate_multi (void *h,
                                        const char *jobs,
                                        queue_adapter_base_t *adapter)
 {
-    return NOT_YET_IMPLEMENTED;
+    errno = ENOSYS;  // not implemented
+    return -1;
 }
 
 int reapi_cli_t::cancel (void *h, const uint64_t jobid, bool noent_ok)
@@ -363,7 +363,8 @@ int reapi_cli_t::stat (void *h,
                        double &max,
                        double &avg)
 {
-    return NOT_YET_IMPLEMENTED;
+    errno = ENOSYS;  // not implemented
+    return -1;
 }
 
 const std::string &reapi_cli_t::get_err_message ()
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -16,6 +16,7 @@ extern "C" {
 #include "config.h"
 #endif
 #include <flux/core.h>
+#include <flux/idset.h>
 }
 
 #include <stdexcept>
@@ -41,6 +42,52 @@ static double get_elapsed_time (timeval &st, timeval &et)
 ////////////////////////////////////////////////////////////////////////////////
 
 std::string reapi_cli_t::m_err_msg = "";
+
+// Helper: parse RFC 22 idset string to std::set<int64_t>
+// Handles "all" as a special case (returns all ranks from rq->db->metadata.by_rank)
+// Unknown ranks are not rejected here (mark() tolerates them to support a
+// future "grow" operation).
+// Returns 0 on success, -1 on error with errno set
+static int parse_ranks (resource_query_t *rq,
+                        std::string_view idset_str,
+                        std::set<int64_t> &ranks_out,
+                        std::string &err_msg)
+{
+    ranks_out.clear ();
+
+    // Check for empty string
+    if (idset_str.empty ()) {
+        errno = EINVAL;
+        err_msg = "empty idset string";
+        return -1;
+    }
+
+    // Handle "all" special case
+    if (idset_str == "all") {
+        for (const auto &kv : rq->db->metadata.by_rank) {
+            if (kv.first >= 0)
+                ranks_out.insert (kv.first);
+        }
+        return 0;
+    }
+
+    // Parse regular idset string using idset_decode_ex with explicit length
+    struct idset *ids = idset_decode_ex (idset_str.data (), idset_str.length (), 0, 0, NULL);
+    if (!ids) {
+        errno = EINVAL;
+        err_msg = std::string ("invalid idset string: ") + std::string (idset_str);
+        return -1;
+    }
+
+    unsigned int rank = idset_first (ids);
+    while (rank != IDSET_INVALID_ID) {
+        ranks_out.insert (static_cast<int64_t> (rank));
+        rank = idset_next (ids, rank);
+    }
+
+    idset_destroy (ids);
+    return 0;
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 // REAPI CLI Class Public API Definitions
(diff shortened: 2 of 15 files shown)


garlick:
A review comment given verbally by @trws and @milroy in our meeting today is that this PR creates a discrepancy between `reapi_cli` and `reapi_module`, which are intended to have identical interfaces.

Since implementing the status API and back end RPCs would be new library code (and tests) with no immediate use case, would it be acceptable to just do stubs that return NOT_YET_IMPLEMENTED for the time being?

This issue applies to the freshly merged clone capability, and will apply to work in progress  to
- add a reader format to `cancel()` so that rv1exec can be used for cancellation on a JGF graph
- implement `update_allocate()` to register allocated jobs during scheduler hello protocol
- I suppose the python reapi bindings if we really want to build out the dual capability

It would let us arrive at a functional python graph scheduler faster if we could just plant stubs in `reapi_module` where interfaces need to be added to get there.

Edit: I'll go ahead and push some stubs.  Also, returning NOT_YET_IMPLEMENTED (-1) without setting errno is not really OK, so I'll add a commit that fixes that for the current unimplemented functions.

garlick:
One self-review comments is that returning `resource_pool_t::status_t &status` is probably undesirable.  Conversion to a string representation is probably better at the API boundary.  Will fix.

garlick:
Rebased on top of
- #1492

and implemented the proposed interface change (strings for status values) and the new stubs.  I'll update the PR description with the new API function signatures.

garlick:
@wihobbs proposal in #1493 reminded me that interfaces for the C++ side are defined and documented in reapi.hpp.  These new ones need to be added.  Coming soon.

garlick:
Also, I happened to notice b60e200e46eab51deb2924a0d6a788dbd7bb6e73 by @zekemorton which makes me question my choice to have these new functions use UP|DOWN strings instead of the enum.  Which do sched developers prefer?   My thinking that using the enum was unnecessarily exposing an internal data structure and that it would be easy to add another state (like DEGRADED or similar) as a string without breaking the API, but the enum is nicer to work with a Zeke points out in the commit message.
```
commit b60e200e46eab51deb2924a0d6a788dbd7bb6e73
Author: Zeke Morton <morton30@llnl.gov>
Date:   Wed Feb 14 09:06:55 2024 -0800

    reapi: use enum to pass match op
    
    Problem: reapi has several layers of bindings that use strings to
    pass the match option. Strings require validation and use more space
    than is required.
    
    Use match_op_t enum for reapi insead. Move match_op_t definition to its
    own file and add function to convert match_op_t to string for resource
    module match call back.
```
Even simpler would be to just use a boolean.  Thoughts?

garlick:
`reapi.hpp` was updated to include prototypes for the (string based) generic status get/set functions, with inline docs.

garlick:
Sorry for the churn - I made another revision on the API being proposed here.  Instead of overloading `get_status()` and `set_status()` for paths and ranks, I broke the rank versions out into `set_rank_status()` and `get_rank_status()`.  

I also had the rank versions take a string idset for ranks with "all" support, which is simpler than what I had before and maps better to what's needed by schedulers.

garlick:
Rebased on master after #1491 was merged (since that created unit tests merge conflicts)

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -363,7 +439,146 @@ int reapi_cli_t::stat (void *h,
...
+        errno = EINVAL;
+        return -1;
+    }
+    rc = rq->traverser->mark (resource_path, rpstatus);
+    if (rc < 0) {
+        if (errno == 0)
+            errno = EINVAL;
+        m_err_msg = __FUNCTION__;
+        m_err_msg += ": " + rq->traverser->err_message ();
+    }
```
jameshcorbett: I think `mark` actually returns `EINVAL` if the path doesn't exist, which contradicts what the doc comment says in the header file.

resource/reapi/bindings/c++/reapi_cli.hpp
```diff
@@ -174,6 +175,12 @@ class reapi_cli_t : public reapi_t {
                      double &min,
                      double &max,
                      double &avg);
+    static int set_status (void *h, const std::string &resource_path, const char *status);
+    static int get_status (void *h, const std::string &resource_path, const char *&status);
+
+    static int set_rank_status (void *h, const char *ranks, const char *status);
+    static int get_rank_status (void *h, const char *rank, const char *&status);
```
jameshcorbett: I don't love that the `*_rank_*` versions take a `const char *` while `get_` and `set_status` take a string. I know that the `ranks` string is eventually passed to the idset library which only operates on `const char *` but I wonder if the cost of eating `std::string` might be worth it? Also a reference to a pointer `const char *&status` gives me conniptions but I don't know what it should be instead. @trws will know.

jameshcorbett (commented):
I think this looks good but will let other sched developers weigh in--I'm missing some of the context. I noticed two things though--

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: You are right, and it seems that mark() always sets errno when it returns -1 so this backstop is unnecessary/wrong.  I'll fix the docs for this function to say EINVAL on not found.

garlick: mark() also ignores unknown ranks for supporting a future "grow" operation so I'll also preserve that behavior in the `set_rank_status ()` behavior.  Fixups coming.

resource/reapi/bindings/c++/reapi_cli.hpp
garlick: it certainly does seem right to fix the `rank`/`ranks` parameters to use `const std::string &`.  Will push those shortly.

What about changing `status` to be a `bool up`?

garlick: I went ahead and pushed fixups for both changes (ranks strings + status bool).

trws:
> Also, I happened to notice [b60e200](https://github.com/flux-framework/flux-sched/commit/b60e200e46eab51deb2924a0d6a788dbd7bb6e73) by @zekemorton which makes me question my choice to have these new functions use UP|DOWN strings instead of the enum. Which do sched developers prefer? My thinking that using the enum was unnecessarily exposing an internal data structure and that it would be easy to add another state (like DEGRADED or similar) as a string without breaking the API, but the enum is nicer to work with a Zeke points out in the commit message.
> 
> ```
> commit b60e200e46eab51deb2924a0d6a788dbd7bb6e73
> Author: Zeke Morton <morton30@llnl.gov>
> Date:   Wed Feb 14 09:06:55 2024 -0800
> 
>     reapi: use enum to pass match op
>     
>     Problem: reapi has several layers of bindings that use strings to
>     pass the match option. Strings require validation and use more space
>     than is required.
>     
>     Use match_op_t enum for reapi insead. Move match_op_t definition to its
>     own file and add function to convert match_op_t to string for resource
>     module match call back.
> ```
> 
> Even simpler would be to just use a boolean. Thoughts?

I'd prefer an enum. In the longer term I really do want to get us using `std::expected` or similar, and their design is to adapt around `errcode` objects that are expected to be implemented as integers or enums with specific translation functions, like possix error codes, so things that stay in that same general shape will make that easier.

If we're not otherwise exposing `resource_t` internals, we could hoist the `status_t` and give it a more meaningful name. Completely agree about not exposing unnecessary internal state but I think status codes are a useful API-boundary level thing to have, and as you said extending with new ones is easier with an enum or string than a boolean.

resource/reapi/bindings/c++/reapi_cli.hpp
trws: `std::string_view` would be preferred if idset can work on pointer and length, since that would work with any string-like thing and carry size, `const std::string&` is fine if it can't.

garlick: Oh indeed it can, will fix.  (I forgot that we we had `idset_decode_ex()` for that until your not reminded me)

garlick:
OK, enum should be fine.

garlick:
OK I think I've addressed the requested changes:
- `std::string:_view` for idset parameters
- `resource_pool_t::status_t` enum for status parameters in C++ (in keeping with this being an internal API)
- new `resource_status_t` enum for status parameters in C (treating this as a future public API)

Edit: the header snippets in the PR description were also updated for quick review

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -42,6 +43,66 @@ static double get_elapsed_time (timeval &st, timeval &et)
...
+    if (!status)
+        return -1;
+    if (strcasecmp (status, "up") == 0)
+        rpstatus = resource_pool_t::status_t::UP;
+    else if (strcasecmp (status, "down") == 0)
+        rpstatus = resource_pool_t::status_t::DOWN;
+    else
+        return -1;
+    return 0;
+}
```
jameshcorbett: Unused?

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -42,6 +43,66 @@ static double get_elapsed_time (timeval &st, timeval &et)
...
+
+// Helper: parse RFC 22 idset string to std::set<int64_t>
+// Handles "all" as a special case (returns all ranks from rq->db->metadata.by_rank)
+// Validates that all ranks exist in rq->db->metadata.by_rank
```
jameshcorbett: This comment (this one line) is misleading, I don't see any validation.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -367,6 +428,139 @@ int reapi_cli_t::stat (void *h,
...
+    // Get the first vertex at this rank
+    if (it->second.empty ()) {
+        errno = EINVAL;
+        m_err_msg = __FUNCTION__;
+        m_err_msg += ": no vertices at rank: ";
+        m_err_msg += rank_str;
+        return -1;
+    }
+
+    vtx_t v = it->second.front ();
```
jameshcorbett: This makes sense to me but is it guaranteed correct to use `.front` here?

jameshcorbett (approved):
LGTM! A couple of minor things below.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: Whoops, from previous implementation!

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: This assumes that's the node vertex (and I _think_ it usually is?) but we probably should explicitly look through all vertices on that rank.

garlick:
Thanks, I added your comments @jameshcorbett and also tried to improve test coverage a bit.

garlick:
I'll go ahead and set MWP since this was already approved.  Thanks!

PR #1491, opened 2026-05-27
reapi: add reapi_cli_match_with_jobid()

garlick:
Problem: the C reapi_cli match functions allocate jobids internally, but when used in a Flux scheduler, Flux allocates the jobids and provides them in the alloc request.

Add a new function variant: `reapi_cli_match_with_jobid()` with a jobid input parameter. 

Note that `reapi_module_match()` already accepts a jobid input parameter.  In an abundance of caution,  I did not force the `reeapi_cli` version to agree which would have been cleaner but would break the API.

This was peeled off of
- #1481

### DIFF
--- a/resource/reapi/bindings/c/test/reapi_cli_test.c
+++ b/resource/reapi/bindings/c/test/reapi_cli_test.c
@@ -40,6 +40,28 @@ static const char *tiny_params =
     "{\"load_format\": \"jgf\", \"matcher_policy\": \"high\", "
     "\"match_format\": \"rv1\", \"matcher_name\": \"CA\"}";
 
+static const char *simple_jobspec =
+    "{"
+    "\"version\": 1,"
+    "\"resources\": ["
+    "{"
+    "\"type\": \"node\","
+    "\"count\": 1,"
+    "\"with\": ["
+    "{"
+    "\"type\": \"slot\","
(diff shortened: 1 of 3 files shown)


garlick:
Rebased on current master and made independent of #1489 (will update PR description)

resource/reapi/bindings/c/test/reapi_cli_test.c
```diff
@@ -65,11 +87,67 @@ static int test_clone ()
...
+                                     simple_jobspec,
+                                     jobid,
+                                     &reserved,
+                                     &R,
+                                     &at,
+                                     &ov);
+    ok (rc == 0, "reapi_cli_match_with_jobid succeeded");
+    ok (R != NULL, "reapi_cli_match_with_jobid returned R string");
+    free (R);
+    R = NULL;
```
jameshcorbett, line 119: Maybe add a test that canceling the jobid works as well?

jameshcorbett (approved):
LGTM, one idea for improvement.

resource/reapi/bindings/c/test/reapi_cli_test.c
garlick: Good idea, just pushed that and I'll set MWP.

PR #1492, opened 2026-05-28
use constant strings where appropriate in some internal functions

garlick:
Problem: get_jobstate_str() fills a std::string& output parameter, which allocates memory, but the returned value could be a constant.

Change the signature to return a const char * pointing to a string literal, and update all existing callers to use simple assignment.

### DIFF
--- a/resource/jobinfo/jobinfo.cpp
+++ b/resource/jobinfo/jobinfo.cpp
@@ -47,27 +47,22 @@ job_info_t::job_info_t (uint64_t j,
 {
 }
 
-void get_jobstate_str (job_lifecycle_t state, std::string &status)
+const char *get_jobstate_str (job_lifecycle_t state)
 {
     switch (state) {
-        case job_lifecycle_t::ALLOCATED:
-            status = "ALLOCATED";
-            break;
-        case job_lifecycle_t::RESERVED:
-            status = "RESERVED";
-            break;
-        case job_lifecycle_t::CANCELED:
-            status = "CANCELED";
-            break;
-        case job_lifecycle_t::ERROR:
-            status = "ERROR";
-            break;
         case job_lifecycle_t::INIT:
+            return "INIT";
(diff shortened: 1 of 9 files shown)


garlick:
The reason I proposed this (which I should have stated in the PR description)  is that a WIP change to return the job state via the reapi C bindings would require another `strdup()` and callers to free the returned strings.  Making them constant improves that API.

However as a general cleanup, this PR should probably go a little further and update the callers to avoid assigning the const string to a `std::string` when it's just coming back out as a `.c_str()`.  And that uncovers a couple more similar "to string" functions that use this pattern.  Bah, may as well fix all of them.  More coming.

garlick:
OK, this now covers 
- `get_jobstate_str()` (INIT|ALLOCATED|RESERVED|CANCELED|ERROR)
- `get_status_string()` (ALLOCATED|RESERVED)
- `status_to_str()` (UP|DOWN)

garlick:
If someone has a quick moment to review this change I'd appreciate it.  It's a pretty simple change.

grondo (approved):
LGTM! Seems like excellent and straightforward cleanup.

garlick:
Thanks!  I'll set MWP and if a sched developer has an objection we can easily revert before the upcoming tag.  This helps reduce my patch stack which is getting kinda deep.

jameshcorbett (commented):
I think maybe this could have used a `std::string_view` instead and I believe using pointer types is generally discouraged but this seems fine!

garlick:
Oh hmm, I wonder if that would've solved my problem of needing to get a const C string out in the C bindings?    E.g. if you do a `.c_str()` on a `std::string_view` do you have to delete it after?  

Edit: I don't think it would have but LMK if you think I went too far in the C direction here and should follow up with something better.

grondo:
Sorry! I guess I should have let the experts review before putting in my ignorant $.02!

TIL about `std::string_view`, which is pretty cool!

Issue #1477, opened 2026-05-07
fluxion encodes some json as strings, resulting message is inconvenient to trace


wihobbs:
Fluxion has a tendency in some callbacks to encode json as a string, which must then later be decoded (possibly taking up unnecessary cycles). As an example:
```
2026-05-06T16:38:43.600 sched-fluxion-qmanager tx > sched-fluxion-resource.match_multi [676]
{
  "cmd": "allocate_with_satisfiability",
  "jobs": "[{\"jobid\": 2398051368960, \"jobspec\": \"{\\\"resources\\\":[{\\\"type\\\":\\\"slot\\\",\\\"count\\\":1,\\\"with\\\":[{\\\"type\\\":\\\"core\\\",\\\"count\\\":1}],\\\"label\\\":\\\"task\\\"}],\\\"tasks\\\":[{\\\"command\\\":[\\\"hostname\\\"],\\\"slot\\\":\\\"task\\\",\\\"count\\\":{\\\"per_slot\\\":1}}],\\\"attributes\\\":{\\\"system\\\":{\\\"duration\\\":0,\\\"cwd\\\":\\\"/g/g0/hobbs17\\\",\\\"shell\\\":{\\\"options\\\":{\\\"rlimit\\\":{\\\"cpu\\\":-1,\\\"fsize\\\":-1,\\\"data\\\":-1,\\\"stack\\\":-1,\\\"core\\\":16384,\\\"nofile\\\":128000,\\\"as\\\":-1,\\\"rss\\\":-1,\\\"nproc\\\":16384}}}}},\\\"version\\\":1}\"}]"
}
2026-05-06T16:38:43.600 sched-fluxion-resource rx > sched-fluxion-resource.match_multi [676]
{
  "cmd": "allocate_with_satisfiability",
[...]

PR #1493, opened 2026-05-28
pass `json_t` objects instead of json strings for `allocate_with_satisfiability`

wihobbs:
This is the beginning of a fix for #1477, and is the initial low hanging fruit. It fixes the display of `allocate_with_satisfiability` RPCs in `flux module trace --full`, but not the `R` responses (yet):

```
2026-05-28T15:15:54.730 sched-fluxion-resource rx > sched-fluxion-resource.match_multi [459]
{
  "cmd": "allocate_with_satisfiability",
  "jobs": [
    {
      "jobid": 473654362112,
      "jobspec": {
        "resources": [
          {
            "type": "slot",
            "count": 1,
            "with": [
              {
                "type": "core",
                "count": 1
              }
            ],
            "label": "task"
          }
        ],
        "tasks": [
          {
            "command": [
              "hostname"
            ],
            "slot": "task",
            "count": {
              "per_slot": 1
            }
          }
        ],
        "attributes": {
          "system": {
            "duration": 0,
            "cwd": "/home/hobbs17.guest",
            "shell": {
              "options": {
                "rlimit": {
                  "cpu": -1,
                  "fsize": -1,
                  "data": -1,
                  "stack": 8388608,
                  "core": 0,
                  "nofile": 524288,
                  "as": -1,
                  "rss": -1,
                  "nproc": 63321
                }
              }
            }
          }
        },
        "version": 1
      }
    }
  ]
}
2026-05-28T15:15:54.731 sched-fluxion-resource tx < sched-fluxion-resource.match_multi [291] success
{
  "jobid": 473654362112,
  "status": "ALLOCATED",
  "overhead": 0.00058957200000000001,
  "R": "{\"version\": 1, \"execution\": {\"R_lite\": [{\"rank\": \"0\", \"children\": {\"core\": \"7\"}}], \"nodelist\": [\"lima-default\"], \"starttime\": 1780006554, \"expiration\": 9223372036}}\n",
  "at": 1780006554
}
2026-05-28T15:15:54.731 sched-fluxion-resource tx < sched-fluxion-resource.match_multi [0] ENODATA
2026-05-28T15:15:54.748 sched-fluxion-resource rx > sched-fluxion-resource.cancel [23]
{
  "jobid": 473654362112
}
2026-05-28T15:15:54.749 sched-fluxion-resource tx < sched-fluxion-resource.cancel [3] success
{}
```

### DIFF
--- a/resource/modules/resource.cpp
+++ b/resource/modules/resource.cpp
@@ -418,44 +418,48 @@ static void match_multi_request_cb (flux_t *h,
 {
     size_t index;
     json_t *value;
-    json_error_t err;
     int saved_errno;
     json_t *jobs = nullptr;
     uint64_t jobid = 0;
     std::string errmsg;
     const char *cmd = nullptr;
-    const char *jobs_str = nullptr;
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
     if (!flux_msg_is_streaming (msg)) {
         errno = EPROTO;
         goto error;
     }
-    if (flux_request_unpack (msg, NULL, "{s:s s:s}", "cmd", &cmd, "jobs", &jobs_str) < 0)
+    if (flux_request_unpack (msg, NULL, "{s:s s:o}", "cmd", &cmd, "jobs", &jobs) < 0)
         goto error;
-    if (!(jobs = json_loads (jobs_str, 0, &err))) {
-        errno = ENOMEM;
+    if (!json_is_array (jobs)) {
+        errno = EINVAL;
         goto error;
     }
 
     json_array_foreach (jobs, index, value) {
-        const char *js_str;
+        json_t *jobspec_obj = nullptr;
+        char *jobspec_str = nullptr;
         int64_t at = 0;
         int64_t now = 0;
         double overhead = 0.0f;
         std::stringstream R;
 
-        if (json_unpack (value, "{s:I s:s}", "jobid", &jobid, "jobspec", &js_str) < 0)
+        if (json_unpack (value, "{s:I s:o}", "jobid", &jobid, "jobspec", &jobspec_obj) < 0)
             goto error;
+        if (!(jobspec_str = json_dumps (jobspec_obj, JSON_COMPACT))) {
+            errno = ENOMEM;
+            goto error;
+        }
         if (is_existent_jobid (ctx, jobid)) {
             errno = EINVAL;
             flux_log_error (h,
                             "%s: existent job (%jd).",
                             __FUNCTION__,
                             static_cast<intmax_t> (jobid));
+            free (jobspec_str);
             goto error;
         }
-        if (run_match (ctx, jobid, cmd, js_str, &now, &at, &overhead, R, NULL) < 0) {
+        if (run_match (ctx, jobid, cmd, jobspec_str, &now, &at, &overhead, R, NULL) < 0) {
             if (errno != EBUSY && errno != ENODEV)
                 flux_log_error (ctx->h,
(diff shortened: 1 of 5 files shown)


resource/reapi/bindings/c++/reapi_module.hpp
```diff
@@ -47,11 +48,11 @@ class reapi_module_t : public reapi_t {
...
-                                     const char *jobs,
+                                     json_t *jobs,
                                      queue_adapter_base_t *adapter);
     static int match_allocate_multi (void *h,
                                      match_op_t match_op,
-                                     const char *jobs,
+                                     json_t *jobs,
                                      queue_adapter_base_t *adapter);
     static int update_allocate (void *h,
                                 const uint64_t jobid,
```
garlick, line 58: Two concerns here: the introduction of a jansson requirement for all callers, and the API change.  Neither of these are big concerns if the API remains internal, but I'm not sure if that's the plan.

Anyway, one way to address both would be to add a new method that accepts "pack" style variadic arguments (similar to what we do in flux-core for the rpc functions, as an example).

I'm not sure if reapi_cli (and reapi.hpp) should be tracking these changes to preserve uniformity across the REAPI.  Probably yes if `match_allocate_multi()`  is directly changed as proposed.  But adding a new method (or an overload) may be OK?

I should let the sched developers weigh in though!

wihobbs: > Two concerns here: the introduction of a jansson requirement for all callers, and the API change. Neither of these are big concerns if the API remains internal, but I'm not sure if that's the plan.

I think last I discussed with @trws the plan was for the API to remain internal. However, I am curious to hear from the other sched developers about if we should: 

> add a new method that accepts "pack" style variadic arguments (similar to what we do in flux-core for the rpc functions, as an example).

As for tracking the changes, what's the proper way to do that? Is the commit log insufficient?

trws: A pack style API for this might make sense, but the structure is an array of objects of shape `{"jobid":<id>, "jobspec":<jobspec>}`. Other than passing a `json_t` array object to the pack function is there a good way to actually get an unknown number of elements packed into an array like that @garlick?

If it came down to it, we could have it take a `std::span<>` of structs of those elements or something too.

resource/modules/resource.cpp
```diff
@@ -418,45 +418,49 @@ static void match_multi_request_cb (flux_t *h,
...
         errno = EPROTO;
         goto error;
     }
-    if (flux_request_unpack (msg, NULL, "{s:s s:s}", "cmd", &cmd, "jobs", &jobs_str) < 0)
+    if (flux_request_unpack (msg, NULL, "{s:s s:o}", "cmd", &cmd, "jobs", &jobs) < 0)
         goto error;
-    if (!(jobs = json_loads (jobs_str, 0, &err))) {
-        errno = ENOMEM;
+    if (!json_is_array (jobs)) {
+        errno = EPROTO;
```
wihobbs: This might be incorrect...

wihobbs: (the switch from ENOMEM to EPROTO, I mean)

trws: Hm, `EINVAL` might be appropriate if it's incorrectly typed.

wihobbs: Ok, fixed.

trws (approved):

wihobbs:
I want to look into that failure, so removing MWP for now.

wihobbs:
This appears to be a hwloc leak, not related to this PR. Retrying tests.

PR #1497, opened 2026-06-03
reapi: add support for partial cancel with rv1exec match format

garlick:
Problem: When used to implement a flux scheduler that initializes with JGF, reapi_cli needs to handle partial cancel with the `rv1exec` match format (since the job manager cannot provide R with scheduling key), but the reapi_cli hard codes the cancel match format to the format used to initialize the graph.

Add a 6-arg C++ `cancel()` method and corresponding C `reapi_cli_cancel_ex()` function that allow the match format to be specified.

This was peeled off of
- #1481

### DIFF
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -985,23 +985,83 @@ int resource_query_t::remove_job (const uint64_t jobid)
 int resource_query_t::remove_job (const uint64_t jobid, const std::string &R, bool &full_removal)
 {
     int rc = -1;
+    json_error_t error;
+    json_t *r_obj = nullptr;
+    json_t *scheduling = nullptr;
+    json_t *writer = nullptr;
+    std::string format;
+    std::string reader_input;
     std::shared_ptr<resource_reader_base_t> reader;
 
     if (jobid > (uint64_t)std::numeric_limits<int64_t>::max ()) {
         errno = EOVERFLOW;
         return rc;
     }
-    if (R == "") {
+
+    // Parse R to determine format and extract appropriate input for reader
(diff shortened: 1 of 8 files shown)


garlick:
Feedback from @trws in today's meeting:  use `scheduling.writer` (or its absence) to infer the match_format instead of adding new API interfaces.  Good idea!  Will go that way!

garlick:
Just pushed the change to select the reader using hints from `R.scheduling.writer`.  It looks like only rv1exec (no schedluing key) actually works with partial cancel presently - fine for now since that's what the job manager will generate.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -799,23 +799,87 @@ int resource_query_t::remove_job (const uint64_t jobid)
...
+            if (writer_uri == "fluxion") {
+                format = "jgf";
+            } else if (writer_uri.compare (0, 8, "fluxion:") == 0) {
+                format = writer_uri.substr (8);
+                if (format.empty ())
+                    format = "jgf";
+            } else {
+                format = writer_uri;
+            }
+        }
```
jameshcorbett: We have this same logic in `parse_R` in `resource_match.cpp` --- how would you feel about making this a utility function and using it in both places?

jameshcorbett (approved):
This looks good --- could you add a test that checks what happens when you pass `scheduling.writer` with things like `fluxion:notaformat` and `notfluxion:somethingelse`?

garlick:
Just pushed an update with the new test and a refactoring that lets reader selection by `scheduling.writer` be shared.  Want to have another look @jameshcorbett ?

jameshcorbett (approved):
LGTM thanks!

garlick:
Thanks, I'll set MWP!

PR #1498, opened 2026-06-03
assorted minor build / test improvements

garlick:
Here some minor build system changes to
- byte compile installed python modules
- add a missing dependency to `install-deps.sh`
- improve workflow for test debs


trws (approved):
Looks good to me. Just to be sure, the debbuild.sh assumes configuration is already done right?

garlick:
Thanks!  Yep, it does assume that.  I'll set MWP.
```
$ mkdir build && cd build
$ cmake ..
$ cmake --build . -t deb
```

PR #1499, opened 2026-06-05
reapi: implement update_allocate()

garlick:
Problem: the reapi_cli `update_allocate()` method is required for a scheduler based on reapi_cli to participate in the job manager hello protocol, but it is not implemented.

The C wrapper already exists so filling in the C++ implementation takes care of both bindings.

Add an implementation and unit tests.

This was peeled off of
- #1481

### DIFF
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -25,6 +25,7 @@ extern "C" {
 #include <boost/algorithm/string.hpp>
 #include "resource/reapi/bindings/c++/reapi_cli.hpp"
 #include "resource/readers/resource_reader_factory.hpp"
+#include "resource/config/system_defaults.hpp"
 
 namespace Flux {
 namespace resource_model {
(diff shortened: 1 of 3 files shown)


resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -171,7 +172,143 @@ int reapi_cli_t::update_allocate (void *h,
...
+    {
+        std::stringstream o;
+        if (rq->writers->emit (o) < 0) {
+            m_err_msg += __FUNCTION__;
+            m_err_msg += ": ERROR: writer emit failed\n";
+            rc = -1;
+            goto out;
+        }
+        R_out = o.str ();
+    }
```
jameshcorbett, line 344: Is blocking this code off within braces necessary? Also, I think `o` should probably be renamed because there is already another `o` declared at function-level scope.

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -171,7 +172,143 @@ int reapi_cli_t::update_allocate (void *h,
...
+                     &expiration,
+                     "scheduling",
+                     &scheduling)
+            == 0
+        && version == 1) {
+        if (expiration == 0.0)
+            expiration = max_exp;
+        at_parsed = static_cast<int64_t> (tstart);
+        uint64_t dur = (expiration > tstart) ? static_cast<uint64_t> (expiration - tstart) : 0;
+        duration = dur ? dur : static_cast<uint64_t> (SYSTEM_MAX_DURATION);
```
jameshcorbett: If `expiration` is 0, then shouldn't duration be `SYSTEM_MAX_DURATION` and not `expiration - tstart`? Also I'm not sure I understand this last line --- if `duration` is 0 why do you switch it back to `SYSTEM_MAX_DURATION`?

jameshcorbett (approved):
LGTM, some small things. Could you maybe add a test that provides R with no `expiration` key?

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: Bug, thanks!  The code is now just
```c++
// An absent or non-positive expiration (dur == 0) falls back to the
// system-wide maximum duration rather than an unbounded span.
uint64_t dur = (expiration > tstart) ? static_cast<uint64_t> (expiration - tstart) : 0;
duration = dur ? dur : static_cast<uint64_t> (SYSTEM_MAX_DURATION);
```

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: The braces are necessary to avoid the goto over initialization (doesn't compile in C++) but the shadowed variable was not intended!  Good catch, fixed.  (Also I added a comment about the goto situation)

garlick:
Fixed those things and added the no-expiration test so I'll set MWP.  Thanks!

resource/reapi/bindings/c++/reapi_cli_impl.hpp
jameshcorbett: Out of curiosity, couldn't you just have moved the `std::stringstream oss;` line to the top of the function and then removed the braces?

jameshcorbett: But I guess now that I'm looking at this, there is a lot of this scoping with braces in the function.

PR #1500, opened 2026-06-05
reapi: add optional match format parameter to cli find() and add C bindings

garlick:
Problem: the reapi_cli `find()` implementation always uses the context's match_format, but schedulers based on reapi_cli will need to generate resource objects in the `rv1_noexec` format in order to answer `resource-status` queries, regardless of the match format configured for `match_allocate()`.

Add an optional format parameter to `find()`.
Fill in missing reapi.hpp declaration and add reapi_module stub for `find()`.
Fix an API inconsistency where `find()` could return > 0 on success.
Add a `reapi_cli_find()` C binding.
Add unit tests.

This is in lieu of the R caching solution initially proposed in
- #1481

This is how the resource broker module builds the query response and is more useful as a debug tool because it samples the live graph state instead of inferring it.

### DIFF
--- a/resource/reapi/bindings/c++/reapi.hpp
+++ b/resource/reapi/bindings/c++/reapi.hpp
@@ -18,10 +18,12 @@ extern "C" {
 #endif
 #include <flux/core.h>
 #include <flux/idset.h>
+#include <jansson.h>
 }
 
 #include <cstdint>
 #include <string>
+#include <optional>
 #include "resource/schema/resource_data.hpp"
 
 namespace Flux {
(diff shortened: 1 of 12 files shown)


resource/reapi/bindings/c++/reapi.hpp
```diff
@@ -17,10 +17,12 @@ extern "C" {
 #include "config.h"
 #endif
 #include <flux/core.h>
+#include <jansson.h>
 }
 
 #include <cstdint>
 #include <string>
+#include <optional>
```
jameshcorbett, line 26: typo in commit message: declare

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -19,6 +19,7 @@ extern "C" {
 #include <cstdlib>
 #include <cstdint>
 #include <cerrno>
+#include <jansson.h>
```
jameshcorbett, line 23: Claude!!! There's a typo in the commit message: "sa"

resource/reapi/bindings/c++/reapi_cli_impl.hpp
```diff
@@ -261,32 +261,49 @@ int reapi_cli_t::cancel (void *h,
...
     if (!rq) {
         errno = EINVAL;
         return -1;
     }
 
-    if ((rc = rq->traverser_find (criteria)) < 0) {
+    std::shared_ptr<match_writers_t> writer;
+    if (format.has_value ()) {
+        // User specified a format - create temporary writer
+        match_format_t fmt = match_writers_factory_t::get_writers_type (format.value ());
```
jameshcorbett, line 478: `get_writers_type` returns the simple writer type if it doesn't recognize the string. You should use `known_match_format` to validate instead. Maybe add a test for this as well if you don't mind?

jameshcorbett (approved):
LGTM, two typos and one substantive thing

resource/reapi/bindings/c/reapi_cli.cpp
garlick: That was probably me actually :-)

resource/reapi/bindings/c++/reapi_cli_impl.hpp
garlick: Great point!  Will do.

garlick:
OK, I made those changes and will set MWP.  Thanks!

PR #1502, opened 2026-06-09
doc: add Internal API Reference via doxygen/breathe

### LINKED ISSUE #51: [question] Incorporate doxygen-based in-line docs for Fluxion
`flux-sched` has a ton of doxygen-based in-line documents designed to help developers to extend our Fluxion graph scheduler. For C++ based project (or even high level object based on like Python), this style of documentations would be more common.

As a test, I have a doxygen configuration for our `fluxion resource` infrastructure piece at https://github.com/flux-framework/flux-sched/blob/master/resource/doxygen/doxy_conf.txt, what would be a feasibility step we should take to see if it is possible to incorporate this into the overall readthedoc?

If I manually build the doxygen document out of it, I get something like the following. 

If this test goes well, I can add a top level doxygen configuration spanning both resource and qmanager components. 

I remember @trws said this would be possible with some tweak. And as I remember he (and @SteVwonder) would have other in-line docs for Python bindings. 

Screen Shot 2020-07-25 at 12 22 29 PM

wihobbs:
Problem: the flux-sched API has inline documentation that doesn't render on the web due to missing config.

Add a shell config and API reference holding pages for Doxygen to render inline documentation. Remove outdated Doxygen configurations.

Fixes flux-framework/flux-docs#51


wihobbs:
It was brought up on the coffee call yesterday that even in a nascent state these internal docs could have an immediate impact for developers looking to hack on Fluxion. So I am opening this up for review. @garlick, please let me know if it meets your expectations!

wihobbs:
Ok, force-pushed with a few changes:

- Updated container dependencies to include doxygen, plus pip and venv for making the HTML and getting the appropriate packages.
- Added the REAPI CLI to the Resource page.
- Fixed the planner commit so that the modifications to the index page were self-contained.

garlick:
Pretty good!  Just looking at the rendered end product:
- it might be nice to have the REAPI C API in a separate section since it's rendered differently than the C++ Resource section it's tacked onto now.
- the REAPI C module API, and both REAPI C++ APIs seem to have gone missing?

wihobbs:
Ok, I broke REAPI out into its own page, with two 2nd-level headings for the C and C++ interfaces.

garlick:
Nice!  Content-wise this seems like a good start.

I did notice that `cmake --build build -t dist` creates a dist tarball that doesn't include the new .rst files (but does include the man page ones).    They may need to be added to the CMakeLists.txt?

I guess there is no cmake equivalent to 'make distcheck'.  I did extract the tarball and do a build/check in there and didn't hit any errors so I guess the docs are not being built by default?

wihobbs:
> so I guess the docs are not being built by default?

No, and that was deliberate. The doxygen apidoc generation, despite my best efforts to suppress noisy output, is still _really_ noisy. It's also an expensive operation, so I decided to make it "only when explicitly requested," like when the website needs to be generated. I _thought_  this was how the Python "Complete API Reference" worked in flux-core. 

> I did notice that cmake --build build -t dist creates a dist tarball that doesn't include the new .rst files (but does include the man page ones). They may need to be added to the CMakeLists.txt?

Good point, especially since that planner README was moved (and I'd like to upcycle other READMEs too). I added those now and tested with `cmake -B build . && cd build/ && make dist`, then extracted the tarball and saw the files there. Here's a concise diff of what I added to make that work:

```diff
diff --git a/doc/CMakeLists.txt b/doc/CMakeLists.txt
index 7044d172..685cc8f4 100644
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -47,6 +47,10 @@ if(SPHINX AND ENABLE_DOCS)
         man5/flux-config-sched-fluxion-qmanager.rst
         man5/flux-config-sched-fluxion-resource.rst
         )
+    set(web_doc_inputs
+        index.rst
+        doc/api/*.rst
+        )
     foreach(file ${man5_inputs})
         string(REPLACE ".rst" ".5" out ${file})
         list(APPEND man5_outputs ${CMAKE_CURRENT_BINARY_DIR}/${out})
@@ -68,7 +72,7 @@ if(SPHINX AND ENABLE_DOCS)
         ${Python_EXECUTABLE} -m sphinx -q -b man ${CMAKE_CURRENT_SOURCE_DIR} ./man5
         )
     add_custom_target(manpages DEPENDS ${man5_outputs})
-    if(DOXYGEN_FOUND)
+    if(DOXYGEN_FOUND AND BREATHE_FOUND AND SPHINX AND ENABLE_DOCS)
         add_dependencies(manpages doxygen)
     endif()
     install(
@@ -85,9 +89,9 @@ if(SPHINX AND ENABLE_DOCS)
         ${Python_EXECUTABLE} -m sphinx -q -b html ${CMAKE_CURRENT_SOURCE_DIR} ./html
         )
     add_custom_target(html DEPENDS ${man5_html_outputs})
-    if(DOXYGEN_FOUND)
+    if(DOXYGEN_FOUND AND BREATHE_FOUND AND SPHINX AND ENABLE_DOCS)
         add_dependencies(html doxygen)
     endif()
     add_custom_target(docs
-        DEPENDS manpages)
+        DEPENDS manpages web_doc_inputs)
 endif()
```

trws (approved):

PR #1507, opened 2026-06-15
cmake: also set normal CMAKE_INSTALL_LIBDIR variable

kush-gupt:
I've been packaging flux-sched for Fedora and hit a build failure on Fedora 44, which ships CMake 4.0. Libraries were being installed under the build tree instead of the prefix. The `%files` section couldn't find any of the `.so` files.

The root cause seems to be CMake policy [CMP0126](https://cmake.org/cmake/help/latest/policy/CMP0126.html), which is NEW by default on CMake 4.0+. Under the new behavior, `set(CACHE FORCE)` no longer removes a same-named normal variable from scope. Since GNUInstallDirs leaves a normal `CMAKE_INSTALL_LIBDIR` that shadows the forced cache value, the override on line 109 is silently ignored.

This fix is a one-liner: also set the non-cache variable so both agree.

Made with [Cursor](https://cursor.com)


garlick (approved):
Thanks!   This LGTM. I tried making a testdeb and libraries seemed to end up in the right places.

garlick:
I'll go ahead and set MWP since no sched developers balked since yesterday :-)

trws:
That's a strange one, thanks for catching it @kush-gupt!

PR #1512, opened 2026-06-23
t: remove reference to flux perilog-run in alloc-check

### LINKED ISSUE #1511: `t1024-alloc-check` uses deprecated `flux perilog-run` command, but is still running (and passing?!?) in CI
In t1024 we're still invoking `flux perilog-run` which is deprecated:
```
# Test for flux-framework/flux-sched#1043
#
test_expect_success 'configure epilog with delay' '
        flux config load <<-EOT &&
        [job-manager]
        epilog.command = [ "flux", "perilog-run", "epilog", "-e", "sleep,2" ]
        EOT
        flux jobtap load perilog.so
'
# Jobs seem to need to be submitted separately to trigger the issue.
test_expect_success 'submit node-exclusive jobs that exceed their time limit' '
        (for i in $(seq 5); do \
            flux run -N1 -x -t1s sleep 30 || true; \
        done) 2>joberr
'
test_expect_success 'some jobs received timeout exception' '
        grep "job.exception" joberr | grep "type=timeout"
'
        test_must_fail grep "job.exception.*type=alloc-check" joberr
'
```

I'm confused on how this is possible when it was deprecated in 2024:
```
2026-06-17T16:25:49.7345975Z  55/130 Test  #53: t1024-alloc-check.t .................................................   Passed   31.52 sec
```

p.s. @jacobtkeio found a bug on his first day back :rooYay:
cc @grondo

wihobbs:
Problem: the alloc-check test maintained a deprecated reference to the removed flux perilog-run command. This wasn't getting caught because by the time the epilog was invoked, the test had stopped checking for the error that was being returned by flux-core. When running the testsuite locally on our clusters, the testsuite fails.

Solution: use the per-rank=true flag which takes a little bit longer, but is supported by flux-core.

Closes #1511

### DIFF
--- a/t/t1024-alloc-check.t
+++ b/t/t1024-alloc-check.t
@@ -29,8 +29,9 @@ test_expect_success 'load fluxion modules' '
 #
 test_expect_success 'configure epilog with delay' '
 	flux config load <<-EOT &&
-	[job-manager]
-	epilog.command = [ "flux", "perilog-run", "epilog", "-e", "sleep,2" ]
+	[job-manager.epilog]
+	per-rank = true
+	command = [ "sleep", "2" ]
 	EOT
 	flux jobtap load perilog.so
 '


t/t1024-alloc-check.t
```diff
@@ -30,7 +30,8 @@ test_expect_success 'load fluxion modules' '
 test_expect_success 'configure epilog with delay' '
 	flux config load <<-EOT &&
 	[job-manager]
-	epilog.command = [ "flux", "perilog-run", "epilog", "-e", "sleep,2" ]
+	per-rank = true
```
garlick, line 33: Doesn't per-rank need to be inside the epilog table (at the same level as `command`)?

wihobbs: Yep, fixed and tested: 
```
$ flux start --config-path=/tmp/config.toml
(s=1,d=0)  corona82 ~/flux-sched (issue-1511)$ flux config get
{"job-manager":{"epilog":{"per-rank":true,"command":["sleep","2"]}}}
(s=1,d=0)  corona82 ~/flux-sched (issue-1511)$ exit
```

garlick (approved):
LGTM!

PR #1514, opened 2026-06-29
testsuite: add Catch2-based REAPI CLI tests

jacobtkeio:
This PR adds simple Catch2-based initialization and match_allocate tests for the C and C++ CLI REAPI.
While these tests are similar to the existing C and C++ CLI REAPI tests, the Catch2 framework will make writing more sophisticated tests easier in the future.

Additionally, this PR fixes t4014's formatting.

### DIFF
--- a/resource/reapi/test/CMakeLists.txt
+++ b/resource/reapi/test/CMakeLists.txt
@@ -1,0 +1,9 @@
+add_executable(reapi_cli_cxx_test reapi_cli_test_cxx.cpp)
+add_sanitizers(reapi_cli_cxx_test)
+target_link_libraries(reapi_cli_cxx_test PRIVATE reapi_cli intern Catch2::Catch2WithMain)
+flux_add_test(NAME reapi_cli_cxx_test COMMAND reapi_cli_cxx_test)
+
+add_executable(reapi_cli_c_test reapi_cli_test_c.cpp)
+add_sanitizers(reapi_cli_c_test)
+target_link_libraries(reapi_cli_c_test PRIVATE reapi_cli intern Catch2::Catch2WithMain)
+flux_add_test(NAME reapi_cli_c_test COMMAND reapi_cli_c_test)
--- a/resource/reapi/test/reapi_cli_test_c.cpp
+++ b/resource/reapi/test/reapi_cli_test_c.cpp
@@ -11,8 +11,10 @@ TEST_CASE ("Initialize REAPI CLI", "[initialize C]")
 {
     const std::string options = "{}";
     std::stringstream buffer;
-    std::ifstream inputFile ("../../../t/data/resource/grugs/tiny.graphml");
+    const char *test_srcdir = std::getenv ("SHARNESS_TEST_SRCDIR");
+    REQUIRE (test_srcdir);
 
+    std::ifstream inputFile (std::string (test_srcdir) + "/data/resource/grugs/tiny.graphml");
     REQUIRE (inputFile.is_open ());
 
     buffer << inputFile.rdbuf ();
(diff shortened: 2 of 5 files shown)


resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+TEST_CASE ("Initialize REAPI CLI", "[initialize C++]")
+{
+    const std::string options = "{}";
+    std::stringstream buffer;
+    std::ifstream inputFile ("../../../t/data/resource/grugs/tiny.graphml");
+
+    if (!inputFile.is_open ()) {
+        std::cerr << "Error opening file!" << std::endl;
+    }
```
jameshcorbett: Here and throughout, use `REQUIRE (file.is_open())` instead? Rather than just printing a message and continuing.

resource/reapi/test/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_executable(reapi_cli_cxx_test reapi_cli_test_cxx.cpp)
+add_sanitizers(reapi_cli_cxx_test)
+target_link_libraries(reapi_cli_cxx_test PRIVATE reapi_cli intern Catch2::Catch2WithMain)
+flux_add_test(NAME reapi_cli_cxx_test COMMAND reapi_cli_cxx_test)
+set_property(TEST reapi_cli_cxx_test APPEND PROPERTY ENVIRONMENT "TESTRESRC_INPUT_FILE=$(CMAKE_SOURCE_DIR)/conf/hype.lua")
```
jameshcorbett: Is this TESTRESRC_INPUT_FILE variable used anywhere? If not remove this?

resource/reapi/test/reapi_cli_test_c.cpp
```diff
@@ -11,7 +11,8 @@ TEST_CASE ("Initialize REAPI CLI", "[initialize C]")
 {
     const std::string options = "{}";
     std::stringstream buffer;
-    std::ifstream inputFile ("../../../t/data/resource/grugs/tiny.graphml");
+    std::string sharness_test_srcdir = std::getenv ("SHARNESS_TEST_SRCDIR");
+    std::ifstream inputFile (sharness_test_srcdir + "/data/resource/grugs/tiny.graphml");
```
jameshcorbett: Add a nullptr check maybe? Like:
```
  const char *p = std::getenv ("SHARNESS_TEST_SRCDIR");                                                                                 
  REQUIRE (p);                                                                                                                          
  std::ifstream inputFile (std::string(p) + "/data/resource/grugs/tiny.graphml");
```

jameshcorbett (dismissed):
Looks great thanks, more tests are always welcome! I had just a few little things--

resource/reapi/test/CMakeLists.txt
jacobtkeio: A quick recursive search for `hype*` didn't find this file. Removed :)

t/t4014-match-feasibility.t
```diff
@@ -103,11 +103,10 @@ test_expect_success 'loading feasibility from non-load-file resource module work
 '
 
 test_expect_success 'loading job-validator conf works' '
-{ cat >conf.tmp << 'EOF'
```
jameshcorbett, line 106: I think the commit message would follow RFC 1 a little more closely if it went something like "testsuite: format heredoc in t4014 with tabs` because I think usually the subject should be more generic than a specific test file.

resource/reapi/test/reapi_cli_test_cxx.cpp
jacobtkeio: Agreed! It's both more correct and more Catch2-y

resource/reapi/test/reapi_cli_test_c.cpp
jacobtkeio: Sounds good. I think my flawed logic was that if SHARNESS_TEST_SRCDIR didn't exist something else went very wrong, but that's no reason not to report it right then and there!

jameshcorbett (approved):
LGTM, looks like the commit validator is failing though. Should be an easy fix and in the meanwhile I have a suggestion--

jameshcorbett:
Never mind, the validator passed, I'm going to set MWP!

PR #1515, opened 2026-06-29
match_op: Expose utils to convert between match_op_t and strings

jacobtkeio:
Problem: match_op.h's utils for converting between match_op_ts and strings are static functions, so they are not available outside of the translation unit. However, these conversions do need to take place elsewhere (the resource module, resource query). As a result, there are several hard-coded converters outside of the translation unit:
* resource_match.cpp:run
* resource_match.cpp:run_match (checking for whether the match option string is valid)
* command.cpp:cmd_match
* command.cpp:cmd_match_multi

These different converters are prone to error because it is easy to misspell duplicated strings.
Additionally, this makes adding a new match_op_t value more difficult because the developer needs to track down all of these places and manually add their new match_op and associated string to each one.

This PR makes a `match_option` static library and links `qmanager`, the `resource` library, and the `REAPI CLI` against it.
Additionally, it makes each of these four hard-coded converters use the conversion utilities in the `match_option` library.

To add a new match option to this scheme, add it to the `match_op_t` enum in resource/policies/base/match_op.h (between the last valid match option and END_MATCH_OP_T) and add a conversion entry to the `match_options` map in resource/policies/base/match_op.cpp. The conversion functions and `match_op_valid` function will work with the new entries.

This PR depends on #1514 for its conversion idempotence test in reapi_test_cxx.

### DIFF
--- a/qmanager/modules/CMakeLists.txt
+++ b/qmanager/modules/CMakeLists.txt
@@ -11,4 +11,5 @@ target_link_libraries(sched-fluxion-qmanager PRIVATE
     flux::schedutil
     PkgConfig::JANSSON
     cppwrappers
+    match_op
     )
--- a/resource/policies/base/match_op.cpp
+++ b/resource/policies/base/match_op.cpp
@@ -1,0 +1,65 @@
+/*****************************************************************************\
+ * Copyright 2026 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+#include <cstring>
+#include <utility>
+#include <array>
+#include <cerrno>
+
+#include "match_op.h"
+
+constexpr auto opt_map = std::to_array<std::pair<match_op_t, const char *>> (
+    {{MATCH_ALLOCATE, "allocate"},
+     {MATCH_ALLOCATE_ORELSE_RESERVE, "allocate_orelse_reserve"},
+     {MATCH_ALLOCATE_W_SATISFIABILITY, "allocate_with_satisfiability"},
+     {MATCH_SATISFIABILITY, "satisfiability"}});
+
+static_assert (opt_map.size () == END_MATCH_OP_T - MATCH_UNKNOWN - 1,
+               "opt_map is missing a match_op_t entry");
+
+const char *match_op_to_string (match_op_t match_op)
+{
+    if (match_op < 0 || match_op >= END_MATCH_OP_T) {
+        errno = EINVAL;
+        return nullptr;
+    }
+
+    for (const auto &[op, op_str] : opt_map)
+        if (op == match_op)
+            return op_str;
+
+    errno = ENOENT;
+    return nullptr;
+}
+
+match_op_t match_op_from_string (const char *str)
+{
+    if (!str) {
+        errno = EINVAL;
+        return MATCH_UNKNOWN;
+    }
+
+    for (const auto &[op, op_str] : opt_map)
+        if (strcmp (op_str, str) == 0)
+            return op;
+
+    errno = ENOENT;
+    return MATCH_UNKNOWN;
+}
+
+bool match_op_valid (match_op_t match_op)
+{
+    bool rc;
+    int saved_errno = errno;
+
+    rc = (match_op_to_string (match_op) != nullptr);
+    errno = saved_errno;
+    return rc;
+}
--- a/resource/policies/base/match_op.h
+++ b/resource/policies/base/match_op.h
@@ -1,5 +1,23 @@
+/*****************************************************************************\
+ * Copyright 2026 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
 #ifndef MATCH_OP_H
 #define MATCH_OP_H
+
+#ifndef __cplusplus
+#include <stdbool.h>
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 typedef enum match_op_t {
     MATCH_UNKNOWN,
@@ -10,30 +28,14 @@ typedef enum match_op_t {
     END_MATCH_OP_T
 } match_op_t;
 
-static const char *match_op_to_string (match_op_t match_op)
-{
-    switch (match_op) {
-        case MATCH_ALLOCATE:
-            return "allocate";
-        case MATCH_ALLOCATE_ORELSE_RESERVE:
-            return "allocate_orelse_reserve";
-        case MATCH_ALLOCATE_W_SATISFIABILITY:
-            return "allocate_with_satisfiability";
-        case MATCH_SATISFIABILITY:
-            return "satisfiability";
-        default:
-            return "error";
-    }
-}
+const char *match_op_to_string (match_op_t match_op);
 
-static bool match_op_valid (match_op_t match_op)
-{
-    if ((match_op != MATCH_ALLOCATE) && (match_op != MATCH_ALLOCATE_W_SATISFIABILITY)
-        && (match_op != MATCH_ALLOCATE_ORELSE_RESERVE) && (match_op != MATCH_SATISFIABILITY)) {
-        return false;
-    }
+match_op_t match_op_from_string (const char *str);
 
-    return true;
+bool match_op_valid (match_op_t match_op);
+
+#ifdef __cplusplus
 }
+#endif
 
 #endif  // MATCH_OP_H
--- a/resource/policies/base/match_op.h
+++ b/resource/policies/base/match_op.h
@@ -6,7 +6,8 @@ typedef enum match_op_t {
     MATCH_ALLOCATE,
     MATCH_ALLOCATE_W_SATISFIABILITY,
     MATCH_ALLOCATE_ORELSE_RESERVE,
-    MATCH_SATISFIABILITY
+    MATCH_SATISFIABILITY,
+    END_MATCH_OP_T
 } match_op_t;
 
 static const char *match_op_to_string (match_op_t match_op)
--- a/resource/reapi/test/reapi_cli_test_cxx.cpp
+++ b/resource/reapi/test/reapi_cli_test_cxx.cpp
@@ -2,6 +2,7 @@
 
 #include <catch2/catch_test_macros.hpp>
 #include <resource/reapi/bindings/c++/reapi_cli.hpp>
+#include <resource/policies/base/match_op.h>
 #include <fstream>
 #include <cstdlib>
 
@@ -23,6 +24,56 @@ TEST_CASE ("Initialize REAPI CLI", "[initialize C++]")
     std::shared_ptr<resource_query_t> ctx = nullptr;
     ctx = std::make_shared<resource_query_t> (rgraph, options);
     REQUIRE (ctx);
+}
+
+TEST_CASE ("Convert between strings and match_op_ts", "[match C++]")
+{
+    errno = 0;
+
+    // MATCH_UNKNOWN is invalid and converts to nullptr + ENOENT
+    CHECK (!match_op_valid (MATCH_UNKNOWN));
+    CHECK (errno == 0);
+
+    CHECK (match_op_to_string (MATCH_UNKNOWN) == nullptr);
+    CHECK (errno == ENOENT);
+    errno = 0;
+
+    // invalid strings convert to MATCH_UNKNOWN + ENOENT
+    CHECK (match_op_from_string ("0xDEADBEEF") == MATCH_UNKNOWN);
+    CHECK (errno == ENOENT);
+    errno = 0;
+
+    // out-of-bounds enum values are invalid and convert to nullptr + EINVAL
+    for (match_op_t OOB_match_op :
+         {(match_op_t)(-1), END_MATCH_OP_T, (match_op_t)(END_MATCH_OP_T + 1)}) {
+        CHECK (!match_op_valid (OOB_match_op));
+        CHECK (errno == 0);
+
+        CHECK (match_op_to_string (OOB_match_op) == nullptr);
+        CHECK (errno == EINVAL);
+        errno = 0;
+    }
+
+    // nullptr converts to MATCH_UNKNOWN + EINVAL (since invalid ptr)
+    CHECK (match_op_from_string (nullptr) == MATCH_UNKNOWN);
+    CHECK (errno == EINVAL);
+    errno = 0;
+
+    // valid match_op_t are valid
+    // and match_op_t <-> string conversion is idempotent
+    for (int op = MATCH_UNKNOWN + 1; op != END_MATCH_OP_T; op++) {
+        CHECK (match_op_valid ((match_op_t)op));
+        CHECK (match_op_from_string (match_op_to_string ((match_op_t)op)) == op);
+    }
+    CHECK (errno == 0);
+
+    // each match_op_t has the correct string representation
+    CHECK (match_op_from_string ("allocate") == MATCH_ALLOCATE);
+    CHECK (match_op_from_string ("allocate_orelse_reserve") == MATCH_ALLOCATE_ORELSE_RESERVE);
+    CHECK (match_op_from_string ("allocate_with_satisfiability")
+           == MATCH_ALLOCATE_W_SATISFIABILITY);
+    CHECK (match_op_from_string ("satisfiability") == MATCH_SATISFIABILITY);
+    CHECK (errno == 0);
 }
 
 TEST_CASE ("Match basic jobspec", "[match C++]")
--- a/resource/CMakeLists.txt
+++ b/resource/CMakeLists.txt
@@ -99,6 +99,7 @@ target_link_libraries(resource PUBLIC
     PkgConfig::UUID
     Boost::graph
     jobspec_conv
+    match_op
     )
 add_sanitizers(resource)
 
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -1460,7 +1460,7 @@ out:
 
 static int run (std::shared_ptr<resource_ctx_t> &ctx,
                 int64_t jobid,
-                const char *cmd,
+                match_op_t op,
                 const std::string &jstr,
                 int64_t *at,
                 flux_error_t *errp)
@@ -1468,19 +1468,7 @@ static int run (std::shared_ptr<resource_ctx_t> &ctx,
     int rc = -1;
     try {
         Flux::Jobspec::Jobspec j{jstr};
-
-        dfu_traverser_t &tr = *(ctx->traverser);
-
-        if (std::string ("allocate") == cmd)
-            rc = tr.run (j, ctx->writers, match_op_t::MATCH_ALLOCATE, jobid, at);
-        else if (std::string ("allocate_with_satisfiability") == cmd)
-            rc = tr.run (j, ctx->writers, match_op_t::MATCH_ALLOCATE_W_SATISFIABILITY, jobid, at);
-        else if (std::string ("allocate_orelse_reserve") == cmd)
-            rc = tr.run (j, ctx->writers, match_op_t::MATCH_ALLOCATE_ORELSE_RESERVE, jobid, at);
-        else if (std::string ("satisfiability") == cmd)
-            rc = tr.run (j, ctx->writers, match_op_t::MATCH_SATISFIABILITY, jobid, at);
-        else
-            errno = EINVAL;
+        rc = ctx->traverser->run (j, ctx->writers, op, jobid, at);
     } catch (const Flux::Jobspec::parse_error &e) {
         errno = EINVAL;
         if (errp && e.what ()) {
@@ -1573,9 +1561,8 @@ int run_match (std::shared_ptr<resource_ctx_t> &ctx,
     bool rsv = false;
 
     start = std::chrono::system_clock::now ();
-    if (strcmp ("allocate", cmd) != 0 && strcmp ("allocate_orelse_reserve", cmd) != 0
-        && strcmp ("allocate_with_satisfiability", cmd) != 0
-        && strcmp ("satisfiability", cmd) != 0) {
+    match_op_t op = match_op_from_string (cmd);
+    if (!match_op_valid (op)) {
         rc = -1;
         errno = EINVAL;
         flux_log (ctx->h, LOG_ERR, "%s: unknown cmd: %s", __FUNCTION__, cmd);
@@ -1584,7 +1571,7 @@ int run_match (std::shared_ptr<resource_ctx_t> &ctx,
 
     epoch = std::chrono::duration_cast<std::chrono::seconds> (start.time_since_epoch ());
     *at = *now = epoch.count ();
-    if ((rc = run (ctx, jobid, cmd, jstr, at, errp)) < 0) {
+    if ((rc = run (ctx, jobid, op, jstr, at, errp)) < 0) {
         elapsed = std::chrono::system_clock::now () - start;
         *overhead = elapsed.count ();
         update_match_perf (*overhead, jobid, false);
--- a/resource/policies/base/CMakeLists.txt
+++ b/resource/policies/base/CMakeLists.txt
@@ -1,2 +1,6 @@
+add_library(match_op STATIC
+    match_op.h
+    match_op.cpp
+    )
 
 add_subdirectory( test )
--- a/resource/reapi/bindings/CMakeLists.txt
+++ b/resource/reapi/bindings/CMakeLists.txt
@@ -12,6 +12,7 @@ target_link_libraries( reapi_cli PRIVATE
 target_link_libraries( reapi_cli PUBLIC
     flux::core
     flux::idset
+    match_op
     )
 install(TARGETS reapi_cli LIBRARY DESTINATION "${FLUX_LIB_DIR}")
 
@@ -24,6 +25,9 @@ add_library (reapi_module STATIC
     )
 target_link_libraries( reapi_module PRIVATE
     flux::core
+    )
+target_link_libraries( reapi_module PUBLIC
+    match_op
     )
 
 if(BUILD_TESTING)
--- a/resource/reapi/bindings/c++/reapi_cli_impl.hpp
+++ b/resource/reapi/bindings/c++/reapi_cli_impl.hpp
@@ -121,8 +121,7 @@ int reapi_cli_t::match_allocate (void *h,
 
     if (!match_op_valid (match_op)) {
         m_err_msg += __FUNCTION__;
-        m_err_msg +=
-            ": ERROR: Invalid Match Option: " + std::string (match_op_to_string (match_op)) + "\n";
+        m_err_msg += ": ERROR: Invalid Match Option: " + std::to_string (match_op) + "\n";
         errno = EINVAL;
         return -1;
     }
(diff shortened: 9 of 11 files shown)


resource/policies/base/match_op.h
```diff
@@ -1,54 +1,28 @@
...
+#endif
+
 typedef enum match_op_t {
     START_MATCH_OP_T = 0,
     MATCH_UNKNOWN,
     MATCH_ALLOCATE,
     MATCH_ALLOCATE_W_SATISFIABILITY,
     MATCH_ALLOCATE_ORELSE_RESERVE,
-    MATCH_SATISFIABILITY
+    MATCH_SATISFIABILITY,
```
jameshcorbett, line 27: This change belongs in the previous commit

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+    std::map<match_op_t, const char *>::const_iterator candidate = match_options.find (match_op);
+
+    if (candidate != match_options.end ())
+        return candidate->second;
+
+    return "error";
+}
+
+const match_op_t string_to_match_op (const char *str)
```
jameshcorbett: Since this function returns by value, the `const` is meaningless.

resource/policies/base/match_op.h
```diff
@@ -1,38 +1,28 @@
...
-            return "allocate_with_satisfiability";
-        case MATCH_SATISFIABILITY:
-            return "satisfiability";
-        default:
-            return "error";
-    }
-}
+const char *match_op_to_string (match_op_t match_op);
+
+const match_op_t string_to_match_op (const char *str);
```
jameshcorbett: Since this function returns by value, the const is meaningless.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <cstring>
+#include <map>
+
+const std::map<match_op_t, const char *> match_options = {{MATCH_UNKNOWN, "error"},
+                                                          {MATCH_ALLOCATE, "allocate"},
+                                                          {MATCH_ALLOCATE_ORELSE_RESERVE,
+                                                           "allocate_orelse_reserve"},
+                                                          {MATCH_ALLOCATE_W_SATISFIABILITY,
+                                                           "allocate_with_satisfiability"},
+                                                          {MATCH_SATISFIABILITY, "satisfiability"}};
```
jameshcorbett: Minor but this could be `constexpr std::array<std::pair<match_op_t,const char*>>`.

resource/reapi/test/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_executable(reapi_cli_cxx_test reapi_cli_test_cxx.cpp)
+add_sanitizers(reapi_cli_cxx_test)
+target_link_libraries(reapi_cli_cxx_test PRIVATE reapi_cli intern Catch2::Catch2WithMain)
```
jameshcorbett, line 3: You may need `match_option` here as well
```suggestion
target_link_libraries(reapi_cli_cxx_test PRIVATE reapi_cli intern match_option Catch2::Catch2WithMain)
```

qmanager/modules/CMakeLists.txt
```diff
@@ -11,4 +11,5 @@ target_link_libraries(sched-fluxion-qmanager PRIVATE
     flux::schedutil
     PkgConfig::JANSSON
     cppwrappers
+    resource
```
jameshcorbett: You may need match_option here as well

jameshcorbett (commented):
Generally looks good!

resource/policies/base/match_op.h
jacobtkeio: So it does!

resource/policies/base/match_op.cpp
jacobtkeio: Could you explain this point to me? My thought was that this const forces this function's users to treat the returned value as a constant instead of a mutable value. Is that already true of return-by-value values, or does this const just get discarded?

I will remove it either way!

jameshcorbett: Basically this `const` just gets discarded. When you return by value, you can't force the caller to treat the returned value as a constant --- they get a copy and it's theirs to do with as they please. You can return a reference or a pointer to something that is const though.

In this case though the return value is just an enum, so basically like an int, so I don't think there's any reason to try to prevent callers from modifying what they get back!

jacobtkeio: That makes sense, thanks! :)

resource/reapi/test/CMakeLists.txt
jacobtkeio: In hindsight I'm surprised this linked in the first place. Is reapi_cli_cxx_test somehow reaching into reapi_cli's private linkage to find match_option? Even if that works, I can see why explicit linking is better (like explicit includes)

jameshcorbett: To be honest I'm not sure!

qmanager/modules/CMakeLists.txt
jacobtkeio: For now, I've linked `qmanager` against `match_option` directly instead of `resource` (I am not sure why I originally linked `resource` instead of `match_op`, but `match_option` is the necessary one), but I linked both `resource` and `reapi_cli` publicly against `match_option` and left `reapi_cli_test_cxx` linked against only `reapi_cli`.
My logic is that users of the `resource` and `reapi_cli` libraries shouldn't need to care about the details of linking against `match_option`, so `match_op` should just come for free with either `resource` or `reapi_cli`. I'm not sure that's actually how public linkage works though.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,37 @@
...
+        if (p.first == match_op)
+            return p.second;
+
+    return "error";
+}
+
+match_op_t string_to_match_op (const char *str)
+{
+    for (const auto p : opt_map)
+        if (strcmp (p.second, str) == 0)
```
jameshcorbett: Claude found a potential problem here: 

 The array is sized (int)(END_MATCH_OP_T)-1. Today that's exactly 5 and matches the 5 initializers. But add a match_op_t before END_
  and forget the map entry: the array pads with a default {START_MATCH_OP_T, nullptr} pair, and string_to_match_op then runs            
  strcmp(nullptr, str) → crash. Guarding the strcmp closes the gap:                                            
  `if (p.second && strcmp (p.second, str) == 0)`

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    return MATCH_UNKNOWN;
+}
+
+bool match_op_valid (match_op_t match_op)
+{
+    return match_op != MATCH_UNKNOWN
+           && count_if (opt_map.begin (), opt_map.end (), [match_op] (auto p) {
+                  return p.first == match_op;
+              }) == 1;
```
jameshcorbett: This took me a sec to understand --- it might be simpler as 

```
      if (match_op == MATCH_UNKNOWN)
          return false;                                                                                                                 
                  
      for (const auto &p : opt_map)                                                                                                     
          if (p.first == match_op)
              return true;                                                                                                              
                  
      return false;
```

But if you agree with my other comment then this function would fall back to just calling `match_op_to_string`.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,37 @@
...
+     std::pair (MATCH_SATISFIABILITY, "satisfiability")};
+
+const char *match_op_to_string (match_op_t match_op)
+{
+    for (const auto p : opt_map)
+        if (p.first == match_op)
+            return p.second;
+
+    return "error";
+}
```
jameshcorbett: I think it would actually make more sense to drop `MATCH_UNKNOWN` from this array and also to have `match_op_to_string` return `nullptr` instead of `"error"`. Also, then `match_op_valid` could just check whether `match_op_to_string` returns `nullptr`.

qmanager/modules/CMakeLists.txt
jacobtkeio: After talking to Dan and Tom, this seems like an acceptable approach. Public linkage is nice because it exposes the enum and three helper functions to users of resource and reapi_cli, who won't care about the match_op details when they use resource or reapi_cli.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,37 @@
+#include "match_op.h"
+#include <cstring>
```
jameshcorbett: Include the usual license and copyright boilerplate here?

resource/policies/base/match_op.cpp
jameshcorbett: In this case you wouldn't need `<algorithm>`. But there would be two new changes you'd need to make:

  1. reapi_cli_impl.hpp:79 — which will need to be something like    
  `": ERROR: Invalid Match Option: " + std::to_string (match_op) + "\n";`                                                           

                                                                       
  2. The idempotence test — it starts the loop at START_MATCH_OP_T + 1, which now has no string, so                   
  string_to_match_op(nullptr) segfaults. Start at + 2.

jacobtkeio:
Dan and Tom have also requested that I rename `string_to_match_op` to `match_op_from_string` for better namespacing, and we decided that `match_op` is a better name for the library object

resource/policies/base/match_op.h
```diff
@@ -1,38 +1,28 @@
 #ifndef MATCH_OP_H
 #define MATCH_OP_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 typedef enum match_op_t {
+    START_MATCH_OP_T = 0,
```
jameshcorbett: I think maybe we could leave this out given that we know that the the enums start at 0 and increment monotonically?

jameshcorbett (commented):
I have some suggestions but overall I think this looks good!

resource/policies/base/match_op.cpp
jacobtkeio: Ah great catch! Do you know of a better way to automate getting the length of this array?
I'll add the guard!

jameshcorbett: Claude has a suggestion ---
```
std::to_array — let the initializer size it, so there's no manual count to keep in sync and no nullptr padding:                       
                                                                                                                                        
  constexpr auto opt_map = std::to_array<std::pair<match_op_t, const char *>> (                                                         
      {{MATCH_ALLOCATE, "allocate"},
       {MATCH_ALLOCATE_ORELSE_RESERVE, "allocate_orelse_reserve"},                                                                      
       {MATCH_ALLOCATE_W_SATISFIABILITY, "allocate_with_satisfiability"},                                                               
       {MATCH_SATISFIABILITY, "satisfiability"}});                                                                                      
                                                                                                                                        
  static_assert (opt_map.size () == END_MATCH_OP_T - START_MATCH_OP_T - 2,                                                              
                 "opt_map is missing a match_op_t entry");                                                                              
                                                                                                                                        
  to_array (C++20, you're on C++23) removes the (int)(END_MATCH_OP_T)-1 magic number and its silent-padding segfault. The static_assert 
  is the actual guard you asked for: add an enum value and forget the map entry → size won't match → compile error, not a 3am crash. The
   -2 accounts for START_ and MATCH_UNKNOWN, neither of which is mapped.  
```

resource/policies/base/match_op.cpp
jacobtkeio: Is it alright if I use a neighboring one with `Copyright 2014`? Or should I update the date? I can't remember the protocol

jameshcorbett: I think the policy is always to start with the current year!

resource/policies/base/match_op.cpp
jacobtkeio: Great suggestion!! Thank you so much for the detailed review :) I agree with this change
I left match_op_is_valid as `return (match_op_to_string (match_op) == nullptr)` for readability even if the `== nullptr` could get dropped

jacobtkeio: I just realized I should have also been checking if `match_op_from_string`'s argument is null!

jacobtkeio: Guarding against a null argument will give the idempotence test back its behavior, but do you think it's preferable to leave the null check out and not guarantee idempotence for MATCH_UNKNOWN?

resource/policies/base/match_op.cpp
jacobtkeio: Yep, I went with your other suggestion! I actually had written all of these using lambdas at one point but in hindsight I feel like they're just not readable enough in C++  (especially with automated formatting) to justify their use when unnecessary after all

resource/policies/base/match_op.cpp
jacobtkeio: Oh very nice! I will add it

jacobtkeio: I didn't realize static_assert worked like a compile-time assertion... I'll have to use it more :)

resource/policies/base/match_op.cpp
jameshcorbett: I think checking for NULL arguments and then setting `errno = EINVAL` and returning `nullptr` is what makes the most sense! And the tests would need to be adjusted accordingly.

resource/policies/base/match_op.cpp
jacobtkeio: I added one retroactively for match_op.h as well using the year it was created. I hope that's okay!

jameshcorbett: Definitely, although I think if you're going to add one anywhere it should be 2026. It's not a big deal though.

resource/policies/base/match_op.cpp
jacobtkeio: I've removed the `p.second` guard now that `const auto p : opt_map` is guaranteed to work correctly!

resource/policies/base/match_op.cpp
jacobtkeio: For now, I've set `errno = EINVAL` only when `match_op_from_string` receives `nullptr` or `match_op_to_string` receives an integer value outside the defined enums (less than START_ or greater than END_).
My idea was that `EINVAL` should mean the argument doesn't make sense at all, so I did not set `EINVAL` when converting a gibberish string (e.g. "" or "jfldjfsj") or converting one of the invalid match_op_t values (START_MATCH_OP_T, END_MATCH_OP_T, and MATCH_UNKNOWN).
Is this the behavior you would expect, or should we set `EINVAL` for all "invalid" conversions?

jameshcorbett: I think if it's an unrecognized string, either EINVAL or ENOENT could work, what do you think @garlick ? You know errno stuff much better than I do.

jameshcorbett:
Is this ready for another / final review?

jacobtkeio:
I think so, although I added ENOENT to the logic and tests, so feel free to wait for me to implement feedback from @.garlick's before the next/final review. Otherwise yes!
Side note I'm sorry for spamming everyone's emails today 😅  It seems like I always want to change just one more thing after pushing!

resource/policies/base/match_op.h
jameshcorbett: Not a big deal but you could actually leave out the entire `START_MATCH_OP_T` entry I think! Because we would know that the first enum value would be 0.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+static_assert (opt_map.size () == END_MATCH_OP_T - START_MATCH_OP_T - 2,
+               "opt_map is missing a match_op_t entry");
+
+const char *match_op_to_string (match_op_t match_op)
+{
+    if (match_op < START_MATCH_OP_T || match_op > END_MATCH_OP_T) {
+        errno = EINVAL;
+        return nullptr;
+    }
```
jameshcorbett: I think this should be `match_op <= START_MATCH_OP_T || match_op >= END_MATCH_OP_T` (you need the or-equal-to part). But if you take my suggestion of dropping `START_MATCH_OP_T` this could be `match_op < 0 || match_op >= END_MATCH_OP_T`.

resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -25,6 +26,51 @@ TEST_CASE ("Initialize REAPI CLI", "[initialize C++]")
...
+        errno = 0;
+    }
+
+    // invalid strings convert to MATCH_UNKNOWN + ENOENT
+    REQUIRE (match_op_from_string ("0xDEADBEEF") == MATCH_UNKNOWN);
+    REQUIRE (errno == ENOENT);
+    errno = 0;
+
+    // out-of-bounds enum values are invalid + EINVAL and convert to nullptr + EINVAL
+    for (match_op_t OOB_match_op : {(match_op_t)(-1), (match_op_t)(END_MATCH_OP_T + 1)}) {
```
jameshcorbett: Could add a check here for END_MATCH_OP_T (not `END_MATCH_OP_T + 1`) also being  invalid!

resource/policies/base/match_op.h
jameshcorbett: I found another reason to not insert `START_MATCH_OP_T` -- it shifts every enum value by 1, and that breaks the go bindings in this repo that have the same enum values manually specified. Wouldn't be a big deal to change, but better not to change them at all.

jameshcorbett (commented):
You may need the following as well in
 resource/reapi/bindings/CMakeLists.txt                                                                                                 
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌                                   
 28 +target_link_libraries( reapi_module PUBLIC                                                                                         
 29 +    match_op                                                                                                                       
 30 +    )

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,59 @@
...
+            return p.first;
+
+    errno = ENOENT;
+    return MATCH_UNKNOWN;
+}
+
+bool match_op_valid (match_op_t match_op)
+{
+    return (match_op_to_string (match_op) != nullptr);
+}
```
jameshcorbett, line 65: Sorry this is on me -- `match_op_to_string` sets errno which is fine but `match_op_valid` probably shouldn't set errno. So I think this function should be like:

```
bool rc;
int saved_errno = errno;

rc = (match_op_to_string (match_op) != nullptr);
errno = saved_errno;
return rc;
```

resource/policies/base/match_op.h
jacobtkeio: That makes sense! I will remove it

resource/policies/base/match_op.cpp
jacobtkeio: Sounds good! I did think about whether match_op_valid should set errno, I just thought the wrong thing 😆 In hindsight it makes sense that conversion functions should set errno but validation functions shouldn't

resource/policies/base/match_op.cpp
jacobtkeio: Do you think MATCH_UNKNOWN should set `errno = EINVAL`? Right now it's setting it to `ENOENT`

jameshcorbett: I think ENOENT is fine probably. I'm not too worried about it either way though.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,64 @@
...
+               "opt_map is missing a match_op_t entry");
+
+const char *match_op_to_string (match_op_t match_op)
+{
+    if (match_op < 0 || match_op >= END_MATCH_OP_T) {
+        errno = EINVAL;
+        return nullptr;
+    }
+
+    for (const auto p : opt_map)
```
jameshcorbett:
```suggestion
    for (const auto &p : opt_map)
```

because why not.

resource/policies/base/match_op.h
```diff
@@ -1,38 +1,37 @@
...
+
+match_op_t match_op_from_string (const char *str);
 
-static bool match_op_valid (match_op_t match_op)
-{
-    if ((match_op != MATCH_ALLOCATE) && (match_op != MATCH_ALLOCATE_W_SATISFIABILITY)
-        && (match_op != MATCH_ALLOCATE_ORELSE_RESERVE) && (match_op != MATCH_SATISFIABILITY)) {
-        return false;
-    }
+bool match_op_valid (match_op_t match_op);
```
jameshcorbett, line 35: Claude had an interesting comment on this:

> extern "C" + bool without <stdbool.h> (match_op.h): only compiles because every includer is C++. If C consumption is actually       
>   intended, add #include <stdbool.h>; if not, the extern "C" wrapper is speculative and could just be dropped.

jameshcorbett: I am actually not sure, do we need this to be `extern C`?

jameshcorbett (dismissed):
I have a couple of final comments but otherwise this LGTM!

Lines 45 and 153 of reapi_module_impl could use an additional check of `|| cmd == nullptr) {`

resource/policies/base/match_op.cpp
trws:
```suggestion
    for (const auto& [op,str] : opt_map)
```

jameshcorbett:
I think commit [3765a74](https://github.com/flux-framework/flux-sched/pull/1515/commits/3765a74acd97570184ca916f951b1d91e85b9384) should be squashed into a prior commit. According to our developer guidelines, a PR should not introduce changes (or issues) that it later updates or fixes in later commits (https://flux-framework.readthedocs.io/en/latest/contributing.html#developer-guidelines).

resource/policies/base/match_op.h
jacobtkeio: My thought is that users of the C bindings to the REAPI CLI may want to use these functions, so ideally they should be C-compatible. IIRC I ran into an issue with linking that led to the `extern "C"` originally (and maybe it was one of the conversion functions that doesn't use bool, so it didn't blow up then?).
I will add a C-only include of `stdbool.h`!

resource/policies/base/match_op.cpp
jacobtkeio: I went with `!cmd` on lines 45 and 153 of reapi_module_impl to match the surrounding style!

jameshcorbett (dismissed):
Great thanks!

jacobtkeio:
Thank you so much for all your feedback! I will do my best to avoid these errors in my future PRs.

Per Hobbs' request, let's see what happens when I comment @.mergifyio queue:

jacobtkeio:
@mergifyio queue

wihobbs:
Thanks @jacobtkeio! That response gives us a specific setting we need to look into if we want to allow guest contributors to set MWP. I'll go ahead and set it on this approved PR.

jameshcorbett:
I believe @milroy still wants to review so I removed MWP

milroy:
I'm adding comments now and will get my review in ASAP.

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+#include <cstring>
+#include <array>
```
milroy, line 13: Include `utility` instead:

```suggestion
#include <utility>
```

This likely compiles because libstdc++/libc++ happen to drag `<utility>` in via `<array>` but that isn't guaranteed by the standard.

resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -25,6 +26,48 @@ TEST_CASE ("Initialize REAPI CLI", "[initialize C++]")
...
+    // nullptr converts to MATCH_UNKNOWN + EINVAL (since invalid ptr)
+    REQUIRE (match_op_from_string (nullptr) == MATCH_UNKNOWN);
+    REQUIRE (errno == EINVAL);
+    errno = 0;
+
+    // valid match_op_t are valid
+    // and match_op_t <-> string conversion is idempotent
+    for (int op = MATCH_UNKNOWN + 1; op != END_MATCH_OP_T; op++) {
+        REQUIRE (match_op_valid ((match_op_t)op));
+        REQUIRE (match_op_from_string (match_op_to_string ((match_op_t)op)) == op);
```
milroy: This is actually testing that `match_op_from_string/match_op_to_string` are mutually consistent.

```suggestion
        REQUIRE (match_op_from_string ("allocate") == MATCH_ALLOCATE);
        REQUIRE (match_op_from_string ("allocate_orelse_reserve") == MATCH_ALLOCATE_ORELSE_RESERVE);
        ... etc...
```

resource/policies/base/match_op.cpp
```diff
@@ -0,0 +1,64 @@
+/*****************************************************************************\
```
milroy, line 1: I don't know where else to put this, but the third commit ("match_op: Query match_op<->string by a defined map") has two concatenated commit message bodies.

milroy: There is no `START_MATCH_OP_T`; I think the enum you mean is `MATCH_UNKNOWN`.

milroy: The function is `match_op_valid`, not `match_op_is_valid`.

resource/policies/base/match_op.h
```diff
@@ -1,6 +1,24 @@
+/*****************************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
```
milroy:
```suggestion
 * Copyright 2026 Lawrence Livermore National Security, LLC
```

milroy (changes requested):
This is really close to being merge-able, but there are a couple of minor things to fix and one moderate problem. I think @jameshcorbett was suggesting to merge commit 1 and 3 as commit 1 introduces changes that commit 3 updates. That needs to get fixed before merging.

resource/policies/base/match_op.cpp
jacobtkeio: \<array> is still required to use `std::make_array`, so I'll add \<utility> and keep \<array>.

resource/reapi/test/reapi_cli_test_cxx.cpp
jacobtkeio: I think it makes sense to have both a mutual consistency test and tests for specific conversions (to save us from having to write the equivalent `REQUIRE (0 == strcmp (match_op_to_string (MATCH_ALLOCATE), "allocate"))` cases for each option).

jacobtkeio: This is orthogonal, but since none of these test expressions depend on any of the others I'm swapping all the REQUIREs for CHECKs so errors get reported all at once instead of one at a time.

resource/policies/base/match_op.cpp
jacobtkeio: I am inclined to keep a lot of the detail, but I've reformatted the commit message into one commit body. Let me know if I should make it shorter!

resource/policies/base/match_op.cpp
trws: Eh, it's debatable. I'm 100% fine with it like this but assuming `errno` remains unset after anything is... fraught.

resource/policies/base/match_op.h
trws: Good catch!

trws (approved):
Looks good to me, good work @jacobtkeio.

resource/policies/base/match_op.cpp
milroy: Looks good; I should have suggested just to add `<utility>`.

resource/policies/base/match_op.cpp
milroy: Looks good now!

milroy (approved):
Nice work; this is good to merge now!

PR #1516, opened 2026-06-30
reader/jgf: vertex equality should ignore properties

### LINKED ISSUE #1513: queue reconfig causes `find_vtx` to error out
```
[root@tuolumne1:tmp]# journalctl -xeu flux | tail
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-qmanager.info[0]: raising fatal exception on running job id=f3F1F7E9iCz3
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-resource.err[0]: run: dfu_traverser_t::run (id=965769565927639040): find_vtx: inconsistent input vertex for 5697.
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-resource.err[0]:  properties=(dat_0619= pall= , pall= pbatch= ).
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-resource.err[0]: run_update: run: Invalid argument
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-resource.err[0]: update_request_cb: update failed (id=965769565927639040): Invalid argument
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-qmanager.err[0]: jobmanager_hello_cb: reconstruct (id=965769565927639040 queue=pbatch): Invalid argument
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-qmanager.info[0]: raising fatal exception on running job id=f3F2MmqjWvwh
Jun 18 15:00:12 tuolumne1 flux[3199643]: sched-fluxion-qmanager.err[0]: error raising fatal exception on f3F2MmqjWvwh: job is inactive: No such file or directory
Jun 18 15:01:07 tuolumne1 flux[3199643]: sched-fluxion-qmanager.err[0]: jobmanager_free_cb: can't find queue for job (id=965769565927639040): No such file or directory
Jun 18 15:01:14 tuolumne1 flux[3199643]: job-exec.info[0]: job-exception: id=f3FC4yYcUQa3: resource allocation expired
```

The error message is reporting a mismatch between the properties it thinks a node has (pall, pdat0619) and the properties a job's JGF is saying the node has (pall, pbatch)

If I'm right, we should be able to reproduce on hetchy fairly easily: submit a job, and remove that job's node from a queue and add it to a different queue, then reload fluxion and let the job complete. The job's JGF, which was generated using the old queue config, won't match Fluxion's new resource representation.

trws:
problem: jgf currently includes potentailly dynamic properties in its notion of equality, and as a result will kill jobs that are running on nodes with non-queue properties set when restored after a restart.

solution: don't consider the properties as part of equality for update/find_vtx in jgf

fixes #1513 

I'm still mulling on the best way to make a meaningful test for this, if anyone has an idea let me know.

### DIFF
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -114,10 +114,10 @@ void fetch_helper_t::scrub ()
 
 bool operator== (const resource_pool_t &r, const fetch_helper_t &f)
 {
+    // we explicitly exclude properties because they may be dynamic
     return (r.type.get () == f.type && r.basename == f.basename
             && r.size == static_cast<unsigned int> (f.size) && r.rank == static_cast<int> (f.rank)
-            && r.id == f.id && r.name == f.name && r.properties == f.properties
-            && r.paths == f.paths);
+            && r.id == f.id && r.name == f.name && r.paths == f.paths);
 }
 
 bool operator!= (const resource_pool_t &r, const fetch_helper_t &f)
--- a/t/t1030-rv1-shorthand.t
+++ b/t/t1030-rv1-shorthand.t
@@ -208,6 +208,37 @@ test_expect_success 'sched-now=allocated is null' '
     flux ion-resource find -q --format=jgf sched-now=allocated | jq -e ". == null"
 '
 
+# test reloading with a job that has one of its nodes' properties updated
+# (see https://github.com/flux-framework/flux-sched/issues/1513)
+test_expect_success 'set a dynamic property on hetchy1002' '
+    NODE_PATH="/hetchy/chassis0/hetchy1002" &&
+    flux ion-resource set-property ${NODE_PATH} maintenance=1 &&
+    flux ion-resource get-property ${NODE_PATH} maintenance \
+        | grep "maintenance = \[.1.\]"
+'
+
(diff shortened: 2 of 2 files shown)


jameshcorbett (approved):
I think it might be helpful to update the `diff` function as well, so that it notes that property mismatches do not affect the equality operator. Thinking about tests...

jameshcorbett:
I came up with a test here: https://github.com/flux-framework/flux-sched/compare/master...jameshcorbett:flux-sched:jgf-vtx-equality

milroy:
The fix LGTM.

> I came up with a test here: https://github.com/flux-framework/flux-sched/compare/master...jameshcorbett:flux-sched:jgf-vtx-equality

I like this test, too. It might be good to force a reconstruction failure and check the graph state for inconsistency, but that's out of scope for this PR and probably best as an addition to `t3022`.

trws:
Pulled in the test and rebased.  I tried reworking it to a multi-host constraint to try and get a case where part of the operation completes before the failure hits, but it isn't surfacing the issue without more investigation, will have to come back to that.

jameshcorbett (approved):
LGTM!

PR #1517, opened 2026-07-06
modules: Merge resource_prop_t into resource_opt_t

jacobtkeio:
This PR is tentative and only open to gather opinions.

Problem: `resource_match_opts.*` has two related classes, `resource_opts_t` and `resource_prop_t`, which both share setters and getters for every option in the resource module.  `resource_opts_t` has a `resource_prop_t` member variable and calls `m_resource_prop.get_...()` in its own `resource_opts_t::get_...()` functions, and likewise for its `set`-ters and `is_set`-ters. This causes half of `resource_match_opts.*` to be boilerplate and makes it more difficult to add new module options.

Solution: `Make resource_opts_t` inherit from `rsrc_prop_t`, and replace all `resource_opts_t` function calls with the inherited `resource_prop_t` implementations. Remove the duplicated functions in `resource_match_opts.*`.

Is there a useful semantic relationship this change destroys? Is the decoupling of the module property data from its parsing logic desirable here?
Or could we take it farther and completely merge `resource_opts_t` and `resource_prop_t`?

### DIFF
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -377,7 +377,6 @@ ret:
 int qmanager_opts_t::parse (const std::string &k, const std::string &v, std::string &info)
 {
     int rc = 0;
-    std::string dflt;
     std::map<std::string, std::string> tmp_mp;
     int key = static_cast<int> (qmanager_opts_key_t::UNKNOWN);
 
--- a/resource/modules/feasibility.cpp
+++ b/resource/modules/feasibility.cpp
@@ -38,7 +38,6 @@ static void set_default_args (std::shared_ptr<resource_ctx_t> &ctx)
     ct_opts.set_match_policy ("first", e);
     ct_opts.set_prune_filters ("ALL:core,ALL:node,ALL:gpu,ALL:ssd");
     ct_opts.set_match_format ("rv1_nosched");
-    ct_opts.set_update_interval (0);
     ctx->opts += ct_opts;
 }
 
@@ -250,6 +249,17 @@ error:
 static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = 0;
+    const resource_opts_t &opts = ctx->opts.get_opt ();
+    const std::optional<std::string> &match_subsystems = opts.get_match_subsystems ();
+    const std::optional<std::string> &match_format = opts.get_match_format ();
+    const std::optional<std::string> &prune_filters = opts.get_prune_filters ();
+
+    // Check for all required module options.
+    rc += required_module_option (ctx, match_subsystems, "subsystems");
+    rc += required_module_option (ctx, match_format, "match-format");
+    rc += required_module_option (ctx, opts.get_load_format (), "load-format");
+    if (rc != 0)
+        return rc;
 
     // The feasibility module should only use "first". Exclusivity has no effect.
     if (!(ctx->matcher = create_match_cb ("first"))) {
@@ -265,26 +275,31 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
                   LOG_ERR,
                   "%s: error processing subsystems %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_match_subsystems ().c_str ());
+                  match_subsystems->c_str ());
         return rc;
     }
 
-    // Create a writers object for matched vertices and edges
-    match_format_t format =
-        match_writers_factory_t::get_writers_type (ctx->opts.get_opt ().get_match_format ());
-    if (!(ctx->writers = match_writers_factory_t::create (format)))
-        return -1;
-
-    if (ctx->opts.get_opt ().is_prune_filters_set ()
-        && ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (),
-                                                   ctx->opts.get_opt ().get_prune_filters ())
-               < 0) {
+    // Create a writers object for matched vertices and edges.
+    if (!(ctx->writers = match_writers_factory_t::create (
+              match_writers_factory_t::get_writers_type (*match_format)))) {
         flux_log (ctx->h,
                   LOG_ERR,
-                  "%s: error setting pruning types with: %s",
+                  "%s: error creating match writer with: %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_prune_filters ().c_str ());
+                  match_format->c_str ());
         return -1;
+    }
+
+    if (prune_filters) {
+        if (ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (), *prune_filters)
+            < 0) {
+            flux_log (ctx->h,
+                      LOG_ERR,
+                      "%s: error setting pruning types with: %s",
+                      __FUNCTION__,
+                      prune_filters->c_str ());
+            return -1;
+        }
     }
 
     // Initialize the DFU traverser
--- a/resource/modules/resource.cpp
+++ b/resource/modules/resource.cpp
@@ -1215,7 +1215,7 @@ static void notify_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
             flux_log_error (h, "%s: flux_msg_route_first", __FUNCTION__);
             goto error;
         }
-        if (ctx->opts.get_opt ().is_load_file_set ()) {
+        if (ctx->opts.get_opt ().get_load_file ()) {
             errno = ENODATA;
             // Since m_acquired_resources is null,
             flux_log_error (ctx->h, "%s: cannot notify when load-file set", __FUNCTION__);
@@ -1361,13 +1361,15 @@ static void status_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
     json_t *R_alloc = nullptr;
     std::chrono::time_point<std::chrono::system_clock> now;
     std::chrono::duration<double> elapsed;
+    std::optional<int> update_interval;
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
     now = std::chrono::system_clock::now ();
     elapsed = now - ctx->m_resources_alloc_updated;
+    update_interval = ctx->opts.get_opt ().get_update_interval ();
     // Get R alloc whenever m_resources_alloc_updated or
     // the elapsed time is greater than configured limit
-    if ((elapsed.count () > static_cast<double> (ctx->opts.get_opt ().get_update_interval ()))
+    if ((update_interval && (elapsed.count () > static_cast<double> (*update_interval)))
         || ctx->m_resources_updated) {
         if (run_find (ctx, "sched-now=allocated", "rv1_nosched", &R_alloc) < 0)
             goto error;
@@ -1572,14 +1574,41 @@ error:
 static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = 0;
+    const resource_opts_t &opts = ctx->opts.get_opt ();
+    const std::optional<std::string> &traverser_policy = opts.get_traverser_policy ();
+    const std::optional<std::string> &match_policy = opts.get_match_policy ();
+    const std::optional<std::string> &match_subsystems = opts.get_match_subsystems ();
+    const std::optional<std::string> &match_format = opts.get_match_format ();
+    const std::optional<std::string> &prune_filters = opts.get_prune_filters ();
+
+    // Check for all required module options.
+    rc += required_module_option (ctx, traverser_policy, "traverser");
+    rc += required_module_option (ctx, match_policy, "match-policy");
+    rc += required_module_option (ctx, match_subsystems, "subsystems");
+    rc += required_module_option (ctx, match_format, "match-format");
+    rc += required_module_option (ctx, opts.get_load_format (), "load-format");
+    if (rc != 0)
+        return rc;
 
     // Select the appropriate traverser based on CLI policy.
-    ctx->traverser =
-        std::make_shared<dfu_traverser_t> (ctx->opts.get_opt ().get_traverser_policy ());
+    try {
+        ctx->traverser = std::make_shared<dfu_traverser_t> (*traverser_policy);
+    } catch (std::bad_alloc &e) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "%s: can't create traverser for policy %s",
+                  __FUNCTION__,
+                  traverser_policy->c_str ());
+        return -1;
+    }
 
     // Select the appropriate matcher based on CLI policy.
-    if (!(ctx->matcher = create_match_cb (ctx->opts.get_opt ().get_match_policy ()))) {
-        flux_log (ctx->h, LOG_ERR, "%s: can't create match callback", __FUNCTION__);
+    if (!(ctx->matcher = create_match_cb (*match_policy))) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "%s: can't create match callback for policy %s",
+                  __FUNCTION__,
+                  match_policy->c_str ());
         return -1;
     }
     if ((rc = populate_resource_db (ctx)) != 0) {
@@ -1591,26 +1620,31 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
                   LOG_ERR,
                   "%s: error processing subsystems %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_match_subsystems ().c_str ());
+                  match_subsystems->c_str ());
         return rc;
     }
 
-    // Create a writers object for matched vertices and edges
-    match_format_t format =
-        match_writers_factory_t::get_writers_type (ctx->opts.get_opt ().get_match_format ());
-    if (!(ctx->writers = match_writers_factory_t::create (format)))
-        return -1;
-
-    if (ctx->opts.get_opt ().is_prune_filters_set ()
-        && ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (),
-                                                   ctx->opts.get_opt ().get_prune_filters ())
-               < 0) {
+    // Create a writers object for matched vertices and edges.
+    if (!(ctx->writers = match_writers_factory_t::create (
+              match_writers_factory_t::get_writers_type (*match_format)))) {
         flux_log (ctx->h,
                   LOG_ERR,
-                  "%s: error setting pruning types with: %s",
+                  "%s: error creating match writer with: %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_prune_filters ().c_str ());
+                  match_format->c_str ());
         return -1;
+    }
+
+    if (prune_filters) {
+        if (ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (), *prune_filters)
+            < 0) {
+            flux_log (ctx->h,
+                      LOG_ERR,
+                      "%s: error setting pruning types with: %s",
+                      __FUNCTION__,
+                      prune_filters->c_str ());
+            return -1;
+        }
     }
 
     // Initialize the DFU traverser
--- a/resource/modules/resource_match.cpp
+++ b/resource/modules/resource_match.cpp
@@ -146,8 +146,8 @@ static int create_reader (std::shared_ptr<resource_ctx_t> &ctx, const std::strin
 {
     if ((ctx->reader = create_resource_reader (format)) == nullptr)
         return -1;
-    if (ctx->opts.get_opt ().is_load_allowlist_set ()) {
-        if (ctx->reader->set_allowlist (ctx->opts.get_opt ().get_load_allowlist ()) < 0)
+    if (auto load_allowlist = ctx->opts.get_opt ().get_load_allowlist ()) {
+        if (ctx->reader->set_allowlist (*load_allowlist) < 0)
             flux_log (ctx->h, LOG_ERR, "%s: setting allowlist", __FUNCTION__);
         if (!ctx->reader->is_allowlist_supported ())
             flux_log (ctx->h, LOG_WARNING, "%s: allowlist unsupported", __FUNCTION__);
@@ -159,19 +159,22 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
 {
     int rc = -1;
     int saved_errno;
+    // load_format and in_file_path are guaranteed to be set
+    std::string load_format = *(ctx->opts.get_opt ().get_load_format ());
+    std::string in_file_path = *(ctx->opts.get_opt ().get_load_file ());
     std::ifstream in_file;
     std::stringstream buffer{};
     graph_duration_t duration;
 
-    if (ctx->reader == nullptr
-        && create_reader (ctx, ctx->opts.get_opt ().get_load_format ()) < 0) {
+    if (ctx->reader == nullptr && create_reader (ctx, load_format) < 0) {
         flux_log (ctx->h, LOG_ERR, "%s: can't create reader", __FUNCTION__);
         goto done;
     }
 
     saved_errno = errno;
     errno = 0;
-    in_file.open (ctx->opts.get_opt ().get_load_file ().c_str (), std::ifstream::in);
+
+    in_file.open (in_file_path.c_str (), std::ifstream::in);
     if (!in_file.good ()) {
         if (errno == 0) {
             // C++ standard doesn't guarantee to set errno but
@@ -179,10 +182,7 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
             // we manually set errno only when it is not set at all.
             errno = EIO;
         }
-        flux_log_error (ctx->h,
-                        "%s: opening %s",
-                        __FUNCTION__,
-                        ctx->opts.get_opt ().get_load_file ().c_str ());
+        flux_log_error (ctx->h, "%s: opening %s", __FUNCTION__, in_file_path.c_str ());
         goto done;
     }
     errno = saved_errno;
@@ -354,18 +354,18 @@ int populate_resource_db (std::shared_ptr<resource_ctx_t> &ctx)
     std::chrono::time_point<std::chrono::system_clock> start;
     std::chrono::duration<double> elapsed;
 
-    if (ctx->opts.get_opt ().is_reserve_vtx_vec_set ())
-        ctx->db->resource_graph.m_vertices.reserve (ctx->opts.get_opt ().get_reserve_vtx_vec ());
+    if (auto reserve_vtx_vec = ctx->opts.get_opt ().get_reserve_vtx_vec ())
+        ctx->db->resource_graph.m_vertices.reserve (*reserve_vtx_vec);
 
     start = std::chrono::system_clock::now ();
-    if (ctx->opts.get_opt ().is_load_file_set ()) {
+    if (auto load_file = ctx->opts.get_opt ().get_load_file ()) {
         if (populate_resource_db_file (ctx) < 0)
             goto done;
         flux_log (ctx->h,
                   LOG_INFO,
                   "%s: loaded resources from %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_load_file ().c_str ());
+                  load_file->c_str ());
     } else {
         if (populate_resource_db_acquire (ctx) < 0) {
             flux_log (ctx->h, LOG_ERR, "%s: populate_resource_db_acquire", __FUNCTION__);
@@ -1250,7 +1250,7 @@ int select_subsystems (std::shared_ptr<resource_ctx_t> &ctx)
      * subsystem1[:relation1[:relation2...]],subsystem2[...
      */
     int rc = 0;
-    std::stringstream ss (ctx->opts.get_opt ().get_match_subsystems ());
+    std::stringstream ss (ctx->opts.get_opt ().get_match_subsystems ().value_or (""));
     subsystem_t subsystem;
     std::string token;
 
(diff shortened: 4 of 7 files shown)


jacobtkeio:
I forgot to mention, resource_prop_t is currently not referenced outside of resource_match_opts.cpp.

resource/modules/resource_match_opts.cpp
```diff
@@ -543,22 +388,22 @@ int resource_opts_t::parse (const std::string &k, const std::string &v, std::str
 
     switch (key) {
         case static_cast<int> (resource_opts_key_t::LOAD_FILE):
```
trws: If you're willing, you could kill off one more thing that makes me sad in this chunk of code.  The `switch` statement in C++ accepts enums and enum classes natively, and can often offer compile-time exhaustiveness checks when using them. For some reason we're adding a bunch of casts to int to remove useful compile-time checks.  So... yeah, if key were a `resource_opts_key_t` it should only need one cast from int on the `m_tab` (which could also be enum typed, but whatever) and then all these other static_casts can go away.

resource/modules/resource_match_opts.cpp
trws: The static_casts made me look over here too.  This is another one you could squash if you're so inclined.

This is the more modern format that's a bit more readable.

```suggestion
m_tab = {
    {"load-file", resource_opts_key_t::LOAD_FILE},
    {"load-format", resource_opts_key_t::LOAD_FORMAT},
    ...
    {last one}
};
```

It could actually just be on the declaration of the member like this, and switch to a default constructor probably.

Really this should just be a constant, since it's never inserted into from anything else, but doing it all as separate inserts at runtime rather than potentially compile-time constructing it is especially nasty.

wihobbs:
> Is there a useful semantic relationship this change destroys? Is the decoupling of the module property data from its parsing logic desirable here?

The lack of documentation for `rsrc_prop_t` (even combing through years of commit history) makes it difficult to reason about why it was separate from the start. The only _guess_ I can make is that `rsrc_opt_t` was designed to parse TOML input only, while `rsrc_prop_t` was meant to support storing options from multiple sources (TOML, command line at initialization, maybe commands too?). 

There's probably a negligible and not-worth-caring-about performance gain from having a smaller class, but as you point out, this is only called once at initialization of the feasibility module, and once at the initialization of the resource module.

> Or could we take it farther and completely merge resource_opts_t and resource_prop_t?

I think you've made a good case for simplifying this code a lot. I'd say go for it. (BTW the LLMs agreed with me. We have the machines on our side, for now.)

resource/modules/resource_match_opts.hpp
wihobbs: Noop, not used anywhere, should be removed?

wihobbs: That didn't render exactly right. I meant that we should remove 
```
    const resource_prop_t &get_resource_prop () const;
```
In the new class since it's a no-op that isn't used anywhere.

resource/modules/resource_match_opts.cpp
jacobtkeio: For now I've made the map a constant static member variable, since in principal I think there should only be one copy of the map. However, since there is only one copy of the class itself do you think it would be more clear just to leave it as a non-static constant member variable?

trws: I think constant static member is appropriate.

resource/modules/resource_match_opts.hpp
jacobtkeio: I removed it in the merger commit!

resource/modules/resource_match_opts.cpp
```diff
@@ -25,204 +25,7 @@ using namespace Flux::resource_model;
...
-                      is_prune_filters_set () ? get_prune_filters ().c_str () : nullptr,
-                      "update-interval",
-                      is_update_interval_set () ? get_update_interval () : 0,
-                      "traverser",
-                      is_traverser_policy_set () ? get_traverser_policy ().c_str () : nullptr);
-}
-
-////////////////////////////////////////////////////////////////////////////////
-// Private API for Resource Match Property Class
+// Private API for Resource Match Option Class
```
wihobbs, line 30: Is one consequence of moving `resource_prop_t` into `resource_opt_t` the loss of certain public API functionality in favor of private functions? I don't think so, but wanted to ask. As long as the API is internal that should be fine.

resource/modules/resource_match_opts.cpp
```diff
@@ -25,204 +25,7 @@ using namespace Flux::resource_model;
 using namespace Flux::opts_manager;
```
wihobbs, line 25: Nit: commit typo in title: s/rsrs/rsrc. Actually, let's fully spell out `resource` just to be clear. (50 char commit titles are a SHOULD not a MUST btw)

resource/modules/resource_match_opts.cpp
```diff
@@ -315,38 +258,38 @@ resource_opts_t &resource_opts_t::operator+= (const resource_opts_t &src)
...
         return k1 < k2;
-    return m_tab.at (k1) < m_tab.at (k2);
+    return opt_key_map.at (k1) < opt_key_map.at (k2);
 }
 
 int resource_opts_t::parse (const std::string &k, const std::string &v, std::string &info)
 {
     int rc = 0;
     std::string dflt;
     std::map<std::string, std::string> tmp_mp;
```
wihobbs:
```suggestion
```

The Machines caught this, existing: both these variables are unused. For that matter, in the same code in `qmanager/modules/qmanager_opts.cpp:311`, the same `std::string dflt` is also unused 🤦

resource/modules/resource_match_opts.hpp
```diff
@@ -43,7 +43,7 @@ class resource_opts_t : public optmgr_parse_t {
     };
 
     // These constructors can throw std::bad_alloc exception
-    resource_opts_t ();
+    resource_opts_t () = default;
```
wihobbs, line 44: I think a `std::bad_alloc` exception is less likely now but this is another thing The Machines say...

wihobbs (commented):
Insofar as I understand what's going on here, this is good with me! ✅ Nice work @jacobtkeio. Someone else should review this before merge-when-passing is added.

resource/modules/resource_match_opts.cpp
wihobbs: This rendered differently in the PR than it did in the code viewer. In the code viewer it looked like Public -> Private not Property -> Option...

jacobtkeio: I don't think any public API functionality gets lost either, since the name `resource_prop_t` doesn't appear anywhere in the repo except inside of `resource_opts_t`. All the public functions except `get_resource_prop()` are preserved, and only `add_to_prune_filters()` is added. Is there any reason to keep it`add_to_prune_filters()` private? It used to be part of `resource_prop_t`, and it only gets used in `resource_opts_t::parse()` now.

resource/modules/resource_match_opts.cpp
jacobtkeio: I've added two commits to remove them!

resource/modules/resource_match_opts.hpp
jacobtkeio: Ah I missed this. I can't figure out whether or not this particular default constructor is `noexcept` or not, but I think either way it makes sense to remove this specific warning about the now-removed old constructor.

jacobtkeio:
This latest push explodes the size of `resource: Merge resource_prop_t into resource_opts_t` to replace the `is_x_set ()` and `RESOURCE_OPTS_UNSET_STR` logic with `std::optional` (also returning optionals from the `get_x` functions). I added it all to one commit to avoid having two commits modifying the same code, but it would be easy to pull them apart again if necessary.
I'm also not sure it's worth touching the function signatures at all, and I can push a version that uses `std::optional` internally but still uses the same `is_x_set` and `get_x` functions without any `std::optional`s. Do you all have thoughts about which interface is nicest?

qmanager/modules/qmanager_opts.cpp
```diff
@@ -308,7 +308,6 @@ int qmanager_opts_t::jsonify (std::string &json_out) const
 int qmanager_opts_t::parse (const std::string &k, const std::string &v, std::string &info)
 {
     int rc = 0;
-    std::string dflt;
```
jameshcorbett, line 311: The two commits for unused variables could be one commit I think! No need for separate commits just because the unused vars are in different functions or files.

resource/modules/resource_match.cpp
```diff
@@ -1243,14 +1246,14 @@ int update_resource_db (std::shared_ptr<resource_ctx_t> &ctx,
...
-int select_subsystems (std::shared_ptr<resource_ctx_t> &ctx)
+int select_subsystems (std::shared_ptr<resource_ctx_t> &ctx, const std::string &subsystems_str)
 {
     /*
      * Format of match_subsystems
      * subsystem1[:relation1[:relation2...]],subsystem2[...
      */
     int rc = 0;
-    std::stringstream ss (ctx->opts.get_opt ().get_match_subsystems ());
+    std::stringstream ss (subsystems_str);
```
jameshcorbett: This change seems uncalled for -- why not just keep the function signature the same as before?

resource/modules/resource.cpp
```diff
@@ -1564,44 +1566,53 @@ static void set_status_request_cb (flux_t *h,
...
     // Select the appropriate matcher based on CLI policy.
-    if (!(ctx->matcher = create_match_cb (ctx->opts.get_opt ().get_match_policy ()))) {
-        flux_log (ctx->h, LOG_ERR, "%s: can't create match callback", __FUNCTION__);
+    if (match_policy
+        && !(ctx->matcher = create_match_cb (*ctx->opts.get_opt ().get_match_policy ()))) {
+        flux_log (ctx->h,
+                  LOG_ERR,
+                  "%s: can't create match callback for policy %s",
+                  __FUNCTION__,
+                  match_policy ? match_policy->c_str () : "(unset)");
```
jameshcorbett: these two new if-checks seem like a definite change of behavior and so I am suspicious of them! What heppens if the values of `traverser_policy` and `match_policy` are unset?

resource/modules/resource.cpp
```diff
@@ -1564,44 +1566,53 @@ static void set_status_request_cb (flux_t *h,
...
-    ctx->traverser =
-        std::make_shared<dfu_traverser_t> (ctx->opts.get_opt ().get_traverser_policy ());
+    if (traverser_policy)
+        ctx->traverser = std::make_shared<dfu_traverser_t> (*traverser_policy);
 
     // Select the appropriate matcher based on CLI policy.
-    if (!(ctx->matcher = create_match_cb (ctx->opts.get_opt ().get_match_policy ()))) {
-        flux_log (ctx->h, LOG_ERR, "%s: can't create match callback", __FUNCTION__);
+    if (match_policy
+        && !(ctx->matcher = create_match_cb (*ctx->opts.get_opt ().get_match_policy ()))) {
```
jameshcorbett: use `match_policy` here instead of the full `*ctx->opts.get_opt ().get_match_policy ()))`

resource/modules/resource.cpp
```diff
@@ -1353,13 +1353,15 @@ static void status_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
...
+    std::optional<int> update_interval;
     std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
 
     now = std::chrono::system_clock::now ();
     elapsed = now - ctx->m_resources_alloc_updated;
+    update_interval = ctx->opts.get_opt ().get_update_interval ();
     // Get R alloc whenever m_resources_alloc_updated or
     // the elapsed time is greater than configured limit
-    if ((elapsed.count () > static_cast<double> (ctx->opts.get_opt ().get_update_interval ()))
+    if (update_interval && (elapsed.count () > static_cast<double> (*update_interval))
```
jameshcorbett:
```suggestion
    if ((update_interval && (elapsed.count () > static_cast<double> (*update_interval)))
```

jameshcorbett: Not a correctness thing, just to emphasize precedence of operators.

resource/modules/resource_match_opts.cpp
```diff
@@ -24,445 +24,155 @@ using namespace Flux;
...
-
-void resource_opts_t::set_reserve_vtx_vec (const int i)
-{
-    m_resource_prop.set_reserve_vtx_vec (i);
-}
-
-void resource_opts_t::set_prune_filters (const std::string &o)
-{
-    m_resource_prop.set_prune_filters (o);
+    m_prune_filters.transform ([&p] (std::string m_pf) { return m_pf + "," + p; });
```
jameshcorbett: According to [the docs](https://en.cppreference.com/cpp/utility/optional/transform), `transform` returns a new value and doesn't modify the original, so this would have no effect. I think this should be something like

```
  if (m_prune_filters) *m_prune_filters += "," + p;
  else                 m_prune_filters = p;
```

this would make `add_to_prune_filters` always safe to call, so the invocation of `add_prune_filters` could drop the prereq check of `get_prune_filters ()`.

jameshcorbett (commented):
Generally looking good! Some comments --

resource/modules/resource_match.cpp
jacobtkeio: I thought it made the combined logic of `init_resource_graph` and `select_subsystems` much clearer since `subsystems_str` is guaranteed to exist in `select_subsystems` and the `std::optional` handling only needs to happen once in `init_resource_graph`.
I'll swap it back for now since I agree it's probably uncalled for in this PR.

resource/modules/resource.cpp
jacobtkeio: It looks like `ctx->traverser` previously would have been constructed with a nullptr `traverser` member variable (without failing, although it seems like it could lead to UB), whereas now `ctx->traverser` itself will get left as a nullptr, leading directly to UB when trying to call `ctx->traverser_initialize`.
Previously, `ctx->matcher` would have been set to nullptr but the if statement would immediately report the error and exit, whereas now `ctx->matcher` would get left as nullptr, leading directly to UB when trying to call `ctx->matcher->set_pruning_types_w_spec`.
Thanks for catching these! Since the resource module needs both a traverser and matcher to function, I'll make sure that when either is unset the module exits with a failure.

resource/modules/resource.cpp
jacobtkeio: I was also wondering, should we use some default value when `update_interval` isn't set? It looks like one previous 'unset' value was `0`, which would cause the update to happen every time instead of never.

resource/modules/resource_match_opts.cpp
jacobtkeio: I love this change! Thanks for catching the `transform` issue :)

resource/modules/resource.cpp
jacobtkeio: This applies to feasibility's `init_resource_graph` too

jacobtkeio: What do you think about this new version? It's more draconian because it requires all the options to be set (or to have a default value), but maybe that's for the best?

resource/modules/resource_match_opts.hpp
```diff
@@ -94,25 +41,23 @@ class resource_opts_t : public optmgr_parse_t {
...
-    const std::string &get_traverser_policy () const;
+    const std::optional<std::string> &get_load_file () const;
+    const std::optional<std::string> &get_load_format () const;
+    const std::optional<std::string> &get_load_allowlist () const;
+    const std::optional<std::string> &get_match_policy () const;
+    const std::optional<std::string> &get_match_format () const;
+    const std::optional<std::string> &get_match_subsystems () const;
+    const std::optional<int> get_reserve_vtx_vec () const;
+    const std::optional<std::string> &get_prune_filters () const;
+    const std::optional<int> get_update_interval () const;
```
jameshcorbett: Since the first and third here are by-value returns, putting `const` on them does nothing. Should be `const std::optional<int> &`, although just `std::optional<int>` would also be fine.

resource/modules/feasibility.cpp
```diff
@@ -260,30 +263,35 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
...
     }
-    if ((rc = select_subsystems (ctx)) != 0) {
+    if (!match_subsystems || (rc = select_subsystems (ctx)) != 0) {
         flux_log (ctx->h,
                   LOG_ERR,
                   "%s: error processing subsystems %s",
                   __FUNCTION__,
-                  ctx->opts.get_opt ().get_match_subsystems ().c_str ());
+                  match_subsystems ? match_subsystems->c_str () : "(unset)");
         return rc;
```
jameshcorbett: If `match_subsystems` is unset this will log an error but then return `0`. There should be an earlier check for `if (!match_subsystems) return -1;`. 

Alternatively, this whole if-block should return -1 instead of `rc`, but if `select_subsystems` has some special return code that's meant to be propagated up, that would be a change in behavior.

resource/modules/feasibility.cpp
```diff
@@ -260,30 +263,35 @@ static int init_resource_graph (std::shared_ptr<resource_ctx_t> &ctx)
...
+                  __FUNCTION__,
+                  match_format ? match_format->c_str () : "(unset)");
         return -1;
+    }
 
-    if (ctx->opts.get_opt ().is_prune_filters_set ()
-        && ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (),
-                                                   ctx->opts.get_opt ().get_prune_filters ())
+    if (!prune_filters
+        || ctx->matcher->set_pruning_types_w_spec (ctx->matcher->dom_subsystem (), *prune_filters)
```
jameshcorbett: The new `if` condition isn't the same as the old one, if the pruning filters are unset this now errors out. How come this change?

resource/modules/resource_match.cpp
```diff
@@ -159,19 +159,22 @@ static int populate_resource_db_file (std::shared_ptr<resource_ctx_t> &ctx)
...
     int saved_errno;
+    std::optional<std::string> in_file_path;
     std::ifstream in_file;
     std::stringstream buffer{};
     graph_duration_t duration;
 
-    if (ctx->reader == nullptr
-        && create_reader (ctx, ctx->opts.get_opt ().get_load_format ()) < 0) {
+    if (auto load_format = ctx->opts.get_opt ().get_load_format ();
+        load_format && ctx->reader == nullptr && create_reader (ctx, *load_format) < 0) {
```
jameshcorbett: I have never seen a semicolon inside of an `if` before, I think maybe you want `if ((auto load_format = ctx->opts.get_opt ().get_load_format ()) && ctx->reader == nullptr && create_reader (ctx, *load_format) < 0)`?

resource/modules/resource.cpp
jameshcorbett: It looks like `init_resource_graph` here is only called after `getctx ()` which in turn calls `set_default_args` so  I don't think these checks should ever succeed (defaults would have been set) but I think it doesn't hurt to keep them.

jameshcorbett (dismissed):
I think this looks good, some minor comments below, and @trws or @milroy should give final review.

resource/modules/resource_match.cpp
trws: This is a new language feature as of C++20 I think, maybe 17, it's a variable declaration scoped to whichever branch is taken. Basically it's like the first portion of a `for` statement, and this is using it correctly

resource/modules/resource_match_opts.hpp
jacobtkeio: I am leaning towards `std::optional<int>`!

trws: Unless the caller is supposed to be able to update it, definitely not a reference return.

jameshcorbett: My suggestion was for a const reference return though?

trws:
Yup, you're right, and that case it wouldn't hurt anything.  In a sane world, the value return should be just as good in this case, but it may technically be slightly slower until we get on a standard with the concept of relocatable types enabled.

resource/modules/resource_match_opts.hpp
trws: Yup, you're right, that would be fine.  Dangers of reading in email on mobile. In principle it shouldn't matter, but because relocatable objects aren't a thing in the C++ version we're using the const reference could in principle be more performant.

jacobtkeio:
I've been messing with the idea of required vs. optional module options: e.g. `match_format` is always required, while `load_file` is optional. However, the nature of `optmgr_composer_t` makes this difficult. I couldn't come up with a simple way to guarantee that some module options will always be set, which is the necessary behavior for "required" options.
The best I came up with was adding a constructor that only accepts "required" options, and failing down the line if any of them is unset. This is not nearly as elegant as making some of the `get_` functions return non-optional values, but it adds a guardrail without needing to change a lot of the option-handling code.
I think some of my struggle with option handling in `init_resource_graph` has come from not understanding which module options are required and which are optional, so hopefully making the distinction explicit will help me get it right.

resource/modules/resource_match_opts.hpp
jacobtkeio: Ah very well then, I'll go with const&!

resource/modules/feasibility.cpp
jacobtkeio: I've added sections at the start of `init_resource_graph` in resource.cpp and feasibility.cpp that check all "required" arguments and returns -1 if any is unset! My list of "required" arguments is:
- match-policy
- subsystems (match_subsystems)
- match-format
- load-format

resource/modules/feasibility.cpp
jacobtkeio: Sorry about that! I took a draconian approach instead of thinking about which options are allowed to be unset. I've replaced this with the original `if (prune_filters && ctx->matcher->set_pruning_types_w_spec...)`

resource/modules/feasibility.cpp
jacobtkeio: I forgot:
- traverser (traverser_policy)

The error messages could look like:
```
  Aug 03 23:03:41.218156 UTC 2026 sched-fluxion-resource.err[0]: init_resource_graph: required module option 'traverser' unset
  Aug 03 23:03:41.218159 UTC 2026 sched-fluxion-resource.err[0]: init_resource_graph: required module option 'subsystems' unset
  Aug 03 23:03:41.218161 UTC 2026 sched-fluxion-resource.err[0]: init_resource_graph: required module option 'match-format' unset
  Aug 03 23:03:41.218164 UTC 2026 sched-fluxion-resource.err[0]: mod_main: can't initialize resource graph database
  Aug 03 23:03:41.218189 UTC 2026 sched-fluxion-resource.crit[0]: module exiting abnormally
```

resource/modules/resource.cpp
```diff
@@ -1564,44 +1566,65 @@ static void set_status_request_cb (flux_t *h,
...
+    std::optional<std::string> match_policy = ctx->opts.get_opt ().get_match_policy ();
+    std::optional<std::string> match_subsystems = ctx->opts.get_opt ().get_match_subsystems ();
+    std::optional<std::string> match_format = ctx->opts.get_opt ().get_match_format ();
+    std::optional<std::string> prune_filters = ctx->opts.get_opt ().get_prune_filters ();
 
     // Select the appropriate traverser based on CLI policy.
-    ctx->traverser =
-        std::make_shared<dfu_traverser_t> (ctx->opts.get_opt ().get_traverser_policy ());
+    if (!traverser_policy
+        || !(ctx->traverser = std::make_shared<dfu_traverser_t> (*traverser_policy))) {
```
jameshcorbett: As you noted in another PR, `make_shared` throws instead of returning `nullptr`, so this has no effect.

jameshcorbett (commented):
@jacobtkeio based on your comments I'm guessing you haven't pushed the most recent version of this PR?

resource/modules/resource.cpp
jacobtkeio: Indeed I have not, but I will in a few minutes! I'll throw a try/catch block around the traverser init. Thanks for catching that!

wihobbs:
I just want to express how happy this makes me:

Screenshot 2026-08-06 at 9 59 28 AM

jacobtkeio:
For posterity's sake, I should mention that I tried to make the option handling more elegant in a few ways, but none of them panned out.
The main issue at hand is that some module options are "required" and should always be set, while some are "optional" and may not be set. Ideally, required module options would not be stored in a `std::optional` at all, and they'd be initialized with a default value. This would cause them to always have a value (regardless of whether or not they had been overwritten), and it would simplify the option-handling code in `init_resource_graph` and elsewhere.
Unfortunately, the `resource_opts_t`-merging design of the option-handling code seems to prevent structural disambiguation of "required" vs. "optional" options:
- `resource_opts_t`'s member variables cannot have in-class initializers because the intermediate `resource_opts_t` objects used for merging in higher-priority options in the pipeline (e.g. in `process_args`) will override previously-merged options from earlier in the pipeline (e.g. from `process_config_file`) with those class defaults. This causes, for example, options from a load-file that have in-class initializers to always be discarded.
- `resource_opts_t` can have a constructor that initializes all "required" module options. However, these member variables must remain `std::optional`s to support the merging behavior since the temporary `resource_opts_t` objects cannot require those variables to be set. Therefore, the option handling logic stays the same.
- It is possible for`resource_opts_t`'s members to remain `std::optionals` but have its `get_` functions return raw types for "required" options. However, the `get_` functions thereafter can throw exceptions when the `std::optionals` are unset. While they may never be unset in practice, this solutions provides structural disambiguation between "required" and "optional" module options at the cost of guaranteed safety.

I opted to leave the current `set_default_args`, every-member-is-optional structure alone and just add runtime checks for unset "required" module options. This is nice because it explicitly divides the module options between "required" and "optional." I hope this approach makes sense in the code and provides at least some more structure than there was before. If any of these alternatives sound better or if you can think of a more elegant way to force some options to be set, I will be more than happy to implement it!

resource/modules/resource_match.cpp
jameshcorbett, line 194: `ctx->reader` used here

resource/modules/resource_match.cpp
jameshcorbett: If `load_format` is unspecified, `ctx->reader` could remain as `nullptr` which is a problem because it is used below. Maybe require `load_format` to be specified?

jameshcorbett (commented):
I think this looks good, one thing, and still need @milroy's review as well

resource/modules/resource_match.cpp
jacobtkeio: `load_format` is required to be specified in the resource module (resource.cpp:1589) and has a default value (resource.cpp:129), but it is not required to be specified in the feasibility module. I think feasibility needs `load_format` both if it's loading from a file and if it's loading from from notify, so I'll require it! Thanks for the catch!
I was able to reproduce a SEGFAULT after removing feasibility's default `load_format` value and setting `load_file` but not `load_format`. The more common error is the disappointingly nondescriptive: `sched-fluxion-feasibility.err[0]: populate_resource_db_file: reader:` (no reader err_msg) due to the default `load_format` not matching the format of the `load_file`. This kind of error will still be able to occur after requiring `load_format` to be set. Would you support adding a hint of some kind, maybe like `"Does the format of your resource load-file match the specified load-format? (default: rv1exec)"`

jameshcorbett: Yeah I think I'd support that!

jacobtkeio: It currently looks like:
```
Aug 10 13:42:55.432307 PDT 2026 sched-fluxion-resource.err[0]: populate_resource_db_file: reader: 
Aug 10 13:42:55.432316 PDT 2026 sched-fluxion-resource.err[0]: Does the load-file (t/data/resource/jgfs/tiny.json) match the specified load-format (rv1exec)?
Aug 10 13:42:55.432320 PDT 2026 sched-fluxion-resource.err[0]: init_resource_graph: can't populate graph resource database
Aug 10 13:42:55.432322 PDT 2026 sched-fluxion-resource.err[0]: mod_main: can't initialize resource graph database
Aug 10 13:42:55.432343 PDT 2026 sched-fluxion-resource.crit[0]: module exiting abnormally
```
Do you think this should be LOG_WARNING or LOG_INFO instead? Theoretically I'd want it to be LOG_INFO, but I think it makes sense to have it visible after this particular `reader:` error message

trws: Info makes sense, but part of this information belongs in the error.  The reader part of the error should probably be "invalid format" or "the file is not valid for selected format: {load_format}" or something like that.

jacobtkeio: Yeah that makes more sense, and it looks much better.
```
[flux-sched]$ flux module load sched-fluxion-resource load-file=t/data/resource/jgfs/tiny.json
Aug 11 15:52:32.820020 PDT 2026 sched-fluxion-resource.err[0]: populate_resource_db_file: reader: invalid file for selected format (rv1exec): 
Aug 11 15:52:32.820032 PDT 2026 sched-fluxion-resource.err[0]: init_resource_graph: can't populate graph resource database
Aug 11 15:52:32.820034 PDT 2026 sched-fluxion-resource.err[0]: mod_main: can't initialize resource graph database
Aug 11 15:52:32.820052 PDT 2026 sched-fluxion-resource.crit[0]: module exiting abnormally
```
The trailing semicolon is reserving a spot for `ctx->reader->err_message ().c_str ()`, which just happens to be empty in this case.

resource/modules/resource_match.hpp
```diff
@@ -174,6 +174,18 @@ int run_remove_subgraph (std::shared_ptr<resource_ctx_t> &ctx, const std::string
 // Resource Graph and Traverser Initialization
 ////////////////////////////////////////////////////////////////////////////////
 
+template<typename T>
+inline int required_module_option (std::shared_ptr<resource_ctx_t> &ctx,
+                                   std::optional<T> opt,
```
jameshcorbett:
```suggestion
                                   const std::optional<T> &opt,
```

jameshcorbett (dismissed):
I'm good with this! Just need @milroy or @trws to approve

jameshcorbett (approved):

zekemorton (approved):
Generally looks really good to me. Looks like the other reviewers brought up some good points and everything seems to be addressed.

PR #1518, opened 2026-07-07
python: add reapi_cli bindings

garlick:
This adds a python class wrapper for reapi_cli. The intent is to use this to implement a ResourcePool class that would let the flux-core python schedulers use fluxion's resource graph, as discussed in

Marking as a WIP while I try to build the pool class and shake out any missing details, but any feedback would be welcome.

I tried to follow the same patterns used for cffi in flux-core, although I'm admittedly in unfamiliar territory and may have done something dumb! Goodness knows claude did many dumb things while this was iterated, and with only me to call out cmake/python/cffi/fluxion issues, heh, maybe you should be afraid.

The original PR could not be reopened because it was force pushed (see isaacs/github#361)
- #1481


garlick:
I've pushed an update after rebasing the python pool class PR (not yet submitted) and getting a everythinging working again.  I'll remove the WIP here as I think this could be rewiewed now.

jameshcorbett:
Do you have thoughts on how we should review this vis-a-vis #1420 ?

garlick:
I guess the main decision point is cffi vs cython!  That was raised in the original PR for this work
- #1481

I honestly don't have a strong opinion either way.

PR #1519, opened 2026-07-07
docs: add python autodoc

garlick:
Problem: python docs are not auto-generated for readthedocs.

There is not much here presently, but the upcoming reapi and pool classes will hopefully have good inline docs that might be worth rendering.

Roughly follow the pattern used in flux-core.

Also: fix a couple of minor docs build issues, and add __init__.py where missing, which simplifies doc generation and probably would be proposed later anyway for consistency.


doc/conf.py
```diff
@@ -47,15 +47,32 @@
...
+
+    BREATHE_IMPORTABLE = True
 except ImportError:
-    print("WARNING: breathe not found. C++ API documentation will not be available.")
-    print("Install breathe with: pip install breathe")
-    BREATHE_AVAILABLE = False
+    BREATHE_IMPORTABLE = False
+
+BREATHE_AVAILABLE = BREATHE_IMPORTABLE and DOXYGEN_AVAILABLE
+if BREATHE_AVAILABLE:
```
jameshcorbett: This global constant is only used in one place -- right here. 
```suggestion
if BREATHE_IMPORTABLE and DOXYGEN_AVAILABLE:
```

doc/conf.py
```diff
@@ -47,15 +47,32 @@
...
+    extensions.append("breathe")
+else:
+    if not BREATHE_IMPORTABLE:
+        print("WARNING: breathe not found. C/C++ API documentation will not be available.")
+        print("Install breathe with: pip install breathe")
+    if not DOXYGEN_AVAILABLE:
+        print("WARNING: doxygen not found. C/C++ API documentation will not be available.")
```
jameshcorbett: Is `print`ing these more appropriate than using https://docs.python.org/3/library/warnings.html ?

jameshcorbett (approved):
Great thank you!

doc/conf.py
garlick: BREATHE_AVAILABLE is used elsewhere, but BREATHE_IMPORTABLE isn't really necessary so I'll drop it.

doc/conf.py
garlick: will fix (and the existing ones)

garlick:
Thanks - fixed those things so I'll set MWP!

doc/conf.py
```diff
@@ -50,15 +50,30 @@
```
wihobbs, line 50: Something in the commit message I wanted to ask about, just for my own education:

> Since the man-page build processes these same pages, this can affect
packaging builds that install man pages, not just the HTML docs.

I thought the manpages were persistent rst files handled by the `doc/CMakeLists.txt` build, and not related to this configuration python file. Did you run into an issue here?

doc/python/index.rst
```diff
@@ -0,0 +1,23 @@
+.. _python-api:
+
+.. include:: /common/stability.rst
+
+Python API
```
wihobbs, line 5: On the web now, we have an "Internal API Reference" h2 header, followed by a "Python API" h2 header at the same level:

Image

Seems like either we should put the Python API under the Internal API or rename the Internal API to C/C++ API. Thoughts?

garlick: The python API can actually be used externally (and is, by flux-coral2).  The C/C++ API is internal only since fluxion doesn't publish the headers (only the DSO object).  So I think that's right for now?

I don't have strong feelings about this however!

doc/conf.py
garlick: Yes, I can't remember how much re-confirmation I went through now but IIRC, the doc target fails if breathe is installed (because that gates building the C++ docs) but doxygen isn't.

PR #1522, opened 2026-07-08
reapi: fix match_satisfy reporting satisfiable on error

### LINKED ISSUE #1521: reapi: returns satisfied if a type is missing entirely
I'm making a setup where I do a bunch of MatchSatisfy in parallel across different clusters and subsystems. The issue is specific to the reapi (and not reproduced in resource-query). Specifically, if I ask for a type that is 1. not in the graph, period and 2. not covered by a prune filter, it will report satisfiable=true. It _does_ return an error (-1) but typically we are checking the satisfied variable, which is true. Satisfied should be false, and the error still present so the cluster (totally missing the resource type) should not be considered a match. To test, I made  a cluster with lammps that has kokkos, and no mention of an MPI vertex. I then ask for:

```yaml
resources:
 - type: slot
   count: 1 
   label: d
   with:
   - type: lammps
     count: 1
     with: 
      - type: mpi
        count: 1

tasks: [{command: [app], slot: d, count: {per_slot: 1}}]
```

It might not be the correct structure for lammps, but we can pretend this is "I want to search the software subsystem for lammps compiled with MPI." In my test I am instantiating the reapi like this to call a satisfy (the variable we don't return is overhead):

```go
// satisfy runs MatchSatisfy and returns both the boolean and the error.
func satisfy(t *testing.T, graph, jobspec, opts string) (bool, error) {
	t.Helper()
	c := fluxcli.NewReapiClient()
	if err := c.InitContext(graph, opts); err != nil {
		t.Fatalf("InitContext failed: %v | errmsg=%q", err, c.GetErrMsg())
	}
	sat, _, err := c.MatchSatisfy(jobspec)
	return sat, err
}

```

Here is my software subsystem:

```json
{"graph":{"nodes":[
 {"id":"0","metadata":{"type":"cluster","basename":"c1","name":"c1","id":0,"uniq_id":0,"rank":-1,"size":1,"unit":"","exclusive":false,"paths":{"containment":"/c1"}}},
 {"id":"1","metadata":{"type":"lammps","basename":"lammps","name":"lammps0","id":0,"uniq_id":1,"rank":-1,"size":1,"unit":"","exclusive":false,"paths":{"containment":"/c1/lammps0"}}},
 {"id":"2","metadata":{"type":"kokkos","basename":"kokkos","name":"kokkos0","id":0,"uniq_id":2,"rank":-1,"size":1,"unit":"","exclusive":false,"paths":{"containment":"/c1/lammps0/kokkos0"}}}
],"edges":[
 {"source":"0","target":"1","metadata":{"subsystem":"containment","name":"contains"}},
 {"source":"1","target":"2","metadata":{"subsystem":"containment","name":"contains"}}
]}}
````
Arguably the top does not need to be cluster, but again, just a test.  My jobspec is shown above, and then the error is triggered:

```console
=== RUN   TestWorkaround_PruneFiltersRejectedByReapi
    repro_test.go:106: as expected, InitContext rejected prune_filters=ALL:...,ALL:mpi: issue initializing resource api client -1 | errmsg="match_allocate: ERROR: Jobspec error for 1: Value of \"type\" must be a resource type known to fluxion\nreapi_cli_initialize: Runtime error: This interner is finalized and must be open to add strings, found new string: 'mpi': Cannot allocate memory\n"
--- PASS: TestWorkaround_PruneFiltersRejectedByReapi (0.00s)
FAIL
FAIL    github.com/converged-computing/fluxion-absent-type-repro        0.012s
FAIL
```

It doesn't happen in flux outside of reapi because of [this check](https://github.com/flux-framework/flux-sched/blob/a0a6817e3a62907f45e7c2312b3c58295a6c9749/resource/libjobspec/jobspec.cpp#L144-L151) I think? But in reapi, it only sets sat to false given a [specific kind of error](https://github.com/flux-framework/flux-sched/blob/a0a6817e3a62907f45e7c2312b3c58295a6c9749/resource/reapi/bindings/c/reapi_cli.cpp#L247-L249) and kind of conflates the two. Is there a case something can even be satisfiable with an error? I think to fix, just determine that section on the retval instead of the errno?

```diff
-    // check for satisfiability
-    if (errno == ENODEV)
+    // Not satisfiable if unsatisfiable (ENODEV) OR any other failure (e.g. a
+    // jobspec referencing an unknown resource type). A nonzero return must
+    // never be reported as satisfiable.
+    if (ret != 0)
         *sat = false;
```

I don't think this would be an issue for most of Fluxion in that the reapi is not commonly used. And when it is used, it's a very limited set of types. But given a project that is created graphs with many different types, differing between clusters (and possibly even unknown) this is almost bound to happen (and this is what I'm working on now, and it did, and with a very simple case). The result would be a false positive that a cluster can satisfy something when it does not even have the defined vertex. For the time being, I will check the satisfy and the error, and require a satisfy with no error.

vsoch:
This will fix https://github.com/flux-framework/flux-sched/issues/1521

Problem: reapi_cli_match_satisfy defaulted sat to true and only cleared it when errno == ENODEV. A jobspec that references a resource type absent from the graph fails in reapi_cli_match with a nonzero return and an errno other than ENODEV, so it was wrongly reported as satisfiable while the function also returned -1. If a user just checks sat, they are going to get back that a cluster can satisfy a type it does not even have.

Solution: I think the retval is closer to the source of truth. If an error happens, I do not see any case where this could be satisfied. So the change here reports sat=true only when the match is successful. I also added a C test.

### DIFF
--- a/resource/reapi/bindings/c/test/reapi_cli_test.c
+++ b/resource/reapi/bindings/c/test/reapi_cli_test.c
@@ -63,6 +63,14 @@ static const char *simple_jobspec =
     "\"attributes\": {\"system\": {\"duration\": 60.0}}"
     "}";
 
+static const char *jobspec_absent_type =
+    "{\"version\": 1,"
+    "\"resources\": [{\"type\": \"node\", \"count\": 1, \"with\": [{"
+    "\"type\": \"slot\", \"count\": 1, \"label\": \"task\","
+    "\"with\": [{\"type\": \"core\", \"count\": 1}, {\"type\": \"mpi\", \"count\": 1}]}]}],"
+    "\"tasks\": [{\"command\": [\"app\"], \"slot\": \"task\", \"count\": {\"per_slot\": 1}}],"
+    "\"attributes\": {\"system\": {\"duration\": 60.0}}}";
+
 static int test_clone ()
 {
     reapi_cli_ctx_t *ctx = reapi_cli_new ();
@@ -483,6 +491,37 @@ static int test_find ()
     return 0;
 }
 
+static int test_match_satisfy_absent_type ()
+{
+    reapi_cli_ctx_t *ctx = reapi_cli_new ();
+    if (!ctx)
+        BAIL_OUT ("reapi_cli_new failed");
+
+    int rc = reapi_cli_initialize (ctx, tiny_jgf, tiny_params);
+    if (rc < 0)
+        BAIL_OUT ("reapi_cli_initialize failed: %s", reapi_cli_get_err_msg (ctx));
+
+    bool sat = false;
+    double ov = 0.0;
+
+    // Sanity: a satisfiable request is reported satisfiable.
+    sat = false;
+    rc = reapi_cli_match_satisfy (ctx, simple_jobspec, &sat, &ov);
+    ok (rc == 0 && sat, "reapi_cli_match_satisfy: satisfiable request -> satisfiable");
+
+    // A request for a resource type absent from the graph must NOT be reported
+    // satisfiable.  reapi_cli_match_satisfy used to default *sat=true and clear
+    // it only for ENODEV, so a jobspec referencing an unknown type (which fails
+    // with errno != ENODEV) was wrongly reported satisfiable.
+    sat = true;
+    errno = 0;
+    rc = reapi_cli_match_satisfy (ctx, jobspec_absent_type, &sat, &ov);
+    ok (rc != 0 && !sat, "reapi_cli_match_satisfy: absent resource type -> unsatisfiable");
+
+    reapi_cli_destroy (ctx);
+    return 0;
+}
+
 int main (int argc, char *argv[])
 {
     plan (NO_PLAN);
@@ -494,6 +533,7 @@ int main (int argc, char *argv[])
     test_null_parameters ();
     test_cancel_ex ();
     test_find ();
+    test_match_satisfy_absent_type ();
 
     done_testing ();
     return EXIT_SUCCESS;
--- a/resource/reapi/bindings/c/reapi_cli.cpp
+++ b/resource/reapi/bindings/c/reapi_cli.cpp
@@ -244,8 +244,10 @@ extern "C" int reapi_cli_match_satisfy (reapi_cli_ctx_t *ctx,
 
     ret = reapi_cli_match (ctx, match_op, jobspec, &jobid, &reserved, &R, &at, ov);
 
-    // check for satisfiability
-    if (errno == ENODEV)
+    // Not satisfiable if the match reported unsatisfiable (ENODEV) or failed
+    // for any other reason (e.g. a jobspec referencing an unknown resource
+    // type).  A nonzero return must never be reported as satisfiable.
+    if (ret != 0)
         *sat = false;
 
     return ret;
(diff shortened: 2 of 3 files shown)


vsoch:
If there are bugs I can check tomorrow - I'm about to fall over because I slept all of 3 hours last night. G'nite!

resource/reapi/bindings/c/test/reapi_cli_test.c
```diff
@@ -483,6 +491,37 @@ static int test_find ()
...
+
+    // A request for a resource type absent from the graph must NOT be reported
+    // satisfiable.  reapi_cli_match_satisfy used to default *sat=true and clear
+    // it only for ENODEV, so a jobspec referencing an unknown type (which fails
+    // with errno != ENODEV) was wrongly reported satisfiable.
+    sat = true;
+    errno = 0;
+    rc = reapi_cli_match_satisfy (ctx, jobspec_absent_type, &sat, &ov);
+    ok (!sat, "reapi_cli_match_satisfy: absent resource type -> unsatisfiable");
+
```
garlick, line 520: This should probably also assert that the return code is nonzero.
```
ok (rc != 0 && !sat, "absent resource type -> error, unsatisfiable");
```

garlick (approved):
Good find, and one I should have caught in my "errno contract" sweep a month ago!

Marking approved since this is a correct fix IMHO, with an optional test improvement suggested.

Taking a step back though, what is the point of this `set` boolean?  A caller that wants to distinguish, say incorrect jobspec, from unsatisfiable still has to:
```c
rc = reapi_cli_match_satisfy(ctx, jobspec, &sat, &ov);
if (rc == 0)
    /* satisfiable */   ;
else if (errno == ENODEV)
    /* unsatisfiable */  ;
else
    /* error */ ;
```
I wonder if we could just nix it?

And if we're not nixing it we should explain in the header that while `set=true` means satisfiable, `set=false` doesn't necessarily mean "unsatisfiable".

vsoch:
> I wonder if we could just nix it?

You mean "sat" right (not set?). That was my intuition too, and I would guess that it is like that for some case where there is an error but the job could be satisfied. I'm not sure what conditions those would be. With error codes, I would assume that nonzero always equals not satisifed, and the point of the error code is just to get details why. 

> set=true means satisfiable, set=false doesn't necessarily mean "unsatisfiable".

Again assuming sat. I think if set=true, we are sure it is satisfiable without error. If sat is false, it is either unsatisfiable OR unknown.

Changes:

- Added notes to header about three cases for sat/error.
- checking return code is also 0 in test.

garlick:
Sorry,`sat` not `set` 🤦 

>I think if set=true, we are sure it is satisfiable without error. If sat is false, it is either unsatisfiable OR unknown.

Exactly, and if that's all the knowledge that a caller wants, they can just check the function return value and ignore errno.  I vote we put that parameter in the stockade for public humiliation at today's sched meeting.

jameshcorbett (approved):
Changes LGTM, and yes I am also in favor of getting rid of `sat`. It looks like this would only require modifying one caller, line 184 of `reapi_cli.go`.

trws (approved):

vsoch:
Do I check the box now that I have three reviews, for the merge queue?

Would you like a follow up PR that just returns the error value (and removes sat)? We will need to update the interface also at https://github.com/flux-framework/fluxion-go and I'll do them at the same time.

jameshcorbett:
I believe Dan is going to review and will set MWP if you're ok with that! And yes a follow up PR for those changes would be most welcome :)

milroy (approved):
This change is good; let's merge it. I'll add an explicit approval (why not four?).

I think we should drop sat, too, but I'm going to advocate for a tri-value return to minimize the likelihood of misinterpretation. @vsoch go ahead and make a follow-up PR and we can debate the merits of the different return approaches.

PR #1524, opened 2026-07-09
dfu: prevent prune errno from overtaking traversal

### LINKED ISSUE #1508: discrepancy between local behavior and Tuolumne behavior regarding feasibility checks with top-level slot resources in jobspec
In the process of investigating flux-framework/flux-sched#1490 , I decided to test submitting a job with a top-level slot resource on Tuolumne and got the behavior in flux-framework/flux-fiction#40. In summary, I noticed that submitting a jobspec containing a top-level slot resource using `flux submit --resources-json=<jobspec> hostname` would return `[19] Unsatisfiable request`. 

I decided to look into this issue further because this is not behavior I get in Flux Fiction when submitting identical jobspecs to Flux. A couple things I realized/tested: 

* I wasn't enabling the feasibility plugin in Flux Fiction. This was easily fixed by toggling it on in the Flux configuration. It also did not change my behavior.
* The real system had 3 frobnicator plugins I wasn't using: `defaults`, `constraints`, and `rabbit`. I understand that `rabbit` comes from flux-framework/flux-coral2 and interprets DW requests, so I figured that was not affecting me. I turned on `defaults` and `constraints` and the only notable changes to my jobspec was that they added the `pbatch` queue requirement to my jobspec and a default timelimit. The only way I was able to get `[19] Unsatisfiable request` from this configuration is by removing pbatch from my system configuration altogether. 
* I tried submitting a normal jobspec to Tuolumne with a node as the top-level resource in the jobspec with the above command and it worked just fine

This makes me think that either there is some sort of hidden modification to the jobspec that is sensitive to top-level slot resources on Tuolumne not shown in `--dry-run` and breaks with top-level slot resources or an explicit filter blocking jobs with top-level slot resources. I would like to know so I can make Flux Fiction have the same behavior as production, or if this is an actual bug, I'm sure that you would like to know about it. Also, I can get the output of `--dry-run` if that would be helpful, but it is very long and cluttered with environment variable attributes.

trws:
problem: errno set as part of pruning can become the overall errno of a dfu traversal if resources are considered after the last resources selected.  This can cause inappropriate error codes to be set which can result in unsatisfiable being returned when the result should be failure to allocate but satisfiable.

solution: save and restore errno in the prune function while we work out how to appropriately aggregate error values across the traversal as a whole.

fixes #1508 

CC @washwor1

### DIFF
--- a/t/t1020-qmanager-feasibility.t
+++ b/t/t1020-qmanager-feasibility.t
@@ -28,47 +28,44 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
         | jq -e ".errnum != 0"
 '
 
+run_sat_unsat() {
+    local label="$1"
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 170 cores" '
+       test_must_fail flux submit -n 170 hostname 2>"err_${label}_cores" &&
+       grep "Unsatisfiable request" err_${label}_cores
+    '
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 2 nodes" '
+       test_must_fail flux submit -N 2 -n2 hostname 2>"err_${label}_nodes" &&
+       grep "Unsatisfiable request" err_${label}_nodes
+    '
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 1 gpu" '
+       test_must_fail flux submit -g 1 hostname 2>"err_${label}_gpu" &&
+       grep "Unsatisfiable request" err_${label}_gpu
+    '
+
+    test_expect_success "feasibility: satisfiable jobs are accepted ($label)" '
+        jobid=$(flux submit -n 8 hostname) &&
+        flux job wait-event -t 10 ${jobid} start &&
+        flux job wait-event -t 10 ${jobid} finish &&
+        flux job wait-event -t 10 ${jobid} release &&
+        flux job wait-event -t 10 ${jobid} clean
+    '
+}
+
 test_expect_success 'feasibility: loading job-ingest with feasibilty works' '
     flux module reload job-ingest validator-plugins=feasibility
  '
 
- test_expect_success 'feasibility: unsatisfiable jobs are rejected' '
-    test_must_fail flux submit -n 170 hostname 2>err1 &&
-    grep "Unsatisfiable request" err1 &&
-    test_must_fail flux submit -N 2 -n2 hostname 2>err2 &&
-    grep "Unsatisfiable request" err2 &&
-    test_must_fail flux submit -g 1 hostname 2>err3 &&
-    grep -i "Unsatisfiable request" err3
- '
-
-test_expect_success 'feasibility: satisfiable jobs are accepted' '
-    jobid=$(flux submit -n 8 hostname) &&
-    flux job wait-event -t 10 ${jobid} start &&
-    flux job wait-event -t 10 ${jobid} finish &&
-    flux job wait-event -t 10 ${jobid} release &&
-    flux job wait-event -t 10 ${jobid} clean
-'
+run_sat_unsat feasibility
 
 test_expect_success 'feasibility: load job-ingest with two validators' '
     flux module reload job-ingest validator-plugins=feasibility,jobspec
  '
 
- test_expect_success 'feasibility: unsatisfiable jobs are rejected' '
-    test_must_fail flux submit -n 170 hostname 2>err4 &&
-    grep "Unsatisfiable request" err4 &&
-    test_must_fail flux submit -N 2 -n2 hostname 2>err5 &&
-    grep "Unsatisfiable request" err5 &&
-    test_must_fail flux submit -g 1 hostname 2>err6 &&
-    grep -i "Unsatisfiable request" err6
- '
-
-test_expect_success 'feasibility: satisfiable jobs are accepted' '
-    jobid=$(flux submit -n 8 hostname) &&
-    flux job wait-event -t 10 ${jobid} start &&
-    flux job wait-event -t 10 ${jobid} finish &&
-    flux job wait-event -t 10 ${jobid} release &&
-    flux job wait-event -t 10 ${jobid} clean
-'
+run_sat_unsat feasibility_jobspec
 
 test_expect_success 'feasibility: cleanup active jobs' '
     cleanup_active_jobs
--- a/t/t3015-resource-basic-jgf.t
+++ b/t/t3015-resource-basic-jgf.t
@@ -7,6 +7,7 @@ test_description='Test Scheduling On Tiny Machine Configuration in JGF'
 cmd_dir="${SHARNESS_TEST_SRCDIR}/data/resource/commands/basics"
 exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/basics"
 jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny.json"
+storage_jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny_storage.json"
 query="../../resource/utilities/resource-query"
 
 #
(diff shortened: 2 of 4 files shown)


t/t3015-resource-basic-jgf.t
```diff
@@ -220,4 +220,79 @@ test_expect_success "${test048_desc}" '
...
+            "data": -1,
+            "stack": 8388608,
+            "core": 0,
+            "nofile": 1048576,
+            "as": -1,
+            "rss": -1,
+            "nproc": -1
+          }
+        }
+      },
```
jameshcorbett: delete?

jameshcorbett (commented):
Given that one of the existing tests is failing how do you want to proceed?

t/t3015-resource-basic-jgf.t
trws: Done.

trws:
I think the rework I just pushed takes care of it.  Turns out we need an EBUSY from prune in some cases, it could be the only thing we ever look at, so restoring isn't always right.  This version actually changes less about the behavior of the code, but as far as I can tell passes all the existing tests and the new one.  The extra commit is because I had a tough time teasing apart which piece of the test in 1020 was failing, so I did a small refactor to make it more apparent.

jameshcorbett (approved):
I think the t3015 tests should probably be in a separate commit but I'm good with this!

milroy (changes requested):
Thanks for the fix, @trws. I turned up a few issues, so I don't think it's mergeable as-is. 

The `prune()` change looks good. The main blocker is the new t`3015 test`. It doesn't look like it's actually exercising the fix. Sorry for the dramatic block, I just wanted to make sure this was discussed before merging since the PR was already approved.

t/t3015-resource-basic-jgf.t
```diff
@@ -220,4 +220,64 @@ test_expect_success "${test048_desc}" '
...
+      "cwd": "/home/user/",
+      "dw": "#DW jobdw capacity=10TiB type=lustre name=project2"
+    }
+  },
+  "version": 1
+}
+END
+    echo -e "m allocate_orelse_reserve jobspec.json\nquit" |
+        ${query} -L ${jgf} -f jgf -S CA -P first -t t49_out |
+        test_must_fail grep 'Unsatisfiable'
```
milroy: I don't think this test can fail. It sends the `resource-query` output into `t49_out`. The test then searches stdout, which is empty.

It also doesn't reach the fix. The bug only occurs when a vertex request is larger than what's physically available (ssd here) but `tiny.json` doesn't have ssds, and I think the default pruning only tracks cores. I'll double check this last point, because I may have configured the pruner to track leaves, but that's not totally relevant here since `tiny` doesn't have ssds.

trws: 100% right. I forgot that tiny had its storage removed, so I added the equivalent including the storage as a new jgf in the first commit, and changed this to go back to outputting to stdout.

t/t3015-resource-basic-jgf.t
```diff
@@ -7,6 +7,7 @@ test_description='Test Scheduling On Tiny Machine Configuration in JGF'
 cmd_dir="${SHARNESS_TEST_SRCDIR}/data/resource/commands/basics"
 exp_dir="${SHARNESS_TEST_SRCDIR}/data/resource/expected/basics"
 jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny.json"
+storage_jgf="${SHARNESS_TEST_SRCDIR}/data/resource/jgfs/tiny_storage.json" &&
```
milroy: `tiny_storage.json` is missing from the commits, and I think you want to drop `&&` here.

t/t3015-resource-basic-jgf.t
milroy: There's another, subtler issue here.  "Unsatisfiable" prints only when `errno` is `ENODEV` or `ENOENT` (command.cpp:271), and the fix swaps `ERANGE` with `EBUSY`.

I recommend creating a test based on the resource module because `resource.cpp` treats `EBUSY` gracefully, but `ERANGE` as a "true" match error.

t/t3015-resource-basic-jgf.t
trws: Oof, can't believe I missed that, thanks @milroy.  Fixed and pushed.

t/t3015-resource-basic-jgf.t
trws: Ok, this one took a bit more tweaking but not too bad.  Again the actual code logic seems to be ok, but I did a side-by-side with this until a pre-commit version failed, updated version succeeded, and we get meaningful output from the test when it fails.  So we don't have to deal with the difference in stderr output between libcs, I have it matching on `ERROR: traverser:` now which should only occur if there's an unsatisfiable error, and gives us slightly broader coverage in case something else leaks out.

trws:
Fixed the run_sat_unsat indent, most of the tests under it were also offset 1 from the normal indent too so I fixed those as well.

t/t1020-qmanager-feasibility.t
```diff
@@ -28,47 +28,42 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
...
+        test_must_fail flux submit -N 2 -n2 hostname 2>err2 &&
+        grep "Unsatisfiable request" err2
+     '
+
+     test_expect_success 'feasibility: unsatisfiable jobs are rejected' '
+        test_must_fail flux submit -g 1 hostname 2>err3 &&
+        grep -i "Unsatisfiable request" err3
+     '
+
+    test_expect_success 'feasibility: satisfiable jobs are accepted' '
```
milroy: A few spacing nits, and one change suggested by Claude:
```suggestion
run_sat_unsat() {
    local label="$1"

    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 170 cores" '
        test_must_fail flux submit -n 170 hostname 2>err_${label}_cores &&
        grep "Unsatisfiable request" err_${label}_cores
    '

    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 2 nodes" '
        test_must_fail flux submit -N 2 -n2 hostname 2>err_${label}_nodes &&
        grep "Unsatisfiable request" err_${label}_nodes
    '

    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 1 gpu" '
        test_must_fail flux submit -g 1 hostname 2>err_${label}_gpu &&
        grep "Unsatisfiable request" err_${label}_gpu
    '

    test_expect_success "feasibility: satisfiable jobs are accepted ($label)" '
```
The argument to `run_sat_unsat()` allows disambiguation of the output of `run_sat_unsat()` with feasibility and its output with feasibility and jobspec. So the later invocations would be like: `run_sat_unsat feasibility` and `run_sat_unsat feasibility,jobspec`.

Also I dropped the `grep -i` from the GPU check. Not clear if that's actually needed here.

trws: Makes sense, got the spacing a bit ago, applied the rest and pushed.

I have to go AFK for ~45, if this is good to go, feel free to set MWP, if not I'll get right on it then.

milroy: I made three more suggestions. Take your time; no rush.

t/t1020-qmanager-feasibility.t
```diff
@@ -28,47 +28,44 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
...
+       test_must_fail flux submit -n 170 hostname 2>err1 &&
+       grep "Unsatisfiable request" err1
+    '
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 2 nodes" '
+       test_must_fail flux submit -N 2 -n2 hostname 2>err2 &&
+       grep "Unsatisfiable request" err2
+    '
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label)" '
```
milroy:
```suggestion
    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 1 gpu" '
```

t/t1020-qmanager-feasibility.t
```diff
@@ -28,47 +28,44 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
...
- '
-
-test_expect_success 'feasibility: satisfiable jobs are accepted' '
-    jobid=$(flux submit -n 8 hostname) &&
-    flux job wait-event -t 10 ${jobid} start &&
-    flux job wait-event -t 10 ${jobid} finish &&
-    flux job wait-event -t 10 ${jobid} release &&
-    flux job wait-event -t 10 ${jobid} clean
-'
+run_sat_unsat
```
milroy:
```suggestion
run_sat_unsat feasibility
```

t/t1020-qmanager-feasibility.t
```diff
@@ -28,47 +28,44 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
...
```
milroy:
```suggestion
run_sat_unsat feasibility_jobspec
```

t/t1020-qmanager-feasibility.t
trws: 🤦 got it, one sec

trws: Done, all 3.

t/t1020-qmanager-feasibility.t
```diff
@@ -28,47 +28,44 @@ test_expect_success 'feasibility: --plugins=feasibility works ' '
         | jq -e ".errnum != 0"
 '
 
+run_sat_unsat() {
+    local label="$1"
+
+    test_expect_success "feasibility: unsatisfiable jobs are rejected ($label): 170 cores" '
+       test_must_fail flux submit -n 170 hostname 2>err1 &&
+       grep "Unsatisfiable request" err1
```
milroy: Minor, non-blocking consideration: making the error output invocation-specific will help in debugging since successive function calls will clobber the outputs.

milroy:
```suggestion
       grep "Unsatisfiable request" err_${label}_cores
```

milroy: (Same for the other two err files.)

milroy (approved):
Thanks for the quick iterations on this! I added an optional change you can make if you want. This is good to merge.

PR #1526, opened 2026-07-10
reapi: drop sat output param from satisfy interfaces

vsoch:
Problem: We currently return a sat boolean along with an error for our reapi satisfy interfaces. This is problematic as it introduces a matrix of possibly outcomes, where an outcome such as a satisfied job with an error becomes confusing.

Solution: return a single error code, where 0 == satisfied, 1 == not satisfied, and -1 == error (and satisfiability is unknown). This update changes the Go and C interfaces. We also update the C unit test, the Go example caller, and add a fluxcli unit test to cover all cases.

I realized that the called function `reapi_cli_match_satisfy` in the Go SDK itself was not returning the granularity we discussed - it was just a 0 or -1, and we could not determine a call that did not error but did not satisfy. I think to fix this we need to look for a -1 with the error `ENODEV` per [here](https://github.com/flux-framework/flux-sched/blob/7b7315a420ead319822028128a50d037de449764/resource/traversers/dfu.cpp#L47-L56) and [here](https://github.com/flux-framework/flux-sched/blob/7b7315a420ead319822028128a50d037de449764/resource/traversers/dfu.cpp#L84-L88).

- Follow up to https://github.com/flux-framework/flux-sched/pull/1522.
- I will keep rebase / push until testing issues pass.

We probably need to look at our go docstrings too. The [convention](https://go.dev/doc/comment#func) I see most often is like:

```go
// FuncName does this thing.
// It returns something else.
func FuncName(input []byte) (name string, size int, err error) {
	// ... function body
	return somethingElse, nil
}
```

e.g.,

```go
// ProcessData executes a specific operation on the provided input payload.
// It returns the newly generated unique identifier string, the calculated 
// total size of the output payload, and any processing error encountered.
// If the input byte slice is empty, it returns 0 and ErrInvalidPayload.
func ProcessData(input []byte) (id string, size int, err error) {
	// ... function body
	return id, size, nil
}
```

And not to have the entire repeated docstring with parameters again. I wrote satisfy for how I want to suggest it can be.

### DIFF
--- a/resource/reapi/bindings/c/reapi_cli.cpp
+++ b/resource/reapi/bindings/c/reapi_cli.cpp
@@ -240,28 +240,19 @@ extern "C" int reapi_cli_match_allocate (reapi_cli_ctx_t *ctx,
     return reapi_cli_match (ctx, match_op, jobspec, jobid, reserved, R, at, ov);
 }
 
-extern "C" int reapi_cli_match_satisfy (reapi_cli_ctx_t *ctx,
-                                        const char *jobspec,
-                                        bool *sat,
-                                        double *ov)
+extern "C" int reapi_cli_match_satisfy (reapi_cli_ctx_t *ctx, const char *jobspec, double *ov)
 {
     match_op_t match_op = match_op_t::MATCH_SATISFIABILITY;
     uint64_t jobid;
     bool reserved;
     char *R;
     int64_t at;
-    int ret;
-    *sat = true;
 
-    ret = reapi_cli_match (ctx, match_op, jobspec, &jobid, &reserved, &R, &at, ov);
-
-    // Not satisfiable if the match reported unsatisfiable (ENODEV) or failed
-    // for any other reason (e.g. a jobspec referencing an unknown resource
-    // type).  A nonzero return must never be reported as satisfiable.
-    if (ret != 0)
-        *sat = false;
-
-    return ret;
+    if (reapi_cli_match (ctx, match_op, jobspec, &jobid, &reserved, &R, &at, ov) == 0)
+        return 0;
+    // The traverser reports an unsatisfiable request as ENODEV; any other
+    // errno is a genuine error (errno is left set for the caller).
+    return (errno == ENODEV) ? 1 : -1;
 }
 
 extern "C" int reapi_cli_update_allocate (reapi_cli_ctx_t *ctx,
--- a/resource/reapi/bindings/c/test/reapi_cli_test.c
+++ b/resource/reapi/bindings/c/test/reapi_cli_test.c
@@ -71,6 +71,14 @@ static const char *jobspec_absent_type =
     "\"tasks\": [{\"command\": [\"app\"], \"slot\": \"task\", \"count\": {\"per_slot\": 1}}],"
     "\"attributes\": {\"system\": {\"duration\": 60.0}}}";
 
+static const char *jobspec_too_many_cores =
+    "{\"version\": 1,"
+    "\"resources\": [{\"type\": \"node\", \"count\": 1, \"with\": [{"
+    "\"type\": \"slot\", \"count\": 1, \"label\": \"task\","
+    "\"with\": [{\"type\": \"core\", \"count\": 2}]}]}],"
+    "\"tasks\": [{\"command\": [\"app\"], \"slot\": \"task\", \"count\": {\"per_slot\": 1}}],"
+    "\"attributes\": {\"system\": {\"duration\": 60.0}}}";
+
 static int test_clone ()
 {
     reapi_cli_ctx_t *ctx = reapi_cli_new ();
@@ -491,7 +499,7 @@ static int test_find ()
     return 0;
 }
 
-static int test_match_satisfy_absent_type ()
+static int test_match_satisfy ()
 {
     reapi_cli_ctx_t *ctx = reapi_cli_new ();
     if (!ctx)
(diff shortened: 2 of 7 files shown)


resource/reapi/bindings/c/test/reapi_cli_test.c
```diff
@@ -501,22 +512,19 @@ static int test_match_satisfy_absent_type ()
...
-    errno = 0;
-    rc = reapi_cli_match_satisfy (ctx, jobspec_absent_type, &sat, &ov);
-    ok (rc != 0 && !sat, "reapi_cli_match_satisfy: absent resource type -> unsatisfiable");
+    // Valid request the graph cannot satisfy (more cores than exist) -> 1.
+    rc = reapi_cli_match_satisfy (ctx, jobspec_too_many_cores, &ov);
+    ok (rc == 1, "reapi_cli_match_satisfy: unsatisfiable request -> 1");
+
+    // Request referencing a resource type absent from the graph -> -1 (error).
+    rc = reapi_cli_match_satisfy (ctx, jobspec_absent_type, &ov);
+    ok (rc == -1, "reapi_cli_match_satisfy: absent resource type -> -1");
```
jameshcorbett: Perhaps check `errno == ENODEV` and `errno != ENODEV` in these two different tests?

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -229,28 +229,19 @@ extern "C" int reapi_cli_match_allocate (reapi_cli_ctx_t *ctx,
...
-
-    // Not satisfiable if the match reported unsatisfiable (ENODEV) or failed
-    // for any other reason (e.g. a jobspec referencing an unknown resource
-    // type).  A nonzero return must never be reported as satisfiable.
-    if (ret != 0)
-        *sat = false;
-
-    return ret;
+    if (reapi_cli_match (ctx, match_op, jobspec, &jobid, &reserved, &R, &at, ov) == 0)
+        return 0;
```
jameshcorbett, line 252: This is an existing problem, not related to this PR, but `R` needs to be freed after calling this function, regardless of its returncode. Would you mind tagging on an additional commit that initializes R to NULL and then unconditionally frees it after calling `reapi_cli_match`? Or I can do that as a follow-up PR, your call.

jameshcorbett (approved):
I'm good with this thanks! But I think @trws and @milroy will want to look as well.

resource/reapi/bindings/c/reapi_cli.cpp
jameshcorbett: #1540 makes this change now, so I think this change could be left to that PR if this goes in first.

resource/reapi/bindings/c/test/reapi_cli_test.c
milroy: I agree!

resource/reapi/bindings/c/reapi_cli.cpp
milroy: I didn't realize my PR overlapped. I'm fine with it either way. May be best to just fix it here and I'll deal with the merge conflict whenever my PR is closer to final.

resource/reapi/bindings/c/test/reapi_cli_test.c
```diff
@@ -71,6 +71,17 @@ static const char *jobspec_absent_type =
     "\"tasks\": [{\"command\": [\"app\"], \"slot\": \"task\", \"count\": {\"per_slot\": 1}}],"
```
milroy, line 71: Adding a comment on this line that really applies to the whole file and `reapi_cli_test.go`: the changes to test files should go in a separate `testsuite: ...` commit. It's an unwritten (I think, and we should codify it) rule that implementation and testing are put in separate commits.

resource/reapi/bindings/c/reapi_cli.cpp
```diff
@@ -229,28 +229,19 @@ extern "C" int reapi_cli_match_allocate (reapi_cli_ctx_t *ctx,
...
+    // The traverser reports an unsatisfiable request as ENODEV; any other
+    // errno is a genuine error (errno is left set for the caller).
+    return (errno == ENODEV) ? 1 : -1;
```
milroy, line 255: I had Opus mull the tristate over, and it recommends an enum-typed return. I disagree with Claude in this case and favor the int tristate return, but I figured I'd pass on the recommendation for discussion.

milroy (changes requested):
Thanks for the quick fix PR @vsoch, and sorry for the harsh block on the merge. This looks quite good and there are only a couple minor issues. I want to make sure they get discussed and addressed before this goes in, though.

resource/reapi/bindings/c/test/reapi_cli_test.c
vsoch: Both fixed.

resource/reapi/bindings/c/reapi_cli.cpp
vsoch: Thanks for including me! :partying_face:

resource/reapi/bindings/c/test/reapi_cli_test.c
vsoch: I separated into two commits (with help from Claude), cloned fresh for the rebase, and reapplied. I removed this comment because it was out of place, generally.

resource/reapi/bindings/c/reapi_cli.cpp
vsoch: Claude does like to "mull" - to grind, to pulverize our code into a soft powder and constitute it into a hot cider. 

https://www.etymonline.com/word/mull

I also disagree, primarily because the convention of the code base is to return integer error codes, and this would be a one-off different case. It isn't a terrible idea, but if we consider it we'd want to make it more of a consistent standard, and have good reason for the change.

jameshcorbett (approved):
The R still isn't freed but that's ok I say let @milroy deal with it! He'll have to deal with merge conflicts anyway I think.

resource/reapi/bindings/c/reapi_cli.cpp
trws: The plan is to make that the standard FWIW, but there's a good bit of work to do to get there and I don't expect it here. The main thing is when we do int returns, we do a negative return value and errno normally.  As long as all failure states are negative, should be fine.

jameshcorbett: Does that mean you think unsatisfiable should be negative (rather than 1 as this PR proposes) or would you not consider that a "failure state"?

trws: It's arguable either way, as long as it's documented and a _true error_ is always -1 it's fine since that will match our conventions for checking errors ok. In an ideal universe I would make it return something like an `expected<void,bool>` so that a successful response is equivalent to `has_value` and you can discriminate the fail case between a true error and unsat, but as long as it's documented and neither are zero it's probably ok.

resource/reapi/bindings/c/reapi_cli.cpp
milroy: Agreeing with @jameshcorbett below- I'll deal with this issue in my PR.

resource/reapi/bindings/c/reapi_cli.cpp
milroy: Ok, let's maintain the integer tristate in the current implementation, then.

resource/reapi/bindings/c/test/reapi_cli_test.c
milroy: Thanks!

milroy (approved):
Thanks for the changes @vsoch; LGTM!

Issue #1424, opened 2026-01-07
Multiple jobs in queue show eta:now while not being the job with highest priority


adamdbertsch:
It appears that if a job has ever been at the head of the queue with eta:now, it will continue to show eta:now even if another job ends up ahead of it. The result is a very confusing display where the next job to run has an eta with some future value, and then jobs below it in priority show eta:now.

ryanday36:
We discussed this some in this week's Flux system instance meeting. Anecdotally, this appears to be happening quite often when a job completion satisfies the dependency condition for a queued job that then has higher priority than the former job at the head of the queue and "takes" the nodes that were being reserved for the `eta: now` job. I believe this is the correct behavior from a fairshare / priority standpoint, but some folks are worried about whether it's wasteful from a utilization standpoint. I.e. are nodes spending excessive amounts of time being idled because the large job that they're reserved for isn't actually able to start when fluxion expects it to. If fluxion knew more about the dependent jobs it might be able to start more jobs via backfill as it would have a more accurate eta for that highest priority job, but I'm not sure how much utilization is actually affected beyond that or if fluxion would be reserving those nodes for the big job anyway.

I've also gotten the suggestion that a possible solution to this would be to not allow new jobs to take the reserved resources of a job with an `eta:now` (or `eta` less than some configurable time). This wouldn't be fully correct from a priority standpoint, but would be less confusing for users and managers.

Finally, I'll get on switching to `conservative` backfill rather than `easy` per @trws suggestion in #1015 to see if that helps at least update the etas.

Also, tagging @milroy for extra visibility.

trws:
I'm digging into this now, and I'm pretty sure I have a handle on what's going on.  Fluxion sets the expected start time as an annotation, and sends that along to core.  Core interprets time of now or earlier as "now." Both of these are fine independently, but sched tries to avoid sending too many annotations because of some old issues with memory ballooning due to too many annotations.

It looks like the most straightforward thing would be to ensure we always update all annotations that have their expected time or position change, rather than just expected time, which may mean that a job that has a time has its time cleared. Would that be sufficient to cover this for you @adamdbertsch?

adamdbertsch:
I think that would be fine for me. Giving empty information is preferable to giving incorrect information.

ryanday36:
[...]

PR #1528, opened 2026-07-10
qmanager: annotate jobs with `sched.selection_type` and update `t_estimate` for jobs

### LINKED ISSUE #1510: annotate jobs started by backfill
we've had some interest in having a better understanding of how much usage on our Flux systems is coming from backfilled jobs and, possibly, handling their usage differently in flux accounting. Would it be possible for flux-sched to add an annotation to jobs that it starts via backfill?

wihobbs:
There is a separate PR open at flux-framework/rfc#529 if you have opinions on what the annotation should be called. I'll rename it in this branch to whatever we decide, but for this prototype, it is called `sched.start_method`.

This is a surprisingly simple amount of code that I believe fixes #1510. Though it may seem like unit tests would be more appropriate, the integration testsuite already has well-commented jobs that are marked "reserved" or "BF" (backfill), so repurposing those tests seems to be a win-win.

### DIFF
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -110,12 +110,19 @@ int qmanager_cb_t::post_sched_loop (
             if (job->schedule.at == job->schedule.old_at)
                 continue;
             job->schedule.old_at = job->schedule.at;
+            // Reserved jobs receive a t_estimate annotation while they wait
+            // for their reserved time.  Report selection_type here too so that
+            // a job that is only ever reserved (and never allocated) still has
+            // its RFC 27 sched.selection_type recorded, rather than only being
+            // reported on the eventual alloc SUCCESS response.
             if (schedutil_alloc_respond_annotate_pack (schedutil,
                                                        job->msg,
-                                                       "{ s:{s:f} }",
+                                                       "{ s:{s:f s:s} }",
                                                        "sched",
                                                        "t_estimate",
-                                                       static_cast<double> (job->schedule.at))) {
+                                                       static_cast<double> (job->schedule.at),
+                                                       "selection_type",
+                                                       job_selection_type_str (job->schedule.selection_type))) {
                 flux_log_error (h, "%s: schedutil_alloc_respond_annotate_pack", __FUNCTION__);
                 goto out;
             }
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -62,12 +62,23 @@ int qmanager_cb_t::post_sched_loop (
         const std::string &queue_name = kv.first;
         std::shared_ptr<queue_policy_base_t> &queue = kv.second;
         while ((job = queue->alloced_pop ()) != nullptr) {
+            // Record how the job was started (immediate, backfill, or
+            // reserved) as a scheduler annotation in the RFC 27 "sched"
+            // namespace.  Annotations carried on the alloc SUCCESS response
+            // are snapshotted into the job's "alloc" event and thus
+            // preserved in the job record.  t_estimate is nulled here to
+            // clear any volatile start-time estimate now that the job has
+            // been allocated.
             if (schedutil_alloc_respond_success_pack (schedutil,
                                                       job->msg,
                                                       job->schedule.R.c_str (),
-                                                      "{ s:{s:n} }",
+                                                      "{ s:{s:n s:s} }",
                                                       "sched",
-                                                      "t_estimate")
+                                                      "t_estimate",
+                                                      // NULL from format string
+                                                      "selection_type",
+                                                      job_selection_type_str (
+                                                          job->schedule.selection_type))
                 < 0) {
                 flux_log_error (h,
                                 "%s: schedutil_alloc_respond_pack (queue=%s)",
--- a/qmanager/modules/qmanager_callbacks.cpp
+++ b/qmanager/modules/qmanager_callbacks.cpp
@@ -106,25 +106,54 @@ int qmanager_cb_t::post_sched_loop (
         for (job = queue->pending_begin (), qd = 0;
              job != nullptr && qd < queue->get_queue_depth ();
              job = queue->pending_next (), qd++) {
-            // if old_at == at, then no reason to send this annotation again.
-            if (job->schedule.at == job->schedule.old_at)
-                continue;
-            job->schedule.old_at = job->schedule.at;
-            // Reserved jobs receive a t_estimate annotation while they wait
-            // for their reserved time.  Report selection_type here too so that
-            // a job that is only ever reserved (and never allocated) still has
-            // its RFC 27 sched.selection_type recorded, rather than only being
-            // reported on the eventual alloc SUCCESS response.
-            if (schedutil_alloc_respond_annotate_pack (schedutil,
-                                                       job->msg,
-                                                       "{ s:{s:f s:s} }",
-                                                       "sched",
-                                                       "t_estimate",
-                                                       static_cast<double> (job->schedule.at),
-                                                       "selection_type",
-                                                       job_selection_type_str (job->schedule.selection_type))) {
-                flux_log_error (h, "%s: schedutil_alloc_respond_annotate_pack", __FUNCTION__);
-                goto out;
+            // only send an annotation when the start-time estimate has
+            // changed since the last successfully transmitted one.
+            if (job->schedule.at != job->schedule.old_at) {
+                if (job->schedule.at > 0) {
+                    // Reserved jobs receive a t_estimate annotation while they
+                    // wait for their reserved time.  Report selection_type
+                    // here too so that a job that is only ever reserved (and
+                    // never allocated) still has its RFC 27
+                    // sched.selection_type recorded, rather than only being
+                    // reported on the eventual alloc SUCCESS response.
+                    if (schedutil_alloc_respond_annotate_pack (schedutil,
+                                                               job->msg,
+                                                               "{ s:{s:f s:s} }",
+                                                               "sched",
+                                                               "t_estimate",
+                                                               static_cast<double> (
+                                                                   job->schedule.at),
+                                                               "selection_type",
+                                                               job_selection_type_str (
+                                                                   job->schedule.selection_type))) {
+                        flux_log_error (h,
+                                        "%s: schedutil_alloc_respond_annotate_pack",
+                                        __FUNCTION__);
+                        goto out;
+                    }
+                } else {
+                    // The job lost its reservation this loop (it could be
+                    // neither allocated nor reserved), so its previously
+                    // reported start-time estimate is now stale.  A JSON null
+                    // value deletes the key in the job manager, so both
+                    // t_estimate and selection_type are cleared rather than
+                    // left displaying an incorrect eta (#1424).
+                    if (schedutil_alloc_respond_annotate_pack (schedutil,
+                                                               job->msg,
+                                                               "{ s:{s:n s:n} }",
+                                                               "sched",
+                                                               "t_estimate",
+                                                               "selection_type")) {
+                        flux_log_error (h,
+                                        "%s: schedutil_alloc_respond_annotate_pack",
+                                        __FUNCTION__);
+                        goto out;
+                    }
+                }
+                // record the transmitted value only after a successful send
+                // so that a failed annotation is retried on the next
+                // post_sched_loop rather than being silently dropped.
+                job->schedule.old_at = job->schedule.at;
             }
         }
         queue->reset_scheduled ();
@@ -515,7 +544,8 @@ void qmanager_cb_t::prep_watcher_cb (flux_reactor_t *r, flux_watcher_t *w, int r
     for (auto &kv : ctx->queues) {
         std::shared_ptr<queue_policy_base_t> &queue = kv.second;
         ctx->pls_sched_loop = ctx->pls_sched_loop || queue->is_schedulable ();
-        ctx->pls_post_loop = ctx->pls_post_loop || queue->is_scheduled ();
+        ctx->pls_post_loop =
+            ctx->pls_post_loop || queue->is_scheduled () || queue->is_annotation_dirty ();
     }
     if (ctx->pls_sched_loop || ctx->pls_post_loop)
         flux_watcher_start (ctx->idle);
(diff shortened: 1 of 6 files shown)


qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -62,12 +62,21 @@ int qmanager_cb_t::post_sched_loop (
...
+            // been allocated.
             if (schedutil_alloc_respond_success_pack (schedutil,
                                                       job->msg,
                                                       job->schedule.R.c_str (),
-                                                      "{ s:{s:n} }",
+                                                      "{ s:{s:n s:s} }",
                                                       "sched",
-                                                      "t_estimate")
+                                                      "t_estimate",
+                                                      "start_mode",
```
trws: I ended up looking at this for ~30 seconds trying to figure out why it was working, maybe a comment?
```suggestion
                                                      "t_estimate",
                                                      // NULL from format string
                                                      "start_mode",
```

wihobbs: Fixed this and the typo in the commit message.

trws:
There are only two things on my mind about this, the implementation looks good

* does it makes sense to include an annotation on reservation to set this as part of setting the t_estimate in that case?
* do we want the "reserved" status to be reported for any job that was ever reserved (current behavior) or a job that was allocated immediately after being reserved without being de-prioritized inbetween?

wihobbs:
> do we want the "reserved" status to be reported for any job that was ever reserved (current behavior) or a job that was allocated immediately after being reserved without being de-prioritized inbetween?

I was actually kind of torn on having "reserved" and "immediate" be different things in the first place, because all we really care about is "backfill" and "not backfill." I wonder if we should make this a boolean instead, just `sched.backfill_selection = [true | false]`...

trws:
That's a fair point.  The advantage to having the reserved boolean tracked here is I think I would need to do the same thing to solve #1424, because we need to know that something was reserved (and thus sent a time estimate) to clear those estimates the next time around.  I suppose we could separate them and set "reserved" along with the time estimate but not track that it was reserved at the end?

wihobbs:
I'll look at #1424, as you point out with this going into the next release it might be good to tackle that as well.

wihobbs:
Looking back through this, I wonder if the behavior is slightly off. @trws's comment that:
> do we want the "reserved" status to be reported for any job that was ever reserved (current behavior) or a job that was allocated immediately after being reserved without being de-prioritized inbetween?

has made me realize that the `was_reserved` being sticky might pose a problem. 

If a job _was_ reserved, another higher-priority job comes along and takes its reservation (possibly while the job itself is also deprioritized), is it possible for the job to end up being backfilled even though `was_reserved` is set?

wihobbs:
Working off of pure vibes, I want to see if this passes CI. Hence the change to draft.

qmanager/policies/queue_policy_bf_base_impl.hpp
```diff
@@ -195,6 +213,24 @@ int queue_policy_bf_base_t<reapi_type>::handle_match_failure (flux_jobid_t jobid
...
+        // actually runs post_sched_loop after this loop ends; otherwise the
+        // clearing annotation would sit untransmitted until the next unrelated
+        // scheduling event.
+        auto &fsched = m_jobs[element_iter->second]->schedule;
+        if (fsched.at != 0) {
+            fsched.at = 0;
+            fsched.selection_type = job_selection_type_t::UNKNOWN;
+            fsched.was_reserved = false;
+            m_scheduled = true;
+        }
```
wihobbs: I'm kinda sweating a bit here because of the reference to the sched loop, I really want someone to check and make sure the machines and I are not proposing something insane.

wihobbs:
@trws, I know we discussed whittling down the scope of this to just backfill/not backfill in last week's Fluxion meeting, and I still can do that if need be. However, I decided to give the `eta:now` issue one more go and I think I solved it.

t/t1004-qmanager-optimize.t
```diff
@@ -112,16 +149,45 @@ test_expect_success 'qmanager: CONSERVATIVE policy conforms to queue-depth=5' '
...
+    flux job wait-event -t 10 ${jobid2} start &&
+    test "$(selection_type ${jobid2})" = "reserved"
+'
+test_expect_success 'qmanager: cancel all active jobs' '
+    active_jobs=$(flux job list --states=active | jq .id) &&
+    for job in ${active_jobs}; do flux cancel ${job}; done &&
+    for job in ${active_jobs}; do flux job wait-event -t 10 ${job} clean; done
+'
+test_expect_success 'cleanup active jobs' '
+    cleanup_active_jobs
```
wihobbs, line 234: This is redundant.

qmanager/modules/qmanager_callbacks.cpp
```diff
@@ -99,14 +110,40 @@ int qmanager_cb_t::post_sched_loop (
             if (job->schedule.at == job->schedule.old_at)
                 continue;
             job->schedule.old_at = job->schedule.at;
```
wihobbs: A review from Codex suggested:

> Failed annotations are not retried
> 
> In post_sched_loop(), old_at is assigned before the annotation is successfully sent:
> 
> `job->schedule.old_at = job->schedule.at;`
> 
> If schedutil_alloc_respond_annotate_pack() fails, the next invocation sees equal values and skips the update. This is especially problematic for stale ETA deletion.
> 
> Update old_at only after success:
> ```
> if (job->schedule.at == job->schedule.old_at)
>     continue;
> 
> if (send_annotation (...) < 0) {
>     flux_log_error (...);
>     goto out;
> }
>
> job->schedule.old_at = job->schedule.at;
> ```

qmanager/policies/queue_policy_bf_base_impl.hpp
wihobbs: Two issues with this came up on a recent review pass using Claude Code:
1. Copying into `fsched` is potentially unsafe, `handle_match_success` guards against this (see lines 141-146)
2. Setting `m_scheduled` can trigger the sched loop to run twice or more often than necessary, recommendation was to use a separate variable when only an annotation update is needed. 

     The proposed changes (currently passes locally) are:

     ```diff
     index d32d971a..d6bc6292 100644
     --- a/qmanager/modules/qmanager_callbacks.cpp
     +++ b/qmanager/modules/qmanager_callbacks.cpp
     @@ -534,7 +534,8 @@ void qmanager_cb_t::prep_watcher_cb (flux_reactor_t *r, flux_watcher_t *w, int r
          for (auto &kv : ctx->queues) {
              std::shared_ptr<queue_policy_base_t> &queue = kv.second;
              ctx->pls_sched_loop = ctx->pls_sched_loop || queue->is_schedulable ();
     -        ctx->pls_post_loop = ctx->pls_post_loop || queue->is_scheduled ();
     +        ctx->pls_post_loop =
     +            ctx->pls_post_loop || queue->is_scheduled () || queue->is_annotation_dirty ();
          }
          if (ctx->pls_sched_loop || ctx->pls_post_loop)
              flux_watcher_start (ctx->idle);
     diff --git a/qmanager/policies/base/queue_policy_base.hpp b/qmanager/policies/base/queue_policy_base.hpp
     index e23cd50a..d77909e5 100644
     --- a/qmanager/policies/base/queue_policy_base.hpp
     +++ b/qmanager/policies/base/queue_policy_base.hpp
     @@ -815,11 +815,22 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {
              return m_scheduled;
          }

     -    /*! Reset this queue's "scheduled" state.
     +    /*! Return true if a job annotation has changed (e.g. a stale
     +     *  t_estimate was cleared) and post_sched_loop must run to transmit
     +     *  it, even though no job was scheduled.  Unlike is_scheduled (),
     +     *  this does not cause the scheduling loop to be restarted.
     +     */
     +    bool is_annotation_dirty ()
     +    {
     +        return m_annotation_dirty;
     +    }
     +
     +    /*! Reset this queue's "scheduled" and "annotation dirty" states.
           */
          void reset_scheduled ()
          {
              m_scheduled = false;
     +        m_annotation_dirty = false;
          }

          /*! Implement queue_adapter_base_t's pure virtual method
     @@ -1192,6 +1203,7 @@ class queue_policy_base_t : public resource_model::queue_adapter_base_t {

          bool m_schedulable = false;
          bool m_scheduled = false;
     +    bool m_annotation_dirty = false;
          bool m_sched_loop_active = false;
          uint64_t m_pq_cnt = 0;
          uint64_t m_rq_cnt = 0;
     diff --git a/qmanager/policies/queue_policy_bf_base_impl.hpp b/qmanager/policies/queue_policy_bf_base_impl.hpp
     index eddc8b22..74998f71 100644
     --- a/qmanager/policies/queue_policy_bf_base_impl.hpp
     +++ b/qmanager/policies/queue_policy_bf_base_impl.hpp
     @@ -213,6 +213,17 @@ int queue_policy_bf_base_t<reapi_type>::handle_match_failure (flux_jobid_t jobid

              // copy iterator before we advance to avoid invalidation
              auto element_iter = m_in_progress_iter;
     +        auto job_it = m_jobs.find (element_iter->second);
     +        if (job_it == m_jobs.end ()) {
     +            // The job is just missing, much worse
     +            errno = ENOENT;
     +            return -1;
     +        }
     +        auto &job = job_it->second;
     +        if (job->id != static_cast<flux_jobid_t> (jobid)) {
     +            errno = EINVAL;
     +            return -1;
     +        }
              // This job could be neither allocated nor reserved this loop, so it no
              // longer has a valid start-time estimate.  Clear it so post_sched_loop
              // detects the change and sends a null sched.t_estimate annotation,
     @@ -220,16 +231,16 @@ int queue_policy_bf_base_t<reapi_type>::handle_match_failure (flux_jobid_t jobid
              // reserved (issue #1424).  was_reserved is also cleared: the job lost
              // its reservation, so its eventual allocation should be categorized by
              // how it actually starts (reserved/backfill/immediate) rather than by
     -        // the reservation it once held.  Flag m_scheduled so the check watcher
     -        // actually runs post_sched_loop after this loop ends; otherwise the
     -        // clearing annotation would sit untransmitted until the next unrelated
     -        // scheduling event.
     -        auto &fsched = m_jobs[element_iter->second]->schedule;
     +        // the reservation it once held.  Flag m_annotation_dirty so the check
     +        // watcher actually runs post_sched_loop after this loop ends; otherwise
     +        // the clearing annotation would sit untransmitted until the next
     +        // unrelated scheduling event.
     +        auto &fsched = job->schedule;
              if (fsched.at != 0) {
                  fsched.at = 0;
                  fsched.selection_type = job_selection_type_t::UNKNOWN;
                  fsched.was_reserved = false;
     -            m_scheduled = true;
     +            m_annotation_dirty = true;
              }
              // if we are allocating and not trying to reserve, as in FCFS for
              // example, EBUSY means the request failed because not enough
     ```
     
I think this solves the issue of "blocked jobs may not recieve an updated annotation, because the sched loop _might_ not run after" (unlikely) but I'm not sure it actually solves the issue of the sched loop running too often or unnecessarily.

trws: It should solve the unnecessary runs, since it isn't requesting the scheduling loop to start anymore.  Need to look the rest over but at least that seems better here.

qmanager/modules/qmanager_callbacks.cpp
trws: That sounds right. Are you also trying to solve stale ETA updates?

qmanager/policies/queue_policy_bf_base_impl.hpp
trws: Ok yeah, the `m_scheduled` update here is definitely not what we want.  Making the switch suggested in the comment here should fix it, otherwise any time something gets reverted from reserved to unknown we could end up re-starting the loop when the current one finishes when we don't need to.

qmanager/policies/queue_policy_fcfs_impl.hpp
```diff
@@ -110,6 +110,9 @@ int queue_policy_fcfs_t<reapi_type>::handle_match_success (flux_jobid_t jobid,
     job->schedule.R = R;
     job->schedule.at = at;
     job->schedule.ov = ov;
+    // fcfs does not backfill or reserve, so any allocation is immediate.
+    if (!job->schedule.reserved)
```
trws: Why the check? Shouldn't this always be immediate regardless?

t/t1004-qmanager-optimize.t
```diff
@@ -112,17 +146,39 @@ test_expect_success 'qmanager: CONSERVATIVE policy conforms to queue-depth=5' '
...
+    test "$(selection_type ${jobid1})" = "immediate"
+'
+# Cancel jobid1 to free its cores so the reserved jobid2 can finally
+# start; because it had previously been reserved it is categorized as
+# reserved.
+test_expect_success 'qmanager: CONSERVATIVE selection_type is reserved' '
+    flux cancel ${jobid1} &&
+    flux job wait-event -t 10 ${jobid2} start &&
+    test "$(selection_type ${jobid2})" = "reserved"
+'
```
trws, line 232: I see the 1424 test above, but do you think we could have a test in here that hits reserved being set and then being cleared and ending up as immediate or backfill? Same for the time estimate actually, it looks like all the ingredients are here to allow for both.

qmanager/modules/qmanager_callbacks.cpp
wihobbs: Fixed this, and also got rid of the `continue` statement because I was having trouble following it.

qmanager/policies/queue_policy_bf_base_impl.hpp
wihobbs: Fixed on the most recent push.

trws (approved):
This looks great @wihobbs. 👍

qmanager/policies/queue_policy_bf_base_impl.hpp
jameshcorbett, line 262: Claude noted that on this code path, the job is removed from m_pending and placed in m_blocked, so its ETA isn't updated. Could be worth a comment and perhaps a test that's expected to fail?

jameshcorbett (approved):
I couldn't find anything wrong with this! Claude noticed one thing though--

qmanager/policies/queue_policy_bf_base_impl.hpp
trws: This might be something for a followup PR.  We had a discussion a while ago about setting an annotation for jobs that go into the blocked state, but had trouble coming up with a term that sounded better than "may never run".  It should clear stale ETAs either way I suppose.

wihobbs: I'll put up a separate PR later today with that! If we don't already have it, that might actually be good contextual info to show in `flux jobs`. Like we show "held" we could also show "blocked" so people know their job can't run under current conditions.

PR #1533, opened 2026-07-19
[bug] request for non-exclusive resource in subtree dropped from R

vsoch:
When one jobspec co-requests compute (node -> core) and something that lives in a sibling subtree under the same parent (ssd, a child of rack alongside node), resource-query match allocate reports allocated but the emitted R doesn't contain the ssd. I think it t is silently dropped, and I'll try to investigate why here. The ssd allocates fine on its own, so I don't think it's an issue of an unsupported type. I discovered this testing quantum devices, and it was based on this test passing, and I realized it was not behaving as I expected.

The "parallel subtree of different types" is the strategy we have to use (at the same level as rack) when we cannot put the slot above the two (it would then be right under the cluster, which I think might be a bad idea) and we cannot put constraints that would prune off an entire tree. This PR has a test that will demonstrate the behavior, meaning that the CI should pass if we do not have the ssd. I will also show it here. First, here is the test that shows that asking for an SSD alone works (so it is not the type):

```bash
cat ssd.out
```
```console
{"graph": {"nodes": [{"id": "85", "metadata": {"type": "ssd", "id": 35, "rank": -1, "exclusive": true, "size": 1024, "paths": {"containment": "/compute0/rack1/ssd35"}}}, {"id": "49", "metadata": {"type": "rack", "id": 1, "rank": -1, "properties": {"rabbit": "kind-worker3", "ssdcount": "36"}, "paths": {"containment": "/compute0/rack1"}}}, {"id": "0", "metadata": {"type": "cluster", "basename": "compute", "id": 0, "rank": -1, "paths": {"containment": "/compute0"}}}], "edges": [{"source": "49", "target": "85"}, {"source": "0", "target": "49"}]}}
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
```

And here is the output when we ask for a core and the ssd. Note there is no ssd.

```bash
cat core_ssd.out 
```
```console
{"graph": {"nodes": [{"id": "48", "metadata": {"type": "core", "id": 9, "rank": 0, "exclusive": true, "paths": {"containment": "/compute0/rack0/compute-01/core9"}}}, {"id": "38", "metadata": {"type": "node", "name": "compute-01", "id": 1, "rank": 0, "paths": {"containment": "/compute0/rack0/compute-01"}}}, {"id": "1", "metadata": {"type": "rack", "id": 0, "rank": -1, "properties": {"rabbit": "kind-worker2", "ssdcount": "36"}, "paths": {"containment": "/compute0/rack0"}}}, {"id": "0", "metadata": {"type": "cluster", "basename": "compute", "id": 0, "rank": -1, "paths": {"containment": "/compute0"}}}], "edges": [{"source": "38", "target": "48"}, {"source": "1", "target": "38"}, {"source": "0", "target": "1"}]}}
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
```

Side note - I notice this jgf that was used for issue 1284 has a kind worker and ssdcount as properties on the node. 

```json
        {
          "id": "1",
          "metadata": {
            "type": "rack",
            "basename": "rack",
            "name": "rack0",
            "id": 0,
            "uniq_id": 1,
            "rank": -1,
            "exclusive": true,
            "unit": "",
            "size": 1,
            "properties": {
              "rabbit": "kind-worker2",
              "ssdcount": "36"
            },
            "paths": {
              "containment": "/compute0/rack0"
            }
          }
        },
```

This observation first confused me because I wondered if it was related to this match at all (I do not think so). Why are they there? Here is the prettified version. I see the cluster, rack, node, and core, but no ssd. 

<details>

<summary> Prettified version with jq </summary>

```json
{
  "graph": {
    "nodes": [
      {
        "id": "48",
        "metadata": {
          "type": "core",
          "id": 9,
          "rank": 0,
          "exclusive": true,
          "paths": {
            "containment": "/compute0/rack0/compute-01/core9"
          }
        }
      },
      {
        "id": "38",
        "metadata": {
          "type": "node",
          "name": "compute-01",
          "id": 1,
          "rank": 0,
          "paths": {
            "containment": "/compute0/rack0/compute-01"
          }
        }
      },
      {
        "id": "1",
        "metadata": {
          "type": "rack",
          "id": 0,
          "rank": -1,
          "properties": {
            "rabbit": "kind-worker2",
            "ssdcount": "36"
          },
          "paths": {
            "containment": "/compute0/rack0"
          }
        }
      },
      {
        "id": "0",
        "metadata": {
          "type": "cluster",
          "basename": "compute",
          "id": 0,
          "rank": -1,
          "paths": {
            "containment": "/compute0"
          }
        }
      }
    ],
    "edges": [
      {
        "source": "38",
        "target": "48"
      },
      {
        "source": "1",
        "target": "38"
      },
      {
        "source": "0",
        "target": "1"
      }
    ]
  }
}
```

</details>

As a sanity check - my jobspec jgf does ask for it.

```bash
cat core_ssd.json
{"version":1,"resources":[{"type":"cluster","count":1,"with":[{"type":"rack","count":1,"with":[{"type":"node","count":1,"with":[{"type":"slot","count":1,"label":"s","with":[{"type":"core","count":1}]}]},{"type":"ssd","count":1}]}]}],"attributes":{"system":{"duration":60}},"tasks":[{"command":["x"],"slot":"s","count":{"per_slot":1}}]}
```

So this test passing:

```bash
test_expect_success 'BUG: the co-requested ssd is silently dropped from the allocation' '
	test_must_fail grep -q "\"type\": \"ssd\"" core_ssd.out
'
```
and not seeing ssd in the above suggests that when we match for a resource in a second subtree, it reports allocated but silently drops the resource from the actual allocation. I don't know what is going on in the actual graph, but I decided to see if I could figure out where this is happening. I think that the entire match and score is OK, but in one of the last steps when it should be emitted from the R, that's where the error is.

I think I found it (i asked Claude for help tracing the code). The resource is matched and its edge stamped for emission, but then in [upd_plan](https://github.com/flux-framework/flux-sched/blob/e35e2db256f46380486d4bd128d54caee6b1bd71/resource/traversers/dfu_impl_update.cpp#L149) there is a check for exclusive that increments n only if something is exclusive, and then it is only emit if n>0 [here](https://github.com/flux-framework/flux-sched/blob/e35e2db256f46380486d4bd128d54caee6b1bd71/resource/traversers/dfu_impl_update.cpp#L264-L268). I can test adding exclusive, and it looks like if I ask for an exclusive device in another subtree, that comes through. Here is my test modified:

```bash
{"graph": {"nodes": [{"id": "37", "metadata": {"type": "ssd", "id": 35, "rank": -1, "exclusive": true, "size": 1024, "paths": {"containment": "/compute0/rack0/ssd35"}}}, {"id": "48", "metadata": {"type": "core", "id": 9, "rank": 0, "exclusive": true, "paths": {"containment": "/compute0/rack0/compute-01/core9"}}}, {"id": "38", "metadata": {"type": "node", "name": "compute-01", "id": 1, "rank": 0, "paths": {"containment": "/compute0/rack0/compute-01"}}}, {"id": "1", "metadata": {"type": "rack", "id": 0, "rank": -1, "properties": {"rabbit": "kind-worker2", "ssdcount": "36"}, "paths": {"containment": "/compute0/rack0"}}}, {"id": "0", "metadata": {"type": "cluster", "basename": "compute", "id": 0, "rank": -1, "paths": {"containment": "/compute0"}}}], "edges": [{"source": "1", "target": "37"}, {"source": "38", "target": "48"}, {"source": "1", "target": "38"}, {"source": "0", "target": "1"}]}}
INFO: JOBID=1
INFO: RESOURCES=ALLOCATED
INFO: SCHEDULED AT=Now
```

I think I might understand this now. If we don't ask for exclusive, we are technically just saying "make sure this matches, but I don't care if other people can use/match it too." The match _does_ fail if I remove the ssd from the jgf entirely. But I still wonder if I should get back a representation of the resource that I requested in the R that is returned? If I get some ssd, for example, that I need to target for my job, how would I know which one I got without that? 

For some context, I am working on an approach to coschedule quantum and classical resources. The classical resources are held (like gating in Kubernetes) with a new abstraction called a hold, which is a `RESERVE_ONLY` call and we essentially hold a reservation until an RPC call is done to release it. The RPC is done by a scout job, which is calling the first quantum call to get into the queue (and then get priority) to pass on the session to the classical resources. Since there is no easy way to represent a partial allocation and gated resources in one job, my strategy is to split into two jobspecs, and use this hold/release pattern. That is handled via a client plugin so its not exposed to the user - they just submit the jobspec that has a request for quantum. That part actually is working OK - what I thought was a potential bug was for the scout job resources, which needs a tiny number of classical (one cpu) and then the quantum. What I found is that the classical allocate, but the other subtree (the quantum) are entirely dropped. I came back to the test with ssd I saw that I was mirroring to sanity check if the same thing was happening, and it appears to be, and I think it comes down to exclusive. 

If this is a bug, we might discuss if it is easily fixable. Arguably, someone requesting something non exclusive might still be informed about what they got. As a workaround I can just set the qpu to exclusive - will test that next. Thanks.


milroy:
> I think I found it (i asked Claude for help tracing the code). The resource is matched and its edge stamped for emission, but then in [upd_plan](https://github.com/flux-framework/flux-sched/blob/e35e2db256f46380486d4bd128d54caee6b1bd71/resource/traversers/dfu_impl_update.cpp#L149) there is a check for exclusive that increments n only if something is exclusive, and then it is only emit if n>0 [here](https://github.com/flux-framework/flux-sched/blob/e35e2db256f46380486d4bd128d54caee6b1bd71/resource/traversers/dfu_impl_update.cpp#L264-L268). I can test adding exclusive, and it looks like if I ask for an exclusive device in another subtree, that comes through.

> If this is a bug, we might discuss if it is easily fixable. 

You've independently derived the changes I was implementing in PR #1408! I don't know if this behavior resulted from a desire to reduce the matched R size, or if it's just a bug. The planner behavior makes me suspect bug, though. This fix is quite straightforward, but the implications aren't.

I need to get PR #1467 merged and then update and merge #1408. The changes create a huge amount of expected output churn, so the review won't be easy.

PR #1534, opened 2026-07-20
dfu: Stop implicit subplan initialization for non-mutating match traversals

jacobtkeio:
Problem: Previously, the traverser used the index operator to access nodes' subplans, like `m_graph[u].subplans[s]`. Because the traversal to prime the pruning filters during graph initialization does not touch every node (it skips leaf nodes, which will never have pruning filters), generalized logic to access node subplans using `[]` in subsequent traversals can implicitly generate elements of the node's subplan list. This is not an issue for correctness since the elements are initialized to NULL, and the traverser always checks whether an element is null before using it.
However, this causes the graph's state to change in cases where it shouldn't. Specifically, this causes infra_data idempotence checks to fail, due to a change in the length of the subplan list, after running a MATCH_SATISFIABILITY request that does not actually add or remove any subplans.
The same is true of MATCH_WITHOUT_ALLOCATING requests (#1387).

Solution: Access node subplans in the traverser using `try_at` instead of `[]`.
Add a regression test that compares the state of the graph before and after running a MATCH_SATISFIABILITY request.

### DIFF
--- a/resource/reapi/test/reapi_cli_test_cxx.cpp
+++ b/resource/reapi/test/reapi_cli_test_cxx.cpp
@@ -3,6 +3,7 @@
 #include <catch2/catch_test_macros.hpp>
 #include <resource/reapi/bindings/c++/reapi_cli.hpp>
 #include <resource/policies/base/match_op.h>
+#include <resource/schema/resource_graph.hpp>
 #include <fstream>
 #include <cstdlib>
 
@@ -101,7 +102,7 @@ TEST_CASE ("Match basic jobspec", "[match C++]")
     ctx = std::make_shared<resource_query_t> (rgraph, options);
     REQUIRE (ctx);
 
-    match_op_t match_op = match_op_t::MATCH_ALLOCATE;
+    match_op_t match_op = MATCH_ALLOCATE;
     bool reserved = false;
     std::string R = "";
     uint64_t jobid = 1;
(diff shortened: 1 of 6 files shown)


jacobtkeio:
Note: it is difficult to change the phrasing of statements like `opt_p.value_or (null_planner)` because none of `boost::optional`'s operators cleanly express "empty or populated with nullptr" or the inverse in a single statement (`opt_x == boost::optional (nullptr)` returns false when empty and true when null), and `try_at` only returns references but `boost::optional` cannot be initialized with an rvalue reference (see [Optional references](https://www.boost.org/doc/libs/1_82_0/libs/optional/doc/html/boost_optional/tutorial/optional_references.html)).

Another option is to use `<=` and `>` for this check, since empty optionals on the LHS will be considered `<` the RHS. This produces the two statements `opt_x > [optional w/ lvalue nullptr]` for populated-and-not-null and `opt_x <= [optional w/ lvalue nullptr]` for not-populated-or-null. While simpler, these statements' meaning is less clear.

If there is clearer way to write this logic please let me know!

resource/traversers/dfu_impl_update.cpp
```diff
@@ -14,12 +14,16 @@ extern "C" {
...
 
+#include <boost/optional/optional.hpp>
+
 #include "resource/traversers/dfu_impl.hpp"
 
 using namespace Flux::Jobspec;
 using namespace Flux::resource_model;
 using namespace Flux::resource_model::detail;
 
+extern planner_multi_t *null_planner;
```
jameshcorbett: if this were kept it should be made const but I don't think it should be kept at all.

jameshcorbett:
> Note: it is difficult to change the phrasing of statements like `opt_p.value_or (null_planner)` because none of `boost::optional`'s operators cleanly express "empty or populated with nullptr" or the inverse in a single statement (`opt_x == boost::optional (nullptr)` returns false when empty and true when null), and `try_at` only returns references but `boost::optional` cannot be initialized with an rvalue reference (see [Optional references](https://www.boost.org/doc/libs/1_82_0/libs/optional/doc/html/boost_optional/tutorial/optional_references.html)).

Why does it have to be a single statement? Why not `opt_x && *opt_x`? I don't love the addition of a new global constant.

resource/schema/infra_data.cpp
```diff
@@ -105,15 +105,30 @@ pool_infra_t &pool_infra_t::operator= (const pool_infra_t &o)
...
+
+pool_infra_t::~pool_infra_t ()
+{
+    for (auto &p : subplans)
+        planner_multi_destroy (&(p));
+    if (x_checker)
+        planner_destroy (&x_checker);
+}
+
+bool pool_infra_t::has_equal_behavior_to (const pool_infra_t &o) const
```
jameshcorbett: I don't feel strongly about this but what about `equal_except_color` as a name instead?

resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -154,4 +155,97 @@ TEST_CASE ("Initialize REAPI CLI and test add and remove", "[initialize-add-remo
...
+
+    // Store the initial state of the graph
+    std::map<vtx_t, pool_infra_t> idata_map;
+    std::map<vtx_t, schedule_t> sched_map;
+    vtx_iterator_t u, end;
+    for (boost::tuples::tie (u, end) = boost::vertices (ctx->db->resource_graph); u != end; u++) {
+        idata_map[*u] = ctx->db->resource_graph[*u].idata;
+        sched_map[*u] = ctx->db->resource_graph[*u].schedule;
+        REQUIRE (idata_map[*u] == ctx->db->resource_graph[*u].idata);
+        REQUIRE (sched_map[*u] == ctx->db->resource_graph[*u].schedule);
```
jameshcorbett, line 191: Are these `REQUIRE`s really accomplishing anything? You just assigned the values.

jameshcorbett (commented):
Generally looks good but as I suggested in my other comment I think the `null_planner` should be dropped.

resource/traversers/dfu_impl_update.cpp
jacobtkeio: Very fair. We were wondering at first glance whether `opt_p && opt_p.get ()` could be replaced with something shorter like `opt_p != nullptr`, and I got tunnel visioned on a one-liner. In hindsight I like the `&&` better too!

resource/reapi/test/reapi_cli_test_cxx.cpp
jacobtkeio: Not semantically, but we define the `operator=` and `operator==` explicitly for both `schedule_t` and `pool_infra_t`, so these `REQUIRE`s are a nice sanity check that could, theoretically, catch a mistake in those four functions.
If some modification to the functions caused this check to fail, I think it's more clear to fail here instead of further down the line (if at all, since this check may not exist elsewhere).
That being said, I admit it's a bit silly and I will be happy to remove these `REQUIRE`s if you see fit!

resource/traversers/dfu_impl_update.cpp
jacobtkeio: By the way, do you feel strongly about `*opt_x` vs. `opt_x.get ()`?

trws: Not really, at least for me what's more important is being consistent about it in a function.

resource/reapi/test/reapi_cli_test_cxx.cpp
jameshcorbett: That's fair actually yeah keep them I guess!

resource/traversers/dfu_impl_update.cpp
milroy: I guess my preference is for the indirection operator since the current code uses it almost exclusively.

jameshcorbett (dismissed):
I'm good with this but @trws or @milroy should give final review.

resource/traversers/dfu_impl.cpp
```diff
@@ -161,9 +165,9 @@ int dfu_impl_t::by_subplan (const jobmeta_t &meta,
     int64_t at = meta.at;
     uint64_t d = meta.duration;
     std::vector<uint64_t> aggs;
-    planner_multi_t *p = (*m_graph)[u].idata.subplans[s];
+    boost::optional<planner_multi_t *&> opt_p = (*m_graph)[u].idata.subplans.try_at (s);
```
milroy, line 168: Putting this comment here, because there's another occurrence outside DFU: https://github.com/flux-framework/flux-sched/blob/3e0582accfa7b6c23fc8ffc9b3bb50b597e07c54/resource/readers/resource_reader_jgf.cpp#L770

Might as well fix that line, too, instead of having a single-line follow-up PR.

resource/schema/infra_data.cpp
```diff
@@ -105,15 +105,24 @@ pool_infra_t &pool_infra_t::operator= (const pool_infra_t &o)
...
+    if (!equal_except_color (o))
+        return false;
+    if (colors != o.colors)
+        return false;
+    return true;
+}
+
+// Two pool_infra_ts with different colors but all else equal will behave the
+// same in a traversal.
+bool pool_infra_t::equal_except_color (const pool_infra_t &o) const
```
milroy, line 118: Noting for a follow-up PR (out of scope here): `planner_multis_equal()` does `*(lhs->plan_multi) == *(rhs->plan_multi)` with no `NULL` guard. https://github.com/flux-framework/flux-sched/blob/3e0582accfa7b6c23fc8ffc9b3bb50b597e07c54/resource/planner/c/planner_multi_c_interface.cpp#L621

This must get fixed in a new PR.

milroy (dismissed):
This looks great @jacobtkeio! Please add the additional `try_at` logic in the non-DFU site, and then you can set MWP: https://github.com/flux-framework/flux-sched/blob/3e0582accfa7b6c23fc8ffc9b3bb50b597e07c54/resource/readers/resource_reader_jgf.cpp#L770

milroy (approved):
Thanks for the quick changes and the great work, @jacobtkeio!

jacobtkeio:
Thank you everyone for your reviews!

Issue #1531, opened 2026-07-14
valgrind test is noisy due to hwloc errors


wihobbs:
With some frequency on the [ci / jammy - test-install (pull_request) check](https://github.com/flux-framework/flux-sched/actions/runs/29121987620/job/86458991413?pr=1517) we're seeing the following issue, which appears to be from hwloc and goes away. Is this a known issue? Is there something we should be doing to suppress the warning?

```
Test project /usr/src
      Start 141: t5000-valgrind.t
  1/1 Test #141: t5000-valgrind.t .................***Failed   24.54 sec
  expecting success: 
  	run_timeout 900 \
  	flux start -s ${VALGRIND_NBROKERS} \
  		--killer-timeout=120 \
  		--wrap=${VALGRIND} \
  		--wrap=--tool=memcheck \
  		--wrap=--leak-check=full \
  		--wrap=--gen-suppressions=all \
  		--wrap=--trace-children=no \
  		--wrap=--child-silent-after-fork=yes \
  		--wrap=--num-callers=30 \
  		--wrap=--leak-resolution=med \
  		--wrap=--error-exitcode=1 \
  		--wrap=--suppressions=$VALGRIND_SUPPRESSIONS \
  		 ${VALGRIND_WORKLOAD}
  
  ==27845== Memcheck, a memory error detector
  ==27845== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
  ==27845== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
  ==27844== Memcheck, a memory error detector
  ==27845== Command: /usr/libexec/flux/cmd/flux-broker --setattr=rundir=/tmp/flux-TaU2Td
  ==27844== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
  ==27845== 
  ==27844== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
  ==27844== Command: /usr/libexec/flux/cmd/flux-broker --setattr=rundir=/tmp/flux-TaU2Td /usr/src/t/valgrind/valgrind-workload.sh
  ==27844== 
  hwloc x86 backend cannot work under Valgrind, disabling.
  May be reenabled by dumping CPUIDs with hwloc-gather-cpuid
  and reloading them under Valgrind with HWLOC_CPUID_PATH.
  hwloc x86 backend cannot work under Valgrind, disabling.
[...]

PR #1535, opened 2026-07-22
doc: relax requirements.txt as documentation allows

### LINKED ISSUE #1529: Python package requirements for docs seem a little too high?
After #1502 from @wihobbs the required versions for Python packages in [`doc/requirements.txt`](https://github.com/flux-framework/flux-sched/blob/master/doc/requirements.txt) seem a little too high, and I am wondering if there was a specific reason all the `<` signs turned into `>=`?

I am building on Rocky9 and it seems to work (with the exception of Sphinx's `suppress_warnings` for `duplicate_declaration` that appear to have been [added in v8.2](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings)) while the current requirements would only be met for RHEL 10 based systems:

rhel version | `sphinx` | `docutils` | `breathe` | `jinja2`
-|-|-|-|-
8 | 1.7.6 | 0.14 | 4.11.1 | 2.10.1
9 | 3.4.3 | 0.16 | 4.34.0 | 2.11.3
10 | 7.2.6 | 0.20.1 | 4.35.0 | 3.1.6

Briefly looking at the features of `breathe` that are actually used in the config, the minimum required version I would come up with would be `4.32.0` for `breathe_show_include` introduced in https://github.com/breathe-doc/breathe/commit/4fb7bd54ab5719d58f237ebd73d34b037b4dfae9.

This would then give [implicit requirements of](https://github.com/breathe-doc/breathe/blob/v4.32.0/requirements/production.txt):
- Sphinx>=3.0,<5
- docutils>=0.12
- Jinja2>=2.7.3

Which would then be satisfied also for RHEL 9, and only limited from RHEL 8 by the sphinx version requirement.

wihobbs:
Problem: the requirements.txt file for the documentation is more stringent than is necessary to build the docs.

Solution: relax the requirements.

Fixes #1529

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,5 +1,5 @@
-sphinx>=6.0.0
-sphinx-rtd-theme>=0.5.2
-docutils>=0.18
-jinja2>=3.1.0
-breathe>=4.35.0
+sphinx
+sphinx-rtd-theme
+docutils
+jinja2
+breathe>=4.32.0


jameshcorbett (approved):
Looks ok!

trws:
The only concern I have would have been fedora40, or now rhel9, passing and building docs.  Assuming they do, this looks great!

Should probably propagate to core if relevant there.

wihobbs:
The current el9 failure appears to be due to #1531 and unrelated to the documentation versions (but I'm not restarting it because I want to work on valgrind suppressions.)

wihobbs:
> The only concern I have would have been fedora40, or now rhel9, passing and building docs.

FYI, I think our CI is set up so that they pass and build the manual pages on `make install`, but not the website -- doxygen was only added to the `bookworm` container when I added it in #1502. 

Do you want the el9 container to also be able to build the internals site? I had sort of determined that to be out of scope for `make install`, and only added the necessary packages to the bookworm container for my own convenience.

trws:
I honestly don't really care, just didn't realize it was a separate issue tickling that one.  It might be worth doing that when we update to the next TOSS release so everything works in our normal production environment but I don't see it being worth all that much right now.  Thanks for this @wihobbs.

trws (approved):

PR #1539, opened 2026-07-27
t: replace `flux run` with `flux submit` in alloc-check

wihobbs:
From a recent local test run:

```
        Start  71: t1024-alloc-check.t
 71/150 Test  #71: t1024-alloc-check.t ..............   Passed   52.16 sec
 ```

### DIFF
--- a/t/t1024-alloc-check.t
+++ b/t/t1024-alloc-check.t
@@ -38,8 +38,9 @@ test_expect_success 'configure epilog with delay' '
 # Jobs seem to need to be submitted separately to trigger the issue.
 test_expect_success 'submit node-exclusive jobs that exceed their time limit' '
 	(for i in $(seq 5); do \
-	    flux run -N1 -x -t1s sleep 30 || true; \
-	done) 2>joberr
+	    flux submit -N1 -x -t1s sleep 30 || true; \
+	done)
+	flux job eventlog $(flux job last) > joberr
 '
 test_expect_success 'some jobs received timeout exception' '
 	grep "job.exception" joberr | grep "type=timeout"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wihobbs** [comment]
Well that didn't work as expected.

**Comment by wihobbs** [comment]
Ok, I've gone around in circles on this test. 

The issue is really that the per-rank epilog has a `sleep 2` command, which means that after all of 5 exclusive jobs, and 10 non-exclusive jobs running two at a time, the system sleeps for 2 seconds. This means that for the duration of the test about 20 seconds is wasted. This is necessary to recreate the test case for the issue the test addresses. 

Using `flux run` or waiting for the `clean` event is blocking on this operation, and waiting for `flux job last` does the same x5. 

tl;dr this test takes a long time, and there's not really a way around it. It's not broken, unless the "submit" in the comment should really be "run," and pedantically changing that doesn't seem like a worthwhile PR.

I did notice that @milroy commented in #1387 recommending a similar test there, and the recommended command is `flux submit` not `flux run` (as the test has verbatim). Here's roughly how you would have to rework this test for `flux submit` to capture the needed information:

```diff
index 18f478cc..e895d3c3 100755
--- a/t/t1024-alloc-check.t
+++ b/t/t1024-alloc-check.t
test_expect_success 'load fluxion modules' '
@@ -38,15 +39,15 @@ test_expect_success 'configure epilog with delay' '
 # Jobs seem to need to be submitted separately to trigger the issue.
 test_expect_success 'submit node-exclusive jobs that exceed their time limit' '
 	(for i in $(seq 5); do \
-	    flux run -N1 -x -t1s sleep 30 || true; \
+	    flux submit -N1 -x -t1s sleep 30 || true; \
 	done)
-	flux job eventlog $(flux job last) > joberr
+	flux job wait-event -v $(flux job last) clean > joberr ## FIXME do not rely on one job, instead subscribe to the journal for all events
 '
 test_expect_success 'some jobs received timeout exception' '
-	grep "job.exception" joberr | grep "type=timeout"
+	grep "exception" joberr | grep "type=\"timeout\""
 '
 test_expect_success 'no jobs received alloc-check exception' '
-	test_must_fail grep "job.exception.*type=alloc-check" joberr
+	test_must_fail grep "exception.*type=alloc-check" joberr
 '
 test_expect_success 'clean up' '
 	flux cancel --all &&
```

And a relatively similar change for the same non-exclusive test on line ~58.

PR #1540, opened 2026-07-27
Fix unguarded dereferences, improve exception safety and planner errno handling

milroy:
Reviewing PR #1534 revealed one case of unguarded pointer dereferencing in `planner`. I had Fable 5 scan the entire codebase for similar problems. Unsurprisingly, it found many and also found issues with exception safety and `errno` reporting.

This PR fixes those cases, and is assisted by Claude. It lays a foundation for planner test improvement and eventual refactor of the planner API.

### DIFF
--- a/resource/planner/c++/planner.cpp
+++ b/resource/planner/c++/planner.cpp
@@ -541,12 +541,18 @@ bool planner::trees_equal (const planner &o) const
 // Public Planner_t methods
 ////////////////////////////////////////////////////////////////////////////////
 
+// The wrapper constructors rethrow so that a planner_t can never be
+// observed with a null inner planner; `new planner_t (...)` either
+// succeeds completely or throws (operator new releases the wrapper
+// allocation automatically when the constructor throws).
+
 planner_t::planner_t ()
 {
     try {
         plan = new planner ();
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
@@ -556,6 +562,7 @@ planner_t::planner_t (const planner &o)
         plan = new planner (o);
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
@@ -568,6 +575,7 @@ planner_t::planner_t (const int64_t base_time,
         plan = new planner (base_time, duration, resource_totals, in_resource_type);
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
--- a/resource/planner/c++/planner.cpp
+++ b/resource/planner/c++/planner.cpp
@@ -104,6 +104,9 @@ planner &planner::operator= (const planner &o)
 {
     int rc = -1;
 
+    if (this == &o)
+        return *this;
+
     if ((rc = erase ()) != 0) {
         throw std::runtime_error ("ERROR erasing *this\n");
     }
@@ -564,6 +567,25 @@ planner_t::planner_t (const planner &o)
         errno = ENOMEM;
         throw;
     }
+}
+
+planner_t &planner_t::operator= (const planner_t &o)
+{
+    if (this != &o) {
+        // Copy-and-swap for the strong exception guarantee: build the
+        // replacement first so *this is untouched if the copy throws
+        // (bad_alloc, or runtime_error from planner's tree/map copies).
+        planner *tmp = nullptr;
+        try {
+            tmp = new planner (*o.plan);
+        } catch (std::bad_alloc &e) {
+            errno = ENOMEM;
+            throw;
+        }
+        delete plan;
+        plan = tmp;
+    }
+    return *this;
 }
 
 planner_t::planner_t (const int64_t base_time,
--- a/resource/planner/c++/planner_multi.cpp
+++ b/resource/planner/c++/planner_multi.cpp
@@ -17,6 +17,7 @@ extern "C" {
 #include <cstdlib>
 #include <cerrno>
 #include <cstring>
+#include <memory>
 #include <vector>
 #include <map>
 
@@ -187,22 +188,26 @@ void planner_multi::add_planner (int64_t base_time,
                                  size_t i)
 {
     std::string type;
-    planner_t *p = nullptr;
+    std::unique_ptr<planner_t> p;
 
+    // Own the new planner via unique_ptr until it is safely inserted so
+    // that a failed map or multi-index insertion cannot leak it.
     try {
         type = std::string (resource_type);
-        p = new planner_t (base_time, duration, resource_total, resource_type);
+        p = std::make_unique<planner_t> (base_time, duration, resource_total, resource_type);
+        m_iter.counts[type] = 0;
+        if (i > m_types_totals_planners.size ())
+            m_types_totals_planners.push_back ({type, resource_total, p.get ()});
+        else {
+            auto it = m_types_totals_planners.begin () + i;
+            m_types_totals_planners.insert (it, planner_multi_meta{type, resource_total, p.get ()});
+        }
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
         throw std::bad_alloc ();
     }
-    m_iter.counts[type] = 0;
-    if (i > m_types_totals_planners.size ())
-        m_types_totals_planners.push_back ({type, resource_total, p});
-    else {
-        auto it = m_types_totals_planners.begin () + i;
-        m_types_totals_planners.insert (it, planner_multi_meta{type, resource_total, p});
-    }
+    // The container now references the planner; release ownership.
+    p.release ();
 }
 
 void planner_multi::delete_planners (const std::unordered_set<std::string> &rtypes)
@@ -334,12 +339,18 @@ void planner_multi::incr_span_counter ()
 // Public Planner_multi_t methods
 ////////////////////////////////////////////////////////////////////////////////
 
+// The wrapper constructors rethrow so that a planner_multi_t can never
+// be observed with a null inner planner_multi; `new planner_multi_t (...)`
+// either succeeds completely or throws (operator new releases the wrapper
+// allocation automatically when the constructor throws).
+
 planner_multi_t::planner_multi_t ()
 {
     try {
         plan_multi = new planner_multi ();
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
@@ -349,6 +360,7 @@ planner_multi_t::planner_multi_t (const planner_multi &o)
         plan_multi = new planner_multi (o);
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
@@ -362,6 +374,7 @@ planner_multi_t::planner_multi_t (int64_t base_time,
         plan_multi = new planner_multi (base_time, duration, resource_totals, resource_types, len);
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+        throw;
     }
 }
 
--- a/resource/planner/c++/planner_multi.cpp
+++ b/resource/planner/c++/planner_multi.cpp
@@ -83,12 +83,19 @@ planner_multi::planner_multi (const planner_multi &o)
     }
     m_iter = o.m_iter;
     m_span_lookup = o.m_span_lookup;
-    m_span_lookup_iter = o.m_span_lookup_iter;
+    // o's iterator points into o's m_span_lookup; copying it would leave
+    // this object holding an iterator into another container.  Reset it to
+    // a defined state instead; users must call avail_time_first () before
+    // avail_time_next () anyway.
+    m_span_lookup_iter = m_span_lookup.end ();
     m_span_counter = o.m_span_counter;
 }
 
 planner_multi &planner_multi::operator= (const planner_multi &o)
 {
+    if (this == &o)
+        return *this;
+
     // Erase *this so the vectors are empty
     erase ();
 
@@ -120,7 +127,8 @@ planner_multi &planner_multi::operator= (const planner_multi &o)
     }
     m_iter = o.m_iter;
     m_span_lookup = o.m_span_lookup;
-    m_span_lookup_iter = o.m_span_lookup_iter;
+    // See the copy constructor: never adopt an iterator into o's container.
+    m_span_lookup_iter = m_span_lookup.end ();
     m_span_counter = o.m_span_counter;
 
     return *this;
@@ -362,6 +370,25 @@ planner_multi_t::planner_multi_t (const planner_multi &o)
         errno = ENOMEM;
         throw;
     }
+}
+
+planner_multi_t &planner_multi_t::operator= (const planner_multi_t &o)
+{
+    if (this != &o) {
+        // Copy-and-swap for the strong exception guarantee: build the
+        // replacement first so *this is untouched if the copy throws
+        // (bad_alloc, or runtime_error from the inner planner copies).
+        planner_multi *tmp = nullptr;
+        try {
+            tmp = new planner_multi (*o.plan_multi);
+        } catch (std::bad_alloc &e) {
+            errno = ENOMEM;
+            throw;
+        }
+        delete plan_multi;
+        plan_multi = tmp;
+    }
+    return *this;
 }
 
 planner_multi_t::planner_multi_t (int64_t base_time,
--- a/resource/planner/c/planner_c_interface.cpp
+++ b/resource/planner/c/planner_c_interface.cpp
@@ -12,6 +12,7 @@
 #include <map>
 #include <list>
 #include <string>
+#include <stdexcept>
 
 #include "resource/planner/c++/planner.hpp"
 
@@ -340,6 +341,11 @@ extern "C" planner_t *planner_copy (planner_t *p)
         ctx = new planner_t (*(p->plan));
     } catch (std::bad_alloc &e) {
         errno = ENOMEM;
+    } catch (std::runtime_error &e) {
+        // planner's copy machinery reports allocation failures in its
+        // tree and map copies as runtime_error; it must not escape this
+        // extern "C" boundary.
+        errno = ENOMEM;
     }
 
     return ctx;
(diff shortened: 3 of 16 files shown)


resource/planner/c/planner_c_interface.cpp
```diff
@@ -342,6 +347,10 @@ extern "C" planner_t *planner_copy (planner_t *p)
 
 extern "C" void planner_assign (planner_t *lhs, planner_t *rhs)
 {
+    if (!lhs || !rhs) {
+        errno = EINVAL;
+        return;
+    }
     (*(lhs->plan) = *(rhs->plan));
 }
```
garlick: A void function setting errno seems like a red flag, since errno should never be tested as the primary signal for an error condition.

milroy: That's a good point; I should have caught that. Refactoring now.

jameshcorbett: If we were writing this interface from scratch, would you have the function return an int then? Or should the function just require that the arguments are valid / not `nullptr`?

milroy: > would you have the function return an int then?

Yes, I think so, and that's what I did for the commit I just added. Eventually we want to get rid of the whole C interface, but we need to address these issues first.

trws: Using an int or discriminated return type or some kind is definitely what we want wherever errno gets touched unless we're using exceptions (which wouldn't be a good idea here).

trws: Actually, adding this to my list of lints we probably ought to have...  Good point @garlick!

resource/planner/c++/planner_multi.cpp
```diff
@@ -364,6 +372,25 @@ planner_multi_t::planner_multi_t (const planner_multi &o)
...
+            tmp = new planner_multi (*o.plan_multi);
+        } catch (std::bad_alloc &e) {
+            errno = ENOMEM;
+            throw;
+        }
+        delete plan_multi;
+        plan_multi = tmp;
+    }
+    return *this;
+}
```
jameshcorbett, line 392: Not introduced by this PR , but here and on planner_t, delete the compiler-generated copy constructor? Destructor would free the `plan` pointer, leading to a double free.

jameshcorbett (approved):
LGTM! Just one little thing.

resource/planner/c++/planner.cpp
```diff
@@ -556,7 +565,27 @@ planner_t::planner_t (const planner &o)
...
+
+planner_t &planner_t::operator= (const planner_t &o)
+{
+    if (this != &o) {
+        // Copy-and-swap for the strong exception guarantee: build the
+        // replacement first so *this is untouched if the copy throws
+        // (bad_alloc, or runtime_error from planner's tree/map copies).
+        planner *tmp = nullptr;
+        try {
+            tmp = new planner (*o.plan);
```
trws: It took me a minute to parse this, because it's a bit different from the normal idiom.  Couple of questions, since you're doing this, is there a reason for plan to remain a pointer rather than a by-value member in planner_t?  I don't think we're refcounting or doing multiple references.

Related to that, why this rather than adding a swap, and using the classic pattern:

```c++
planner_t &planner_t::operator= (planner_t o)
{
  using std::swap;
  swap (*this, other);
  return *this;
}
```

Or the equivalent by-reference version that needs to use a by-value temp of the current type?

resource/planner/c++/planner_multi.cpp
trws: I think it's implicitly deleted because of the custom destructor etc., but yeah it should either be custom or deleted.  Good old rule of 5.

milroy:
While rebasing onto PR #1526 I noticed the REAPI module `reapi_module_match_satisfy` didn't get the tri-state. #1526 only touched `reapi_cli.cpp` and the CLI test, so the module still reports satisfiability the old way.

That leaves the two backends on different contracts for the same call, which is even more problematic for the module. `-1` now conflates "definitively unsatisfiable" with "couldn't reach the resource module."

I opened issue #1554 and give @vsoch first dibs on fixing it. However, I can also fix it here (probably not the right place) or in a new PR.

resource/planner/c++/planner_multi.cpp
milroy: I confirmed this problem empirically. Claude and I added a case that copy-constructs a `planner_t` from a `planner_t` and ran it against the tree without the rule-of-5 fix: `planner_errno_test` segfaults, and ASan reports "heap-use-after-free" with the free in `~planner_t` and the allocation in `planner_new`. So the copy constructor is not implicitly deleted. A user-declared destructor or copy assignment operator only deprecates it. With a deep copy constructor in place the case passes.

jameshcorbett:
Is this PR ready to go you think?

> I opened issue #1554 and give @vsoch first dibs on fixing it. However, I can also fix it here (probably not the right place) or in a new PR.

Should this fix go in the release that's about to go out?

milroy:
> Should this fix go in the release that's about to go out?

Ideally, yes, but I haven't gotten a chance to work on it yet.

milroy:
Claude and I updated the PR significantly.

The main change answers @jameshcorbett. The compiler-supplied copy constructor is not deleted. A user-declared destructor or copy assignment operator only deprecates it ([depr.impldec]); only a user-declared move operation deletes it, and neither wrapper has one. Both wrappers thus had a shallow copy constructor that caused a double free, which a test reproduces as a heap-use-after-free under AddressSanitizer. Both now have a deep copy constructor, a noexcept swap as a hidden friend, and copy-and-swap assignment, which is also @trws' suggestion.

That work revealed two more problems. First, no constructor of planner assigns `m_span_lookup_iter`, so the iterator is singular, and `erase ()` clears the map without a reset. Commit 5 fixes this. Second, `planner_span_next ()` and `planner_multi_span_next ()` increment that iterator before they test it against the end of the map, which is undefined in three states.  Commit 6 fixes this and adds a test for all three states. The work also added `catch (...)` at the C boundaries, because `reserve ()` throws `length_error` and not `bad_alloc`. I also added the `ENODEV` assertions @jameshcorbett asked for in #1526.

**Where to pay attention in a follow-up review:**
Commit 5 is the most important, in particular the decision not to declare move operations. Commit 6  keeps the `errno` that each function already reports, so `planner_span_next` still reports `EINVAL` where the `planner_multi` functions report `ENOENT`; `ENOENT` is correct, but a change is a change of contract. Commit 12 makes `schedule_t` and `pool_infra_t` throw where a failed `planner_assign` was previously ignored. This matches the behavior for `planner_copy`, but it has an effect outside the planner. Commit 4 is unchanged and is subject to a state where a stale span vector now gives `EINVAL` instead of an exception, so a span can become impossible to remove.

resource/planner/c++/planner.cpp
milroy: > is there a reason for plan to remain a pointer rather than a by-value member

Not in `planner_t`, and there's no refcounting. It matters in `planner_multi_meta`, which holds `planner_t *` inside a `Boost.MultiIndex` container whose elements are `const` except through `modify ()`, so the planners must be outside the container.

I don't think `planner_t` should inherit the guarantees of `planner::operator=`. That operator erases and then copies, so it gives the basic guarantee only. To keep the strong exception guarantee with a value member, planner first needs a noexcept swap. However, planner holds two intrusive trees, an `m_p0` that points into both, and an iterator into its own map. A swap must move all of these.

milroy: Updated comment to fix some inaccuracies.
> why this rather than adding a swap, and using the classic pattern

Implemented for both `planner_t` and `planner_multi_t`. It needed the rule of three first: copy-and-swap copy-constructs its by-value parameter, so the wrappers had to have a deep copy constructor rather than the implicit shallow one. Claude and I added swap as a hidden friend, which is what the unqualified s`wap (*this, o)` in the assignment body resolves to through ADL.

There are deliberately no move operations, so an rvalue deep-copies through the copy constructor. Declaring a move constructor would introduce a moved-from wrapper holding a null inner object. If you want moves, I'd flatten `plan` to a value member, which returns to your first point.

resource/planner/c/planner_multi_c_interface.cpp
```diff
@@ -652,12 +760,24 @@ extern "C" int planner_multi_update (planner_multi_t *ctx,
         }
         rtypes.insert (resource_types[i]);
```
jameshcorbett, line 761: This could throw `bad_alloc`, also will raise if `resource_types[i]` is NULL.

resource/planner/c/planner_c_interface.cpp
```diff
@@ -708,6 +708,14 @@ extern "C" int64_t planner_span_next (planner_t *ctx)
         errno = EINVAL;
         return -1;
     }
+    // Test before the increment.  The iterator sits at end () until
+    // planner_span_first sets it, after the previous call ran off the end,
+    // and after a copy, an assignment, or an erase.  Incrementing an end
+    // iterator is undefined.
```
jameshcorbett, line 714: This could be `// test before incrementing, incrementing an end iterator is undefined`

resource/planner/c++/planner_multi.cpp
jameshcorbett, line 59: codex had a good suggestion here, which I'm happy to implement in a separate PR. It noted that if a constructor throws, the destructor is not run. But the copy constructor gradually puts owning raw pointers into m_types_totals_planners. If an exception is raised during this method, the container itself is destroyed, but its raw planner_t * elements are not deleted. If push_back() throws, even the current np leaks because it was never inserted.

  The simplest robust fix is:

  1. Hold each new planner in a std::unique_ptr until insertion succeeds.
  2. Wrap the whole constructor body so previously inserted planners are deleted if any later operation throws.

  Conceptually:

```C++
  planner_multi::planner_multi (const planner_multi &o)
  {
      try {
          for (const auto &item : o.m_types_totals_planners) {
              std::unique_ptr<planner_t> np;

              if (item.planner)
                  np = std::make_unique<planner_t> (*item.planner);
              else
                  np = std::make_unique<planner_t> ();

              m_types_totals_planners.push_back ({
                  item.resource_type,
                  item.resource_total,
                  np.get (),
              });

              // The container now owns this pointer by convention.
              np.release ();
          }

          m_iter = o.m_iter;
          m_span_lookup = o.m_span_lookup;
          m_span_lookup_iter = m_span_lookup.end ();
          m_span_counter = o.m_span_counter;
      } catch (...) {
          // The class destructor will not run when construction fails.
          // Explicitly free planners inserted before the failure.
          erase ();
          throw;
      }
  }
```

  The unique_ptr and outer cleanup cover different failure points:

  - make_unique() fails: nothing leaks.
  - push_back() fails: np deletes the current planner.
  - A later iteration fails: erase() deletes earlier inserted planners.
  - Copying m_iter or m_span_lookup fails: erase() deletes every inserted planner.

jameshcorbett (approved):
LGTM! @wihobbs will love me for this but I do find that a lot of these (presumably Claude-generated) comments and commit messages are long and convoluted.

milroy:
Here's a Claude generated, edited-by-me summary of the recent changes and corrections to earlier statements:

#### Review comments
A `NULL` element of `resource_types[]` is now rejected with `EINVAL` in `planner_multi_new`, `_update`, and `_reduce_span` before any state changes. The iterator comment uses @jameshcorbett's wording. The `planner_multi` copy-constructor leak codex found is fixed here rather than in a separate PR: each new planner is held in a `unique_ptr` until inserted, and the inserted ones are freed before the exception leaves the constructor.

#### A crash the approved series still had
`planner`'s span-lookup iterator was initialized, but `planner_multi`'s wasn't. The new guard compares the iterator with `end ()`, and an uninitialized iterator is not equal to `end ()`, so `planner_multi_span_next` on an empty `planner_multi` still incremented an invalid iterator and crashed. The test covered` planner_span_next` only. Both are now fixed and tested.

#### Guards the earlier commits missed
`reapi_cli_info`, `reapi_cli_set/get_rank_status`, `reapi_cli_update_allocate`, `reapi_cli_partial_cancel`, and `reapi_module_update_allocate` now check every pointer they use. Two of them used to crash on `NULL`. Two `planner_multi` getters returned -1 without setting `errno`. Handlers that assigned an error message after an allocation failure are wrapped the same way as in `reapi_cli_initialize`.

#### Removed
Three of the four throw sites added earlier for a failed `planner_assign` could never run (a copy constructor starts with a null `planner`; `pool_infra_t::operator=` destroys its `planner` first). The only remaining check is in `schedule_t::operator=`. The inner `planner::operator=` and `planner_multi::operator=` are derelict now that `planner_assign` moved to the wrapper. Both are deleted. `erase ()` now resets `m_p0`.

#### Corrections from previous statements
The wrappers follow the rule of three, not five. Move operations are deliberately absent, so an rvalue deep-copies. The `std::swap` recursion I described earlier couldn't have happened. There is no `using std::swap` in the `planner` and `planner_t` is in the global namespace, so an unqualified `swap` had no candidate at all until the hidden friend existed; the hidden friend is simply what the call resolves to. `reserve ()` throws `length_error` only when the request exceeds `max_size ()`, while a failed allocation throws `bad_alloc`.

#### TODOs for follow-up PRs
`dfu_impl.cpp` ignores `planner_multi_update`'s return value; after this PR a stale span fails to remove instead of throwing. We also need a unit test for the module binding's argument guards.

milroy:
> @wihobbs will love me for this but I do find that a lot of these (presumably Claude-generated) comments and commit messages are long and convoluted.

I definitely agree. I regenerated and condensed the commit messages, so they should be intelligible now.

resource/planner/c/planner_c_interface.cpp
milroy: Good catch. This statement has been greatly simplified in the current version.

resource/planner/c/planner_multi_c_interface.cpp
milroy: Good catches. These problems are fixed in the updated commits.

resource/planner/c++/planner_multi.cpp
milroy: Also a good suggestion. The behavior you identified above is fixed in the updated commit.

milroy:
@jameshcorbett, thanks for your reviews! I think this is ready for a final pass, and it may be good to have someone else review these changes, or at least the updates, too.

resource/planner/c/planner_multi_c_interface.cpp
```diff
@@ -560,6 +623,10 @@ extern "C" int64_t planner_multi_span_planned_at (planner_multi_t *ctx,
         errno = ENOENT;
         return -1;
     }
+    // A type added after the span was created holds no allocation in it; report 0 rather
```
zekemorton: Test pending review comment — staged for preview only.

zekemorton, line 626: The new guards make invalid `m_span_lookup` entries fail safely, but `planner_multi_update()` can still add, remove, or reorder resource types without updating the existing positional span vectors. This can make an old span ID refer to the wrong planner and cause removal to target the wrong allocation. Could existing spans be remapped by resource type when the planner list changes?

This appears to be pre-existing, so a follow-up would be fine.

resource/planner/c/planner_multi_c_interface.cpp
```diff
@@ -381,14 +411,30 @@ extern "C" int64_t planner_multi_add_span (planner_multi_t *ctx,
...
+    try {
+        auto res = ctx->plan_multi->get_span_lookup ().insert (
+            std::pair<int64_t, std::vector<int64_t>> (mspan, std::vector<int64_t> ()));
+        if (!res.second) {
+            errno = EEXIST;
+            return -1;
+        }
+        // Reserve up front so the push_back below cannot throw after
+        // underlying planner spans have already been allocated.
+        res.first->second.reserve (len);
```
zekemorton, line 429: This `reserve()` prevents `push_back()` from failing after spans have been allocated. There is still a pre-existing partial-failure case below: if `planner_add_span()` succeeds for an earlier planner and fails for a later one, line 449 erases the lookup without removing the spans already added. Could that rollback be handled in a follow-up?

resource/planner/c/planner_c_interface.cpp
```diff
@@ -676,6 +699,11 @@ extern "C" int64_t planner_span_next (planner_t *ctx)
         errno = EINVAL;
         return -1;
     }
+    // Test before incrementing; incrementing an end iterator is undefined.
+    if (ctx->plan->get_span_lookup_iter () == ctx->plan->get_span_lookup ().end ()) {
```
zekemorton, line 703: This handles an unset or exhausted iterator, but the iterator can still dangle if the current span is removed. The next call would then compare an invalid iterator with `end()`. Could the erase paths reset the cursor before erasing the selected span, like `planner::erase()` now does?

resource/planner/c++/planner.cpp
```diff
@@ -556,9 +530,28 @@ planner_t::planner_t (const planner &o)
...
         errno = ENOMEM;
+        throw;
     }
 }
 
+// Deep copy; o.plan is non-null by the constructors' invariant.
+planner_t::planner_t (const planner_t &o)
+{
+    try {
+        plan = new planner (*o.plan);
```
zekemorton: **Blocking:** This new deep-copy path relies on `planner::copy_trees()`, which does not preserve minimum-time-tree membership during an active availability iteration. Tracked candidate points have been removed from that tree and have `in_mt_resource_tree == 0`, but `copy_trees()` copies the flag and then unconditionally inserts every point into the destination tree, setting it true. `copy_maps()` also recreates the availability tracker, so restoration can try to insert an already-linked intrusive node.

Could `copy_trees()` insert a point into `m_mt_resource_tree` only when the source point is a member? A regression test should start availability iteration, copy before exhaustion, and independently continue and reset or destroy both planners.

resource/planner/c++/planner_multi.cpp
```diff
@@ -56,73 +58,24 @@ planner_multi::planner_multi (int64_t base_time,
...
-        } else {
-            try {
-                np = new planner_t ();
-            } catch (std::bad_alloc &e) {
-                errno = ENOMEM;
-                throw std::bad_alloc ();
-            }
+    try {
+        for (const auto &iter : o.m_types_totals_planners) {
+            auto np = iter.planner ? std::make_unique<planner_t> (*(iter.planner->plan))
```
zekemorton, line 63: Should the regular constructor at lines 48–54 use the same unique pointer either here or in a follow-up? It still holds the new planner in a raw pointer while inserting into two containers, so an exception—or a rejected duplicate resource type—can leak it.

resource/planner/c++/planner.cpp
zekemorton, line 404: The new copy constructor can copy a planner during an active availability iteration. Some points are temporarily absent from `m_mt_resource_tree` in that state, but this inserts them into the copy unconditionally. Could we add a check to preserve their current state?

  ```suggestion
              if (point->in_mt_resource_tree
                 && (rc = m_mt_resource_tree.insert (new_point)) != 0)
  ```

zekemorton (approved):
Overall, this looks really good and is a major improvement. I did another pass with help from Claude and Codex and left a few inline comments for consideration.

jameshcorbett (approved):
Have looked over again and am still good with this!

PR #1541, opened 2026-07-28
traverser: support fractional durations

jameshcorbett:
Problem: submitting a jobspec with a fractional duration
(attributes.system.duration) is mishandled. jobmeta_t::build ()
converts the double duration into the planner's integer-second
representation with a C-style cast, which truncates toward zero. This
silently discards the fractional part of any non-integer duration
(e.g. 1.9 becomes 1). In the sub-second case it is fatal: a duration
such as 0.1 truncates to 0, which the planner rejects with EINVAL
(it requires duration >= 1), surfacing as a "match failed ... Invalid
argument" error. An integer duration such as 1 works.

Round the duration up with std::ceil () so a fractional duration is
never truncated away and any duration in (0, 1] maps to the planner's
minimum granularity of 1 second rather than to 0.

### DIFF
--- a/resource/traversers/dfu_impl.hpp
+++ b/resource/traversers/dfu_impl.hpp
@@ -13,6 +13,7 @@
 
 #include <cstdlib>
 #include <cstdint>
+#include <cmath>
 #include <memory>
 #include "resource/libjobspec/jobspec.hpp"
 #include "resource/config/system_defaults.hpp"
@@ -72,7 +73,8 @@ struct jobmeta_t {
                                  graph_duration.graph_end - graph_duration.graph_start)
                                  .count ();
 
-        if (g_duration <= 0) {
+        // check for valid durations, use ! >= to catch NaN
+        if (g_duration <= 0 || !(jobspec.attributes.system.duration >= 0)) {
             errno = EINVAL;
             return -1;
         }
@@ -80,14 +82,17 @@ struct jobmeta_t {
         // int64_t max () for comparison with at in dfu_traverser_t::run
         if ((jobspec.attributes.system.duration > static_cast<double> (g_duration))
             || (jobspec.attributes.system.duration
-                > static_cast<double> (std::numeric_limits<int64_t>::max ()))) {
+                > static_cast<double> (std::numeric_limits<int64_t>::max () - 1))) {
             errno = EINVAL;
             return -1;
         }
         if (jobspec.attributes.system.duration == 0.0f) {
             duration = g_duration;
         } else {
-            duration = (int64_t)jobspec.attributes.system.duration;
+            // Round up to the nearest second: better to pad out the duration a tiny bit, and
+            // a fractional duration in (0, 1) must not truncate to 0 (which the planner rejects
+            // with EINVAL)
+            duration = static_cast<int64_t> (std::ceil (jobspec.attributes.system.duration));
         }
         if (jobspec.attributes.system.queue != "") {
             m_queue = jobspec.attributes.system.queue;
--- a/resource/traversers/test/jobmeta_test.cpp
+++ b/resource/traversers/test/jobmeta_test.cpp
@@ -1,0 +1,180 @@
+/*****************************************************************************\
+ * Copyright 2026 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, LICENSE)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\*****************************************************************************/
+
+extern "C" {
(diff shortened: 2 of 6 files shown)


resource/traversers/dfu_impl.hpp
```diff
@@ -87,7 +88,9 @@ struct jobmeta_t {
         if (jobspec.attributes.system.duration == 0.0f) {
             duration = g_duration;
         } else {
-            duration = (int64_t)jobspec.attributes.system.duration;
+            // Round up to the nearest second: better to pad out the duration a tiny bit, and
+            // duration < 0 must not truncate to 0 (which the planner rejects with EINVAL)
+            duration = static_cast<int64_t> (std::ceil (jobspec.attributes.system.duration));
```
milroy, line 95: Safer from an IEEE754 perspective to check for NaN and casting UB:

```suggestion
            float_dur = jobspec.attributes.system.duration;
            constexpr double int64_hi = 0x1p63; // 0x1p63 is a hexadecimal float literal (C++17+) for exactly 2^63
            if (float_dur >= 0 && float_dur < int64_hi) { // >= 0 is an ordered comparison, so it's already false for NaN
                duration = static_cast<int64_t> (std::ceil (float_dur));
            } else {
                errno = EINVAL;
                return -1;
            }
```

t/data/resource/jobspecs/duration/test003.yaml
```diff
@@ -0,0 +1,30 @@
...
+                      - type: socket
+                        count: 1
+                        with:
+                          - type: core
+                            count: 1
+# a sub-second (fractional) duration must round up to 1 second rather
+# than truncate to 0, which the planner rejects with EINVAL
+attributes:
+  system:
+    duration: 0.1
```
milroy, line 25: Given the previous suggestion, may want to add a test with `duration > std::numeric_limits<int64_t>::max ()` and `duration < std::numeric_limits<int64_t>::min ()`.

milroy (commented):
This looks good already, but we do need to guard against exceeding expressible limits.

t/data/resource/jobspecs/duration/test003.yaml
trws: This is a good idea. Since this is yaml though, it needs to be a floating point number rather than an integer at those scales, anything over a 32-bit int is allowed to fail to parse per the yaml/json specs.

resource/traversers/dfu_impl.hpp
jameshcorbett: The code already has a check for `jobspec.attributes.system.duration > static_cast<double> (std::numeric_limits<int64_t>::max ()`, I added an additional check for `!(jobspec.attributes.system.duration >= 0)` to catch negatives and NaN.

t/data/resource/jobspecs/duration/test003.yaml
jameshcorbett: I added tests, please re-review when you have time.

resource/traversers/test/jobmeta_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+TEST_CASE ("NaN duration is rejected with EINVAL", "[jobmeta]")
+{
+    jobmeta_t m;
+    auto js = make_jobspec (std::numeric_limits<double>::quiet_NaN ());
+    auto gd = default_graph_duration ();
+    errno = 0;
+    REQUIRE (m.build (js, alloc_type_t::AT_ALLOC, 1, 0, gd) == -1);
+    REQUIRE (errno == EINVAL);
+}
+
```
trws, line 117: This reminded me for some reason, we should catch it because of the greater than int-max check, but probably worth adding a version of this for inf.


```suggestion
TEST_CASE ("NaN duration is rejected with EINVAL", "[jobmeta]")
{
    jobmeta_t m;
    auto js = make_jobspec (std::numeric_limits<double>::quiet_NaN ());
    auto gd = default_graph_duration ();
    errno = 0;
    REQUIRE (m.build (js, alloc_type_t::AT_ALLOC, 1, 0, gd) == -1);
    REQUIRE (errno == EINVAL);
}

TEST_CASE ("inf duration is rejected with EINVAL", "[jobmeta]")
{
    jobmeta_t m;
    auto js = make_jobspec (std::numeric_limits<double>::infinity ());
    auto gd = default_graph_duration ();
    errno = 0;
    REQUIRE (m.build (js, alloc_type_t::AT_ALLOC, 1, 0, gd) == -1);
    REQUIRE (errno == EINVAL);
}
```

jameshcorbett: Done and force-pushed.

trws: Excellent, thanks!

trws (approved):

PR #1542, opened 2026-07-30
traverser: skip expansion when next is empty

washwor1:
@milroy and I noticed that a bare `node[1]` jobspec with the firstnodex match policy produced 28 traversal events in resource-query for its first match on a Tuolumne-size resource graph, compared with only 6 for `node[1]` with `core[1]`, despite the bare node already satisfying the request. When the node matched, `test()` returned an empty next vector, but `dom_dfv()` still called `dom_exp()`, which interpreted that empty vector by iterating over every outgoing core edge. Each core failed to match but still incremented the preorder counter, and the normal early-exit logic was never reached because it only applies after a successful child match. The fix changes the unconditional `else` to `else if (!next.empty())`, so dynamic expansion occurs only when the matcher actually requests another resource level. The one-node `firstnodex` case now requires 4 traversal events instead of 28, and larger bare-node requests show similarly large traversal reductions without changing which jobs match. There was also a positive effect on other match policies with `lonodex` seeing 3818 pre-order visits in traversal for the first match, down from 119042 before the change. 

This change results in a significant performance improvement for the bare `node[1]` jobspec:

Before Patch:
image

After patch:
image

### DIFF
--- a/resource/traversers/dfu_impl.cpp
+++ b/resource/traversers/dfu_impl.cpp
@@ -808,7 +808,7 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta,
     (*m_graph)[u].idata.colors[dom] = m_color.gray ();
     if (sm == match_kind_t::SLOT_MATCH)
         dom_slot (meta, u, next, nslots, check_pres, &x_inout, dfu);
-    else
+    else if (!next.empty ())
         dom_exp (meta, u, next, check_pres, &x_inout, dfu);
     *excl = x_in;
     (*m_graph)[u].idata.colors[dom] = m_color.black ();


jameshcorbett (commented):
This looks reasonable, the commit needs a message though.

resource/traversers/dfu_impl.cpp
```diff
@@ -808,7 +808,7 @@ int dfu_impl_t::dom_dfv (const jobmeta_t &meta,
     (*m_graph)[u].idata.colors[dom] = m_color.gray ();
     if (sm == match_kind_t::SLOT_MATCH)
         dom_slot (meta, u, next, nslots, check_pres, &x_inout, dfu);
-    else
+    else if (!next.empty ())
```
trws, line 811: I would have guessed this might have issues with exclusivity, if a core in the node is already allocated for example, but everything passes.  If you would, I would really appreciate a test that allocates a single core from a one-node topology, then tries to allocate a bare node. I very much like the performance improvement, but am a bit wary when it comes to pruning out checks in this area of the code.

Issue #1536, opened 2026-07-22
qmanager: support RFC 33 virtual queues


grondo:
[Virtual queues](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html#virtual-queues) (vqueues) were recently added to RFC 33. In summary: a queue with a `parent` key inherits its parent queue's resources and policy, and its jobs SHALL be scheduled as part of the parent queue in a single priority order.

Fluxion qmanager currently creates one internal queue per `[queues]` entry and assigns jobs by the jobspec `queue` attribute. Under the RFC 33 amendment, a vqueue must not get its own internal queue: jobs whose queue is a vqueue belong in the parent queue's internal queue. Fluxion will need to check for a `parent` key in a queue and effectively treat jobs with the vqueue name as belonging to the parent.

Notes:
- qmanager already reads `[queues]` config directly (commented in-tree as a "workaround to satisfy RFC 33"), so it can read `parent` the same way for now. Longer-term the queue set will be delivered by the job-manager (to support runtime queues flux-framework/flux-core#7737), which removes the config read entirely.
- Per-queue scheduler policy (`queue-policy-per-queue`) cannot be set for a vqueue (the RFC forbids `policy.scheduler` on vqueues); the parent's policy governs the shared internal queue.

PR #1543, opened 2026-07-31
qmanager: support RFC 33 virtual queues

grondo:
This adds support for RFC 33 virtual queues to qmanager, closing #1536.

A virtual queue is declared by a `parent` key in a `[queues]` entry: an alternate submission name over the parent queue's resources with different policy at ingest. Jobs keep the virtual queue name, but must be scheduled as part of the parent queue's single priority order. Without this change, qmanager would create a separate internal queue for each virtual queue.

`process_config_file()` now detects `parent` keys in the `[queues]` table, excludes virtual queues from the internal queue list, and records a vqueue->parent map that `resolve_queue_name()` consults in the hello and alloc callbacks to route a virtual queue's jobs to the parent's internal queue.

The vqueue->parent map travels into `qmanager_opts_t` as a synthesized `vqueue-parents` option, following the precedent of the synthesized `queues` option that already carries the internal queue list derived from the same `[queues]` table. This keeps both halves of the derived queue topology in one place and follows existing code precedent. If extending that pattern is not acceptable for the vqueues support, an alternative is to hold the map on the qmanager context directly with a plain lookup call in the callbacks. This would seem to break the options abstraction but removes the synthetic option used only to carry vqueues; happy to implement that instead if preferred (understandable because it would eliminate a lot of glue code I think).

Since `vqueue-parents` is only ever populated from a config table containing `parent` keys, this is a strict no-op with any flux-core that predates virtual queues. Associated tests will be skipped until the flux-core PR is merged.

### DIFF
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -141,13 +141,34 @@ static int process_config_file (std::shared_ptr<qmanager_ctx_t> &ctx)
     optmgr_kv_t<qmanager_opts_t> opts_store;
     std::string info_str = "";
     if (queues_conf) {
-        // workaround to satisfy RFC 33
-        std::string queues = classify_queues (queues_conf);
+        // workaround to satisfy RFC 33: split the queues config table into
+        // real (internal) queues and RFC 33 virtual queues. A malformed
+        // config is rejected here. Because ctx->opts is only updated on
+        // success at the end of this function, returning an error leaves
+        // the currently-loaded config in place -- matching flux-core,
+        // where a failed config reload is reported via the reload RPC and
+        // the running config is retained.
+        std::string queues;
+        std::string vqueue_parents;
+        std::string err;
+
+        if (classify_queues (queues_conf, queues, vqueue_parents, err) < 0) {
+            flux_log_error (ctx->h, "%s: %s", __FUNCTION__, err.c_str ());
+            return -1;
+        }
         if ((rc = opts_store.put ("queues", queues)) < 0) {
             flux_log_error (ctx->h,
                             "%s: optmgr_kv_t::put ('queues', %s)",
                             __FUNCTION__,
                             queues.c_str ());
+            return rc;
+        }
+        if (!vqueue_parents.empty ()
+            && (rc = opts_store.put ("vqueue-parents", vqueue_parents)) < 0) {
+            flux_log_error (ctx->h,
+                            "%s: optmgr_kv_t::put ('vqueue-parents', %s)",
+                            __FUNCTION__,
+                            vqueue_parents.c_str ());
             return rc;
         }
     }
--- a/qmanager/modules/qmanager.cpp
+++ b/qmanager/modules/qmanager.cpp
@@ -142,15 +142,12 @@ static int process_config_file (std::shared_ptr<qmanager_ctx_t> &ctx)
     std::string info_str = "";
     if (queues_conf) {
         // workaround to satisfy RFC 33
-        std::ostringstream queues;
-        json_object_foreach (queues_conf, k, v) {
-            queues << std::string (k) << " ";
-        }
-        if ((rc = opts_store.put ("queues", queues.str ())) < 0) {
+        std::string queues = classify_queues (queues_conf);
+        if ((rc = opts_store.put ("queues", queues)) < 0) {
             flux_log_error (ctx->h,
                             "%s: optmgr_kv_t::put ('queues', %s)",
                             __FUNCTION__,
-                            queues.str ().c_str ());
+                            queues.c_str ());
             return rc;
         }
     }
--- a/qmanager/modules/qmanager_opts.cpp
+++ b/qmanager/modules/qmanager_opts.cpp
@@ -14,10 +14,24 @@ extern "C" {
 #endif
 }
 
+#include <sstream>
+
 #include "qmanager_opts.hpp"
 
 using namespace Flux;
 using namespace Flux::opts_manager;
+
+std::string Flux::opts_manager::classify_queues (json_t *queues_conf)
+{
+    const char *k = nullptr;
+    json_t *v = nullptr;
+    std::ostringstream queues;
+
+    json_object_foreach (queues_conf, k, v) {
+        queues << k << " ";
+    }
+    return queues.str ();
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 // Private API for Queue Manager Option Class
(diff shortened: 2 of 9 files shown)


t/t1032-qmanager-vqueue.t
```diff
@@ -0,0 +1,148 @@
...
+test_expect_success HAVE_VQUEUES 'qmanager: job can be submitted to virtual queue' '
+	flux run -n1 --queue=expedite true
+'
+
+test_expect_success HAVE_VQUEUES 'qmanager: virtual queue job did not create a phantom queue' '
+	flux module stats sched-fluxion-qmanager >stats2.out &&
+	test_must_fail jq -e ".queues.expedite" stats2.out
+'
+
+test_expect_success HAVE_VQUEUES 'qmanager: fill the parent queue with a running job' '
```
jameshcorbett: maybe remove all the `qmanager:` prefixes from the test cases? Not sure what purpose they serve.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -336,6 +361,11 @@ int qmanager_opts_t::parse (const std::string &k, const std::string &v, std::str
             break;
 
         case static_cast<int> (qmanager_opts_key_t::QUEUE_POLICY_PER_QUEUE):
+            // m_per_queue_prop only contains queues that have an internal
+            // queue (i.e., RFC 33 virtual queues are excluded), so this
+            // also rejects a queue-policy-per-queue entry that names a
+            // virtual queue, since a virtual queue cannot have its own
+            // scheduler policy.
```
jameshcorbett, line 412: worth a test in `qmanager_opts_test.cpp`?

jameshcorbett (approved):
LGTM!

t/t1032-qmanager-vqueue.t
grondo: This is following the pattern of labeling in other sharness tests. It sometimes helps when reading the output of 'make check', but agreed it probably isn't necessary here.

qmanager/modules/qmanager_opts.cpp
grondo: Yes! Will add.

qmanager/modules/test/qmanager_opts_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+TEST_CASE ("resolve_queue_name maps a virtual queue to its parent", "[qmanager_opts]")
+{
+    qmanager_opts_t opts;
+    std::string info;
+
+    REQUIRE (opts.parse ("vqueue-parents", "expedite:batch", info) == 0);
+    REQUIRE (opts.resolve_queue_name ("expedite") == "batch");
+    REQUIRE (opts.resolve_queue_name ("batch") == "batch");
+    REQUIRE (opts.resolve_queue_name ("debug") == "debug");
```
trws: Thank you for using catch for these!  As an intent question, do you want all of these to stop the test on the first failure, or should it report as many problems as it finds in one run? `REQUIRE` will stop the test immediately, like an assert, where `CHECK` will report every bad result and continue.

qmanager/modules/qmanager.cpp
```diff
@@ -143,8 +143,31 @@ static int process_config_file (std::shared_ptr<qmanager_ctx_t> &ctx)
...
+            json_t *parent = json_object_get (v, "parent");
+            if (parent && json_is_string (parent))
+                candidates[k] = json_string_value (parent);
+        }
+        // Second pass: a candidate is only a valid RFC 33 virtual queue if
+        // its named parent is a configured queue that is not itself a
+        // virtual queue (a parent is never itself virtual -- flux-core
+        // validation enforces this, but tolerate malformed config here by
+        // falling back to treating the entry as a real queue instead of
+        // crashing).
```
trws: Does core tolerate this case or stop?  I'm vaguely in favor of failing to load if the config is invalid rather than continuing at the moment.

qmanager/modules/qmanager_opts.cpp
```diff
@@ -48,6 +48,19 @@ int qmanager_opts_t::parse_queues (const std::string &queues)
...
+    int rc = 0;
+    try {
+        m_vqueue_parents.clear ();
+        rc = parse_multi_options (vqueue_parents, ' ', ':', m_vqueue_parents);
+    } catch (std::bad_alloc &e) {
+        errno = ENOMEM;
+        rc = -1;
+    }
+    return rc;
+}
```
trws: Is it valid to continue if the string or map couldn't be allocated here? If not, allowing the bad_alloc to propagate up to the module main might make sense.  Alternately, is this the only exception that can throw in parse_multi_options?

trws (commented):
A few questions, overall looks good. 👍

qmanager/modules/qmanager.cpp
grondo: Core doesn't tolerate it. On initial startup the instance will fail to start with an invalid config. On reload though, the config is rejected and the previous config stays active.

Your preference sound reasonable to me. I'll make that change.

qmanager/modules/test/qmanager_opts_test.cpp
grondo: The intent is to keep going.  I just don't know Catch well enough to know the difference. Will fix.

qmanager/modules/qmanager_opts.cpp
grondo: `parse_multi_options` already catches `bad_alloc` so it should have just been propagated here.

qmanager/modules/qmanager.cpp
grondo: This has been fixed in the latest push. A new `classify_queues()` function for traversing the `[queue]` config and marking vqueues->parent is now included, so failure can be unit tested. Since flux-core rejects bad config, this was the best way to get coverage.

A new commit on top first extracts the current code into `classify_queues()` which will hopefully make the re-review easier.

grondo:
I pushed a new version of this branch to address comments above:
1. queue classification is now pulled out into a `classify_queues()` function that can be easily unit tested
2. bad config is rejected outright with human readable error message (which is then logged)
3. Removed the unnecessary bad_alloc catch

This could probably use a re-review. We'll wait to merge until the vqueue support PR lands in flux-core anyway.

qmanager/modules/qmanager.cpp
trws: That sounds great, thanks!

trws (approved):
This looks great, thanks @grondo!

grondo:
The actual sharness tests will skipped here until the vqueues support lands in flux-core. I say we wait to merge until that happens.

grondo:
Ok, vqueues support was merged in flux-core, so I'll set MWP here now.

Issue #1544, opened 2026-08-07
Rabbit jobs unsatisfiable on ElCap; degradation of state


jameshcorbett:
Offline, @behlendorf noted:

> there's a ~4000 node DAT on El Cap today which is trying to use the Rabbits. When requesting a ~4000 nodes jobs which requires rabbits it's failing with "Unsatisfiable", without rabbits it allocates fine. @kkier mentioned this is the same symptom as https://github.com/flux-framework/flux-sched/issues/1503 which should be fixed in this release. The Rabbit's themselves look healthy.

I hypothesized that this could be due to the `badrabbit` property--If there are exactly 4k nodes in the DAT queue and the job requests exactly 4k nodes, but there is at least one node with the badrabbit property in that queue, the job would be rejected as unsatisfiable. However, @kkier effectively ruled out this hypothesis when he noted that only 16 nodes in the queue had the `badrabbit` property and the largest jobs were 256 nodes smaller than the queue size.

@behlendorf noted it got worse as he was trying to run and even smaller jobs started failing: some just sat in `SCHED` forever [@kkier verified after the fact that the `priority` event was posted a few minutes before the job was canceled) and some failed with unsatisfiable, even at 1k scale.

This behavior is consistent with what I noted during a previous TOSS update on ElCap: larger jobs were rejected outright and smaller ones were stuck in SCHED, with the state degrading as more jobs ran.

@grondo had a question that he wasn't able to ask on the video call, so tagging him.

trws:
Ok, clearly we have some digging to do here.  The sitting in sched forever issue would normally mean something like there being too many elements in `down` state to schedule the job, or otherwise held onto.  It would be good to get a snapshot of the state of the agfilters and rabbits if we can, I think `find` can do that.  Anything that could let us see if there's some inconsistent state would be good, because while the hung sched state could be things being down, the unsatisfiable result should only be if there flat isn't enough hardware accessible at this point.

milroy:
Quick status update: my investigations into the reported behaviors are pointing to a bug in partial cancel. I'll make a full post ASAP with tests and findings.

milroy:
The partial cancel bug explains the reduced allocation space that caused ever smaller job requests to fail. It has a very simple reproducer.

`rabbit-test.yaml`:
```yaml
version: 9999
resources:
  - type: slot
    count: 2
    label: rabbit
    with:
    - type: node
      count: 1
      exclusive: true
      with:
      - type: slot
        label: task
[...]

PR #1551, opened 2026-08-13
Sweep vertices by jobid to ensure partial cancel correctness

milroy:
A recent incident on El Capitan (#1544) featured a complex interaction between unsatisfiable job requests and collapsing scheduler state. Debugging identified a cause of the behavior: unsuccessful deallocation of SSD vertices after job cancellation/completion (see #1544 for reproducer).

The current problem us that vertices never named in a freed R fragment (e.g., rankless `ssd`s, ranked-but-never-freed rabbit vertices) permanently stranded their per-job state, because release completeness depended on rank indexes, a tag path that partial cancels sever due to assumptions about agflters tracking available counts. This causes the final cleanup cancel issued during a `.free` to fail to remove job metadata from low-level resources.

This PR fixes the reproducer and improves regression and CI testing by introducing a `by_jobid` map that associates an `unordered_set` of vertices with their jobid. This container allows full cancel to clear job metatada via `by_jobid[jobid] `(the tag-pruned DFS `mod_dfv()`.

   #### Rough estimate of memory footprint of `by_jobid`: 
`by_jobid` stores a 16 byte hash node + about 8 byte amortized bucket slot per (jobid, vertex) pair plus about 120–220 bytes for one outer map node + the set object. That could be about 40 bytes/pair resident, with the fixed cost of the empty container amortized by a larger job.

For comparison, Fluxion already stores roughly 1000 bytes per pair resident: red-black-tree map entries (allocations/reservations, tags, x_spans, job2span at about 64 bytes each) plus at least two planner spans with their lookup entries (about 200 bytes each), plus the scheduled points (fudge factor). The index therefore represents a single-digit percent overhead beyond what the graph already has to store. 

For a 4,000 node job:
  - by_jobid: 400k (4,000 nodes and 100 vertices per node, ignoring negligible ancestors and ssds) × 40 bytes is approximately 16 MB resident
  - Existing job-keyed state: 400k × 1000 bytes is about 400 MB.

~This PR is WIP as I need to double-check a few implementation details generated by Claude Fable 5 and do a human rewrite of the commit messages.~


milroy:
Please hold off on initial reviews. I've discovered a few minor problems that Claude and I need to correct. I'll add a comment when the PR is ready for an informal review.

milroy:
Ok, Claude and I added more thorough reconstruction tests and fixed a problem in the initial implementation where a failed cancel clears a jobid from `by_jobid`.

Note that the justification for a set container is visible in the need for adding a vertex to the map in both the reader and traverser. I don't think `std::bitset` will work here because its size is fixed. The justification for `unordered_set` comes from its expected `O(1)` erase by value for integers binned by a prime modulus.

This is ready for an initial review.

trws (approved):
I like the way this turned out, and the testing on it.  Everything here is good.  My one meaningful concern is that it would be pretty easy for a new failure of a similar type, something missing an `add_job_vertex` call, to sneak in without us noticing.

Given that this fixes the issue, I don't think it's a blocker for moving this forward necessarily, but in the immediate future I think we might do well to add something for regular sanity checking.  I'm debating a bit whether it should be something a user can invoke to check for consistency, and we register checks to it to report different things, or if we want something that runs regularly on a timer, but some way that we can invoke `verify_job_index` on a production system to get a sanity check seems like something we're going to want.

jameshcorbett (approved):
I can't find anything wrong with this

milroy:
> it would be pretty easy for a new failure of a similar type, something missing an add_job_vertex call, to sneak in without us noticing.

That is definitely a risk.

> some way that we can invoke verify_job_index on a production system to get a sanity check seems like something we're going to want.

Agreed. We need a check that's more powerful, less verbose, and less prone to misinterpretation than `find`.

milroy:
Claude and I added two more tests for general storage topologies to guard against regressions that are not specifically rabbits but are exemplified by potential rabbit configurations: storage under a rack vertex, and an ssd collection with no rabbit vertex.

I think this is good to merge.

milroy:
Thanks for the reviews @trws and @jameshcorbett!

Issue #1472, opened 2026-05-05
at "initialization" when a bunch of jobs are queued 1-2 extremely large messages are transmitted by sched-fluxion-resource


wihobbs:
Perhaps this is expected behavior (feel free to close if so), but as part of my testing for #1471, I noticed that when I queued 16,000 jobs, Fluxion transmits two extremely large messages that block before any jobs can be scheduled. 
```
(s=1,d=0)  hetchy1005 ~/content-test $ seq 1 16000 | flux bulksubmit --quiet -N 40 --setattr=exec.test.run_duration=\"1\" hostname &
[1] 190165
(s=1,d=0)  hetchy1005 ~/content-test $ flux module trace sched-fluxion-resource
2026-05-05T15:00:34.749 sched-fluxion-resource tx > log.append [117]
2026-05-05T15:00:34.774 sched-fluxion-resource tx > log.append [137]
2026-05-05T15:00:37.796 sched-fluxion-resource tx > log.append [121]
2026-05-05T15:00:38.258 sched-fluxion-resource tx > log.append [123]
2026-05-05T15:00:38.259 sched-fluxion-resource tx > log.append [110]
2026-05-05T15:00:38.259 sched-fluxion-resource tx > module.status [13]
# these are the ? ones
2026-05-05T15:01:02.355 sched-fluxion-resource tx < sched-fluxion-resource.notify [184.35091M] success
2026-05-05T15:01:02.355 sched-fluxion-resource tx < sched-fluxion-resource.notify [35] success
2026-05-05T15:01:02.500 sched-fluxion-resource rx > sched-fluxion-resource.match_multi [828]
2026-05-05T15:01:29.492 sched-fluxion-resource tx < sched-fluxion-resource.notify [184.35091M] success
2026-05-05T15:01:29.492 sched-fluxion-resource tx < sched-fluxion-resource.notify [35] success
# barage of match_multis begins here
2026-05-05T15:01:32.757 sched-fluxion-resource tx < sched-fluxion-resource.match_multi [12.509766K] success
2026-05-05T15:01:32.757 sched-fluxion-resource tx < sched-fluxion-resource.match_multi [0] ENODATA
2026-05-05T15:01:32.776 sched-fluxion-resource rx > sched-fluxion-resource.match_multi [828]
2026-05-05T15:01:32.842 sched-fluxion-resource tx < sched-fluxion-resource.match_multi [13.314453K] success
[...]

PR #1552, opened 2026-08-26
modules: Prevent unnecessary resource graph transmissions

jacobtkeio:
Problem: As described by #1472, the resource module currently sends the entire resource graph to any module that makes a `sched-fluxion-resource.notify` RPC request, whether that module needs the graph or not. This behavior was introduced by #1285 as a side effect of getting resource graph information into the feasibility module. This causes a significant extra delay on fluxion instance initialization when the graph is large enough.

Add an API for encoding and decoding desired `.notify` fields in flags like `NOTIFY_RESOURCES`, `NOTIFY_DOWN`.
Let modules use these flags to specify what data they want from `.notify`. Respond to `.notify` requests only with the requester's specified fields.
Specify `NOTIFY_NONE` in qmanager.


jacobtkeio:
To more naturally express the find "all resources" query, I've tried adding some regex support to the `names` constraint under Dan and Tom's advisement. Now, the resource module can run `find ("names=.*")` to get the set of all resources to send through `.notify` RPCs. It uses the extended POSIX regex grammar sans square-bracket expressions, which are currently reserved for the optional idlist (which is nice to have because regex cannot handle proper number ranges like `11-33`). See the end of t3025 for some usage examples.

@jameshcorbett No pressure to reply, but I think I heard you were once interested in regex resource queries. Is this the sort of thing you had in mind? Is not supporting square brackets a deal breaker?

I'll be happy to split the regex change into a new PR if anyone would prefer it!

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+    const std::string prefix = brackets.prefix ().str ();
+    const std::string suffix = brackets.suffix ().str ();
+
+    // Count '('s to calculate the submatch index representing the hostlist brackets.
+    const size_t submatch_idx = 1 + std::count (prefix.cbegin (), prefix.cend (), '(');
+
+    // Check if name matches the regex string, but ignore everything in the brackets.
+    // Example lcx becomes: "core1[0-9+].*"
+    // The example name matches the new regex, so we continue.
+    if (!std::regex_match (name,
```
vsoch, line 103: In Python it's possible to compile a regex and then use many times. I'm not sure if you are compiling it here for each call to the function, and if there might be performance implications to that. I maintain a library called deid for medical images  (dicom) and a contributor did this one tweak and the library got *hugely* faster. I would be interested in the performance of compiling up to two of these objects per vertex on every find, versus maybe compiling elsewhere and using across operations. I'm also curious about the pattern itself - if doing `.*` over 100k vertices is efficient, and if we could compile the regex maybe on init or validate and then cache it.

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+    std::string hlist_name = name;
+
+    // Example name: "core11"
+    // Example lcx: "core1[0].*"
+    // Check if the regex string contains brackets. Otherwise, it's a simple regex match.
+    // The example name contains brackets, so we continue.
+    if (!std::regex_search (lcx, brackets, std::regex ("\\[.*\\]"))) {
+        result = std::regex_match (name, std::regex (lcx, std::regex_constants::extended));
+        return 0;
+    }
```
vsoch, line 92: This matching schema maybe should be another pull request? I'm wondering if we can still match `names=core1,core2` or `core1,core2[0-3]`. I think we'd also want to add these to test cases, or if for C++ there is an existing library or algorithm that can reliably parse different variants of hostlist.

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+
+    const std::string prefix = brackets.prefix ().str ();
+    const std::string suffix = brackets.suffix ().str ();
+
+    // Count '('s to calculate the submatch index representing the hostlist brackets.
+    const size_t submatch_idx = 1 + std::count (prefix.cbegin (), prefix.cend (), '(');
```
vsoch, line 98: What happens if there is a paren in the target string here as a literal character? Would that break the intended functionality by returning the wrong index? And is that even a possible case? Please disregard if not.

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+
+    // Check if name matches the regex string, but ignore everything in the brackets.
+    // Example lcx becomes: "core1[0-9+].*"
```
vsoch, line 101:
```suggestion
    // Example lcx becomes: "core1([0-9]+).*"
```

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+        result = false;
+        return 0;
+    }
+
+    // Generate hlist_name, the format for hostlist checking.
+    // Example hlist_name: "core1[0]"
+    hlist_name.replace (hostids.position (submatch_idx),
+                        hostids.length (submatch_idx),
+                        brackets.str ());
+    if (!(hlist = hostlist_decode (hlist_name.c_str ()))) {
```
vsoch, line 116: Do we need to set the errno here?

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -131,16 +184,12 @@ int expr_eval_vtx_target_t::evaluate (const std::string &p,
             result = false;
         }
     } else if (p == "names") {
-        if (!(hlist = hostlist_decode (lcx.c_str ()))) {
+        try {
+            rc = evaluate_regex_and_hostlist ((*m_g)[m_u].name, lcx, result);
+        } catch (std::regex_error e) {
```
vsoch, line 189: Is there a way to catch the specifics of the error here so it's not just "I got -1 and I don't know why" (speaking from personal experience...) 😆

resource/evaluators/expr_eval_vtx_target.cpp
```diff
@@ -71,14 +73,65 @@ int expr_eval_vtx_target_t::validate (const std::string &p, const std::string &x
...
+    // Generate hlist_name, the format for hostlist checking.
+    // Example hlist_name: "core1[0]"
+    hlist_name.replace (hostids.position (submatch_idx),
+                        hostids.length (submatch_idx),
+                        brackets.str ());
+    if (!(hlist = hostlist_decode (hlist_name.c_str ()))) {
+        return -1;
+    }
+
+    // Check if name matches hlist_name.
```
vsoch, line 120: For this entire function, I want to suggest we start from the explicit list of what is allowed. I don't think right now we can tell the difference between "nothing matched your query" vs. "your query is not supported." Does flux-core or flux-sched not already have a library to parse these strings?

Another thought - doesn't flux core have `flux/hostlist.h` ? Maybe we can figure out the missing functionality, add it there, and then use it here. If these operations are needed in other parts of flux it is nice to have all that in one place.

resource/modules/resource.cpp
```diff
@@ -1222,6 +1227,15 @@ static void notify_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
...
 
+        // Traverse all nodes to calculate the current resource set
+        json_t *resource_graph_json = nullptr;
+        if (m->get_notify_flags () & NOTIFY_RESOURCES) {
+            if (run_find (ctx, "names=.*", "rv1", &resource_graph_json) < 0) {
+                flux_log_error (h, "%s: run_find", __FUNCTION__);
+                goto error;
+            }
+        }
+
```
vsoch, line 1238: Just wondering on like 1243 if we intend to copy the entire Boost resource graph on each find request? E.g., could we do:

```cpp
auto &rg = ctx->db->resource_graph;
```

resource/modules/resource_match.cpp
```diff
@@ -222,7 +228,19 @@ static void update_resource (flux_future_t *f, void *arg)
 
     std::shared_ptr<resource_ctx_t> &ctx = *(static_cast<std::shared_ptr<resource_ctx_t> *> (arg));
 
-    if ((rc = flux_rpc_get_raw (f, (const void **)&acquire_str, &acquire_str_size)) < 0) {
+    if ((rc = flux_rpc_get_unpack (f,
```
vsoch, line 231: Would this make the module exit on a decode failure?

resource/modules/resource_notify.hpp
```diff
@@ -0,0 +1,86 @@
...
+                      NOTIFY_DOWN_KEY,
+                      flags & NOTIFY_DOWN,
+                      NOTIFY_SHRINK_KEY,
+                      flags & NOTIFY_SHRINK,
+                      NOTIFY_EXPIRATION_KEY,
+                      flags & NOTIFY_EXPIRATION);
+}
+
+/*! Decode notification flags from a json object.
+ *  /param json     Input json object.
```
vsoch, line 56:
```suggestion
 *  \param json     Input json object.
```

resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -371,6 +372,49 @@ TEST_CASE ("Test the graph idempotence of certain match operations", "[match C++
     }
 }
 
+TEST_CASE ("Test the graph idempotence of loading from the output of find()", "[find C++]")
+{
+    const std::string options = "{\"load_format\": \"jgf\"}";
```
vsoch, line 377: Do we want to test rv1?

resource/modules/resource.cpp
```diff
@@ -1258,43 +1271,34 @@ static void notify_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
...
+                                         ctx->db->metadata.graph_duration.graph_end)
+                                   : -1.)
             < 0) {
             flux_log_error (ctx->h, "%s: flux_respond_pack", __FUNCTION__);
             goto error;
         }
 
         free (up_str);
         free (down_str);
-        free (lost_str);
```
vsoch, line 1297: What happens if we jump to error and we don't run these, do we not free the memory?

resource/modules/resource_match.cpp
```diff
@@ -281,34 +269,28 @@ static void update_resource (flux_future_t *f, void *arg)
...
+                                   m->get_notify_flags () & NOTIFY_UP ? up : nullptr,
+                                   NOTIFY_DOWN_KEY,
+                                   m->get_notify_flags () & NOTIFY_DOWN ? down : nullptr,
+                                   NOTIFY_SHRINK_KEY,
+                                   m->get_notify_flags () & NOTIFY_SHRINK ? lost : nullptr,
+                                   NOTIFY_EXPIRATION_KEY,
+                                   m->get_notify_flags () & NOTIFY_EXPIRATION ? expiration : -1.)
                 < 0) {
-                flux_log_error (ctx->h, "%s: flux_respond_raw", __FUNCTION__);
+                flux_log_error (ctx->h, "%s: flux_respond_pack", __FUNCTION__);
```
vsoch, line 289: If there are a bunch of messages here and one is wonky, would we want to go to done, or could we set rc to -1 and allow the rest to process? But then are we adding the error code to rc, or the result of the comparison < 0, so if it is less than 0 we would actually get 1, and if rc was set to -1 the error would be undone I think? 

Do we need to update the rc there? Can we just do like and then not have to worry about the addition / subtraction?

```cpp
for (const auto &[_, m] : ctx->notify_msgs) {
    if (flux_respond_pack (ctx->h, m->get_msg (), ...) < 0) {
        flux_log_error (ctx->h, "%s: flux_respond_pack", __FUNCTION__);
        rc = -1;
    }
}
```

resource/reapi/test/reapi_cli_test_cxx.cpp
```diff
@@ -371,6 +372,49 @@ TEST_CASE ("Test the graph idempotence of certain match operations", "[match C++
...
+
+    std::shared_ptr<resource_query_t> ctx = nullptr;
+    ctx = std::make_shared<resource_query_t> (rgraph, options);
+    REQUIRE (ctx);
+
+    // Store the initial state of the graph
+    std::map<vtx_t, pool_infra_t> idata_map;
+    std::map<vtx_t, schedule_t> sched_map;
+    vtx_iterator_t u, end;
+    for (boost::tuples::tie (u, end) = boost::vertices (ctx->db->resource_graph); u != end; u++) {
```
vsoch, line 397: If we are looping based on the new graph index, what if the new graph has 100 vertices and the original had 116 - we are not checking them all? And if the new graph is larger won't we get an out of range error? Can we index on something else like the containment path?

resource/modules/resource.cpp
```diff
@@ -1201,11 +1203,14 @@ static void notify_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
         const char *route;
         std::shared_ptr<resource_ctx_t> ctx = getctx ((flux_t *)arg);
         std::shared_ptr<msg_wrap_t> m = std::make_shared<msg_wrap_t> ();
+        json_t *requested = NULL;
 
-        if (flux_request_decode (msg, NULL, NULL) < 0) {
-            flux_log_error (h, "%s: flux_request_decode", __FUNCTION__);
+        if (flux_request_unpack (msg, NULL, "{s?:o}", NOTIFY_REQUEST_KEY, &requested) < 0) {
```
vsoch, line 1208: What if you provide requested as a valid Json object here, but the structure (e.g., container type/type) is wrong. Do we silently return `NOTIFY_NONE`? E.g.,

```
{"requested": {"resources": 1}}
{"requested": ["resources", "shrink"]}
```

vsoch (commented):
Taking a look! :eyes: 

To walk through what I understand, the `sched-fluxion-resource.notify` is sending a massive payload (the entire resource graph) to every module that subscribes regardless of needing it or not. Hobbs reported a slowdown (which makes sense). This PR is saying "let's send only what we need, and when it is asked for" to de-chonkify the size. I'll add specific notes via comments, and put high level ones here. 

Your last commit has:

> Problem: No test exists for the behavior of `find ("exists=true")`, which should output the loaded graph verbatim.

I didn't find the exists=true anywhere - is the commit message referencing something that was changed or removed? Some other quick comments:

- Should we test a NOTIFY_NONE gets no resources key? I think we could use a Flux Python RPC test.
- Do we test add_subgraph and remove_subgraph per the original use case?
- Whatever we decide for hostlist formats (maybe in another PR) I think we should document it somewhere.

I am new to reviewing C++ so do not jump into any changes until @milroy has been able to take a look. I will learn from his review too. Thanks for your hard work @jacobtkeio! 👍

jacobtkeio:
After discussing it more based on your feedback, @milroy and I decided to remove the regex parsing and open an issue in flux-core to support a globstar hostname `*` that matches everything. In the meantime, I'll switch back to using `status=up or status=down` in this PR.
Thank you so much for your detailed review! I am looking forward to addressing all your comments :)

resource/evaluators/expr_eval_vtx_target.cpp
jacobtkeio: I just opened issue https://github.com/flux-framework/rfc/issues/551 to propose adding the "match anything" functionality there!

resource/modules/resource.cpp
jacobtkeio: Oops! I'll replace it with a const reference. Thanks for catching that!

resource/modules/resource_match.cpp
jacobtkeio: Yes. I think that's fine for the first decode, where a failure should immediately exit so the user can try something else to load the module correctly. On subsequent decodes, maybe it should be more lenient and just issue a warning? I'll add a commit for now and if we decide we don't need it later we can drop it.

jacobtkeio: The main argument in favor of failing no matter what is that if this decode fails we can't update the resource db and the receiving module will desynchronize with the sending module. In that case, shutting down and restarting the module to resynchronize may be the best option in all cases. @milroy what do you think? In production, is shutting modules down better than letting them desynchronize but stay active (in the context of sched-fluxion-resource and feasibility)?

resource/modules/resource.cpp
jacobtkeio: Indeed we do not! It looks like that error case is not covered by our tests, so we haven't been able to automatically detect the leak. I'll add this set of frees and destroys to the error case!

resource/modules/resource_match.cpp
jacobtkeio: I agree that's way better! Thanks :)

resource/reapi/test/reapi_cli_test_cxx.cpp
jacobtkeio: I've added a prerequisite check that both graphs have the same number of vertices. I think that should prevent any out of range errors and catch all size mismatches!

resource/modules/resource.cpp
jacobtkeio: Yes, these would both silently return `NOTIFY_NONE`. I see why you'd want to catch that. Do you think just logging the final, interpreted request once it's subscribed would be sufficient? The notify request would complete silently, but the user would get an `INFO` log saying what was actually requested, like 
```
2026-09-11T15:59:24.880661-07:00 sched-fluxion-resource.info[0]: received request for notifications: {
2026-09-11T15:59:24.880673-07:00 sched-fluxion-resource.info[0]:     "resources": false,
2026-09-11T15:59:24.880685-07:00 sched-fluxion-resource.info[0]:     "up": false,
2026-09-11T15:59:24.880688-07:00 sched-fluxion-resource.info[0]:     "down": false,
2026-09-11T15:59:24.880690-07:00 sched-fluxion-resource.info[0]:     "shrink": false,
2026-09-11T15:59:24.880692-07:00 sched-fluxion-resource.info[0]:     "expiration": false
2026-09-11T15:59:24.880695-07:00 sched-fluxion-resource.info[0]: }
```

resource/modules/resource.cpp
```diff
@@ -1222,6 +1227,15 @@ static void notify_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_
             goto error;
         }
 
+        // Traverse all nodes to calculate the current resource set
+        json_t *resource_graph_json = nullptr;
+        if (m->get_notify_flags () & NOTIFY_RESOURCES) {
+            if (run_find (ctx, "names=.*", "rv1", &resource_graph_json) < 0) {
```
jameshcorbett, line 1233: Codex caught a leak here:

> 
> run_find() returns a newly owned json_t * with refcount 1. The O* formatter gives the response its own reference,
>   but the original reference remains owned by the callback and is never released.
> 
>   I’d use the repository’s json::value RAII wrapper so every success and error path is covered:

```
  diff --git a/resource/modules/resource.cpp b/resource/modules/resource.cpp
  @@ -1228,12 +1228,15 @@ static void notify_request_cb (...)

           // Traverse all nodes to calculate the current resource set
  -        json_t *resource_graph_json = nullptr;
  +        json_t *resource_graph_json_raw = nullptr;
  +        json::value resource_graph_json;
           if (m->get_notify_flags () & NOTIFY_RESOURCES) {
  -            if (run_find (ctx, "names=.*", "rv1", &resource_graph_json) < 0) {
  +            if (run_find (ctx, "names=.*", "rv1", &resource_graph_json_raw) < 0) {
                   flux_log_error (h, "%s: run_find", __FUNCTION__);
                   goto error;
               }
  +            resource_graph_json = json::value::take (resource_graph_json_raw);
           }

  @@ -1282,7 +1285,7 @@ static void notify_request_cb (...)
                                  msg,
                                  "{s:O* s:s* s:s* s:f}",
                                  NOTIFY_RESOURCES_KEY,
  -                               resource_graph_json,
  +                               resource_graph_json.get (),
                                  NOTIFY_UP_KEY,
```

>   The reference lifecycle then becomes:
> 
>   1. run_find() creates refcount 1.
>   2. json::value::take() assumes ownership without incrementing it.
>   3. O* gives the packed response its own reference.
>   4. json::value releases the callback’s reference when the scope exits.
>   5. The response releases its reference through its normal lifecycle.
> 
>   This also cleans up correctly if status processing, response packing, or subscriber insertion fails after
>   run_find(). Simply adding json_decref() after the successful response would still leak on those error paths.

jacobtkeio: Do you think I should include an `Assisted-by:` trailer in the commit message when I accept this patch? If so, what name and model version should I report?

jameshcorbett: You could do Codex:gpt-5.6 but I think it's totally optional.

PR #1557, opened 2026-09-10
resource: add R properties to JGF nodes

### LINKED ISSUE #1556: R-JGF property agreement validation
A common workflow among the system administrators for deploying dedicated application time (DAT)/reservations is to: 

- stop all queues, 
- unload Fluxion, 
- unload the core resource module, 
- update the configuration of resources (i.e. by adding a property to a bunch of nodes that didn't have it before and a corresponding new queue), 
- and then reload the core resource module and Fluxion modules.

A problem comes up on our systems where Fluxion is using JGF if the node queue properties were added in the core resource module's _R_ but _not_ in the JGF graph representation on which Fluxion relies. 

Initially, in the meeting we proposed having Fluxion (or in the interim the admin script) validate that the nodes in both R and JGF agreed on all properties for each node. However, @jameshcorbett pointed out that there are some valid reasons to have properties that are in JGF but not in R. 

@garlick suggested that JGF consider R's "node properties" like queue to be authoritative, then have JGF just add properties to those in its representation (possibly as a custom writer in Fluxion).

As a corollary to this, @ryanday36 urged us to add a test to our testsuite (maybe an issues test?) that incorporates [the workflow the admins use in their script](https://izlc.llnl.gov/gitlab/hobbs17/ansible/-/blob/main/admin-scripts/scripts/flux_reconfig.sh?ref_type=heads) to set up DATs since it's caused a few bugs to come to light over the last year.

jameshcorbett:
Problem: when an R payload contains a complete scheduling graph, the JGF
reader builds the graph but ignores node properties carried by the R
execution section.

Decode execution properties by rank and configure the JGF reader to merge
them into node vertices after namespace remapping. Preserve explicit JGF
property values when the same property is present in both sources.

Fixes #1556


wihobbs:
On both my test lima vm (ubuntu) and corona, I got a segfault when I tried to build and load the module compiled off of this branch:
```
(s=1,d=1)  corona186 $ flux start
(s=1,d=0)  corona186 $ flux --version
commands:    		0.88.0
libflux-core:		0.88.0
libflux-security:	0.15.0
broker:  		0.88.0
FLUX_URI:		local:///var/tmp/hobbs17/flux-D1P2FN/local-0
build-options:		+ascii-only+systemd+hwloc.api==2.12.0+zmq==4.3.4
(s=1,d=0)  corona186 $ flux module unload sched-fluxion-resource
(s=1,d=0)  corona186 $ flux module load $(pwd)/sched-fluxion-resource.so
flux-module: load /g/g0/hobbs17/flux-sched/build/resource/modules/sched-fluxion-resource.so: Invalid argument
Segmentation fault (core dumped)
$ flux: flux_open: Connection refusedat james
flux: flux_open: Connection refused
(s=,d=)  corona186 ~/flux-sched/build/resource/modules $ exit
Sep 10 17:22:53.764141 PDT 2026 broker.err[0]: rc2.0: /bin/bash Exited (rc=139) 54.2s
flux: flux_open: Connection refused
flux: flux_open: Connection refused
[detached: session exiting]

**Comment by jameshcorbett** [comment]
> On both my test lima vm (ubuntu) and corona, I got a segfault when I tried to build and load the module compiled off of this branch:
> 
> ```
> (s=1,d=1)  corona186 $ flux start
> (s=1,d=0)  corona186 $ flux --version
> commands:    		0.88.0
> libflux-core:		0.88.0
> libflux-security:	0.15.0
> broker:  		0.88.0
> FLUX_URI:		local:///var/tmp/hobbs17/flux-D1P2FN/local-0
> build-options:		+ascii-only+systemd+hwloc.api==2.12.0+zmq==4.3.4
> (s=1,d=0)  corona186 $ flux module unload sched-fluxion-resource
> (s=1,d=0)  corona186 $ flux module load $(pwd)/sched-fluxion-resource.so
> flux-module: load /g/g0/hobbs17/flux-sched/build/resource/modules/sched-fluxion-resource.so: Invalid argument
> Segmentation fault (core dumped)
> $ flux: flux_open: Connection refusedat james
> flux: flux_open: Connection refused
> (s=,d=)  corona186 ~/flux-sched/build/resource/modules $ exit
> Sep 10 17:22:53.764141 PDT 2026 broker.err[0]: rc2.0: /bin/bash Exited (rc=139) 54.2s
> flux: flux_open: Connection refused
> flux: flux_open: Connection refused
> [detached: session exiting]
> ```

I believe this may be because the ABI changes with this PR and so if you have the feasibility module loaded (did you?) there is an incompatibility between the new `sched-fluxion-resource` module's expectation and the `libsched-fluxion-resource-module.so` kept loaded by feasibility. At least that is my hypothesis so far. If that's right maybe @milroy will have thoughts on how best to resolve or if that's an acceptable danger/limitation to have.

**Comment by wihobbs** [comment]
> I believe this may be because the ABI changes with this PR and so if you have the feasibility module loaded (did you?) there is an incompatibility between the new sched-fluxion-resource module's expectation and the libsched-fluxion-resource-module.so kept loaded by feasibility. At least that is my hypothesis so far. 

I did have feasibility loaded, but feasibility exits with an error whenever `sched-fluxion-resource` is unloaded. I just checked and if I start Flux with sched-simple, unload that and then load Fluxion off your branch, there's no segmentation fault initially. 

Can we purge the loaded `libsched-fluxion-resource-module.so` whenever `sched-fluxion-resource` is unloaded so that Fluxion can be unloaded and reloaded with version changes? Perhaps an issue or separate PR.

If we keep this limitation we should have a note somewhere warning users/sysadmins about the breakage.

Edit: for anyone else who wants to test by hand with an installed binary, use `$ FLUX_MODPROBE_DISABLE=sched-fluxion-resource flux start` to force Flux to start with sched-simple.

File: resource/modules/resource_match.cpp [hunk 1]
```diff
@@ -949,6 +960,11 @@ static int grow_resource_db_jgf (std::shared_ptr<resource_ctx_t> &ctx, json_t *r
         rc = 0;
     }
     if (db.metadata.roots.find (containment_sub) == db.metadata.roots.end ()) {
+        auto jgf_reader = std::dynamic_pointer_cast<resource_reader_jgf_t> (ctx->reader);
+        if (!jgf_reader || jgf_reader->set_node_properties (properties) < 0) {
+            flux_log_error (ctx->h, "%s: invalid execution properties", __FUNCTION__);
+            goto done;
```
wihobbs, line 972: Should errno be set here? Failure to initialize `jgf_reader` could be `ENOMEM`, no?

resource/readers/resource_reader_jgf.cpp
```diff
@@ -29,6 +29,39 @@ extern "C" {
...
 using namespace Flux::resource_model;
 
+int resource_reader_jgf_t::set_node_properties (json_t *properties)
+{
+    const char *name = nullptr;
+    json_t *ranks_json = nullptr;
+    struct idset *ranks = nullptr;
+
+    m_node_properties.clear ();
+    if (!properties)
```
wihobbs, line 39: This feels too-pythonic-to-be-true for C++. Is this checking for if `properties` is `{}, or if it's just `nullptr`?

t/t1030-rv1-shorthand.t
```diff
@@ -11,6 +11,12 @@ SIZE="$(flux hostlist -c ${HOSTLIST})"
 
 test_under_flux ${SIZE}
 
+test_expect_success 'add overlapping property to JGF' '
```
wihobbs, line 14: I didn't see any updates to the JGF shorthand reader, so why the update to the test?

wihobbs (commented):
I poked at this by hand and verified:

- queue properties in JGF are ignored when they exist in R (prior to this PR, wonky output showed up in flux resource list)
- putting no properties in R but having some in JGF, they are also ignored by `flux resource` 
- JGF with mismatched properties can still be loaded, but the R takes precedence

Sorry my other comments slid through. I thought I was in a review.

resource/modules/resource_match.cpp
```diff
@@ -28,6 +28,7 @@ extern "C" {
 #include "resource_match.hpp"
 #include "resource/schema/resource_graph.hpp"
 #include "resource/readers/resource_reader_factory.hpp"
+#include "resource/readers/resource_reader_jgf.hpp"
```
wihobbs, line 31: Commit nit: assisted-by line should include the minor version for GPT (5.4, 5.6) and the model (sol, luna, etc.)

resource/readers/resource_reader_jgf.cpp
jameshcorbett: It's just a `NULL / nullptr` check! Check for whether it's an object is down a few lines.

t/t1030-rv1-shorthand.t
jameshcorbett: The JGF shorthand reader is the JGF reader plus some add-ons, and we want to ensure the new property functionality works with JGF shorthand, so I think it's appropriate.

resource/modules/resource_match.cpp
jameshcorbett: `set_node_properties` sets `errno` so I think this is OK! `jgf_reader` is just being cast, not being initialized, so although that case doesn't get its own error code I think that's all right.

resource/modules/resource_match.cpp
jameshcorbett: Fixed momentarily!

resource/readers/resource_reader_jgf.cpp
wihobbs: I see that, I was just saying you could check to see if the JSON object is empty `{}` and bail out here. But that's not a huge performance improvement by any means.

resource/modules/resource_match.cpp
wihobbs: Fair. Should `rc` be set though if you're going to `done` after?

resource/readers/resource_reader_jgf.cpp
jameshcorbett: Oh that's true! I think it's ok the way it is though, the `json_object_foreach` loop will just exit immediately.

resource/modules/resource_match.cpp
jameshcorbett: Good point! Done.

wihobbs (approved):
I tried several ways to make this break and couldn't find any!

jameshcorbett:
Updated the last commit to add a test for the `storage_node` type.

Issue #1558, opened 2026-09-01
`duplicate housekeeping allocation` error seen on elcap


grondo:
According to @kkier, a node was stuck in housekeeping for > 1 week. After a config change and reload (of scheduler and `resource` modules), commands that fetch resource status were returning `duplicate housekeeping allocation`. A reboot of the affected node fixed the error.

There isn't much data now on this, but maybe @kkier can come back around and add any extra available detail (e.g. what the config change was). 

If this happens again, the output of `flux housekeeping list` might be interesting to see if it shows that two jobs really were in housekeeping on the same node.

grondo:
I was finally able to reproduce this. The reproducer requires `match-foramt = "rv1_shorthand"` and housekeeping still _partially_ active for a previous job (i.e. some nodes have completed housekeeping before the reconfigure)

I had Claude help write a comprehensive sharness for the reconfigure queues + reload modules use case, which I'll attach here. This test executes a change queues + reconfigure/reload + run test job scenario over a matrix of possible configurations:


The test exercises a 4-node instance with three property-selected queues (`pall` and `pbatch` over all ranks, `pdat` over ranks 1-2), and runs the site reconfigure sequence (stop queues `--nocheckpoint`, unload feasibility/qmanager/fluxion-resource/resource, `flux config reload`, reload in reverse order, restart queues) while resources are held. Rank 1 is the held node in every case and is in both `pbatch` and `pdat`.

Matrix:

* Fluxion match-format:
  * `rv1`
  * `rv1_shorthand`
  * `rv1_nosched`
* What holds resources across the reconfigure:
  * `housekeeping-partial` — 4-node job completes, `release-after=0` frees ranks 0,2,3, rank 1 left in housekeeping
  * `housekeeping-whole` — 1-node job on rank 1 completes, whole allocation passes to housekeeping (no partial release)
  * `running` — 2-node job still running, no housekeeping active
  * `both` — partial-release straggler plus a separate running job
* Queue configuration change, applied in both R and `queues.toml`:
  * add `pdat` (overlapping `pbatch`, including the held node)
  * remove `pdat` again

After each reconfigure the test asserts that the scheduler's free node count still matches reality, that a job needing a held node stays pending (both from `pbatch` and from `pdat`), that nothing was drained, and that the queue and property changes actually took effect.

The test reproduces this bug for the `rv1_shorthand` and `housekeeping-partial` case.

[...]

PR #1559, opened 2026-09-15
  resource: fix lost allocations on reload with rv1_shorthand

grondo:
With `match-format=rv1_shorthand`, reloading the scheduler modules while a job's housekeeping task holds a subset of that job's nodes causes fluxion to lose the allocation entirely. The resources may then be allocated to a new job.

This PR ensures that vertices removed from a shorthand JGF because their rank was already released are skipped rather than treated as a silent error, so the ranks the job still holds are processed and are marked allocated.

A few memory leaks were also discovered during testing and are fixed here  Also included:

- Fix a leak on partial release found during testing 
- Propagate return codes in `update_vertices()` so a failed update reports failure and rolls back.
- Initialize `vmap_val_t` members so a `vmap` lookup miss yields a detectable value
- sharness: drop the `libtool e` prefix so `FLUX_TEST_VALGRIND` works in this project.
- add a comprehensive reconfigure/reload sharness test which originally reproduced this bug, and ensures that reconfigure with queues and reload of scheduler modules works over several scenarios and match formats.
 
This PR was heavily assisted by Claude, so needs careful attention from a developer. Also, feel free to take one or all of these commits to create a new PR if this one isn't quite right.


t/sharness.d/flux-sharness.sh
```diff
@@ -248,7 +248,7 @@ test_under_flux() {
 
     if test -n "$FLUX_TEST_VALGRIND" ; then
         VALGRIND_SUPPRESSIONS=${SHARNESS_TEST_SRCDIR}/valgrind/valgrind.supp
-        valgrind="--wrap=libtool,e"
+        valgrind="--wrap="
         valgrind="$valgrind,valgrind,--leak-check=full"
```
jameshcorbett: This yields `--wrap=,valgrind,--leak-check=full`. Maybe combine two of the lines into one, like `--wrap=valgrind,--leak-check=full`?

resource/readers/resource_reader_jgf.cpp
```diff
@@ -992,12 +992,12 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
...
-                                    fetcher,
-                                    additional_vertices,
-                                    update_data.sequence_number)
+        if ((rc = fetch_additional_edges (g,
+                                          m,
+                                          vmap,
+                                          fetcher,
+                                          additional_vertices,
+                                          update_data.sequence_number))
             < 0) {
```
jameshcorbett: while you're at it make this != 0 to be consistent with the others?

resource/readers/resource_reader_jgf_shorthand.cpp
```diff
@@ -105,6 +105,12 @@ int resource_reader_jgf_shorthand_t::update_additional_edges (
     vtx_t v;
     std::map<std::string, vmap_val_t> empty_vmap{};  // so `find_vtx` doesn't error out
     std::string vertex_id = std::to_string (fetcher.uniq_id);
+    // update_vtx () skips a vertex whose rank was released by a previous
+    // partial free, so it has no vmap entry and no edges to update here.
+    // update_edges () tolerates this for a complete JGF via unpack_edge ();
+    // do the same for the vertices elided from a shorthand JGF.
+    if (vmap.find (vertex_id) == vmap.end ())
+        return 0;
```
jameshcorbett, line 113: This change looks good to me. However, I think the following would be equivalent and more efficient, since it would prevent the recursive searching for child vertices entirely:

```
diff --git a/resource/readers/resource_reader_jgf.cpp b/resource/readers/resource_reader_jgf.cpp
index b5eb0277..643bf515 100644
--- a/resource/readers/resource_reader_jgf.cpp
+++ b/resource/readers/resource_reader_jgf.cpp
@@ -857,6 +857,7 @@ int resource_reader_jgf_t::update_vtx (resource_graph_t &g,
     vtx_t v = boost::graph_traits<resource_graph_t>::null_vertex ();
     std::pair<std::map<std::string, vmap_val_t>::iterator, bool> ptr;
 
+    update_data.skipped = false;
     if ((rc = find_vtx (g, m, vmap, fetcher, v)) != 0)
         goto done;
     if ((rc = check_root (v, g, root_checks)) != 0)
@@ -864,6 +865,7 @@ int resource_reader_jgf_t::update_vtx (resource_graph_t &g,
     // Check if skipping due to previous partial free
     if (update_data.isect_ranks && !update_data.ranks.empty ()) {
         if (update_data.ranks.find (fetcher.rank) != update_data.ranks.end ()) {
+            update_data.skipped = true;
             rc = 0;
             goto done;
         }
@@ -983,6 +985,12 @@ int resource_reader_jgf_t::update_vertices (resource_graph_t &g,
             goto done;
         if ((rc = update_vtx (g, m, vmap, fetcher, update_data)) != 0)
             goto done;
+        if (update_data.skipped) {
+            // The rank was released by a previous partial free, so this
+            // vertex is not in vmap.  Anything a reader would collect below
+            // it shares its rank and would be skipped too, so don't walk it.
+            continue;
+        }
         if ((rc = fetch_additional_vertices (g, m, fetcher, additional_vertices)) != 0)
             goto done;
         for (auto &additional_fetcher : additional_vertices) {
```

jameshcorbett (approved):
Looks great to me, thanks @grondo! Some nonessential suggestions below--

t/sharness.d/flux-sharness.sh
grondo: Ok, but `--wrap` tolerates the leading `,` (just fyi)

resource/readers/resource_reader_jgf_shorthand.cpp
grondo: Claude did suggest keeping the existing conditional as defense in depth, even with this efficiency improvement.
The reasoning:

>  the continue handles the common node-exclusive case and avoids the subtree walk, but it can't cover an exclusive subtree rooted above node level (rank -1, never skipped) whose elided node descendants are on freed ranks — those reach update_additional_edges() through the additional-vertices loop, which the continue doesn't guard.

I'm not sure I understand this fully, so I'll let you decide. I'll add this patch with your authorship and leave the existing guard alone for now. Easy to drop later if it is truly dead code.

grondo:
I've force pushed here after addressing @jameshcorbett's comments.

For the suggested patch in [this comment](https://github.com/flux-framework/flux-sched/pull/1559#discussion_r4031820855), I added a separate commit authored by @jameshcorbett. I believe the existing guard is still good defensive code, but if that's not true then I'm happy to drop it!

resource/readers/resource_reader_jgf_shorthand.cpp
milroy: > the continue handles the common node-exclusive case and avoids the subtree walk, but it can't cover an exclusive subtree rooted above node level (rank -1, never skipped) whose elided node descendants are on freed ranks

This is true, and is related to a pathological case in partial cancel that I commented in the code but didn't fix because I didn't think it would occur in production. Inspecting this change with Fable's help has revealed that this case is possible in production. I created issue #1561 to describe the problem and suggest a fix and tests.

milroy: >   // update_vtx () skips a vertex whose rank was released by a previous
    // partial free, so it has no vmap entry and no edges to update here.
    // update_edges () tolerates this for a complete JGF via unpack_edge ();
    // do the same for the vertices elided from a shorthand JGF.

After the continue in `update_vertices` in @jameshcorbett's change above, the `vmap.find (vertex_id)` guard does not execute for any JGF that the shorthand writer emits. The comment in the code gives the wrong reason: a freed node is the target of the edge from its parent, and `update_tgt_edge` has no target guard. Please update the comment to state that the guard protects a vertex that is fetched but is not in `vmap`.

t/t1035-reconfig-reload-sweep.t
```diff
@@ -0,0 +1,438 @@
...
+		fi
+	} >conf.d/queues.toml
+}
+
+# Number of nodes the scheduler believes are free, as opposed to the
+# number the resource module believes are free. The difference is the
+# whole point of the test.
+sched_free_nnodes () {
+	FLUX_RESOURCE_LIST_RPC=sched.resource-status \
+	    flux resource list -s free -no {nnodes}
```
milroy: Under a policy that is not `firstnodex`, free node count is not a valid check. Suggestion: `-s allocated`.

t/t1035-reconfig-reload-sweep.t
```diff
@@ -0,0 +1,438 @@
...
+# leaving it running invalidates every later test in the scenario, and its
+# own housekeeping task would collide with the one under test.
+stays_pending () {
+	id=$(flux submit --queue=$1 -N$2 sleep inf) || return 1
+	test_must_fail flux job wait-event -t $PENDING_TIMEOUT $id alloc
+	rc=$?
+	flux cancel $id >/dev/null 2>&1
+	flux job wait-event -t 30 $id clean >/dev/null 2>&1
+	return $rc
+}
```
milroy, line 334: A job that dies also passes this check. Suggestion: add `flux jobs -no {state} $id` and check that the state is `SCHED`.

t/t1035-reconfig-reload-sweep.t
```diff
@@ -0,0 +1,438 @@
+#!/bin/sh
```
milroy, line 1: Overall comments on this commit:

- The test holds housekeeping on rank 1. If it holds rank 0, the test passes on the old code. Hold rank 1 on purpose and write a comment that explains why.
- Add `skip_all for `HAVE_PARTIAL_OK` not set, and for ASAN builds as t3044 does.
- The test run time is 5–7 minutes. Please reduce the timeouts where possible.

milroy (commented):
This looks really good, and we should get this in soon. Fable and I did come up with some changes, including increasing the test coverage:

1. **Reload of a node-exclusive job with `free_ranks` (resource-query).**
   Add this test to `t/t3041-jgf-shorthand-load-format.t`. Do not add a new
   file. That file already builds the JGF with `flux ion-R encode`, matches
   the 20-node jobspec under `-P lonodex`, and writes `shorthand.json`. Add
   one `test_expect_success` after the `update graph to allocate a job using
   jgf_shorthand` test:
   - Run `jq '. + {free_ranks: "0-4"}' shorthand.json > shorthand_free.json`.
   - Run `update allocate jgf_shorthand shorthand_free.json 0 0 5`.
   - Run `find jobid-alloc=0` with `-F jgf`. Check that the output has 240
     cores and 60 gpus (15 nodes), not 320 and 80.
   - Run `match allocate` with a one-node jobspec. Check that it succeeds.

   This is a `resource-query` unit test for issue #1558. Without it, only the live test
   `t1035` covers the reader change.

2. **Reload of a node-exclusive job with `free_ranks`, `jgf` reader.** Add
   this test to the same file. Run the `generate jgf_shorthand` match again
   with `-F jgf` to get a full JGF of the same job. Add the same `free_ranks`
   key with `jq`. Run `update allocate jgf <file> 0 0 5` and make the same
   checks. The `continue` in `update_vertices` is in the base reader, so both
   readers need coverage.

t/t1035-reconfig-reload-sweep.t
grondo: Even if the job runs and fails, `flux job wait-event` will still return success when the `alloc` event is posted. So the statement here is incorrect. The only way to get a false positive here (`flux job wait-event` fails) is if the job gets a `clean` event before `alloc` is posted (event never seen), which means the job was canceled. Since there's nothing canceling the job I don't think that could happen unless the instance was shutting down, in which case there would be other obvious failures.

t/t1035-reconfig-reload-sweep.t
grondo: It is too dangerous to remove the timeouts, they are only 5s each and on a busy system a smaller timeout might cause false positive failures.

The test runs long because it is sweeping many configurations. I think the runtime is probably worth getting the extra coverage, which was completely missing in the tree until now. Perhaps there are other ways to run this test faster (i.e. run some sweeps in parallel?) However, given the severity of this issue and the benefit of this test, I would argue merging as-is (it is a _proven_ reproducer of a real bug), and improve the tests later.

grondo: > The test holds housekeeping on rank 1. If it holds rank 0, the test passes on the old code. Hold rank 1 on purpose and write a comment that explains why.

I don't understand this comment. This is what the test does (explicitly holds rank 1)

t/t1035-reconfig-reload-sweep.t
grondo: Though I guess it doesn't hurt to add this check, so I'll add it.

t/t1035-reconfig-reload-sweep.t
grondo: Good catch. However, this test suite requires a node-exclusive policy (or other tests would fail, like attempting to allocate resources on an already (partially) allocated node would succeed).

However, the check for allocated resources is more direct, so I'll make this change.

t/t1035-reconfig-reload-sweep.t
milroy: > I don't understand this comment. This is what the test does (explicitly holds rank 1)

Sorry, that statement was confusing. I only meant to request a comment in the test that specifies that holding rank 1 is required.

The test detects issue #1558 if the held node comes after a freed node in the `graph.nodes` array of the JGF that qmanager sends to the reader ([resource/readers/resource_reader_jgf.cpp#L981](https://github.com/flux-framework/flux-sched/blob/bf17181e51304918c9fbc32c4b9643a7416a3ee8/resource/readers/resource_reader_jgf.cpp#L981)).

The match writer produced that JGF when the job was allocated. The update traversal visits the cluster's out-edges in insertion order, and the resource module inserted them in `R_lite` rank order when it built the graph, so `graph.nodes` lists the nodes in ascending rank. Holding rank 1 and freeing rank 0 therefore places the held node at a larger index than a freed node, and the old code left the held node unallocated. Holding rank 0 puts the held node at index 0, so the old code allocated it before exiting. If the traversal emitted vertices in a different order, the rank to test would change with it.

grondo: Thanks! Adding to the test description.

grondo:
I just force-pushed a new version with some updates to address above comments:
- added the requested comment about forcing delayed housekeeping on rank 1
- check for state == SCHED in `stays_pending()` after `flux job wait-event` failure 
- switch from using nodes free to nodes allocated in the test for directness (but note that tests require node-exclusive policy)
- Reduced runtime of the sweep test by decreasing timeout to 3s, and removing 1 redundant `stays_pending` check per loop
- added the suggested unit tests (fully authored by Claude). Proved that 2/4 tests failed before the fixes in this PR.

grondo:
I forgot to add the `skip_all` for `ASAN` to the test. Just added that and force-pushed again. Sorry for the noise.

I don't think we need to worry about the `partial-ok` check. This was added to flux-core in v0.70.0 and the minimum required version of flux-core in flux-sched is v0.78.0.

milroy (approved):
The new tests look good, as do the other updates. This is good to merge as far as I'm concerned, @grondo!

t/t3041-jgf-shorthand-load-format.t
```diff
@@ -55,4 +56,64 @@ test_expect_success 'JGF output shows all cores and gpus allocated' '
...
+test_expect_success 'update graph with free_ranks using jgf' '
+    ${query} -L JGF.json -f jgf_shorthand -F jgf -t full1.out -P lonodex < match_jobspec_cmd &&
+    head -n1 full1.out | jq -S . > full.json &&
+    jq ". + {free_ranks: \"0-4\"}" full.json > full_free.json &&
+    cat > update_full_cmd <<-EOF &&
+    update allocate jgf full_free.json 0 0 5
+    find jobid-alloc=0
+    quit
+EOF
+    jq -S .scheduling R_JGF.json > JGF.json &&
```
jameshcorbett: This test case assumes `JGF.json` exists (see the first command in this test case) but then regenerates it.

grondo:
Thanks @milroy and @jameshcorbett. I'll set MWP.

Issue #1560, opened 2026-09-17
shrink error when node lost in el capitan subinstance


garlick:
Problem: during el capitan testing yesterday, we had a bad node that dropped out of a subinstance.  This triggered a `resource.acquire` shrink response, which cause fluxion to log the following (from hand written note, may be a bit off):

```
partial cancel by ranks ("3") failed
mod_agfilter: planner_multi_reduce_span returned -1
cluster0: Invalid argument
remove cluster0: cancel_vertex failed
update_resource_db shrink (lost): Invalid argument
```

grondo:
I was able to reproduce this only by running multiple jobs on the node that is affected by the shrink, though I don't think we were running in that manner when we saw these errors production. However, there does seem to be a definite bug here.

Here's the reproducer, meant to be saved as `t/issues/t1560-shrink-multiple-jobs-on-lost-rank.sh`:

Example run:
```console
$  T5100_ISSUES_GLOB="t1560*" ./t5100-issues-test-driver.t
# ./t5100-issues-test-driver.t: flux session size will be 2
issue#1560: Re-launching test script under flux-start
issue#1560: Loading fluxion...
issue#1560: Submitting two single-core jobs, both on rank 3...
fLjXSZD RUN 3
fJY2WKZ RUN 3
issue#1560: Forcing rank 3 down to trigger a shrink...
Sep 17 23:10:43.971218 UTC 2026 sched-fluxion-resource.err[0]: partial cancel by ranks ("3") failed: mod_agfilter: planner_multi_reduce_span returned -1.
Sep 17 23:10:43.971226 UTC 2026 sched-fluxion-resource.err[0]: cluster0 Invalid argument.
Sep 17 23:10:43.971227 UTC 2026 sched-fluxion-resource.err[0]: remove: cluster0: cancel_vertex failed.
Sep 17 23:10:43.971230 UTC 2026 sched-fluxion-resource.err[0]: update_resource_db: shrink (lost): Invalid argument
Sep 17 23:10:43.971230 UTC 2026 sched-fluxion-resource.err[0]: update_resource: update_resource_db: Invalid argument
issue#1560: FAIL: fluxion reported errors during shrink:
[Sep17 23:10] sched-fluxion-resource.err[0]: partial cancel by ranks ("3") failed: mod_agfilter: planner_multi_reduce_span returned -1.
[  +0.000009] sched-fluxion-resource.err[0]: remove: cluster0: cancel_vertex failed.
[...]

PR #1562, opened 2026-09-22
resource: fix stale `sched.resource-status` response after shrink

grondo:
`shrink_resources()` does not invalidate the `sched.resource-status` cache (`m_r_all`/`m_r_down`), unlike `grow_resource_db()` and `mark()`. Any status request issued before a shrink populates that cache, and every request afterward is served the stale pre-shrink view, so fluxion keeps reporting nodes that are no longer in the resource set. The impact is limited since `sched.resource-status` is not the typical resource status endpoint, and this only affects subinstances AFAIK.

However, while testing this issue it was also apparent that the existing `t2317-resource-shrink.t` test has some flaws. Fluxion wasn't even loaded in the subinstance, so the test was merely duplicating testing with sched-simple already implemented in flux-core. Therefore, this PR also rewrites that test to 1. ensure Fluxion is loaded in subinstances, 2. check that resource-status from the subinstance reflects the shrink, and 3. tests with JGF in the enclosing instance and a couple different configurations for enhanced coverage.


grondo:
Hold off on a review of this one as I'm having trouble getting the new test to run reliably in CI.

grondo:
I think the failures here are related to #1560, possibly the same bug. The test failure is caused by Fluxion modules exiting in the subinstance after a shrink is requested, and for some reason are not 100% reproducible in this test. Once #1560 is resolved, we can try these tests again.

cc: @milroy

